PR #8, opened 2023-04-12
fix(docs): physics.blur.gaussian_blur

jscanvic:


tachella:
many thanks for spotting this - it has been fixed in a recent push

PR #26, opened 2023-05-24
Sphinx gallery

tachella:
# Update

## Changes to Physics
- Added single pixel camera operator.
- Added single photon lidar operator.
- fixed blurFFT to be real SVD.
- small fixes to MRI.
- Decolorize has now SVD.

## Changes to demos
- Fixed and documented all existing demos.
- Added new sampling and basic demos.

## Changes to test
- Added a sampling test using a posterior which is available in closed form

## Changes to docs
- Improved organization of docs (welcome page)
- added docs for plotting utils
- sphinx gallery integration

- [x] Added an entry to the [CHANGELOG.md](../../CHANGELOG.md) (as applicable) (to come).


deepinv/tests/test_sampling.py
```diff
@@ -17,84 +16,109 @@ def device():
...
+            sigma=1,
+        )
+    elif algo == "DDRM":
+        diff = dinv.sampling.DDRM(
+            denoiser=GaussianDenoiser(sigma_prior),
+            sigma_noise=sigma,
+            eta=1,
+            sigmas=np.linspace(1, 0, 100),
         )
+        out = dinv.sampling.DiffusionSampler(diff, clip=(-100, 100), max_iter=100)
```
matthieutrs, line 57: This test failed here in the last push with no apparent reason; maybe we can set a random seed to avoid having some random failuers?

deepinv/loss/regularisers.py
```diff
@@ -87,7 +98,7 @@ def forward(self, y, x):
                 v.detach_()
                 u.detach_()
 
-        return z.view(-1).sqrt()
+        return self.weight * z.view(-1).sqrt()
```
matthieutrs, line 101: I would not do this: this functions gives the spectral norm of the Jacobian, so if we multiply the max eigenvalue by `self.weight`, it is confusing. I would rather add the weight somewhere else. If we do not find anything clearer, then I would move the compute function inside a `def compute(self, ...)` method (or any other name) returning the proper computation, and the forward would just call `self.compute(...)` and multiply the output with `self.weight`. What do you think is best?

deepinv/optim/optimizers.py
```diff
@@ -243,11 +243,13 @@ def update_params_fn(self, it):
 
     def init_params_fn(self):
         r"""
-        Get initialization parameters and re-initialized the parameters dictionary to its initial value. This is necessary if the parameters have been updated during optimizaiton, for example via backtracking.
+        Get initialization parameters and re-initialized the parameters' dictionary to its initial value.
```
matthieutrs, line 246: Needs clarification, can assign myself to that.

examples/sampling/demo_ddrm.py
```diff
@@ -0,0 +1,159 @@
...
+# We can now reconstruct a noisy measurement using the diffusion method.
+# We use the DDRM method from :class:`deepinv.sampling.DDRM`, which works with inverse problems that
+# have a closed form singular value decomposition of the forward operator.
+# The diffusion method requires a schedule of noise levels ``sigmas`` that are used to evaluate the denoiser.
+
+diff = dinv.sampling.DDRM(
+    denoiser=denoiser,
+    etab=1.0,
+    sigma_noise=sigma,
+    sigmas=np.linspace(1, 0, 100),  # uses 100 evaluations of the denoiser
```
matthieutrs, line 94: Time issues on CPU

```suggestion
    sigmas=np.linspace(1, 0, 10),  # uses 10 evaluations of the denoiser; set to sigmas=np.linspace(1, 0, 10) for better results
```

matthieutrs (changes requested):
Great PR! I'm requesting changes because I am afraid sphinx-gallery examples won't run with current parameters (e.g. number of iterations) that are too optimistic for CPU.

deepinv/tests/test_sampling.py
tachella: Yes, good idea

deepinv/loss/regularisers.py
tachella: good point. Maybe we can add the weights somewhere in the train function, and avoid adding them to the losses one by one.

deepinv/physics/blur.py
```diff
@@ -460,74 +467,92 @@ class BlurFFT(DecomposablePhysics):
...
+    plt.show()
+    plt.imshow(y.squeeze(0).permute(1, 2, 0).cpu().numpy())
+    plt.show()
+    plt.imshow(xhat.squeeze(0).permute(1, 2, 0).cpu().numpy())
+    plt.show()
+    plt.imshow(xhat2.squeeze(0).permute(1, 2, 0).cpu().numpy())
+    plt.show()
+
+    plt.imshow(physics.A(xhat).squeeze(0).permute(1, 2, 0).cpu().numpy())
+    plt.show()
```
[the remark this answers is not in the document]
tachella: this code is just for debugging the blur operators. We can just comment it

examples/sampling/demo_sampling.py
```diff
@@ -0,0 +1,152 @@
...
+#
+# Here we use the Unadjusted Langevin Algorithm (ULA) to sample from the posterior defined in
+# :class:`deepinv.sampling.ULA`.
+# The hyperparameter ``step_size`` controls the step size of the MCMC sampler,
+# ``regularization`` controls the strength of the prior and
+# ``iterations`` controls the number of iterations of the sampler.
+
+regularization = 0.9
+step_size = 0.01 * (sigma**2)
+iterations = int(5e3)
```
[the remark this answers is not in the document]
tachella: we can also reduce the image size

deepinv/__init__.py
```diff
@@ -48,10 +48,14 @@
...
 __all__ += ["sampling"]
 
-from deepinv.loss import metric as metric
+from deepinv.loss import metric
 
 __all__ += ["metric"]
 
+__all__ += ["unfolded"]
+
+from deepinv import unfolded
```
tomMoral, line 57: Reverse the two lines?
```suggestion
from deepinv import unfolded

__all__ += ["unfolded"]
```

docs/source/deepinv.loss.rst
```diff
@@ -38,6 +38,18 @@ about the forward measurement process.
     deepinv.loss.SurePGLoss
     deepinv.loss.TVLoss
 
+Transforms
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
tomMoral, line 42:
```suggestion
Transforms
^^^^^^^^^^
```

docs/source/deepinv.loss.rst
tomMoral, line 1: Usually, this module level documentation is put as a docstring in the `__init__.py` file of the module, so here in `deepinv/loss/__init__.py`, so it can be accessed as

```python
import deepinv.loss
deepinv.loss.__doc__
```

It will also be generated with autodoc in the API description page.

docs/source/index.rst
```diff
@@ -22,27 +34,35 @@ Here quick guide
...
+============================
+
+The preferred way to contribute to ``deepinv`` is to fork the `main
+repository <https://github.com/deepinv/deepinv/>`__ on GitHub,
+then submit a "Pull Request" (PR).
+
 
 Lead Developers
-==================
+============================
```
tomMoral, line 61: Usually, rst linters prefer that the length of the `====` is the same as the one of the text, for easier reading in text mode (in particular in docstrings).

examples/basics/demo_physics.py
```diff
@@ -0,0 +1,165 @@
...
+#     :meth:`deepinv.physics.LinearPhysics.adjointness_test()`,
+#     and that it has a unit norm using :meth:`deepinv.physics.LinearPhysics.compute_norm()`.
+
+print(f"The linear operator has norm={physics.compute_norm(x):.2f}")
+
+if physics.adjointness_test(x) < 1e-5:
+    print("The linear operator has a well defined transpose")
+
+
+# %%
```
tomMoral, line 88: I think this should be a second example, maybe exposing simply more of the possibilities of the physic operator and having a specific decomposable operator example along side, maybe not in basics.

examples/basics/demo_physics.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.mask = 0.447
+        self.noise_model = dinv.physics.GaussianNoise(sigma=0.1)
+
+    def V_adjoint(self, x):
+        y = x[:, 0, :, :] * 0.6687 + x[:, 1, :, :] * 1.3132 + x[:, 2, :, :] * 0.2550
+        return y.unsqueeze(1)
```
tomMoral, line 114: Not sure this is the best example as here  it is very similar to the original one.
So if you don't know what this is, it might be complex to get it.

Maybe doing a transform in Fourier where everyone knows the SVD or something a little bit more didactic changing from RGB to another encoding  (with a matrix) and computing the SVD explicitly might give better insights.

examples/basics/demo_physics.py
```diff
@@ -0,0 +1,165 @@
...
+    Converts RGB images to grayscale.
+
+    Signals must be tensors with 3 colour (RGB) channels, i.e. [*,3,*,*]
+    The measurements are grayscale images.
+
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.mask = 0.447
```
tomMoral, line 109: seems unused? It is not clear what you need this so this should be explained in the example.

examples/basics/demo_physics.py
```diff
@@ -0,0 +1,165 @@
...
+end = time.time()
+print(f"Elapsed time for LinearPhysics: {end - start:.2f} seconds")
+
+start = time.time()
+for i in range(10):
+    xlin2 = physics2.A_dagger(x)
+    xprox2 = physics2.prox_l2(x, y2, 0.1)
+
+end = time.time()
+print(f"Elapsed time for DecomposablePhysics: {end - start:.2f} seconds")
```
tomMoral, line 165: If this goes faster than `0.01`, I recommend using higher precision or `.2e`.

tomMoral (commented):
A small round of comments but this looks pretty good!

examples/sampling/demo_ddrm.py
```diff
@@ -0,0 +1,159 @@
...
+    "Lionel-Messi-Argentina-2022-FIFA-World-Cup_%28cropped%29.jpg"
+)
+res = requests.get(url)
+x = imread(BytesIO(res.content)) / 255.0
+
+x = torch.tensor(x, device=dinv.device, dtype=torch.float).permute(2, 0, 1).unsqueeze(0)
+x = torch.nn.functional.interpolate(
+    x, scale_factor=0.5
+)  # reduce the image size for faster eval
+
```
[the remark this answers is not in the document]
matthieutrs: Added in demo_sampling as well.

docs/source/conf.py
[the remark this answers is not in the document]
matthieutrs: Ok, thanks! Not updating for now, FYI @tachella

docs/source/deepinv.loss.rst
matthieutrs: Might be good indeed @tachella. Leaving as is for this merge but might be worth doing that in the future?

tachella: yes, definitely. thanks for the recommendation!

examples/basics/demo_physics.py
matthieutrs: We should probably split this example in two and add a `Physics` section to the example.

matthieutrs (approved):

PR #56, opened 2023-06-29
fix: wrong return variable

jscanvic:


samuro95 (approved):
Thanks for spotting the bug !

PR #73, opened 2023-07-21
fix: log epoch loss instead of batch loss

jscanvic:
When called, the function `deepinv.train` should log the average batch loss of each epoch to wandb and to the disk. However, for both records, what is actually logged in the current implementation is the latest batch loss of each epoch. In this PR, we fixed the two bugs by logging, in both cases, the value `loss_meter.avg` instead of `loss_total`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ New Features
 
 Fixed
 ^^^^^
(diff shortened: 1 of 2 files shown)


tachella (approved):
thanks for spotting this bug

PR #76, opened 2023-08-02
feat: add SwinIR model

jscanvic:

### DIFF
--- a/deepinv/models/swinir.py
+++ b/deepinv/models/swinir.py
@@ -1,0 +1,867 @@
+# -----------------------------------------------------------------------------------
+# SwinIR: Image Restoration Using Swin Transformer, https://arxiv.org/abs/2108.10257
+# Originally Written by Ze Liu, Modified by Jingyun Liang.
+# -----------------------------------------------------------------------------------
+
+import math
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.utils.checkpoint as checkpoint
+from timm.models.layers import DropPath, to_2tuple, trunc_normal_
+
+
+class Mlp(nn.Module):
+    def __init__(self, in_features, hidden_features=None, out_features=None, act_layer=nn.GELU, drop=0.):
+        super().__init__()
+        out_features = out_features or in_features
+        hidden_features = hidden_features or in_features
+        self.fc1 = nn.Linear(in_features, hidden_features)
+        self.act = act_layer()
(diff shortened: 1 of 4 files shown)


deepinv/models/swinir.py
```diff
@@ -0,0 +1,1111 @@
...
+
+        for layer in self.layers:
+            x = layer(x, x_size)
+
+        x = self.norm(x)  # B L C
+        x = self.patch_unembed(x, x_size)
+
+        return x
+
+    def forward(self, x, sigma=None):
```
matthieutrs, line 1049: I would suggest to rename this forward

```
def forward_patch(self, x, sigma=None):
      ...
```

and to add a forward containing this small [snippet of code](https://github.com/JingyunLiang/SwinIR/blob/main/main_test_swinir.py#L68) (enabling to run on images of any size)

deepinv/models/swinir.py
```diff
@@ -0,0 +1,1111 @@
+# This file is taken (with only mild modifications) from the SwinIR repository:
```
matthieutrs, line 1: added

deepinv/tests/test_models.py
```diff
@@ -104,6 +104,25 @@ def test_denoiser(imsize, device, denoiser):
     assert x_hat.shape == x.shape
 
 
+def test_swinir_denoiser(device):
```
matthieutrs, line 107: Modified because of shape constraints, but the changes I've requested should make it possible to move it back to how it was previously.

matthieutrs (changes requested):

jscanvic:
Hi! Thanks for the review!

If I'm not mistaken, the `forward` method should already accept arbitrary-sized images, although I got confused about that when making the PR. I fixed the model initialization in the last commit and it should be fine now.

I also made sure the model is using pretrained weights by default.

deepinv/models/swinir.py
jscanvic: What should be an equivalent snippet of code is already called at the beginning of the method `forward`.

https://github.com/deepinv/deepinv/blob/ac5ebe093a8d08575317058f93da22f78efbb418/deepinv/models/swinir.py#L1057

matthieutrs:
Ok, thanks for the clarification! Code looks good to me now.

Minor updates: updated docstrings + added option on the noise level for downloading the pretrained model.

matthieutrs (approved):

PR #78, opened 2023-08-09
fix: call wandb.init only when needed

jscanvic:
Currently, when the functions `deepinv.train` and `deepinv.test` are called, the function `wandb.init` is called no matter if a run is already active or not. The current behavior can lead to unexpected effects. Instead, a more predictable behavior would be to call the function `wandb.init` only if no run is already active, i.e. if either the function `wandb.init` was never called, or if the function `wandb.finish` was called after the last call to the function `wandb.init`; or equivalently if the variable `wandb.run` is equal to the value `None`. In this PR, we implemented the latter behavior.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ New Features
 
 Fixed
 ^^^^^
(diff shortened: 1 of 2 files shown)


tachella (approved):
many thanks for fixing this issue, it looks good to me.

PR #83, opened 2023-09-20
Fix get_freer_gpu grep statement to work for different versions of nvidia-smi 

alexmehta:
In this PR I introduce a simple bug fix for the `get_freer_gpu` function. 

The grep statement in this function is problematic:

```nvidia-smi -q -d Memory |grep -A4 GPU|grep Free >tmp```

On some versions of nvidia-smi, this results in the "Free" label that is later referenced to be cut off. I just changed the `-A4` to `-A5` to address this. 


matthieutrs (approved):
Thanks for spotting this and PRing! lgtm, works on my side

PR #86, opened 2023-10-04
Anderson acc

samuro95:
Add Anderson Acceleration for acceleration of fixed-point resolution. Particularly useful for Deep Equilibrium.
It is a boolean argument `anderson_acceleration` in the class FixedPoint. With the parameters history_size / beta_anderson_acc / eps_anderson_acc. In order to use not only in the forward pass but also in the backward pass of DEQ, it is also an argument of the BaseDEQ class. I show how to use these arguments in the example demo_DEQ.py 


deepinv/tests/pyvenv.cfg
tachella, line 1: This seems like a spurious push

deepinv/optim/fixed_point.py
```diff
@@ -101,13 +109,101 @@ def __init__(
...
+        :param torch.Tensor x_hist: history of last ``history-size`` iterates.
+        :param torch.Tensor T_hist: history of T evlauaton at the last ``history-size``, where T is the fixed-point operator.
+        :param torch.Tensor H: H in the Anderson acceleration linear system Hp = q .
+        :param torch.Tensor q: q in the Anderson acceleration linear system Hp = q .
+        :param deepinv.optim.DataFidelity cur_data_fidelity: Instance of the DataFidelity class defining the current data_fidelity.
+        :param deepinv.optim.prior cur_prior: Instance of the Prior class defining the current prior.
+        :param dict cur_params: Dictionary containing the current parameters of the algorithm.
+        :param args: arguments for the iterator.
+        """
+        TX = self.iterator(X_prev, cur_data_fidelity, cur_prior, cur_params, *args)
```
matthieutrs, line 178: Maybe take this out of the function? Like that anderson is the same thing as regular but it's just an additional computation after the `self.iterator` call. Seems possible to me?

matthieutrs (commented):
Do you think we can leave the call to the iterator outside the Anderson acceleration?

deepinv/optim/fixed_point.py
samuro95: sounds good

matthieutrs (approved):

PR #91, opened 2023-10-17
Learned primal dual and training refactoring

samuro95:
The objective of this PR was to create an example file for replicating the results from the paper "Learned Primal-dual Reconstruction" (Adler et. al) (https://arxiv.org/pdf/1707.06474.pdf).  This paper proposes a unfolded version of the Primal-Dual algorithm. One of the specificity of the approach is that the Prox of data-fidelity term is also trained with a network. 
The overall approach is explained in the example file "learned_primal_dual.py".  

Above this particular problem, this PR contains the following more general updates : 
- mild refactoring of the training function, for better logging : add progress bar and the different metrics (losses and psnr) on the progress bar. 
- small refactoring in the BaseOptim class : the get_primal_updates and get_auxilliary_updates functions are removed and replaced by a function called get_output. By default, this function takes the dictionary representing the current iterate X = { "est" : (x,z), "cost" : F(x) } and output the primal update x . It can be replaced by a custom function if the desired output is an auxiliary variable or something else. For instance, for the PDNet example, the ouput is one specific dimension in x . 
- changed the wandb plotting code in order to handle the case of y and x with different shapes. 


deepinv/models/PDNet.py
tachella, line 1: did you add the PDNet to the init file in models? I cannot find it in the modified files

tachella: same for the docs - we should add it there too

deepinv/training_utils.py
```diff
@@ -73,130 +70,196 @@ def train(
...
 
     loss_history = []
 
+    log_dict = {}
+
     for epoch in range(epochs):
+        ### Evaluation
+
+        # perform evaluation every eval_interval epoch
+        perform_eval = (
```
tachella, line 127: the evaluation shouldn't happen after the training step inside the epoch?

tachella (changes requested):
the PR looks great - I just have some comments that I would be nice checking before the push

deepinv/training_utils.py
samuro95: I think it is usual to do an eval step before all so that it tests the initialized network and also be sure that there is no error in the eval loop without wiating for a whole trianing epoch

deepinv/training_utils.py
```diff
@@ -73,130 +70,196 @@ def train(
...
+                    "epoch": epoch,
+                    "learning rate": last_lr,
+                }
+            )
+            if perform_eval:
+                wandb.log({"eval psnr": test_psnr})
+
+        ### Training
+
+        model.train()
```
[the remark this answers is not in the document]
samuro95: discussed in previous PR ?

deepinv/training_utils.py
```diff
@@ -73,130 +70,196 @@ def train(
...
-                device,
-                verbose=False,
-                plot_images=plot_images,
-                plot_metrics=plot_metrics,
-                wandb_vis=wandb_vis,
-                step=epoch,
-                n_plot_max_wandb=n_plot_max_wandb,
+            wandb_vis
+        ):  # Note that this may not be 16 images because the last batch may be smaller
+            y_reshaped = torch.nn.functional.interpolate(y, size=x.shape[2])
```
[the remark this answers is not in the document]
samuro95: Ok, i'll change to resize with same scale + 0-pad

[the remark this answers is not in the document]
samuro95: I changed the wandb plotting code to use the fig already created with the matplotlib plotting.

deepinv/tests/test_models.py
```diff
@@ -243,34 +242,37 @@ def __init__(self, model, *args, **kwargs):
             self.model = model
 
         def prox(self, x, w):
-            return self.model(x, w)
+            return self.model(x, w[:, 1, :, :].unsqueeze(1))
```
matthieutrs, line 245: Shouldn't it be `w[:, 0, :, :].unsqueeze(1)`? Also, maybe use slicing: `w[:, 0:1, :, :]` without unsqueeze (or `w[:, 1:2, :, :]` in the case where you want channel index 1 and not 0)

deepinv/tests/test_models.py
```diff
@@ -243,34 +242,37 @@ def __init__(self, model, *args, **kwargs):
...
 
     class PDNetDataFid(DataFidelity):
         def __init__(self, model, *args, **kwargs):
             super().__init__(*args, **kwargs)
             self.model = model
 
         def prox(self, x, w, y):
-            return self.model(x, w, y)
+            return self.model(x, w[:, 2, :, :].unsqueeze(1), y)
```
matthieutrs, line 253: Same: shouldn't it be `w[:, 1, :, :].unsqueeze(1)`? Also, maybe use slicing: `w[:, 1:2, :, :]` without unsqueeze (or `w[:, 2:3, :, :]` in the case where you want channel index 2 and not 1)

matthieutrs (approved):

deepinv/tests/test_models.py
samuro95: Well seen, thanks

deepinv/training_utils.py
tachella: good point, I agree

deepinv/tests/test_models.py
tachella: @samuro95 what are your thoughts?

samuro95: yes it is correct, I changed accordingly. Thanks Matthieu !

PR #95, opened 2023-10-24
MTN move all config to pyproject.toml and simplify CI install

tomMoral:
This PR move all config to the modern `pyproject.toml` config file (this is the latest build system in Python, it avoids maintaining many files).
I also did a pass on the dependencies, to remove unused ones (hopefully I only removed the ones that were not needed).
Now the ci workflow also relies on extra requirements to install the dependencies, this avoids having to maintain them in two places.

For Pytorch-wavelet, I am not sure why it is installed separately. You don't want to depend on it?

I will continue a pass on simplify maintenance and CI.


deepinv/tests/test_models.py
```diff
@@ -25,31 +25,46 @@ def imsize_1_channel():
...
+            "installed with `pip install "
+            "git+https://github.com/fbcotter/pytorch_wavelets.git`"
+        )
+    if name == "bm3d":
+        pytest.importorskip(
+            "bm3d",
+            reason="This test requires bm3d. It should be "
+            "installed with `pip install bm3d`"
+        )
+    if name == "swinir":
```
matthieutrs, line 60:
```
if name in ["swinir", "scunet"]:
```

matthieutrs (approved):
Thanks a lot for the PR! @tachella see the optional dependencies in toml

deepinv/tests/test_models.py
matthieutrs: also: would be clean to add a test when additional dependencies are not installed (pytorch_wavelets, timm, bm3d...). Would follow a similar pattern as here

PR #99, opened 2023-10-25
TST improve optional deps test+MTN use conftest

tomMoral:
- Use `conftest.py` to define fixtures across test files.
- More consistent tests for optional dependencies
- Add a test to check that the reported error without optional dependencies indicates how to install the dep.
- Add a CI entry to check that the package can run and passes the test without optional deps for `denoisers`.

~I will add a  CI entry to test the package without optional dependencies.~

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,4 +1,4 @@
-name: docs
+name: Build docs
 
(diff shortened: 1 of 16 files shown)


matthieutrs (approved):
Thanks a lot for the PR, it's much cleaner now.

PR #108, opened 2023-11-03
Fix bugs 

samuro95:
- In all the Prox calculations in the DataFidelity and Prior classes, the stepsize gamma parameter is set to 1 by default. So as to compute the prox of a function without needing to specify a stepsize.
- The d function, argument in the DataFidelity class, took only one argument as input and then the distance was computed by sending x-y to d() . This is not desirable because most of the distances don't take x-y as input. For instance the KL divergence. I changed d to take both x and y as inputs (as it was actually written in the docs). Changed the examples and tests accodringly.
- DRUNet would not be used in unfolded because the sigma_denoiser parameter sent is a tensor (with no dimensions). Changed the forward implementation to use sigma.item() in this case.
- Add L1 prior in the prior.py file. 

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,7 @@ Current (v0.0.2)
 
 New Features
 ^^^^^^^^^^^^
+- Added L1 prior, change distance in DataFidelity (:gh:`108` by `Samuel Hurault`_) - 03/11/2023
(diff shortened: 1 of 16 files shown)


matthieutrs (approved):
great PR thanks a lot!

PR #120, opened 2023-12-01
DataParallel compatibility

matthieutrs:
In this PR we propose minor updates:
- allowing`torch.nn.DataParallel` compatibility, implying some device changes at the level of the DRUNet;
- condensating the calls to `wandb.log` and allowing to decrease frequency of image logging for memory.

### DIFF
--- a/deepinv/training_utils.py
+++ b/deepinv/training_utils.py
@@ -82,6 +82,8 @@ def train(
     :param bool plot_measurements: Plot the measurements y. default=True.
     :param bool check_grad: Check the gradient norm at each iteration.
     :param str ckpt_pretrained: path of the pretrained checkpoint. If None, no pretrained checkpoint is loaded.
+    :param list fact_losses: List of factors to multiply the losses. If None, all losses are multiplied by 1.
+    :param int freq_plot: Frequency of plotting images to wandb. If 1, plots at each epoch.
     :returns: Trained model.
     """
     save_path = Path(save_path)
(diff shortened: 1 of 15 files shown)


deepinv/training_utils.py
samuro95, line 1: - no doc for the argument fact_loss

deepinv/training_utils.py
```diff
@@ -183,7 +203,12 @@ def train(
                     )  # In this case the dataloader outputs also a class label
                     x = x.to(device)
                     physics_cur = physics[g]
-                    physics_cur.reset()
+
+                    if isinstance(physics_cur, torch.nn.DataParallel):
+                        physics_cur.module.noise_model.__init__(sigma_max=0.2, x=x)
```
samuro95, line 208: I do not get why this ?

deepinv/training_utils.py
```diff
@@ -254,23 +296,32 @@ def train(
                         "From top to bottom : Input, Backprojection, Output, Target"
                     )
                 else:
-                    imgs = [physics_cur.A_adjoint(y), x_net, x]
+                    if isinstance(physics_cur, torch.nn.DataParallel):
+                        back = physics_cur.module.A_adjoint(y)
```
samuro95, line 300: Can we reset physics as physic.module for once somewhere ?

deepinv/training_utils.py
matthieutrs: In this case, `physics_cur` needs to be re-instanciated because it's not possible to update physics parameters within the forward pass in a `nn.DataParallel` module, they need to be reset instead.

deepinv/training_utils.py
matthieutrs: no, because of the above phenomenon: we need to keep track of whether it is a `nn.DataParallel` module and potentially re-init the module during training, which cannot be done if setting `physics_cur.module` above.

samuro95 (approved):

PR #122, opened 2023-12-04
Fix prior

matthieutrs:
In this PR we fix a bug that was raised when using priors from the `Prior` class, namely Tikhonov and L1Prior.

EDIT: there remains a convergence issue that fails when the regularization parameter is not 1, I still need to fix this.


tachella (approved):
thanks for fixing the bug!

PR #124, opened 2023-12-06
prerelease fixes

tachella:
- safeguarded bm3d and odl imports
- removed A_adjoint requirement for training
- added test only datasets 
- added gaussian blur to docs
- minor fixes in sampling docs
- added L1Prior to docs
- added UniformGaussianNoise to docs

### DIFF
--- a/deepinv/models/bm3d.py
+++ b/deepinv/models/bm3d.py
@@ -5,8 +5,8 @@ import torch.nn as nn
 # Compat for optional dependency on BM3D
 try:
     import bm3d
-except ImportError as e:
-    bm3d = e
+except:
+    bm3d = ImportError
 
 
 class BM3D(nn.Module):
--- a/deepinv/models/bm3d.py
+++ b/deepinv/models/bm3d.py
@@ -6,7 +6,7 @@ import torch.nn as nn
 try:
     import bm3d
 except:
-    bm3d = ImportError
+    bm3d = ImportError("The bm3d package is not installed.")
 
 
 class BM3D(nn.Module):
--- a/deepinv/physics/mri.py
+++ b/deepinv/physics/mri.py
@@ -115,7 +115,7 @@ class MRI(DecomposablePhysics):
             image_size[0] // 2 - num_lines_center // 2,
             image_size[0] // 2 + num_lines_center // 2 + 1,
             steps=50,
-            dtype=torch.int,
+            dtype=torch.long,
         )
         mask[:, center_line_indices] = 1
         random_line_indices = np.random.choice(
--- a/deepinv/training_utils.py
+++ b/deepinv/training_utils.py
@@ -2,11 +2,10 @@ import torchvision.utils
 from deepinv.utils import (
     save_model,
     AverageMeter,
-    ProgressMeter,
     get_timestamp,
     cal_psnr,
 )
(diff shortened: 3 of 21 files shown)


deepinv/models/bm3d.py
```diff
@@ -5,8 +5,8 @@
 # Compat for optional dependency on BM3D
 try:
     import bm3d
-except ImportError as e:
-    bm3d = e
+except:
+    bm3d = ImportError("The bm3d package is not installed.")
```
matthieutrs, line 9: So `except ImportError as e:` is unstable but this way it is stable?

deepinv/physics/mri.py
```diff
@@ -115,7 +115,7 @@ def sample_mask(self, image_size=(320, 320), acceleration_factor=4, seed=None):
             image_size[0] // 2 - num_lines_center // 2,
             image_size[0] // 2 + num_lines_center // 2 + 1,
             steps=50,
-            dtype=torch.int,
+            dtype=torch.long,
```
matthieutrs, line 118: Just curious, why?

deepinv/training_utils.py
```diff
@@ -473,10 +490,10 @@ def test(
                         ):
                             if plot_measurements and len(y.shape) == 4:
                                 imgs = [y, x_init, x1, x]
-                                name_imgs = ["Input", "Linear", "Recons.", "GT"]
+                                name_imgs = ["Input", "No learning", "Recons.", "GT"]
```
matthieutrs, line 493: Why not stick to "Linear"? "No learning" looks confusing to me because it may seem like the output of the network at initialization or some method that has frozen parameters

docs/source/index.rst
```diff
@@ -16,7 +16,7 @@ Featuring
 * |:boomerang:| Many :ref:`pretrained deep denoisers <models>` which can be used for :ref:`plug-and-play restoration <pnp>`.
 * |:book:| Framework for :ref:`building datasets <datasets>` for inverse problems.
 * |:building_construction:| Easy-to-build :ref:`unfolded architectures <unfolded>` (ADMM, forward-backward, deep equilibrium, etc.).
-* |:microscope:| :ref:`Sampling algorithms <sampling>` for uncertainty quantification (Langevin, diffusion, etc.).
+* |:microscope:| :ref:`Diffusion algorithms <sampling>` for uncertainty quantification (Langevin, diffusion, etc.).
```
matthieutrs, line 19:
```suggestion
* |:microscope:| :ref:`Diffusion algorithms <sampling>` for image restoration and uncertainty quantification (Langevin, diffusion, etc.).
```

matthieutrs (commented):
Thanks a lot! I have few minor questions, see above

deepinv/models/bm3d.py
tachella: it is to avoid the case where it fails during the import, but raises an exception that is not necessary an ImportError (it was actually happening to me)

deepinv/physics/mri.py
tachella: it was giving errors when running the pytest on a gpu

deepinv/training_utils.py
tachella: because for non-linear operators (e.g. blind deblurring), the reconstruction is not necessarily linear...

PR #137, opened 2023-12-21
Small improvements to unet.UNet, shift.Shift & noise.PoissonNoise

Andrewwango:
- UNet: option to select nn.BatchNorm2d as batch norm instead of BFBatchNorm2d
- Shift: add option for max percentage shift
- PoissonNoise: clip input to non-negative

### DIFF
--- a/deepinv/models/unet.py
+++ b/deepinv/models/unet.py
@@ -60,6 +60,10 @@ class UNet(nn.Module):
     can be controlled with the `scales` parameter. The number of trainable parameters increases with the number of
     scales.
 
+    .. warning::
+        When using the bias-free batch norm `BFBatchNorm2d` via `batch_norm=biasfree`, NaNs may be encountered
+        during training, causing the whole training procedure to fail.  
+
     :param int in_channels: input image channels
     :param int out_channels: output image channels
     :param bool residual: use a skip-connection between output and output.
@@ -67,7 +71,7 @@ class UNet(nn.Module):
     :param bool cat: use skip-connections between intermediate levels.
     :param bool bias: use learnable biases.
     :param bool|str batch_norm: if False, no batchnorm applied, if True, use `nn.BatchNorm2d`, if `biasfree`, use
-        `BFBatchNorm2d`.
+        `BFBatchNorm2d` from `"Robust And Interpretable Blind Image Denoising Via Bias-Free Convolutional Neural Networks" by Mohan et al. <https://arxiv.org/abs/1906.05478>`_.
     :param int scales: Number of downsampling steps used in the U-Net. The options are 2,3,4 and 5.
         The number of trainable parameters increases with the scale.
     """
--- a/deepinv/models/unet.py
+++ b/deepinv/models/unet.py
@@ -66,7 +66,7 @@ class UNet(nn.Module):
     :param bool circular_padding: circular padding for the convolutional layers.
     :param bool cat: use skip-connections between intermediate levels.
     :param bool bias: use learnable biases.
-    :param bool|str batch_norm: if False, no batchnorm applied, if True, use `BFBatchNorm2d`, if `noiseless`, use 
+    :param bool|str batch_norm: if False, no batchnorm applied, if True, use `BFBatchNorm2d`, if `noiseless`, use
         `nn.BatchNorm2d`.
     :param int scales: Number of downsampling steps used in the U-Net. The options are 2,3,4 and 5.
         The number of trainable parameters increases with the scale.
@@ -93,9 +93,9 @@ class UNet(nn.Module):
         self.cat = cat
         self.compact = scales
         self.Maxpool = nn.MaxPool2d(kernel_size=2, stride=2)
-        
+
         noiseless = batch_norm != "noiseless"
-        
+
         def conv_block(ch_in, ch_out):
             if batch_norm:
                 return nn.Sequential(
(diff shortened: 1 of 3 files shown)


deepinv/models/unet.py
```diff
@@ -92,6 +94,8 @@ def __init__(
         self.compact = scales
         self.Maxpool = nn.MaxPool2d(kernel_size=2, stride=2)
 
+        noiseless = batch_norm != "noiseless"
```
matthieutrs, line 97: This is a bit confusing as it kind of reads `noiseless != "noiseless"`.

Maybe a solution would be to change the `batch_norm` variable for `batch_norm_type` which could be either `None`/`False`, `biasfree` or `True`?

deepinv/physics/noise.py
```diff
@@ -114,7 +114,7 @@ def forward(self, x):
         :param torch.Tensor x: measurements
         :returns: noisy measurements
         """
-        y = torch.poisson(x / self.gain)
+        y = torch.poisson(torch.clip(x / self.gain, min=0.0))
```
matthieutrs, line 117: Not sure I agree with this: the noise should not impact the values of the measurement; what prevents one to add poisson noise to negative measurements?

Maybe with your usecase you could add this in the measurement operator?

deepinv/transform/shift.py
```diff
@@ -8,11 +8,13 @@ class Shift(torch.nn.Module):
     Generates n_transf randomly shifted versions of 2D images with circular padding.
 
     :param n_trans: number of shifted versions generated per input image.
+    :param shift_max: maximum shift as fraction of total height/width
```
matthieutrs, line 11:
```suggestion
    :param float shift_max: maximum shift as fraction of total height/width.
```

matthieutrs (changes requested):
Hi, thanks a lot for the suggestions! See my comments above :)

deepinv/physics/noise.py
Andrewwango:
```python
>>> import torch
>>> torch.poisson(torch.tensor([-1.]))

---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[5], [line 2](vscode-notebook-cell:?execution_count=5&line=2)
      [1](vscode-notebook-cell:?execution_count=5&line=1) import torch
----> [2](vscode-notebook-cell:?execution_count=5&line=2) torch.poisson(torch.tensor([-1.]))

RuntimeError: invalid Poisson rate, expected rate to be non-negative
```

The other option is to only apply noise to a mask of non-negative values?

deepinv/models/unet.py
Andrewwango: That is a good idea, although it would result in breaking changes for those who already have `batch_norm=True` now won't be bias free?

matthieutrs: Yes but this is not an issue since when loading a model trained in with a different BN than the one used for training it should raise an error.

deepinv/physics/noise.py
matthieutrs: My bad I misunderstood the torch notation, the input should indeed be positive. Let's wait for @tachella's input :)

deepinv/models/unet.py
Andrewwango: See [440e06b](https://github.com/deepinv/deepinv/pull/137/commits/440e06b0d4c6cd802ade7ad8d6ceecb5fc46b3b1)

deepinv/physics/noise.py
tachella: I agree that poisson inputs should be positive but adding the clip might hide bugs - I think we should leave it to the user to pass positive rates

deepinv/transform/shift.py
```diff
@@ -8,11 +8,13 @@ class Shift(torch.nn.Module):
     Generates n_transf randomly shifted versions of 2D images with circular padding.
 
     :param n_trans: number of shifted versions generated per input image.
+    :param float shift_max: maximum shift as fraction of total height/width.
     """
 
-    def __init__(self, n_trans=1):
+    def __init__(self, n_trans=1, shift_max=1):
```
tachella, line 14: a pedantic comment, but it would be good to set it as 1. to stress the float type

deepinv/models/unet.py
```diff
@@ -104,12 +108,16 @@ def conv_block(ch_in, ch_out):
...
-                    BFBatchNorm2d(ch_out, use_bias=bias),
+                    BFBatchNorm2d(ch_out, use_bias=bias)
+                    if biasfree
+                    else nn.BatchNorm2d(ch_out),
                     nn.ReLU(inplace=True),
                     nn.Conv2d(
                         ch_out, ch_out, kernel_size=3, stride=1, padding=1, bias=bias
                     ),
-                    BFBatchNorm2d(ch_out, use_bias=bias),
+                    BFBatchNorm2d(ch_out, use_bias=bias)
```
tachella, line 118: I had removed the BF batch norm from Mohan et al since it often produces NaNs during training. Unless you found a solution to this, I wouldn't add this modle

tachella (changes requested):
Sorry for the very late review!

The PR looks good, I just worry that the BFBatchNorm2d module from Mohan et al sometimes produces NaNs during training which causes the whole training procedure to fail. 

@Andrewwango did you observe this as well? Maybe there is a solution

deepinv/models/unet.py
Andrewwango: @tachella This PR means UNet defaults to nn.BatchNorm2d but if the user really wants, they can pass `batch_norm=biasfree` to use BF batch norm to maintain compatibility.

The BF batch norm did give me problems with NaNs during training, yes, btw.

deepinv/physics/noise.py
Andrewwango: How about a new parameter `PoissonNoise(clip_positive=False)`? Then users can easily choose this if they want to save them reimplementing PoissonNoise themselves. I see this as a common use-case when your NN might have negative hidden values (e.g. when using LeakyReLU) and the Poisson Noise applies the physics of clipping it and adding noise.

deepinv/transform/shift.py
Andrewwango: Done

deepinv/physics/noise.py
tachella: ok, good idea, let's go with that optional parameter!

deepinv/models/unet.py
tachella: ok, I agree. 

Could you please add a link to the arxiv paper when mentioning the biasfree batchnorm option in the params (and maybe a warning about the nans using [sphinx's warning](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html))?

Andrewwango: Done, thanks @tachella

tachella: thanks!

PR #141, opened 2023-12-23
CG for complex vectors

matthieutrs:
Very minor modifications to support complex data in conjugate gradient.

### DIFF
--- a/deepinv/optim/utils.py
+++ b/deepinv/optim/utils.py
@@ -1,3 +1,4 @@
+import torch
 from deepinv.utils import zeros_like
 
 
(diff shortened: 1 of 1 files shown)


matthieutrs:
Moving from `(s1 * s2).flatten().sum()` to `torch.vdot(s1.flatten(), s2.flatten())` allows to handle automatically the dot product for complex variables, but turns out to be problematic in one of the physics examples where `s1` and `s2` don't have the same dimension (`(1,3,w,h)` vs `(1,1,w,h)` which is bypassed by the `*` operator), how should we solve?

matthieutrs:
Addressed in #165

PR #147, opened 2024-01-29
`training_utils.train()` -> `training_utils.Trainer`

Andrewwango:
It would be nice to have a Trainer class to run deep inverse experiments, rather than just a train() function, and have modular methods inside for each train step. This is so that for custom experiments (e.g. with new datasets or physics that require different processing) one can simply override the appropriate train step instead of reimplementing the whole training/visualisation pipeline.

One example use-case is that I am currently training a GAN and have overridden `Trainer.backward_pass` to allow simultaneous updating of 2 optimisers, while reusing all the same training code from deepinv.

In this PR I have
- Moved all the training logic such that `train(...)` is now `Trainer(...).train()`
- Separated the training code into modules such as `epoch_eval()` and `backward_pass()`
- Left `training_utils.test()` as is
- Added `img_interval` as argument for controlling frequency of plotting "Test images" 

Note that `Trainer` uses `@dataclass` to avoid a very lengthy constructor.

Happy to rename methods or reshuffle. Thanks!


tachella:
Hi Andrew,

The idea looks quite interesting - the train function is indeed getting out of control in terms of complexity. 

Could you please add a small usage example in the comments to illustrate the new usage?
Thanks!

deepinv/training_utils.py
```diff
@@ -316,37 +222,159 @@ def train(
...
+                    y = next(self.iterators[g])
+                    x = None
+                else:
+                    x, y = next(self.iterators[g])
+                    if type(x) is list or type(x) is tuple:
+                        x = [s.to(self.device) for s in x]
+                    else:
+                        x = x.to(self.device)
+
+                physics_cur = self.physics[g]
```
matthieutrs, line 321: Would putting this in a method clarify things?

deepinv/training_utils.py
```diff
@@ -316,37 +222,159 @@ def train(
...
-            optimizer,
-            ckp_interval,
-            epochs,
-            loss_history,
-            str(save_path),
-            eval_psnr=eval_psnr if perform_eval else None,
-        )
+    def batch_eval(self, x, x_net):
+        # training psnr and logging
+        if not self.unsupervised:
```
matthieutrs, line 237: Maybe `assert` instead of `if`

deepinv/training_utils.py
```diff
@@ -316,37 +222,159 @@ def train(
...
-            epoch,
-            model,
-            optimizer,
-            ckp_interval,
-            epochs,
-            loss_history,
-            str(save_path),
-            eval_psnr=eval_psnr if perform_eval else None,
-        )
+    def batch_eval(self, x, x_net):
```
matthieutrs, line 235: not sure if `batch_eval` as such is the best name (for me it would refer to doing forward in eval mode). What about `batch_eval_metric` instead?

Andrewwango:
Thanks @tachella ! `Trainer` can be used in the same way as the original function:
```python
trainer = Trainer(
    model=model,
    epochs=epochs,
    losses=losses,
    eval_dataloader=test_dataloader,
    scheduler=scheduler,
    physics=physics,
    optimizer=optimizer,
    device=device,
    D=D,
)

trainer.train()
```

This allows us to for example write custom Trainers without reimplementing all training logic. For example to implement GAN training only requires the following:

```python
class GANTrainer(Trainer):
    """Overrides Trainer with custom backward pass.
    """
    optimizer: GANOptimizer = None # Custom optimizer that wraps optimizers for both G and D
    D: Module = None # Discriminator network

    def backward_pass(self, g, x, y, x_net):
        """Overrides Trainer.backward_pass
        """
        # Train G
        loss_g = self.losses[0](x=x, x_net=x_net, y=y, physics=self.physics[g], model=self.model, D=self.D)
    
        loss_g.backward(retain_graph=True)
        
        self.check_clip_grad()

        self.optimizer.G.step()

        # Train D
        self.optimizer.D.zero_grad()

        loss_d = self.losses[1](x=x, x_net=x_net, y=y, physics=self.physics[g], model=self.model, D=self.D)

        loss_d.backward()

        self.optimizer.D.step()

if __name__ == "__main__":
    trainer = GANTrainer(
        model=model,
        epochs=epochs,
        losses=losses,
        eval_dataloader=test_dataloader,
        scheduler=scheduler,
        physics=physics,
        optimizer=optimizer,
        device=device,
        D=D,
    )

    trainer.train()
```

I realise there's a lot of docstrings/typing etc. to do, but wanted to see what you guys thought before committing to it all!

tachella:
Thanks for the demo!
I really like the idea tbh  - it provides a much cleaner way of editing the training without copy-pasting all of the code.

Would you be okay a minimally working example (ie passing tests)? 
I know we would have to update many example files and some tests, but we are happy to give you a hand with those if necessary.

Andrewwango:
@eliottbourrigan @matthieutrs @tachella any progress on this PR?

deepinv/training_utils.py
```diff
@@ -487,15 +485,9 @@ def validate(self, epoch):
...
-
-                    x, y, physics_cur = (
-                        samples["x"],
-                        samples["y"],
-                        samples["physics_cur"],
-                    )
+                    x, y, physics_cur = self.get_samples(self.val_iterators, g)
 
-                    x_net = self.model_inference(samples)
+                    x_net = self.model_inference(y=y, physics_cur=physics_cur)
```
Andrewwango, line 490: I like this. However, if the user wants to add another variable (e.g. `z`, or `labels`) to samples, they have to override `train_step`, `get_samples` and `model_inference`? I'm wondering if we could have a kwargs approach to not have to override `train_step`, but I don't want to go back to the dict approach, maybe like:
```
x, y, physics, kwargs = self.get_samples(...)
x_net = self.model_inference(y=y, physics_cur=physics, **kwargs)
``` 
... but this is quite ugly, so happy to have the user have to override `train_step` too if they need this functionality.

deepinv/training_utils.py
```diff
@@ -529,6 +736,22 @@ def test(
     return test_psnr, test_std_psnr, linear_psnr, linear_std_psnr
 
 
+def train(*args, **kwargs):
+    """
+    WARNING: DEPRECATED.
```
tachella, line 741: we can use

.. warning: 
   
    This usage is deprecated and might be removed in future versions of deepinv.

deepinv/training_utils.py
```diff
@@ -68,285 +46,507 @@ def train(
...
+                str(self.save_path),
+                eval_psnr=self.eval_psnr if self.perform_eval else None,
+            )
+
+        if self.wandb_vis:
+            wandb.save("model.h5")
+
+        return self.model
+
+    def validate(self, epoch):
```
Andrewwango, line 465: Rename to e.g. `validation_step`? Currently `validate` sounds like it simply performs data validation like [`sklearn.utils.check_X_y`](https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_X_y.html)

deepinv/training_utils.py
```diff
@@ -68,285 +46,507 @@ def train(
...
+
+        :param torch.Tensor x: Ground truth.
+        :param torch.Tensor x_net: Network reconstruction.
+        :param bool train: Whether the model is being trained. (Necessary to attribute to relevant log key)
+        :param bool log: Whether to log the metrics. (Only if train is True)
+        """
+
+        assert (
+            not self.unsupervised
+        ), "batch_eval_metric should not be called when self.unsupervised is True."
```
Andrewwango, line 229: `batch_metric `or `batch_eval_metric`?

matthieutrs:
Thanks for all your feedbacks! 

The current implementation allows to do something like this to override the Trainer class (for example if you have a denoiser instead of a backbone image restoration algorithm). 
To do so without re-writing the training loop (and hence also the validation loop) I've introduced a `sef.forward_pass` method. What do you think?

```python
from deepinv.training_utils import Trainer

class MyTrainer(Trainer):
    def __init__(self, *args, **kwargs):
        super(MyTrainer, self).__init__(*args, **kwargs)

    def get_samples_online(self, iterators, g):
        r"""
        Get the samples for the online measurements.

        In this setting, a new sample is generated at each iteration by calling the physics operator.
        This function returns a dictionary containing necessary data for the model inference. It needs to contain
        the measurement, the ground truth, and the current physics operator, but can also contain additional data.

        :param list iterators: List of dataloader iterators.
        :param int g: Current dataloader index.
        :returns: a dictionary containing at least: the ground truth, the measurement, and the current physics operator.
        """
        x, _ = next(
            iterators[g]
        )  # In this case the dataloader outputs also a class label
        x = x.to(self.device)
        physics_cur = self.physics[g]

        y, sigma = physics_cur(x)

        return x, y, physics_cur, sigma

    def model_inference(self, y, physics_cur, sigma):
        r"""
        Perform the model inference.

        It returns the network reconstruction given the samples.

        :param torch.Tensor y: Measurement.
        :param deepinv.physics.Physics physics_cur: Current physics operator.
        :returns: The network reconstruction.
        """
        y = y.to(self.device)
        x_net = self.model(y, sigma)
        return x_net


    def forward_pass(self, iterator, g):
        r"""
        Perform the forward pass.

        It returns the ground truth, the measurement, the current physics operator, and the network reconstruction.
        """
        x, y, physics_cur, sigma = self.get_samples(iterator, g)

        y = y.to(self.device)

        # Run the forward model
        x_net = self.model_inference(y=y, physics_cur=physics_cur, sigma=sigma)

        return x, y, physics_cur, x_net
        
```

Andrewwango:
> Thanks for all your feedbacks!
> 
> The current implementation allows to do something like this to override the Trainer class (for example if you have a denoiser instead of a backbone image restoration algorithm). To do so without re-writing the training loop (and hence also the validation loop) I've introduced a `sef.forward_pass` method. What do you think?
> 
> ```python
> from deepinv.training_utils import Trainer
> 
> class MyTrainer(Trainer):
>     def __init__(self, *args, **kwargs):
>         super(MyTrainer, self).__init__(*args, **kwargs)
> 
>     def get_samples_online(self, iterators, g):
>         r"""
>         Get the samples for the online measurements.
> 
>         In this setting, a new sample is generated at each iteration by calling the physics operator.
>         This function returns a dictionary containing necessary data for the model inference. It needs to contain
>         the measurement, the ground truth, and the current physics operator, but can also contain additional data.
> 
>         :param list iterators: List of dataloader iterators.
>         :param int g: Current dataloader index.
>         :returns: a dictionary containing at least: the ground truth, the measurement, and the current physics operator.
>         """
>         x, _ = next(
>             iterators[g]
>         )  # In this case the dataloader outputs also a class label
>         x = x.to(self.device)
>         physics_cur = self.physics[g]
> 
>         y, sigma = physics_cur(x)
> 
>         return x, y, physics_cur, sigma
> 
>     def model_inference(self, y, physics_cur, sigma):
>         r"""
>         Perform the model inference.
> 
>         It returns the network reconstruction given the samples.
> 
>         :param torch.Tensor y: Measurement.
>         :param deepinv.physics.Physics physics_cur: Current physics operator.
>         :returns: The network reconstruction.
>         """
>         y = y.to(self.device)
>         x_net = self.model(y, sigma)
>         return x_net
> 
> 
>     def forward_pass(self, iterator, g):
>         r"""
>         Perform the forward pass.
> 
>         It returns the ground truth, the measurement, the current physics operator, and the network reconstruction.
>         """
>         x, y, physics_cur, sigma = self.get_samples(iterator, g)
> 
>         y = y.to(self.device)
> 
>         # Run the forward model
>         x_net = self.model_inference(y=y, physics_cur=physics_cur, sigma=sigma)
> 
>         return x, y, physics_cur, x_net
>         
> ```

@matthieutrs I like it, I think we'll need a few notebook examples of different ways to implement custom Trainers as it could seem daunting for novices, e.g. training a denoiser, a GAN, custom eval metric... happy to help

tachella (changes requested):
great PR!
I will do some extra iterations soon

deepinv/training_utils.py
```diff
@@ -6,43 +7,20 @@
...
-    wandb_vis=False,
-    wandb_setup={},
-    online_measurements=False,
-    plot_measurements=True,
-    check_grad=False,
-    ckpt_pretrained=None,
-    fact_losses=None,
-    freq_plot=1,
-):
+@dataclass
```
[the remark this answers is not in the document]
Andrewwango: `dataclass` automatically initialises all the class attributes so the code doesn't look as bulky. Completely optional though.

deepinv/training_utils.py
```diff
@@ -6,43 +7,20 @@
...
+class Trainer:
     r"""
-    Trains a reconstruction network.
-
+    Trainer class for training a reconstruction network.
```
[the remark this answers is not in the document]
Andrewwango: To highlight the OOP nature of it:
> ...such that the user can only modify a few methods without redoing the whole training code, 

... by inheriting from Trainer and overriding only the methods you want to modify.

Also I think it's worth mentioning that it's super flexible and can be reused for many inverse problems and deep learning methodologies!

tachella:
This PR has been moved to Last Trainer #181.

PR #150, opened 2024-01-30
CLN avoid border effect from importing deepinv

tomMoral:
Right now, simply importing `deepinv` mess up with the `matplotlib` config, and can result in `matplotlib` not working anymore (independently from `deepinv`) with incomplete latex installation.
As a rule of thumb, it is usually better to not change global library config on import.

I propose a small helper that sets up the `matplotlib` config with `deepinv` settings, only when needed. (hopefully I watched all the places where it was needed but this might need to be checked).


matthieutrs:
Hi Thomas, thanks a lot for spotting this and sorry for the late reply! I think that this fails because you are running with `black==23.x`. We recently upgraded to `24.x`, so if you upgrade black this should fix it :) For the codecov this is fine

PR #154, opened 2024-02-08
Fix decomposable physics noise

### LINKED ISSUE #153: DecomposablePhysics forward
When solving #136, the way noise was handled was not correct for all instances of the `DecomposablePhysics` class, in particular BlurFFT.

[This](https://github.com/deepinv/deepinv/blob/main/deepinv/physics/forward.py#L630):
```python
 noise = self.U(self.V_adjoint(self.V(self.U_adjoint(self.noise_model(x)) * self.mask)))
```
only works for U = V = identity.

In particular this issue impacts the BlurFFT class.

matthieutrs:
Fixes #153. We remove the application of noise by default from the `DecomposablePhysics` class and move it to the appropriate classes instead (for now, only `Inpainting`).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ New Features
 
 Fixed
 ^^^^^
+- Fixed issue in noise forward of Decomposable class (:gh:`154` by `Matthieu Terris`_) - 08/02/2024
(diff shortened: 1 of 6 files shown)


tachella (approved):
thanks for fixing this bug. it looks good to me!

PR #156, opened 2024-02-12
Add a new Prior TVPrior + minor fix on index type

Nils-Laurent:
This PR implements the TVPrior as a Prior class. The proximal operator of TV is inherited from the TV denoiser.

Minor side changes: 
1. introduces a [`html-pattern`](https://github.com/deepinv/deepinv/blob/b069182024dc3129d5f39ed11b669d1b127a67aa/docs/Makefile#L28) function to build docs of a single example at a time. The resulting function can be used as 
```bash
PATTERN=../examples/path/to/example_file.py make html-pattern
```
2. use type "tensor.long" for indices in transform/scale, adds the l21 norm.


CHANGELOG.rst
```diff
@@ -10,6 +10,7 @@ New Features
 ^^^^^^^^^^^^
 - Added r2r loss (:gh:`148` by `Brayan Monroy`_) - 30/01/2024
 - Added scale transform (:gh:`135` by `Jérémy Scanvic`_) - 19/12/2023
+- Added priors for total variation and l21 mixed norm - 09/02/2023
```
tachella, line 13: please add your name Nils

deepinv/models/tgv.py
```diff
@@ -163,70 +168,60 @@ def forward(self, y, ths=None):
...
-
-
-# # ADJOINTNESS TEST
-# u = torch.randn((2, 3,100,100,2)).type(torch.DoubleTensor)
-# Au = epsilon(u)
-# v = torch.randn(*Au.shape).type(Au.dtype)
-# Atv = epsilonT(v)
-# e = v.flatten()@Au.flatten()-Atv.flatten()@u.flatten()
-# print('Adjointness test (should be small): ', e)
+    def nabla(self, x):
```
tachella, line 171: nabla seems to be already defined in TV, I would avoid duplicating code (just import it)

deepinv/models/tgv.py
```diff
@@ -163,70 +168,60 @@ def forward(self, y, ths=None):
...
+        """
+        b, c, h, w = x.shape
+        u = torch.zeros((b, c, h, w, 2), device=x.device).type(x.dtype)
+        u[:, :, :-1, :, 0] = u[:, :, :-1, :, 0] - x[:, :, :-1]
+        u[:, :, :-1, :, 0] = u[:, :, :-1, :, 0] + x[:, :, 1:]
+        u[:, :, :, :-1, 1] = u[:, :, :, :-1, 1] - x[..., :-1]
+        u[:, :, :, :-1, 1] = u[:, :, :, :-1, 1] + x[..., 1:]
+        return u
+
+    def nabla_adjoint(self, x):
```
tachella, line 183: same for as nabla

deepinv/models/tv.py
```diff
@@ -2,7 +2,7 @@
 import torch.nn as nn
 
 
-class TV(nn.Module):
+class proxTV(nn.Module):
```
tachella, line 5: I would call it TVDenoiser instead of proxTV

deepinv/models/tgv.py
```diff
@@ -3,7 +3,7 @@
 import torch.nn as nn
 
 
-class TGV(nn.Module):
+class proxTGV(nn.Module):
```
tachella, line 6: I would call it TGVDenoiser

examples/sampling/demo_diffpir.py
```diff
@@ -25,7 +25,6 @@
 # Generate an inverse problem
 # ---------------------------
 #
-
```
tachella, line 28: why is this?

tachella (changes requested):
Very nice PR! I think everything looks great overall. Many thanks for adding a simple way of generating a single example!

I would recommend the following changes:

- rename proxTV to TVDenoiser, as it would be a simpler name for users interested only in denoising. Do the same for proxTGV -> TGVDenoiser
- Add a ... warning:: in the TVDenoiser docstrings saying that TVPrior is recommended for plug-and-play tasks.
- remove the duplicate nabla and nabla_adjoint in tv.py and tgv.py (just keep the tv one and import it from tgv).
- add the instructions for using make html-pattern in the contributing page of the docs.

examples/sampling/demo_diffpir.py
matthieutrs: Because the note was not showing in the docs, see https://deepinv.github.io/deepinv/auto_examples/sampling/demo_diffpir.html#generate-an-inverse-problem

deepinv/transform/scale.py
```diff
@@ -9,7 +9,7 @@ def sample_from(values, shape=(1,), dtype=torch.float32, device="cpu"):
     values = torch.tensor(values, device=device, dtype=dtype)
     N = torch.tensor(len(values), device=device, dtype=dtype)
     indices = torch.floor(N * torch.rand(shape, device=device, dtype=dtype)).to(
-        torch.int
+        torch.long
```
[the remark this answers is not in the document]
Nils-Laurent: I changed the type for the variables because torch gave me an error when running `pytest`. But I think that was because I used an older version of `torch` (I had also errors on other tests), when I performed the update, I was able to run the validation. I included this in the PR to be robust w.r.t. different versions of `torch`.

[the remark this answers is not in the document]
tachella: I think it is fine

Issue #113, opened 2023-11-09
normalization of l2 loss


matthieutrs:
The normalization factor seems to be missing in the cost function here https://github.com/deepinv/deepinv/blob/main/deepinv/optim/data_fidelity.py#L259; it seems to be used correctly in other methods however.

PR #158, opened 2024-02-14
fix normalization l2

### LINKED ISSUE #157: small bug in optim.L2.forward()
Hi guys,

It seems there's a small bug in the `L2.forward()`, as the `sigma` (or `L2.norm` as it is stored in the `L2` class) of the Gaussian likelihood is not taken into account in the forward function or the `L2.d()`. 

It could be fixed by, for example, replacing [this line](https://github.com/deepinv/deepinv/blob/08944ac02c89fcc3ca229a2565bdcdf19da517c7/deepinv/optim/data_fidelity.py#L260):
`return d`
by:
`return d * self.norm`

matthieutrs:
Fixes #157 (= #113)


tachella (approved):
thanks for fixing this, I wonder why we missed it there.

PR #164, opened 2024-02-28
Add Patch priors (EPLL and PatchNR)

johertrich:
I implemented two patch priors, namely expected patch log likelihood (EPLL) and patch normalizing flow regularizers (PatchNR). They require a few other additions.

In detail, I added:
- a model representing EPLL including a reconstruction method using the (approximated) half-quadratic splitting as proposed by Zoran and Weiss.
- a model representing PatchNR
- an example for using EPLL for denoising and inpainting
- an example for using EPLL and PatchNR for low-dose limited-angle CT
- Log-Poisson noise and a corresponding likelihood term used in low-dose CT

Moreover, I added the following, because they are required by the above models:
- a class for Gaussian mixture models including a method for parameter estimation via the EM algorithm.
- a class for represent the dataset of all patches given a tensor of images
- a helper function which selects patches from a tensor of images either randomly or by index
- a PatchPrior class, which takes a prior on the space of patches and creates a prior on the space of images
- a training method for normalizing flows


deepinv/training_utils.py
```diff
@@ -527,3 +527,52 @@ def test(
         wandb.log({"Test PSNR": test_psnr}, step=step)
 
     return test_psnr, test_std_psnr, linear_psnr, linear_std_psnr
+
+
+def train_normalizing_flow(
```
matthieutrs, line 532: I would leave it here for now, it may be the most convenient. However we are heavily refactoring the training class atm and we would probably want to make this training function compatible with the (upcoming) Trainer. For now I would not change what you've done if this is convenient. Only thing is I would not load it by default in the global `deepinv/__init__.py` file since this is going to change.

deepinv/__init__.py
```diff
@@ -55,7 +55,7 @@
 
 __all__ += ["unfolded"]
 
-from deepinv.training_utils import train, test
+from deepinv.training_utils import train, test, train_normalizing_flow
```
matthieutrs, line 58:
```suggestion
from deepinv.training_utils import train, test

```

I would remove for now because we are currently doing a big refactoring of the `training_utils.py` file (see my comment over there)

deepinv/utils/demo.py
```diff
@@ -169,3 +169,16 @@ def load_url_image(
     transform = transforms.Compose(transform_list)
     x = transform(img).unsqueeze(0).to(device)
     return x
+
+
+def load_torch_url(url):
```
matthieutrs, line 174: Comment: maybe we could unify our loading from url, to do some day.

matthieutrs (changes requested):
Hi Johannes, thanks for the great PR! It's really cool to now have patch methods! I've done a first batch of minor comments regarding the organization. Among general questions/comments I have:
- for now let's keep your training function as is, it's the simplest, but we will refactor that some day once #147 is merged (don't try to match it as it is still WIP and subject to change)
- maybe move the GMM inside EPLL because it is not a restoration model (as other models are) but a class used specificaly for EPLL
- for the CI to run with an optional dependency (FrEIA), you'll need to update the .toml file [here](https://github.com/deepinv/deepinv/blob/main/pyproject.toml#L66): just like we added a `ptwt` line, there should be a `FrEIA` line (the thing that will be running after `pip install ...`). 
- lastly we will need to add some tests in `deepinv/tests` to check that (i) the method are running for all types of tensors and (ii) that they behave correctly on very small toy examples.

I can definitely help with that later this week as well as putting things on HF.

deepinv/datasets/patch_dataset.py
```diff
@@ -0,0 +1,34 @@
+from torch.utils import data
+import torchvision.transforms.v2
+
+class PatchDataset(data.Dataset):
+    r"""
+    Builds the dataset of all patches from a tensor of images.
+
+    :param torch.Tensor imgs: Tensor of images, size: batch size x channels x height x width
+    :param int patch_size: size of patches
+    :param callable or str transforms: data augmentation. None for no augmentation, 'flip and rotate' for including random flips and rotations or callable object.
```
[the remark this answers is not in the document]
johertrich: I removed this predefined transform entirely and set the default value to None (= no data augmentation). Even though it can give some gain for training a patch normalizing flow, it is not used in the examples and particularly for estimating a GMM one must not use data augmentation (since one has to pass the whole dataset in a single iteration of the EM algorihtm).

deepinv/models/epll.py
```diff
@@ -0,0 +1,162 @@
...
+                    assert channels == 1
+                    url = "https://drive.google.com/uc?export=download&id=17d40IPycCf8Cb5RmOcrlPTq_AniBlYcK"
+                else:
+                    raise ValueError("Pretrained weights not found!")
+                weights = load_torch_url(url)
+            self.GMM.load_parameter_dict(weights)
+
+            self.GMM.load_parameter_dict(weights)
+
+    def forward(self, x):
```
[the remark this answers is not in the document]
johertrich: I renamed the previous forward method in negative_log_likelihood and added a new forward method which calls self.reconstruction with a identity forward operator (a denoiser should not require a physics argument) and the noisy observation as initialization.

Moreover, I changed the argument from the squared noise level to the noise level to be consistent with the other denoisers...

deepinv/models/gmm.py
```diff
@@ -0,0 +1,237 @@
+import torch
+from tqdm import tqdm
+
+
+class GaussianMixtureModel:
```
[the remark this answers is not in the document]
johertrich: I have done that. However, for me the negative_log_likelihood appeared to be the more natural choice as forward.

deepinv/models/epll.py
```diff
@@ -0,0 +1,457 @@
...
+    r"""
+    Patch prior via Gaussian mixture models.
+
+    The forward method evaluates the negative log likelihood of the GMM.
+    The reconstruction function implements the approximated half-quadratic splitting method as in Zoran, D., and Weiss, Y.  "From learning models of natural image patches to whole image restoration." (ICCV 2011).
+
+    :param deepinv.models.GaussianMixtureModel GMM: Gaussian mixture defining the distribution on the patch space.
+        None creates a GMM with n_components components of dimension accordingly to the arguments patch_size and channels.
+    :param int n_components: number of components of the generated GMM if GMM is None.
+    :param str pretrained: Path to pretrained weights of the GMM with file ending .pt. None for no pretrained weights, "download" for pretrained weights on the BSDS500 dataset, "GMM_lodopab_small" for the weights from the limited-angle CT example.
```
matthieutrs, line 20: If you can add a pointer to `deepinv.models.rst` that specifies the refs for the pretrained models, that would be amazing!

deepinv/models/epll.py
```diff
@@ -0,0 +1,457 @@
...
+        :param float stopping_criterion: stop when objective decrease is smaller than this number.
+            None for performing exactly max_iters iterations
+        :param bool data_init: True for initialize mu by the first data points, False for using current values as initialization
+        :param bool verbose: Output progress information in the console
+        """
+        if data_init:
+            first_data = next(iter(dataloader))[0][: self.n_components].to(self.mu)
+            if first_data.shape[0] == self.n_components:
+                # self.mu = first_data
+                self.mu.copy_(first_data)
```
matthieutrs, line 388: I've replaced line 387 with the following, is this ok? (it raised a bug in the demo file)

docs/source/deepinv.models.rst
```diff
@@ -134,4 +155,5 @@ associated reference and relevant details. All pretrained weights are hosted on
    * -
      - from `Dhariwal and Nichol <https://arxiv.org/abs/2105.05233>`_ trained on ImageNet128 (256 hidden channels per layer).
        `weights <https://huggingface.co/deepinv/diffunet/resolve/main/diffusion_openai.pt?download=true>`_.
-
+   * - :meth:`deepinv.models.EPLL`
+     - Default: parameters estimated with deepinv on 50 mio patches from the training/validation images from BSDS500 for grayscale and color images.
```
matthieutrs, line 159: Can you please add refs for each pretrained model available? (if refs exist?)

deepinv/training_utils.py
```diff
@@ -527,3 +527,52 @@ def test(
...
+    epochs=10,
+    learning_rate=1e-3,
+    device="cpu",
+    jittering=1 / 255.0,
+    verbose=False,
+):
+    """
+    Trains a normalizing flow.
+
+    Uses the forward KL (maximum likelihood) loss function and the Adam optimizer.
```
matthieutrs, line 544: Maybe add 1 sentence and 1 equation stating a bit more explicitely the training objective?

matthieutrs (changes requested):

matthieutrs (approved):

PR #170, opened 2024-03-07
Lambda optim

samuro95:
With this PR, the regularization parameter lambda is now in front of the regularization function g. 


tachella (changes requested):
amazing refactoring work.

I just spotted one issue in the unfolded demo (why theta instead of sigma?) and we would need to also 
- change the README example that uses lambda
- change the open in colab example?
- report the changes in the docs

tachella (approved):
many thanks for incorporating the changes

PR #173, opened 2024-03-15
Projective transform for equivariant imaging

Andrewwango:
This PR introduces the projective transformation, or homography, an image transformation available for self-supervised learning using equivariant imaging. The transformation is implemented at `deepinv.transform.Homography`. 

I've also implemented a few subgroups of transformations, including `deepinv.transform.projective.Affine`, `deepinv.transform.projective.Similarity` and `deepinv.transform.projective.Euclidean` transformations, which may also be useful in some use-cases.

This was introduced in our recent paper [Perspective-Equivariant Imaging: an Unsupervised Framework for Multispectral Pansharpening](https://arxiv.org/abs/2403.09327), as the homography provides a richer group of transformations and so provides better performance for equivariant imaging in camera-based imaging scenarios such as satellite imaging.

See further demos of this functionality [at this webpage notebook](https://andrewwango.github.io/perspective-equivariant-imaging/demo.html#appendix-subgroup-transforms).

Note this adds `kornia` as a dependency, but happy to make it optional dependency instead if needed.


tachella:
Hey Andrew, 

This would be a nice contribution! I think it would be good to document all of the transforms, and add them in the  [documentation](https://deepinv.github.io/deepinv/deepinv.loss.html#transforms).

I can have a look once the tests are passing!

samuro95:
I am in favor of making the kornia dependency optional, especially because it is used only once, and only in case the transform is applied to tensors. @tachella what do you think ?

Andrewwango:
> I am in favor of making the kornia dependency optional, especially because it is used only once, and only in case the transform is applied to tensors. @tachella what do you think ?

How do I add it as an optional dependency?

tachella:
> > I am in favor of making the kornia dependency optional, especially because it is used only once, and only in case the transform is applied to tensors. @tachella what do you think ?
> 
> How do I add it as an optional dependency?

You can do it in the same way as bm3d for example, see https://github.com/deepinv/deepinv/blob/311f98ee3e30d1a95762b6b8f19d5c536e2e5921/deepinv/models/bm3d.py#L6

deepinv/training/testing.py
```diff
@@ -165,18 +165,13 @@ def test(
...
-            else:
-                save_folder_im = None
-            if plot_metrics:
-                save_folder_curve = (
-                    (save_folder / ("G" + str(g))) if G > 1 else save_folder
-                ) / "curves"
-                save_folder_curve.mkdir(parents=True, exist_ok=True)
-
-            if plot_images:
+                    (
```
tachella, line 168: why did you apply these changes?

deepinv/training/trainer.py
```diff
@@ -717,7 +726,7 @@ def train(
     **kwargs,
 ):
     """
-    Alias function for training a model using :class:`deepinv.training_utils.Trainer` class.
+    Alias function for training a model using :class:`deepinv.training.Trainer` class.
```
tachella, line 729: I think this should be directly deepinv.Trainer

deepinv/transform/projective.py
```diff
@@ -0,0 +1,330 @@
+from dataclasses import dataclass
+from typing import Union
+
+import numpy as np
+import torch
+from PIL import Image
+
+from scipy.spatial.transform import Rotation
```
tachella, line 8: Do you think it is possible to remove this dependency?
If I understood well, it is used for generating a rotation matrix from euclidean angles on a single line of code.

deepinv/transform/projective.py
```diff
@@ -0,0 +1,330 @@
...
+    Subgroup transformations include :class:`deepinv.transform.Affine`, :class:`deepinv.transform.Similarity`,
+    :class:`deepinv.transform.Euclidean` along with the basic :class:`deepinv.transform.Shift`,
+    :class:`deepinv.transform.Rotation` and semigroup :class:`deepinv.transform.Scale`.
+
+    Transformations with perspective effects (i.e. pan+tilt) are recovered by setting
+    theta_max > 0.
+
+    Generates n_trans random transformations concatenated along the batch dimension.
+
+    Example:
```
tachella, line 143: the standard way of adding examples is adding first a separator via |sep| 
and then using >>> (otherwise the doctests are not run,  and the automatic copy-paste button will not work) as in [here](https://github.com/deepinv/deepinv/blob/0d0c174f4e5fa9031439bcac82631420c745ed28/deepinv/physics/blur.py#L323)

deepinv/transform/projective.py
```diff
@@ -0,0 +1,330 @@
...
+
+    Special case of homography which corresponds to the actions of the affine subgroup
+    Aff(3). Affine transformations include translations, rotations, reflections,
+    skews, and stretches. These transformations are parametrised using geometric parameters in the pinhole camera model. See :class:`deepinv.transform.Homography` for more details.
+
+    Generates n_trans random transformations concatenated along the batch dimension.
+
+    Example:
+
+    ::
```
tachella, line 232: same here

deepinv/transform/projective.py
```diff
@@ -0,0 +1,330 @@
...
+class Similarity(Homography):
+    """Random 2D similarity image transformations.
+
+    Special case of homography which corresponds to the actions of the similarity subgroup
+    S(2). Similarity transformations include translations, rotations, reflections and
+    uniform scale. These transformations are parametrised using geometric parameters in the pinhole camera model. See :class:`deepinv.transform.Homography` for more details.
+
+    Generates n_trans random transformations concatenated along the batch dimension.
+
+    Example:
```
tachella, line 256: same here

deepinv/transform/projective.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+class Similarity(Homography):
+    """Random 2D similarity image transformations.
+
+    Special case of homography which corresponds to the actions of the similarity subgroup
+    S(2). Similarity transformations include translations, rotations, reflections and
+    uniform scale. These transformations are parametrised using geometric parameters in the pinhole camera model. See :class:`deepinv.transform.Homography` for more details.
+
+    Generates n_trans random transformations concatenated along the batch dimension.
```
tachella, line 254: n_trans -> ``n_trans`` using double `

deepinv/transform/projective.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+class Affine(Homography):
+    """Random affine image transformations.
+
+    Special case of homography which corresponds to the actions of the affine subgroup
+    Aff(3). Affine transformations include translations, rotations, reflections,
+    skews, and stretches. These transformations are parametrised using geometric parameters in the pinhole camera model. See :class:`deepinv.transform.Homography` for more details.
+
+    Generates n_trans random transformations concatenated along the batch dimension.
```
tachella, line 228: n_trans -> ``n_trans`` using double `

deepinv/transform/projective.py
```diff
@@ -0,0 +1,330 @@
...
+
+    Special case of homography which corresponds to the actions of the Euclidean subgroup
+    E(2). Euclidean transformations include translations, rotations and reflections. These transformations are parametrised using geometric parameters in the pinhole camera model.
+    See :class:`deepinv.transform.Homography` for more details.
+
+    Generates n_trans random transformations concatenated along the batch dimension.
+
+    Example:
+
+    ::
```
tachella, line 285: same here, use >>> for doctests and copybutton

deepinv/transform/projective.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+class Euclidean(Homography):
+    """Random Euclidean image transformations.
+
+    Special case of homography which corresponds to the actions of the Euclidean subgroup
+    E(2). Euclidean transformations include translations, rotations and reflections. These transformations are parametrised using geometric parameters in the pinhole camera model.
+    See :class:`deepinv.transform.Homography` for more details.
+
+    Generates n_trans random transformations concatenated along the batch dimension.
```
tachella, line 281: same here with n_trans -> ``n_trans`` using double `

deepinv/transform/projective.py
```diff
@@ -0,0 +1,330 @@
...
+    The transformations simulate panning, tilting or rotating the camera, leading to a
+    "perspective" effect. The subgroup is isomorphic to SO(3).
+
+    See :class:`deepinv.transform.Homography` for more details.
+
+    Generates n_trans random transformations concatenated along the batch dimension.
+
+    Example:
+
+    ::
```
tachella, line 317: Same here with >>>

examples/self-supervised-learning/demo_ei_transforms.py
```diff
@@ -0,0 +1,158 @@
...
+# masked with a random mask, without ground truth, using EI. For this
+# example we use the Urban100 images of natural urban scenes. As these
+# scenes are imaged with a camera free to move and rotate in the world,
+# all of the above transformations are valid invariances that we can
+# impose on the unknown image set :math:`x\in X`.
+#
+
+physics = dinv.physics.Inpainting((3, 256, 256), mask=0.6, device=device)
+
+download_and_extract_archive(
```
tachella, line 87: can you please update to the Urban100HR class added in a recent PR?

examples/self-supervised-learning/demo_ei_transforms.py
```diff
@@ -0,0 +1,158 @@
...
+
+train_dataloader, test_dataloader = DataLoader(train_dataset, shuffle=True), DataLoader(
+    test_dataset
+)
+
+
+# %%
+# For training, use a small UNet, Adam optimizer, EI loss with homography
+# transform, and the ``deepinv.Trainer`` functionality:
+#
```
tachella, line 109: maybe add a 

.. note:: 

that clarifies that we only train for a single epoch in the demo, but it is recommended to train x epochs for training the network from scratch.

tachella (changes requested):
The PR looks great, I just found a few details in the docstrings to correct: at the moment the example snippets will not run since the :: indicator for code is not recognized by doctest.

deepinv/training/testing.py
Andrewwango: I believe that was @samuro95 ?

deepinv/transform/projective.py
Andrewwango: Done!

examples/self-supervised-learning/demo_ei_transforms.py
Andrewwango: This increases demo runtime by ~1min as the current deepinv Urban100HR is much bigger than the HF Urban100 I'm using. I could create an issue?

tachella (approved):
many thanks @Andrewwango for this PR - everything looks good to me

PR #184, opened 2024-04-03
Plot inset util code

Andrewwango:
Util code for plotting images with insets (often useful for highlighting restoration/reconstruction results). Thanks to @jscanvic for helping with this. Example of plots produced using this code: https://andrewwango.github.io/perspective-equivariant-imaging/demo.html#results

### DIFF
--- a/deepinv/tests/test_utils.py
+++ b/deepinv/tests/test_utils.py
@@ -82,3 +82,13 @@ def test_plot():
         deepinv.utils.plot(imgs, titles=["a", "b"], show=False)
         deepinv.utils.plot(x, titles="a", show=False)
         deepinv.utils.plot(imgs, show=False)
+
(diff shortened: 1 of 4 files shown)


samuro95:
Could plot_inset be an option of the previous plot function ?

Andrewwango:
> Could plot_inset be an option of the previous plot function ?

It could do, but it will mean `plot` has a lot of unnecessary parameters for most users (unless we put all the inset locations etc. in a dict or some kwargs)

samuro95:
I think the dict could be the best solution indeed !

tachella (approved):
cool addition to the plotting utils of the library, thanks for suggesting this.
I think it is best to keep it separate from plot, to avoid having a monster plot function

PR #200, opened 2024-04-25
Additions to the Tomography Operator

johertrich:
I added two extensions to the tomography operator:
- an option parallel_computation. If it is set to True, then all projections are performed in parallel (instead of using a for loop over the angles). In my tests this is (on a GPU) faster than the non-parallel computation, but requires more memory.
- an option which allows to use a fan beam geometry instead of parallel beam geometry. I only implemented the Radon transform (and not its inverse) for fan beam. For calling the adjoint and filtered backprojection, it uses the deepinv.physics.adjoint_function method (which works fine, but maybe a direct implementation could be more efficient).

During implementing these changes, I observed that the A_adjoint method of the tomography operator is not the adjoint in the linear algebra sense, but a rescaled version of it. I also added this rescaling for the fan beam case to keep it consistent, but I am not sure if this is the intended behaviour.

### DIFF
--- a/deepinv/physics/functional/radon.py
+++ b/deepinv/physics/functional/radon.py
@@ -128,15 +128,21 @@ class Radon(nn.Module):
 
     :param int in_size: the size of the input image. If None, the size is inferred from the input image.
     :param torch.Tensor theta: the angles at which the Radon transform is computed. Default is torch.arange(180).
-    :param bool circle: if True, the input image is assumed to be a circle. Default is False.
-    :param bool parallel_computation: if True, all projections are performed in parallel. Requires more memory but is faster on GPUs.
-    :param bool fan_beam: if True, use fan beam geometry, if False use parallel beam
-    :param dict fan_parameters: only used if fan_beam is True. Contains the parameters defining the scanning geometry. The dict should contain the keys
+    :param bool circle: if ``True``, the input image is assumed to be a circle. Default is ``False``.
+    :param bool parallel_computation: if ``True``, all projections are performed in parallel. Requires more memory but is faster on GPUs.
+    :param bool fan_beam: If ``True``, use fan beam geometry, if ``False`` use parallel beam
+    :param dict fan_parameters: Only used if fan_beam is ``True``. Contains the parameters defining the scanning geometry. The dict should contain the keys:
+
         - "pixel_spacing" defining the distance between two pixels in the image, default: 0.5 / (in_size)
+
         - "source_radius" distance between the x-ray source and the rotation axis (middle of the image), default: 57.5
+
         - "detector_radius" distance between the x-ray detector and the rotation axis (middle of the image), default: 57.5
+
         - "n_detector_pixels" number of pixels of the detector, default: 258
+
         - "detector_spacing" distance between two pixels on the detector, default: 0.077
+
         The default values are adapted from the geometry in `https://doi.org/10.5281/zenodo.8307932 <https://doi.org/10.5281/zenodo.8307932>`_,
         where pixel spacing, source and detector radius and detector spacing are given in cm.
         Note that a to small value of n_detector_pixels*detector_spacing can lead to severe circular artifacts in any reconstruction.
--- a/deepinv/physics/functional/radon.py
+++ b/deepinv/physics/functional/radon.py
@@ -132,11 +132,12 @@ class Radon(nn.Module):
     :param bool parallel_computation: if True, all projections are performed in parallel. Requires more memory but is faster on GPUs.
     :param bool fan_beam: if True, use fan beam geometry, if False use parallel beam
     :param dict fan_parameters: only used if fan_beam is True. Contains the parameters defining the scanning geometry. The dict should contain the keys
-        - "pixel_spacing" defining the distance between two pixels in the image
-        - "source_radius" distance between the x-ray source and the rotation axis (middle of the image)
-        - "detector_radius" distance between the x-ray detector and the rotation axis (middle of the image)
-        - "n_detector_pixels" number of pixels of the detector
-        - "detector_spacing" distance between two pixels on the detector
+        - "pixel_spacing" defining the distance between two pixels in the image, default: 0.077
+        - "source_radius" distance between the x-ray source and the rotation axis (middle of the image), default: 57.5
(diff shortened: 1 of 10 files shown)


deepinv/physics/functional/radon.py
```diff
@@ -95,6 +129,15 @@ class Radon(nn.Module):
     :param int in_size: the size of the input image. If None, the size is inferred from the input image.
     :param torch.Tensor theta: the angles at which the Radon transform is computed. Default is torch.arange(180).
     :param bool circle: if True, the input image is assumed to be a circle. Default is False.
+    :param bool parallel_computation: if True, all projections are performed in parallel. Requires more memory but is faster on GPUs.
+    :param bool fan_beam: if True, use fan beam geometry, if False use parallel beam
+    :param dict fan_parameters: only used if fan_beam is True. Contains the parameters defining the scanning geometry. The dict should contain the keys
+        - "pixel_spacing" defining the distance between two pixels in the image, default: 0.077
+        - "source_radius" distance between the x-ray source and the rotation axis (middle of the image), default: 57.5
```
tachella, line 136: this value is in pixels right?

deepinv/physics/functional/radon.py
```diff
@@ -113,9 +159,29 @@ def __init__(
...
+        self.parallel_computation = parallel_computation
+        self.fan_beam = fan_beam
+        self.fan_parameters = fan_parameters
+        if fan_beam:
+            if self.fan_parameters is None:
+                self.fan_parameters = {}
+            if not "pixel_spacing" in self.fan_parameters.keys():
+                self.fan_parameters["pixel_spacing"] = 0.077
+            if not "source_radius" in self.fan_parameters.keys():
+                self.fan_parameters["source_radius"] = 57.5
```
tachella, line 171: should this be somehow relative to the number of pixels in the image?

deepinv/physics/tomography.py
```diff
@@ -85,18 +101,57 @@ def __init__(
...
+                not self.img_width is None
+            ), "Image size unknown. Apply forward operator or add it for initialization."
+            # lazy implementation for the adjoint...
+            adj = adjoint_function(
+                self.A,
+                (y.shape[0], y.shape[1], self.img_width, self.img_width),
+                device=self.device,
+                dtype=self.dtype,
+            )
+            # the tomography operator does not exactly implement the adjoint but a rescaled version of it...
```
tachella, line 154: I think we should use the exact adjoint and not a rescaled version.

tachella (commented):
Hi Johannes, many thanks for this PR!

Parallel computation and beam geometry are two very nice feature upgrades.

Regarding the scaled adjoint, I think we should apply the true 'mathematical' adjoint, to be coherent with the rest of the library. Could you please modify both the fan beam and parallel adjoints for this?

Going back to a point that you raised in the past, we should probably add an option for normalizing the operator to have norm approx 1, would it be possible to add this as well?

I've also added a few minor comments on default values for the parallel beam geometry parameters.

deepinv/physics/functional/radon.py
johertrich: No, the default values are adapted from the scanner used in 
https://doi.org/10.5281/zenodo.8307932
where pixel spacing, source radius and detector radius where given in centimeter, I added this in the description. Actually, rescaling all the values (pixel spacing, source/detector radius, detector spacing) by the same constant does not change the operator. So, one could decide to fix the pixel spacing as 1 and give all the other values in pixels. However, if you have data from some scanner you always have to convert cm into pixels...

deepinv/physics/functional/radon.py
johertrich: You are right that the values should adapt to the image size. I changed the default value of the pixel spacing to 1/im_with.

johertrich:
Ok, I am done with checking and adapting the examples to the change in the scaling. The adaption of the learned primal dual is not really necessary (the neural network can learn to compensate the scaling), but it accelerates the training a lot.

deepinv/physics/functional/radon.py
tachella: ok, thanks for adding the explanation!

deepinv/physics/tomography.py
```diff
@@ -32,6 +36,19 @@ class Tomography(LinearPhysics):
     :param int img_width: width/height of the square image input.
     :param bool circle: If ``True`` both forward and backward projection will be restricted to pixels inside a circle
         inscribed in the square image.
+    :param bool parallel_computation: if True, all projections are performed in parallel. Requires more memory but is faster on GPUs.
+    :param bool normalize: if True, the outputs are normlized by the image size (i.e. it is assumed that the image lives on [0,1]^2 for the computation of the line integrals).
```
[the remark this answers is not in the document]
tachella: is this value different for the fan beam case?

tachella (changes requested):
Many thanks for applying the normalization fixes and fixing the examples!

I would 

1. add to the docs the value of the operator norm under the normalization (this is quite handy when setting up optim algorithms) 
2. wait for confirmation of @samuro95 regarding the demo_learned_primal_dual 

and then we can merge!

deepinv/physics/tomography.py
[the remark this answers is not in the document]
johertrich: As long as source and detector are not too close to the measured sample (i.e. detector radius and source radius are large enough), the value should be approximately the same.

examples/unfolded/demo_learned_primal_dual.py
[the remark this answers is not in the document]
samuro95: I agree that the network should automatically adapt, there is not stepsize to rescale.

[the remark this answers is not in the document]
johertrich: I guess the main reason is that the filtered backprojection is closer to an inverse of the forward operator and has already the correct scaling. I added a short note and a reference to the example.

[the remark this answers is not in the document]
johertrich: Also, when one runs the training long enough, the reconstruction becomes reasonable with the adjoint as well (but still worse than with the FBP).

samuro95 (approved):

PR #216, opened 2024-05-15
fix drunet diffunet

matthieutrs:
### Checks to be done before submitting your PR
- [ ] `python3 -m pytest tests/` runs successfully.
- [ ] `black .` runs successfully.
- [ ] `make html` runs successfully (in the `docs/` directory).
- [ ] Updated docstrings related to the changes (as applicable).
- [ ] Added an entry to the [CHANGELOG.rst](../CHANGELOG.rst).

PR #228, opened 2024-05-29
adapt sigma when batch size > 1

Nils-Laurent:
When using a batch size greater than 1 and a noise model using a parameter sigma (which is a tensor of the length equals to the batch size when using a generator), then an error happens when multiplying `torch.rand_like(x) * sigma`.

The fix proposed here completes the missing dimensions of sigma to perform the operation.

(based on https://github.com/Nils-Laurent/deepinv/commits/main/)

### DIFF
--- a/deepinv/physics/noise.py
+++ b/deepinv/physics/noise.py
@@ -47,7 +47,7 @@ class GaussianNoise(torch.nn.Module):
         if sigma is not None:
             self.sigma = to_nn_parameter(sigma)
 
-        return x + torch.randn_like(x) * self.sigma
+        return x + torch.randn_like(x) * self.sigma[(...,) + (None,) * (x.ndim - 1)]
 
 
 class UniformGaussianNoise(torch.nn.Module):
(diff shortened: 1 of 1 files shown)


Nils-Laurent:
This code may be invalid, locally, I encountered a problem of incompatible devices.

tachella (changes requested):
Hi Nils,

Many thanks for proposing this fix.

Does the code work for the following cases?

- sigma.size(0) = x.size(0) but not the other dims
- sigma is a float/a single element tensor
- sigma and x are on the gpu

matthieutrs:
Hey @Nils-Laurent, the current implementation of `Physics` is not extremely friendly with multi-gpu computations (which your problem suggest) when one samples a different `sigma` at each step of the training (for example). If what you are trying to do is to train on mutliple levels of noise simultaneously (is this the case?) on multiple gpus (is this the case?) I would maybe suggest to rewrite your training physics so they output the training noise level together with the noisy image, and input both in the model. In this vein, you may take a look at what is being done in this #216 (more precisely [here](https://github.com/deepinv/deepinv/pull/216/files#diff-7b4519f1e26eb4aae3631b613f2cfc863a49b1d84fc14f938b20aeb4a71ff491R186)). Would such a solution work for you?

tachella:
should we close this PR then?

Nils-Laurent:
I think PR #280 solved the issue, for Gaussian noise. Should we use the same fix for PoissonGaussianNoise ? Otherwise, I think we can close the PR.

tachella:
I see, we can close it I believe, and add other noise distributions later.

PR #229, opened 2024-05-30
Add fastMRI pytorch dataset

MaximeSongIdris:
Adding support for
- [x] single-coil MRI
- [x] multi-coil TODO


MaximeSongIdris:
In the actual iteration, since kspace of different mris have different (height, width) sizes, the dataset can't be feed to a dataloader. One solution is to filter mri slices by (height, width) size to keep only one size of mri slice.

deepinv/datasets/div2k.py
```diff
@@ -30,65 +33,60 @@ class DIV2K(torch.utils.data.Dataset):
...
+                        f"""
+                    The train folder already exists,
+                    thus the download is aborted.
+                    Please set `download=False`
+                    OR remove `{train_dir_path}`."""
+                    )
+
+                if os.path.exists(valid_dir_path):
+                    raise ValueError(
+                        f"""
```
tachella, line 64: I think there is a newline in the string that should be removed 

f"""text here 

instead of

f""" (new line)
text here

tachella:
> In the actual iteration, since kspace of different mris have different (height, width) sizes, the dataset can't be feed to a dataloader. One solution is to filter mri slices by (height, width) size to keep only one size of mri slice.

@matthieutrs do you know more about this?
- Do they have different sizes because they correspond to different acceleration factors of the same underlying image size, or because the are related to different image sizes?
- can we incorporate a torchvision transform that resizes/normalizes?

deepinv/datasets/fastmri.py
```diff
@@ -0,0 +1,248 @@
...
+    :param callable, optional transform_kspace: A function/transform that takes in the
+        kspace and returns a transformed version. E.g, ``torchvision.transforms.RandomCrop``
+    :param callable, optional transform_target: A function/transform that takes in the
+        target and returns a transformed version. E.g, ``torchvision.transforms.RandomCrop``
+
+    |sep|
+
+    :Examples:
+
+        Instanciate dataset without transform ::
```
tachella, line 84: 1) Are we not testing on purpose here?

2) is this showing right on sphinx? I feel that the :: should be on a new line.

deepinv/datasets/fastmri.py
```diff
@@ -0,0 +1,248 @@
...
+        self.root --- file1000005.h5
+                   |
+                   -- xxxxxxxxxxx.h5
+
+    | 0) To download raw data, please go to the bottom of the page `https://fastmri.med.nyu.edu/`
+    | 1) Each MRI scan is stored in a HDF5 file and can be read with the h5py package.
+    | Each file contains the k-space data, ground truth and some meta data related to the scan.
+    | 2) MRI scans can either be single-coil or multi-coil with each coil in
+    | a multi-coil MRI scan focusses on a different region of the image.
+    | 3) In multi-coil MRIs, k-space has the following shape:
```
tachella, line 52: k-space --> the k-space data

deepinv/datasets/fastmri.py
```diff
@@ -0,0 +1,248 @@
...
+    | (number of slices, number of coils, height, width)
+    | 4) For single-coil MRIs, k-space has the following shape:
```
tachella, line 54: same here

deepinv/datasets/fastmri.py
```diff
@@ -0,0 +1,248 @@
...
+
+    :param Union[str, Path] root: Path to the dataset.
+    :param bool test: Whether the split is the "test" set.
+    :param str challenge: "singlecoil" or "multicoil" depending on the type of mri scan.
+    :param bool load_metadata_from_cache: Whether to load dataset metadata from cache.
+    :param bool save_metadata_to_cache: Whether to cache dataset metadata.
+    :param Union[str, Path] metadata_cache_file: A file used to cache dataset
+        information for faster load times.
+    :param callable, optional sample_filter: A callable object that takes a
+        SliceSampleFileIdentifier as input and returns a boolean indicating
```
tachella, line 65: add a ref to the class with :meth:`deepinv.datasets.SliceSampleFileIdentifier` (check that this works)

deepinv/datasets/fastmri.py
```diff
@@ -0,0 +1,248 @@
...
+    | Each file contains the k-space data, ground truth and some meta data related to the scan.
+    | 2) MRI scans can either be single-coil or multi-coil with each coil in
+    | a multi-coil MRI scan focusses on a different region of the image.
+    | 3) In multi-coil MRIs, k-space has the following shape:
+    | (number of slices, number of coils, height, width)
+    | 4) For single-coil MRIs, k-space has the following shape:
+    | (number of slices, height, width)
+
+    :param Union[str, Path] root: Path to the dataset.
+    :param bool test: Whether the split is the "test" set.
```
tachella, line 58: This is a bit non-standard, I would add a train flag rather than test

deepinv/physics/blur.py
```diff
@@ -380,6 +380,9 @@ def update_parameters(self, filter=None, **kwargs):
         if filter is not None:
             self.filter = torch.nn.Parameter(filter, requires_grad=False)
 
+        if hasattr(self.noise_model, "update_parameters"):
```
tachella, line 383: why is this here? It should be part of another PR

deepinv/physics/mri.py
```diff
@@ -22,86 +24,206 @@ class MRI(DecomposablePhysics):
...
+                  [ 0.4564, -0.0000]],
+        <BLANKLINE>
+                 [[ 0.0000,  1.8622],
+                  [ 0.0603, -0.0000]]]])
+        >>> physics = MRI(img_size=x.shape) # No subsampling
         >>> physics(x)
-        tensor([[-0.5305,  0.0351,  0.3326,  2.1730,  1.7072,  0.0418]])
+        tensor([[[[ 2.2908, -1.4290],
+                  [ 0.4564, -0.1814]],
+        <BLANKLINE>
```
tachella, line 54: why the BLANKLINE?

deepinv/physics/mri.py
```diff
@@ -22,86 +24,206 @@ class MRI(DecomposablePhysics):
...
-            )
-
-    def U_adjoint(self, x):
-        _, c, h, w = self.mask.shape
-        out = torch.zeros((x.shape[0], c, h, w), device=x.device)
+    def U(self, x: Tensor) -> Tensor:
+        if not torch.all((self.mask == 0) | (self.mask == 1)):
+            warnings.warn("The mask must have values 0 or 1")
 
-        if self.mask.size(0) == 1:
```
tachella, line 88: why did you remove the size reduction? @matthieutrs

tachella (commented):
I have a few minor comments. I worry that there are some changes that correspond to old PRs, is this expected?

deepinv/datasets/fastmri.py
MaximeSongIdris: yes, here I want to separate the behavior between train/val vs test

deepinv/physics/mri.py
MaximeSongIdris: Duno, that's the result we get when testing with doctest :

python -m doctest -v deepinv/physics/mri.py

deepinv/datasets/fastmri.py
tachella: why not a 'mode' which equals 'train' 'val' or 'test?

deepinv/datasets/fastmri.py
tachella: perfect!

deepinv/datasets/fastmri.py
MaximeSongIdris: train and val have return the target, while test doesn't
and we don't use mode for other

tachella (approved):

PR #237, opened 2024-06-05
Urban100

MaximeSongIdris:

### DIFF
--- a/deepinv/datasets/urban100.py
+++ b/deepinv/datasets/urban100.py
@@ -11,7 +11,7 @@ class Urban100HR(torch.utils.data.Dataset):
     """Dataset for `Urban100 <https://paperswithcode.com/dataset/urban100>`_.
 
     The Urban100 dataset contains 100 images of urban scenes.
-    It commonly used as a test set to evaluate the performance of super-resolution models.
+    It is commonly used as a test set to evaluate the performance of super-resolution models.
 
 
     **Raw data file structure :** ::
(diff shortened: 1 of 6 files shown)


deepinv/datasets/urban100.py
```diff
@@ -0,0 +1,135 @@
...
+                   -- readme.txt
+                   -- source_selected.xlsx
+                   -- Urban100_SR.zip
+
+    For each image in the dataset, several variations of the same image exist.
+    For this class, we are dealing with only high resolution images in `image_SRF_4` folder.
+    For more information about the different versions, look at `readme.txt`.
+
+    >>> root = "/path/to/dataset/Urban100"
+    ... dataset = Urban100HR(root=root, download=True)  # will download dataset at root
```
tachella, line 35: the ... should be >>> I believe, since they represent a new line.

deepinv/datasets/urban100.py
```diff
@@ -0,0 +1,135 @@
+from typing import Any, Callable
+import os
+
+from PIL import Image
+import torch
+
+from .utils import calculate_md5_for_folder, download_zipfile, extract_zipfile
+
+
+class Urban100HR(torch.utils.data.Dataset):
```
tachella, line 10: I would remove the HR from the name

deepinv/datasets/urban100.py
tachella: Also please add a separator |sep| and Example as in [here](https://github.com/deepinv/deepinv/blob/01a7c863438117c7ab52f20be780bbec0b6b749e/deepinv/physics/blur.py#L156).

deepinv/datasets/urban100.py
```diff
@@ -0,0 +1,135 @@
...
+    For each image in the dataset, several variations of the same image exist.
+    For this class, we are dealing with only high resolution images in `image_SRF_4` folder.
+    For more information about the different versions, look at `readme.txt`.
+
+    >>> root = "/path/to/dataset/Urban100"
+    ... dataset = Urban100HR(root=root, download=True)  # will download dataset at root
+    ... dataset.check_dataset_exists()                  # check that raw data has been downloaded correctly
+
+    :param str root: Root directory of dataset. Directory path from where we load and save the dataset.
+    :param bool download: If True, downloads the dataset from the internet and puts it in root directory.
```
tachella, line 39: True -> ``True``

deepinv/datasets/urban100.py
```diff
@@ -0,0 +1,135 @@
...
+                    )
+                    # extract local zipfile
+                    extract_zipfile(os.path.join(self.root, filename), self.root)
+
+                    if self.check_dataset_exists():
+                        print("Dataset has been successfully downloaded.")
+            # stop the execution since the dataset is not available and we didn't download it
+            else:
+                raise RuntimeError(
+                    f"""
```
tachella, line 93: Here be careful that this prints without new lines.

deepinv/datasets/urban100.py
tachella: also in the example I would print the dataset length and check that is 100

deepinv/datasets/urban100.py
```diff
@@ -0,0 +1,135 @@
...
+                   |
+                   -- image_SRF_4 --- img_001_SRF_4_A+.png
+                   |               |
+                   |               -- img_100_SRF_4_SRCNN.png
+                   -- readme.txt
+                   -- source_selected.xlsx
+                   -- Urban100_SR.zip
+
+    For each image in the dataset, several variations of the same image exist.
+    For this class, we are dealing with only high resolution images in `image_SRF_4` folder.
```
tachella, line 31:
```
For each image in the dataset, several variations of the same image exist.
For this class, we are dealing with only high resolution images in `image_SRF_4` folder.
```

rewrite to 

```
This dataset wrapper gives access to the 100 high resolution images in the `image_SRF_4` folder.
```

tachella (changes requested):
Hey Maxime,

The PR looks very good, I have a few comments for changes.

deepinv/datasets/urban100.py
MaximeSongIdris: I have a preference on HR since we are not exactly using the whole dataset (with all 11 variations).
Putting HR explicitly will make them know that they are not using the whole dataset.

deepinv/datasets/urban100.py
MaximeSongIdris: it prints a new lines, i didn't found an elegant way to avoid the new line and having a clean str msg

deepinv/datasets/urban100.py
tachella: ok

deepinv/datasets/urban100.py
tachella: I think it is better to remove the new lines (eg torch's standard printing doesn't use newlines)

MaximeSongIdris: ok

Issue #239, opened 2024-06-07
Averaging total loss


tachella:
At the moment, the dinv.Trainer losses are being averaged over epochs.

PR #241, opened 2024-06-08
Update params for physics

tachella:
1- I've added an update_parameters() method to the physics class that modifies the physics parameters without needing to call the forward operator or the adjoint.

So far the only way of changing parameters on the go was via 
`y = physics(x, params)`
which results in an unnecessary forward call if y is no necessary.

whereas now we can also do:
`physics.update_parameters(params)`

2- I've fixed issue #239 by cleaning the meters in the training loop after each epoch. I also cleaned up a bit the header of the Trainer docs.

3- I've removed the environment.yml file which was not used anymore.

4- I've applied a minor fix to the conjugate gradient code.

5- I've added functions to change the eval mode of the models trained via SplittingLoss or R2RLoss


deepinv/optim/utils.py
```diff
@@ -57,7 +57,7 @@ def conjugate_gradient(
     def dot(a, b):
         ndim = a[0].ndim if isinstance(a, TensorList) else a.ndim
         dot = (a.conj() * b).sum(
-            dim=tuple(range(1, ndim)), keepdim=False
```
matthieutrs, line 60: I agree with this but this is already the case in the main branch, did you merge from main?

deepinv/physics/forward.py
```diff
@@ -741,6 +757,15 @@ def A_dagger(self, y, **kwargs):
 
         return self.V(self.U_adjoint(y) * mask)
 
+    def update_parameters(self, mask=None, **kwargs):
```
Andrewwango, line 760: For the most generic update_parameters, should we make it purely kwargs:
```
def update_parameters(self, **kwargs):
for kwarg in kwargs...
    if hasattr(self, kwarg)...
```
This way, we can call `self.update_parameters(mask=mask)` in `A` without fear that its behaviour will be different when `update_parameters` is overriden.

tachella: good idea!

deepinv/optim/utils.py
tachella: I've merged from main and the issue seems to be gone

deepinv/physics/forward.py
Andrewwango: Can we also replace `if self.mask is None: self.mask = nn.Parameter(...` with `self.update_parameters(mask=mask)` in `DecomposablePhysics.__init__`, `DecomposablePhysics.A`, `A_adjoint`, `A_dagger`?

Andrewwango: Or I can do this in #242

Andrewwango: Likewise replacing the `self.filter = ...` in `Blur.__init__`, `Blur.A`...

tachella: done!

PR #244, opened 2024-06-11
Remove unnecessary double download of same dataset during test

MaximeSongIdris:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,12 +36,10 @@ venv
 build
 deepinv.egg-info
 dist
-examples/datasets
 wandb/
 kernels
 measurements/
 degradations/
-examples/*/datasets
 results/
 tmp/
 sg_execution_times.rst
--- a/.gitignore
+++ b/.gitignore
@@ -36,12 +36,12 @@ venv
 build
 deepinv.egg-info
 dist
-datasets
+examples/datasets
 wandb/
 kernels
 measurements/
 degradations/
-datasets/
+examples/*/datasets
 results/
 tmp/
 sg_execution_times.rst
--- a/deepinv/datasets/div2k.py
+++ b/deepinv/datasets/div2k.py
@@ -1,16 +1,32 @@
+from typing import Any, Callable
 import os
 
 from PIL import Image
 import torch
 
-from .utils import calculate_md5_for_folder, download_zipfile, extract_zipfile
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -36,12 +36,12 @@ venv
 build
 deepinv.egg-info
 dist
-datasets
+examples/datasets
```
tachella, line 39: Is this removing the examples from git tracking?

deepinv/datasets/div2k.py
```diff
@@ -30,81 +50,70 @@ class DIV2K(torch.utils.data.Dataset):
...
 
+        # download a split of the dataset, we check first that this split isn't already downloaded
+        if not self._verify_split_dataset_integrity():
+            if download:
+                if not os.path.isdir(self.root):
+                    os.makedirs(self.root)
+                if os.path.exists(self.img_dir):
+                    raise ValueError(
+                        f"""
+                    The {self.mode} folder already exists,
```
tachella, line 80: can you print all the string in one line, without any space at the beginning?

.gitignore
```diff
@@ -36,12 +36,12 @@ venv
...
 wandb/
 kernels
 measurements/
 degradations/
-datasets/
+examples/*/datasets
```
tachella, line 44: same question here?

tachella (changes requested):
the PR looks good, but I think there might be an issue with the gitignore.

.gitignore
MaximeSongIdris: initially it removed every "datasets" folder in the git tracking
notably the "deepinv.datasets" folder
and I changed it to examples/datasets, which contains raw dataset

MaximeSongIdris: i though that the initial behavior was intended for "examples/datasets" -> do i need to change this line ?

.gitignore
MaximeSongIdris: same as above, 
I inferred that the expected behavior was : 
ignoring all "datasets" subfolders in "examples" folder

deepinv/datasets/div2k.py
MaximeSongIdris: Done :
every multi line strings  ->  one-line string

tachella (approved):
thanks for the PR, looks good to me!

Issue #238, opened 2024-06-07
`DecomposablePhysics`+`DecomposablePhysics`=`LinearPhysics`


Andrewwango:
Performing arithmetic on `DecomposablePhysics` should result in an object that retains methods and attributes of the original `DecomposablePhysics`. The resultant mask should also be `mask1*mask2` too.

```python
>>> import deepinv as dinv
>>> p1 = dinv.physics.Inpainting((2, 2, 2), 0.3)
>>> p2 = dinv.physics.Inpainting((2, 2, 2), 0.4)
>>> p3 = p1 * p2
>>> p3.mask

AttributeError: 'LinearPhysics' object has no attribute 'mask'
```

tachella:
Ok, I'm proposing the following solution for inpainting, which seems to be the most useful case:

```
    def __mul__(self, other):
        r"""
        Concatenates two forward operators :math:`A = A_1\circ A_2` via the mul operation

        If the second operator is an Inpainting or MRI operator, the masks are multiplied elementwise,
        otherwise the default implementation of LinearPhysics is used (see :meth:`deepinv.physics.LinearPhysics.__mul__`).

        :param deepinv.physics.Physics other: Physics operator :math:`A_2`
        :return: (deepinv.physics.Physics) concantenated operator

        """
        if isinstance(other, Inpainting):
            return Inpainting(
                tensor_size=self.tensor_size,
                mask=self.mask * other.mask,
                noise_model=other.noise_model,
                pixelwise=self.pixelwise,
                device=self.device,
            )
        elif isinstance(other, MRI):
            return MRI(
                mask=self.mask * other.mask,
                noise_model=other.noise_model,
                img_size=other.img_size,
                device=other.device,
            )
        else:
            return self.super().__mul__(other)
```

Andrewwango:
- Do we need to add a similar thing to `MRI.__mul__` too?
- And can we replace 
```
        if isinstance(other, Inpainting):
            return Inpainting(
```
with
```
        if isinstance(other, self.__class__):
            return self.__class__(
```
so that if it's a child class (e.g. some future multidimensional inpainting) it will still work?

tachella:
no for MRI, since I think the concatenation can only happen as Inpainting * MRI = MRI, but MRI * Inpainting \neq MRI.

Yes, for the second suggestion! Actually I've added demosaicing who falls into that category.

Andrewwango:
Great!
[...]

PR #255, opened 2024-06-25
changes to splitting and r2r + demosaicing + trainer

tachella:
- I've modified the R2R and Splitting self-supervised losses, such that they can be more easily combined with other losses (eg the EI loss), and they avoid an additional model call during training (they now use x_net directly).
- I've added the Demosaicing inverse problem, which is a trivial (but cool) adaptation of inpainting
- I've added a few fixes to the Trainer


deepinv/training/trainer.py
```diff
@@ -641,30 +675,44 @@ def train(
...
-                l.reset()
+            self.loss_history.append(self.logs_total_loss_train.avg)
+
+            if self.scheduler:
+                self.scheduler.step()
 
             ## Evaluation
             perform_eval = self.eval_dataloader and (
-                (epoch + 1) % self.eval_interval == 0 or epoch + 1 == self.epochs
+                epoch % self.eval_interval == 0 or epoch == self.epochs
```
Andrewwango, line 707: Surely `epoch` will never `==self.epochs`? I think both should be kept as `epoch+1`

deepinv/training/trainer.py
```diff
@@ -253,6 +257,17 @@ def setup_train(self):
 
         self.loss_history = []
 
+    def reorder_losses(self):
```
Andrewwango, line 260: Why is this needed? seems a bit hacky

tachella: I agree, but I don't see another solution at the moment. Both SplittingLoss and R2R need to save the mask/noise they use in the forward eval `x_net = model(y, physics)` to use it in the loss computation. If another loss is called before them, eg EILoss, it will generate a new mask/noise during the model's call, and this might cause supoptimal training.

deepinv/training/trainer.py
tachella: I've changed this to always perform an eval on the first iteration (it is good to check where the model starts in my opinion)

deepinv/training/trainer.py
Andrewwango: It makes sense, but I think it's risky (and not very OOP) to have the mask/noise rely on no forward pass being called in between the forward pass and the loss calculation. Could we put `self.mask = get_mask(...` in the `SplittingModel.__init__` so the mask definitely stays the same for forward pass and the loss calculation? For the `tsize` we could ask for it in the `SplittingLoss` definition and pass it to the `SplittingModel`

tachella: maybe we can use a fix that is robust and doesn't require changing that much?

```
x_net=model(y, physics)
if hasattr(model, update_parameters)
       x_net=model(y, physics, update_params=True)
else: 
      x_net = model(y, physics)
```

deepinv/loss/measplit.py
```diff
@@ -52,96 +87,77 @@ def forward(self, y, physics, model, **kwargs):
...
+        self.MC_samples = MC_samples
+        self.mask = 0
+
+    def forward(self, y, physics):
+        MC = 1 if self.training else self.MC_samples
+        tsize = y.size()[1:]
+        out = 0
+        with torch.set_grad_enabled(self.training):
+            for i in range(MC):
+                mask = get_mask(
```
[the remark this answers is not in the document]
tachella: I've changed this so that now we can pass a physics generator!

deepinv/loss/measplit.py
```diff
@@ -21,18 +33,41 @@ class SplittingLoss(Loss):
...
+    |sep|
+
+    :Example:
+
+    >>> import torch
+    >>> import deepinv as dinv
+    >>> physics = dinv.physics.Inpainting()
+    >>> model = dinv.models.MedianFilter()
+    >>> loss = dinv.loss.SplittingLoss(split_ratio=0.9)
+    >>> model = loss.adapt_model(model, MC_samples=2) # important step!
```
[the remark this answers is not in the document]
tachella: done

deepinv/training/trainer.py
Andrewwango: Looks good to me!

deepinv/loss/measplit.py
```diff
@@ -52,96 +87,77 @@ def forward(self, y, physics, model, **kwargs):
...
-            mask[..., start::stride, start::stride] = 0.0
+        mask = model.get_mask()
 
         # create inpainting masks
-        inp = Inpainting(tsize, mask, device=y.device)
         inp2 = Inpainting(tsize, 1 - mask, device=y.device)
 
         # concatenate operators
-        physics1 = inp * physics  # A_1 = P*A
         physics2 = inp2 * physics  # A_2 = (I-P)*A
```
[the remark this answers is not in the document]
tachella: I've applied a 'manual' fix see #238

deepinv/loss/measplit.py
```diff
@@ -52,96 +75,92 @@ def forward(self, y, physics, model, **kwargs):
...
-                out = 0
-                with torch.no_grad():
-                    for i in range(self.MC_samples):
-                        if not self.regular_mask:
-                            mask[torch.rand_like(mask) > self.split_ratio] = 0
-                        else:
-                            stride = int(1 / (1 - self.split_ratio))
-                            start = np.random.randint(stride)
-                            mask[..., start::stride, start::stride] = 0.0
+        if model.__class__.__name__ == "SplittingModel":
```
Andrewwango, line 118: `if isinstance(model, SplittingModel)`?

tachella: oh yes, that's a more elegant solution

deepinv/loss/measplit.py
```diff
@@ -21,28 +21,51 @@ class SplittingLoss(Loss):
...
-        :math:`\hat{x} = \frac{1}{N}\sum_{i=1}^N \inversef{y_1^{(i)}}{A_1^{(i)}}`. This can be achieved using
-        :meth:`deepinv.loss.splitting_eval`.
+        :math:`\hat{x} = \frac{1}{N}\sum_{i=1}^N \inversef{y_1^{(i)}}{A_1^{(i)}}`.
 
     :param torch.nn.Module metric: metric used for computing data consistency,
         which is set as the mean squared error by default.
     :param float split_ratio: splitting ratio, should be between 0 and 1. The size of :math:`y_1` increases
         with the splitting ratio.
-    :param bool regular_mask: If ``True``, it will use a regular mask, otherwise it uses a random mask.
+    :param deepinv.physics.generator.PhysicsGenerator, None mask_generator: function to generate the mask. If
```
Andrewwango, line 38: Can we keep the `regular_mask` parameter and deprecate it for backwards compatibility? And can we have a generator for regular masks?

tachella: I am happy to keep regular_mask for backwards compatibility. However, I think if a regular mask is needed (for denoising) then it is much better to use Neighbor2Neighbor

tachella: I would actually prefer to get rid of it if that's okay

Andrewwango: Ok no problem, I'll let our student know

deepinv/physics/inpainting.py
```diff
@@ -71,7 +72,10 @@ def __init__(self, tensor_size, mask, pixelwise=True, device="cpu", **kwargs):
             else:
                 mask[:, aux[0, :, :] > mask_rate] = 0
 
-        self.mask = torch.nn.Parameter(mask.unsqueeze(0), requires_grad=False)
+        if len(mask.shape) == len(tensor_size):
```
Andrewwango, line 75: I think at the top of `__init__` we need `if tensor_size == 3: tensor_size = (1, *tensor_size)`, and then here we need `while len(mask.shape) < len(tensor_size): mask = mask.unsqueeze(0)` to catch all mask shape inputs. Also change 
`mask[:, aux[0, :, :] > mask_rate] = 0` to `mask[:, :, aux[:, 0, ...] > mask_rate] = 0`

tachella: the issue is that some tensor sizes might not have the standard C x H x W dimensions (some measurements have arbitrary sizes)

Andrewwango: Ok, although I thought `if len(mask.shape) == len(tensor_size):` would fail on if `mask` is BCHW and `tensor_size` is also BCHW? Like in https://github.com/deepinv/deepinv/blob/main/deepinv/optim/data_fidelity.py#L232

tachella: I don't get why this wasn't failing already?

tachella: I fixed that demo

deepinv/loss/measplit.py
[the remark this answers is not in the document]
Andrewwango: Closes #238

PR #263, opened 2024-06-27
fix the gradient stopping in EILoss

### LINKED ISSUE #262: Failing to stop the gradient in EILoss
The gradient stopping in `deepinv.loss.EILoss` is currently implemented using `torch.no_grad()`. Depending on how the transformations are implemented, this might not be enough and stopping the gradient might silently fail. More precisely, if the transformation function returns its own input, using `torch.no_grad()` has no effect and the resulting tensor is not marked as being independent of the learnable weights in the backpropagation algorithm. Instead, the transformed signal should be explicitely detached from the computational graph.

**Reproducing steps**

```python
import torch
from torch.nn import Module
from deepinv.transform import Shift
from deepinv.physics import Denoising
from deepinv.models import AutoEncoder, ArtifactRemoval

import torch
import torch.nn as nn


# the current EILoss with an additional assert statement
class EILoss(Loss):
    def __init__(
        self,
        transform,
        metric=torch.nn.MSELoss(),
        apply_noise=True,
        weight=1.0,
        no_grad=False,
    ):
        super(EILoss, self).__init__()
        self.name = "ei"
        self.metric = metric
        self.weight = weight
        self.T = transform
        self.noise = apply_noise
        self.no_grad = no_grad

    def forward(self, x_net, physics, model, **kwargs):
        if self.no_grad:
            with torch.no_grad():
                x2 = self.T(x_net)
            # make sure the gradient stopping has succeeded
            assert x2.is_leaf
        else:
            x2 = self.T(x_net)

        if self.noise:
            y = physics(x2)
        else:
            y = physics.A(x2)

        x3 = model(y, physics)

        loss_ei = self.weight * self.metric(x3, x2)
        return loss_ei

# unlike other transforms it may directly return its input which makes the gradient stopping fail
class Transform(Module):
    def __init__(self):
        super().__init__()
        self.T = Shift()

    def forward(self, x):
        if torch.rand(1) < .5:
            x = self.T(x)
        return x

loss = EILoss(Transform(), no_grad=True)

backbone = AutoEncoder(dim_input=1 * 32 * 32, dim_mid=128, dim_hid=32)
model = ArtifactRemoval(backbone)
physics = Denoising()
x = torch.randn((1, 1, 32, 32))
y = physics(x)
x_net = model(y, physics)

torch.manual_seed(0)
loss(y=y, x_net=x_net, physics=physics, model=model)
print("stopping the gradient has succeeded")

try:
    torch.manual_seed(1)
    loss(y=y, x_net=x_net, physics=physics, model=model)
except:
    print("stopping the gradient has failed")
```

**Expected behavior**

Both calls to the `loss` function should **not** fail with an exception.

**Actual behavior**

The first call to the `loss` function does not fail with an exception, but the second one does.

jscanvic:
Fixes #262 

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,7 @@ New Features
 
 Fixed
 ^^^^^
+- Fixed the gradient stopping in EILoss (:gh:`263` by `Jérémy Scanvic`_) - 27/06/2024
 - Fixed averaging loss over epochs Trainer (:gh:`241` by Julian Tachella) - 11/06/2024
 - Fixed Trainer save_path timestamp problem on Windows (:gh:`245` by `Andrew Wang`_)
 
--- a/deepinv/datasets/datagenerator.py
+++ b/deepinv/datasets/datagenerator.py
@@ -156,7 +156,7 @@ def generate_dataset(
         else:
             y = physics[g](x)
 
-        #TODO save params if physics_generator is not None
+        # TODO save params if physics_generator is not None
         torch.save(physics[g].state_dict(), f"{save_dir}/physics{g}.pt")
 
         if train_dataset is not None:
--- a/deepinv/datasets/utils.py
+++ b/deepinv/datasets/utils.py
@@ -9,6 +9,7 @@ from tqdm.auto import tqdm
 from torch.utils.data import Dataset
 from torch import randn
 
+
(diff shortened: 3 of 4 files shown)


tachella (approved):
thanks for spotting this issue Jérémy - the fix looks good to me!

PR #266, opened 2024-06-27
Minor trainer docs addition

Andrewwango:
In Trainer docs, add example to show `CustomTrainer` which uses `update_parameters` in `get_samples_offline`


deepinv/training/trainer.py
```diff
@@ -70,7 +70,25 @@ def compute_loss(self, physics, x, y, train=True):
     If the user wants to change the way the metrics are computed, they can rewrite the
     :meth:`deepinv.Trainer.compute_metrics` method.
     The user can also change the way the samples are generated by rewriting the :meth:`deepinv.Trainer.get_samples_online`
-    or :meth:`deepinv.Trainer.get_samples_offline` methods.
+    or :meth:`deepinv.Trainer.get_samples_offline` methods, e.g. to change the physics parameters on-the-fly with parameters
+    from your dataset (useful for MRI where each mask has its own, different, mask):
```
matthieutrs, line 74:
```suggestion
    The user can also change the way samples are generated by overriding either the :meth:`deepinv.Trainer.get_samples_online` method or the :meth:`deepinv.Trainer.get_samples_offline` method,
    e.g. to change the physics parameters on-the-fly with parameters from the dataset.

    - Use :meth:`deepinv.Trainer.get_samples_online` when measurements are simulated from a ground truth returned by the dataloader.
    - Use :meth:`deepinv.Trainer.get_samples_offline` when both the ground truth and measurements are returned by the dataloader.

    For instance, in MRI, the dataloader often returns both the measurements and the mask associated with the measurements.
    In this case, to update the :meth:`deepinv.physics.Physics` parameters accordingly, a potential implementation would be:
```

matthieutrs (commented):
Thanks for adding this, it clarifies things! I would even be more precise in the description maybe, what do you think?

deepinv/training/trainer.py
Andrewwango: Thanks @matthieutrs, done!

tachella (approved):

PR #267, opened 2024-06-27
SplittingLoss fixes

Andrewwango:
Fixes
- Fix BernoulliMaskGenerator which was only returning zeros
- Remove noise from mask during measurement splitting
- fix demosaicing docs

Added
- SplittingLoss now handles physics with masks (Inpainting, MRI), where you want your splitting masks to be subset of your physics mask. See docs for equations
- SplittingLoss also has option `eval_split_output` which passes output through output mask during eval (Mike's idea). This makes sense for when physics input and output are same domain, e.g. Denoising. This empirically gives better results on MRI Denoising. 
- Add measurement splitting example


tachella:
@Andrewwango this looks good to me, except for a missing black fix

examples/self-supervised-learning/demo_splitting_loss.py
```diff
@@ -0,0 +1,101 @@
+r"""
+Self-supervised learning with measurement splitting
+===================================================
+
```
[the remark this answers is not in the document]
Andrewwango: Yep will do, didn't want to spend too much time doing it in case it was all wrong anyway

deepinv/physics/generator/inpainting.py
```diff
@@ -4,23 +4,23 @@
...
+    :param np.random.Generator, torch.Generator rng: random number generator.
     """
 
     def __init__(
         self,
         tensor_size: Tuple,
         split_ratio: float,
         device: torch.device = torch.device("cpu"),
-        rng: np.random.Generator = np.random.default_rng(),
+        rng: torch.Generator = torch.Generator().manual_seed(0),
```
tachella, line 23: this is generating some device issues when we test with gpu

deepinv/loss/measplit.py
```diff
@@ -121,42 +139,84 @@ def adapt_model(self, model, MC_samples=5):
...
+                        1.0 if not hasattr(physics, "mask") else physics.mask
+                    ) - mask["mask"]
+                    inp2 = Inpainting(tsize, mask=mask2, device=y.device)
+                    x_hat = inp2.A(x_hat)
+                    normaliser += mask2
+
+                out += x_hat
+
+            if eval_split_output:
+                out[normaliser != 0] /= normaliser[normaliser != 0]
```
tachella, line 213: if normaliser == 0, it means that out will have a hole...
we should raise some warning at least I guess

deepinv/physics/generator/inpainting.py
```diff
@@ -30,15 +30,31 @@ def __init__(
...
         Create a bernoulli mask.
 
         :param int batch_size: batch_size.
+        :param torch.Tensor, None input_mask: optional mask to be split. If None, all pixels are considered. If not None, only pixels where mask==1 are considered.
         :return: dictionary with key **'mask'**: tensor of size (batch_size, *tensor_size) with values in {0, 1}.
         :rtype: dict
         """
-        mask = torch.zeros((batch_size, *self.tensor_size), device=self.device)
-        mask[torch.rand_like(mask) > self.split_ratio] = 0
+        if not isinstance(input_mask, torch.Tensor) or tuple(input_mask.shape) != (
```
tachella, line 42: the second condition is not right I believe, what if there is a single mask (batch dim=1)?

deepinv/loss/measplit.py
Andrewwango: What do you mean? Wherever normaliser==0, out also ==0 so the output will just be 0.

tachella: yes this is what I meant - if the output is zero it will be probably be a bad output

Andrewwango:
Note: `eval_split_input` option added as [Yaman](https://github.com/byaman14/SSDU/blob/main/test.py#L33) doesn't perform measplit at eval-time

deepinv/loss/measplit.py
```diff
@@ -111,52 +141,126 @@ def adapt_model(self, model, MC_samples=5):
...
-            for i in range(MC):
-                mask = self.mask_generator.step(y.size(0))
-                y1 = inp.A(y, **mask)
-                physics1 = inp * physics  # A_1 = P*A
-                out += self.model(y1, physics1)
+
+            if not self.eval_split_input and not self.training:
+                return self.model(y, physics)
+            elif self.eval_split_output and self.eval_split_input and not self.training:
+                return self._forward_split_output(y)
```
tachella, line 220: _forward_split_output(y, physics) should solve the test issue

deepinv/loss/measplit.py
```diff
@@ -111,52 +141,126 @@ def adapt_model(self, model, MC_samples=5):
...
+                    # Forward pass
+                    out += self.model(y1, physics1) / MC_samples
 
             if self.training and update_parameters:
                 self.mask = mask["mask"]
-            out = out / MC
+
+        return out
+
+    def _forward_split_output(self, y: torch.Tensor):
```
tachella, line 239: _forward_split_output(self, y: torch.Tensor, physics: Physics):

examples/self-supervised-learning/demo_splitting_loss.py
```diff
@@ -0,0 +1,177 @@
...
+
+# %%
+# Furthermore, we can disable measurement splitting at evaluation
+# altogether by setting ``eval_split_input`` to False (this is done in
+# `SSDU <https://pubmed.ncbi.nlm.nih.gov/32614100/>`__):
+#
+
+model.eval_split_input = False
+trainer.test(test_dataloader)
```
tachella, line 177: should we add a warning saying that we expect that the difference between these options is reduced if we work with bigger images?

deepinv/loss/measplit.py
tachella, line 220:
```suggestion
                return self._forward_split_output(y, physics)

```

deepinv/loss/measplit.py
tachella, line 239:
```suggestion
    def _forward_split_output(self, y: torch.Tensor, physics: Physics):

```

PR #268, opened 2024-06-27
Solve problem with non-float parameters for Unfolding

samuro95:
- Very simple correction to change the input params to float tensors
- Add a code to load an image from its path. 


tachella (changes requested):
The PR looks good but there is an intruder file 'IntroDeepInverse.ipynb' to remove,

It would also be good to add the new load_image function in the docs [here](https://github.com/deepinv/deepinv/blob/4a4b8ebf40a7470a12f25e171135e27d69828cd5/docs/source/deepinv.utils.rst?plain=1#L88).

tachella (approved):
thanks for the fixes!

PR #269, opened 2024-06-28
Resolve bug in data_fidelity

johertrich:
As mentioned by @zeljkozeljko: Currently the code
```
import torch
import deepinv

phantom=torch.randn(1,1,128,128)
physics=deepinv.physics.Tomography(100,128)
obs=physics(phantom)
forward=deepinv.optim.L2()
out=forward(phantom,obs,physics)
print(out)
```
raises the following error:
```
  File ".../deepinv/optim/data_fidelity.py", line 269, in d
    d = 0.5 * torch.norm(x.view(x.shape[0], -1), p=2, dim=-1) ** 2
                         ^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: view size is not compatible with input tensor's size and stride (at least one dimension spans across two contiguous subspaces). Use .reshape(...) instead.

```
The same happens for the data-fidelity terms `L1`, `AmplitudeLoss` and `IndicatorL2`. The reason is that the output of `physics` might not be contiguous.
The PR fixes it by replacing the `view` command by `reshape` (which can take non-contiguous tensors).

### DIFF
--- a/deepinv/optim/data_fidelity.py
+++ b/deepinv/optim/data_fidelity.py
@@ -266,7 +266,7 @@ class L2(DataFidelity):
         :return: (torch.Tensor) data fidelity :math:`\datafid{u}{y}` of size `B` with `B` the size of the batch.
         """
         x = u - y
-        d = 0.5 * torch.norm(x.view(x.shape[0], -1), p=2, dim=-1) ** 2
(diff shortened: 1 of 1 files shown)


samuro95 (commented):
Thanks for spotting the mistake. Looks all good.

samuro95 (approved):

PR #270, opened 2024-06-28
LIDC-IDRI dataset

MaximeSongIdris:

### DIFF
--- a/deepinv/datasets/lidc_idri.py
+++ b/deepinv/datasets/lidc_idri.py
@@ -5,10 +5,14 @@ from typing import (
     Optional,
 )
 import os
-from pydicom import dcmread
 import torch
 import numpy as np
 import pandas as pd
+
+try:
+    from pydicom import dcmread
+except:
+    error_import = ImportError("dicom is not available. Please install the dicom package with `pip install dicom`.")
 
 
 class LidcIdriSliceDataset(torch.utils.data.Dataset):
@@ -17,10 +21,11 @@ class LidcIdriSliceDataset(torch.utils.data.Dataset):
     | The Lung Image Database Consortium image collection (LIDC-IDRI) consists
     | of diagnostic and lung cancer screening thoracic computed tomography (CT)
     | scans with marked-up annotated lesions.
-    | It is a web-accessible international resource for development, training,
-    | and evaluation of computer-assisted diagnostic (CAD) methods for lung
-    | cancer detection and diagnosis.
-    | To download the dataset, you will need to install the `NBIA Data Retriever <https://wiki.cancerimagingarchive.net/display/NBIA/Downloading+TCIA+Images>`_.
+
+    .. warning::
+        To download the raw dataset, you will need to install the `NBIA Data Retriever <https://wiki.cancerimagingarchive.net/display/NBIA/Downloading+TCIA+Images>`_,
+        then download the manifest file (.tcia file)`here <https://www.cancerimagingarchive.net/collection/lidc-idri/>`_, and open it by double clicking. 
+
 
     **Raw data file structure:** ::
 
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -31,6 +31,8 @@ dependencies = [
     "wandb",
     "scipy",
     "datasets",
+    "pandas",
+    "pydicom",
```
tachella, line 35: I would move pydicom as an optional dependency, similar to [here](https://github.com/deepinv/deepinv/blob/7c282240a8088a56e304708731516a2132ac1946/deepinv/models/bm3d.py#L6)

deepinv/datasets/lidc_idri.py
```diff
@@ -0,0 +1,139 @@
...
+import pandas as pd
+
+
+class LidcIdriSliceDataset(torch.utils.data.Dataset):
+    """Dataset for `LIDC-IDRI <https://www.cancerimagingarchive.net/collection/lidc-idri/>`_ that provides access to CT image slices.
+
+    | The Lung Image Database Consortium image collection (LIDC-IDRI) consists
+    | of diagnostic and lung cancer screening thoracic computed tomography (CT)
+    | scans with marked-up annotated lesions.
+    | It is a web-accessible international resource for development, training,
```
tachella, line 20: I would remove this paragraph "It is a web-accessible international resource for dev ..."

since the usage in the library is mostly focused on reconstructing the CT scans

deepinv/datasets/lidc_idri.py
```diff
@@ -0,0 +1,139 @@
...
+    | and evaluation of computer-assisted diagnostic (CAD) methods for lung
+    | cancer detection and diagnosis.
+    | To download the dataset, you will need to install the `NBIA Data Retriever <https://wiki.cancerimagingarchive.net/display/NBIA/Downloading+TCIA+Images>`_.
```
tachella, line 23: I would give more detailed instructions regarding the download and set up, and add a warning box (.. warning: ) saying that this class does not download the dataset, and it is up to the user to download the raw files.

deepinv/datasets/lidc_idri.py
```diff
@@ -0,0 +1,139 @@
...
+
+        self.root --- LIDC-IDRI --- LICENCE
+                   |             -- LIDC-IDRI-0001 --- `STUDY_UID` --- `SERIES_UID` --- xxx.xml
+                   |             |                                                   -- 1-001.dcm
+                   |             -- LIDC-IDRI-1010                                   |
+                   |                                                                 -- 1-xxx.dcm
+                   -- metadata.csv
+
+    :param str root: Root directory of dataset. Directory path from where we load and save the dataset.
+    :param callable, optional transform: TODO
```
tachella, line 35: finish TODO

deepinv/datasets/lidc_idri.py
```diff
@@ -0,0 +1,139 @@
...
+
```
tachella, line 36: How are the slices loaded? Please give more information here. What are the sizes of the slices?

tachella (changes requested):
The PR looks good, but I think we need to give a more detailed explanation on 

i) how to download the raw data 
ii) what is the dataloader giving (in terms of slices, image sizes, etc.)

pyproject.toml
MaximeSongIdris: Under a list of optional dependencies for everything related to datasets ?

deepinv/datasets/lidc_idri.py
MaximeSongIdris: we open a dicom file with pydicom and we retrieve the "pixel_array" field

MaximeSongIdris: ct scan folders are found thanks to the metadata.csv
and in these folders, we have the dicom files

pyproject.toml
tachella: yes good idea

tachella (approved):

PR #279, opened 2024-07-17
Splitting loss new demo, SSDU, Artifact2Artifact, Phase2Phase

Andrewwango:
- Change splitting loss demo to use Tomography
- Add Gaussian splitting mask for 2D and 2D+t splitting (for Yaman et al. (SSDU) and Acar et al.)
- Add Artifact2Artifact (from [RARE](https://arxiv.org/abs/1912.05854)) and [Phase2Phase](https://journals.lww.com/investigativeradiology/pages/articleviewer.aspx?year=2021&issue=12000&article=00004&type=Fulltext)
- Fix metrics bug in Trainer
- Add time-agnostic network wrapper


deepinv/tests/test_models.py
```diff
@@ -560,3 +560,11 @@ def test_optional_dependencies(denoiser, dep):
 #     mse_out = (x_net - x).pow(2).mean()
 #
 #     assert mse_out < mse_in
+
+
+def test_time_agnostic_net():
+    backbone_net = dinv.models.UNet(scales=2)
+    net = dinv.models.TimeAgnosticNet(backbone_net)
+    y = torch.rand(1, 1, 4, 8, 8)  # B,C,T,H,W
```
tachella, line 568:
```suggestion
    y = torch.rand(1, 1, 2, 4, 4)  # B,C,T,H,W
```

deepinv/training/testing.py
```diff
@@ -143,18 +145,27 @@ def test(
...
-            if hasattr(physics_cur, "A_adjoint"):
-                if isinstance(physics_cur, torch.nn.DataParallel):
-                    x_init = physics_cur.module.A_adjoint(y)
-                else:
-                    x_init = physics_cur.A_adjoint(y)
-            elif hasattr(physics_cur, "A_dagger"):
-                if isinstance(physics_cur, torch.nn.DataParallel):
-                    x_init = physics_cur.module.A_dagger(y)
+            phys = (
+                physics_cur.module
```
tachella, line 149: I'm going to integrate all this in future PR

tachella (changes requested):
Many thanks for the PR!

I think most of my comments were about esthetic changes, except for

- Cloning the mask in the splittingloss to avoid torch backward issues
- Adapting splitting loss to different measurement sizes at eval time
- We should also test the new splitting-based losses (Phase2Phase and Artifact2Artifact).

deepinv/training/testing.py
Andrewwango: Should I just leave this as is and then this will be overwritten in #280

deepinv/loss/measplit.py
```diff
@@ -209,8 +242,6 @@ def forward(
                 device=y.device,
             )
```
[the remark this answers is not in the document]
Andrewwango: I've added `self.mask_generator.tensor_size != y.size()[1:]` as a condition:
`if self.mask_generator is None or self.mask_generator.tensor_size != y.size()[1:]`
but not added the rest of the suggested code as it's the same as before. Can you check this is ok @bemc22 ?

tachella (approved):

PR #280, opened 2024-07-19
Tester integration, Noise2Score, python version 3.9

tachella:
- homogeinizing python versions to 3.9+
- small fixes to gaussian noise to use with SigmaGenerator
- Added Gamma Noise
- Added Score Loss (there was one existing version but it was hidden from the API and didn't work)
- Moved all testing to Trainer
- Added no-learning reconstruction to Trainer.
- SURE improvements
- Added ideal sinc filter with 'optimal' kaiser windowing.
- Added windows machine in github actions testing 
- fix wandb logging issues

### DIFF
--- a/deepinv/loss/measplit.py
+++ b/deepinv/loss/measplit.py
@@ -108,6 +108,7 @@ class SplittingLoss(Loss):
         """
         mask = model.get_mask()
 
+        mask *= getattr(physics, "mask", 1.0)
         # create inpainting masks
         mask2 = getattr(physics, "mask", 1.0) - mask
         inp2 = Inpainting(y.size()[1:], mask=mask2, device=y.device)
--- a/deepinv/loss/measplit.py
+++ b/deepinv/loss/measplit.py
@@ -235,7 +235,9 @@ class SplittingModel(torch.nn.Module):
                     out += self.model(y1, physics1) / MC_samples
 
             if self.training and update_parameters:
-                self.mask = mask["mask"].clone() # clone is needed to avoid in-place operation
+                self.mask = mask[
+                    "mask"
+                ].clone()  # clone is needed to avoid in-place operation
 
         return out
 
--- a/deepinv/loss/measplit.py
+++ b/deepinv/loss/measplit.py
@@ -210,8 +210,6 @@ class SplittingModel(torch.nn.Module):
                 device=y.device,
             )
 
-        inp = Inpainting(y.size()[1:], device=y.device)
-
         with torch.set_grad_enabled(self.training):
 
             if not self.eval_split_input and not self.training:
@@ -221,6 +219,8 @@ class SplittingModel(torch.nn.Module):
 
             else:
                 MC_samples = 1 if self.training else self.MC_samples
+
+                inp = Inpainting(y.size()[1:], device=y.device)
 
                 for _ in range(MC_samples):
                     # Perform input masking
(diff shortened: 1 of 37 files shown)


deepinv/loss/measplit.py
```diff
@@ -108,6 +108,7 @@ def forward(self, x_net, y, physics, model, **kwargs):
         """
         mask = model.get_mask()
 
+        mask *= getattr(physics, "mask", 1.0)
```
Andrewwango, line 111: What is the reasoning here? If a `physics.mask` is present (e.g. in MRI), `mask` will already be subsampled from `physics.mask` (in the `SplittingModel.forward`), so this line does nothing.

tachella: it is a safety line for the case where the second mask is not subsampled.

Andrewwango: I have added this in #279 now.

deepinv/loss/score.py
```diff
@@ -1,41 +1,161 @@
...
 from deepinv.loss.loss import Loss
 
 
 class ScoreLoss(Loss):
     r"""
-    Approximates score of a distribution.
+    Learns score of noise distribution.
 
-    It approximates the score of the measurement distribution :math:`S(y)\approx \nabla \log p(y)`
+    Approximates the score of the measurement distribution :math:`S(y)\approx \nabla \log p(y)`
```
samuro95, line 10: isn't it p_\sigma rather than p ?

deepinv/loss/score.py
```diff
@@ -1,41 +1,161 @@
...
     https://proceedings.neurips.cc/paper_files/paper/2021/file/077b83af57538aa183971a2fe0971ec1-Paper.pdf.
 
     The score loss is defined as
 
     .. math::
 
-        \| \epsilon - \sigma S(y+ \sigma \epsilon) \|^2
+        \| \epsilon + \sigma S(y+ \sigma \epsilon) \|^2
```
samuro95, line 17: This is the same loss for all noise models ? or is it only valid for Gaussian noise ?

deepinv/loss/score.py
```diff
@@ -1,41 +1,161 @@
...
+        :param torch.nn.Module model: Backbone model approximating the score.
+        :return: (torch.nn.Module) Adapted reconstruction model.
+        """
+        if isinstance(model, ScoreModel):
+            return model
+        else:
+            return ScoreModel(model, self.noise_model, self.delta, self.total_batches)
+
+
+class ScoreModel(torch.nn.Module):
```
samuro95, line 106: Missing some docstring for the arguments

deepinv/loss/score.py
tachella: yes, but they anneal sigma to zero during training.

deepinv/loss/score.py
```diff
@@ -1,41 +1,161 @@
...
-        :return: (torch.Tensor) loss.
+        :param torch.Tensor y: Measurements.
+        :param deepinv.physics.Physics physics: Forward operator associated with the measurements.
+        :param torch.nn.Module model: Reconstruction model.
+        :return: (torch.Tensor) Score loss.
         """
-        std = np.randn() * self.delta
-        noise = torch.randn_like(y)
-        return self.metric(noise, std * model(std * noise + y))
+        return model.get_error()
```
samuro95, line 91: Can we add a test to ensure that we indeed computer the error beforehand ?

tachella: do you mean a pytest? or a flag that makes sure that the error has been recomputed?

deepinv/loss/score.py
tachella: now it should be fine right?

Issue #283, opened 2024-07-19
`datasets` ImportError


Andrewwango:
If `datasets` is to be an optional requirement, then its import should be made optional in the datasets that use it. Otherwise it should be made a necessary requirement (although I would rather not, to keep requirements to a minimum).
https://github.com/deepinv/deepinv/blob/b22094cddb431294db3ed0ae90fa121c33436aed/deepinv/datasets/cbsd68.py#L4

MaximeSongIdris:
datasets is necessary in cbsd68, but it is very specific to it
so the idea was to make `datasets` a optional module which can be installed along with other optional modules with
"pip install .[datasets]"

MaximeSongIdris:
https://github.com/deepinv/deepinv/pull/286

I made an explicit error message if there is an import error, what do you think ?

PR #286, opened 2024-07-24
add explicit msg for import error for optional module needed in some …

MaximeSongIdris:
[…datasets](https://github.com/deepinv/deepinv/issues/283)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -48,6 +48,16 @@ You can also install the latest version of ``deepinv`` directly from github:
 
     pip install git+https://github.com/deepinv/deepinv.git#egg=deepinv
 
+You can also install additional dependencies needed for some modules in deepinv.datasets and deepinv.models:
+
+.. code-block:: bash
+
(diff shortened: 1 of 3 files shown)


Andrewwango:
Can we not just make `datasets` optional completely e.g. by putting it in the CBSD68 init? Because this will still result in an error if I want to use deepinv but not these datasets and don't want to install `datasets`

tachella (commented):
@Andrewwango is this solving the issues you were having?

MaximeSongIdris:
@Andrewwango 
yes I prefer to have the import at the start of the file, the diff is that I stored the import error in a variable which will be called in __init__ (so the import error will be raised when instantiating the class) and it shouldn't cause any issue with the library if you don't call the class

also added in README, command line to install the library with the optional dependencies

Andrewwango (approved):
Looks good, thanks

Andrewwango:
Closes #283

PR #302, opened 2024-09-11
Remove unecessary gradient computation

### LINKED ISSUE #301: Memory leak
The memory usage explodes over the iterations on some of the examples from the documentation.
The issue can notably be reproduced on the [random phase retrieval example](https://deepinv.github.io/deepinv/auto_examples/basics/demo_phase_retrieval.html#sphx-glr-auto-examples-basics-demo-phase-retrieval-py), and makes it impossible to run some algorithms past a few iterations on larger images.

<details>
<summary>Click to reveal code used to reproduce the issue</summary>
<br>

```python
import deepinv as dinv
import torch
from deepinv.models import DRUNet
from deepinv.models.complex import to_complex_denoiser
from deepinv.optim.data_fidelity import L2
from deepinv.optim.optimizers import optim_builder
from deepinv.optim.prior import PnP
from deepinv.utils.demo import get_image_url, load_url_image

# Set global random seed to ensure reproducibility.
torch.manual_seed(0)
device = "cpu" # Issue is present both on cpu and CUDA devices

img_size = 32
url = get_image_url("SheppLogan.png")
x = load_url_image(
    url=url, img_size=img_size, grayscale=True, resize_mode="resize", device=device
)
x_phase = torch.exp(1j * x * torch.pi - 0.5j * torch.pi)

# Define physics information
oversampling_ratio = 5.0
m = int(oversampling_ratio * torch.prod(torch.tensor(x.shape[1:])))
noise_level_img = 0.05  # Gaussian Noise standard deviation for the degradation
n_channels = 1  # 3 for color images, 1 for gray-scale images

# Create the physics
physics = dinv.physics.RandomPhaseRetrieval(
    m=m,
    img_shape=x.shape[:1],
    noise_model=dinv.physics.GaussianNoise(sigma=noise_level_img),
    device=device,
)

# Generate measurements
y = physics(x_phase)

# Load the pre-trained denoiser
denoiser = DRUNet(
    in_channels=n_channels,
    out_channels=n_channels,
    pretrained="download",
    device=device,
)
# The original denoiser is designed for real-valued images, so we need to convert it to a complex-valued denoiser for phase retrieval problems.
denoiser_complex = to_complex_denoiser(denoiser, mode="abs_angle")

# Instantiate the algorithm class to solve the IP problem.
model = optim_builder(
    iteration="PGD",
    prior=PnP(denoiser=denoiser_complex),
    data_fidelity=L2(),
    early_stop=True,
    max_iter=500,
    verbose=True,
    params_algo={"stepsize": 0.30, "g_param": 0.04},
)
model.fixed_point.show_progress_bar = True

# Run the algorithm
x_phase_pnp, metrics = model(y, physics, x_gt=x_phase, compute_metrics=True)
```

</details>

leak
_Memory usage of the random phase retrieval example with PnP & DRUNet, measured using memory-profiler_

The issue seems to be caused by gradients leaking : the problem isn't present when running the algorithm within a `no_grad` context manager.

I suspect the problem is caused by [line 245](https://github.com/deepinv/deepinv/blob/e0f479209c509651f202e705397e15e25668a132/deepinv/optim/fixed_point.py#L245) of `FixedPoint.forward`, as replacing the code there with the following seems to solve the problem. 
```python
X_prev = X 
# New code below 
X_prev["est"] = [x.detach() for x in X_prev["est"]]
```
no_leak
_Memory usage when changing the code to the snippet above_

I'm using `deepinv==0.2.1`, `python==3.12.5` and `torch==2.4.1`.

Seon82:
Fixes #301. As far as I can tell, the autograd graph is never needed anywhere after `OptimIterator.forward`.

Would there be a better way to do this? It would seem simpler and more efficient to wrap the entirety of `OptimIterator.forward` in a `no_grad` context manager, but I'm not familiar enough with the codebase to know whether that could have some unwanted side effects.

- [X] `python3 -m pytest tests/` runs successfully.
- [X] Added an entry to the [CHANGELOG.rst](../CHANGELOG.rst).


tachella:
Thanks for spotting this - @samuro95 / @matthieutrs can you have a look?

samuro95:
Thanks ! I think that model.eval() should solve the issue.

samuro95:
Thanks for noticing this important issue. I propose the following workaround: add a torch.no_grad() in the forward of the class BaseOptim and keep it unchanged in the forward of the class BaseUnfold

examples/unfolded/demo_custom_prior_unfolded.py
```diff
@@ -249,7 +249,8 @@ def g(x, *args, **kwargs):
 
 # Get the measurements and the ground truth
 y = physics(test_sample)
-rec = model(y, physics=physics)
+with torch.no_grad():
```
tachella, line 252: why do we need this?

samuro95: because in the forward of unfolded, we can not put a torch.no_grad(), so we need a workaround...

examples/plug-and-play/demo_vanilla_PnP.py
```diff
@@ -133,19 +133,19 @@
...
-    imgs,
-    titles=["Input", "GT", "Linear", "Recons."],
-    save_dir=RESULTS_DIR / "images",
-    show=True,
-)
-
-# plot convergence curves. Metrics are saved in RESULTS_DIR.
-if plot_convergence_metrics:
-    plot_curves(metrics, save_dir=RESULTS_DIR / "curves", show=True)
+# # compute PSNR
```
tachella, line 136: why is this code commented out? Should we remove it?

samuro95: oups my bad

PR #308, opened 2024-09-17
Refactor PatchDataset class

apolychronou:
Refactor PatchDataset class to calculate patches taking into consideration stride too.
Let transform be used on the returned tensor if not None (def None).
Be able to choose the shape of the returned tensor. None returns same C x H x W as input (def (-1,) ).

New init args:
1.  -  shape arg for the shape of the output tensor. 
2. -  stride arg for the window stride.


tachella:
@johertrich

Since you contributed the PatchDataset, I was wondering if you could review this PR or give your opinion?
Many thanks!

deepinv/datasets/patch_dataset.py
```diff
@@ -7,28 +7,31 @@ class PatchDataset(data.Dataset):
 
     :param torch.Tensor imgs: Tensor of images, size: batch size x channels x height x width
     :param int patch_size: size of patches
-    :param callable: data augmentation. callable object, None for no augmentation.
+    :param callable transforms: data augmentation. callable object, None for no augmentation.
+    :param tuple shape: shape of the returned tensor. None returns original C x H x W
```
johertrich, line 11: "None returns original C x H x W" I find this comment a bit confusing. We have H=W=patch_size right? Maybe you could write "None returns C x patch_size x patch_size"?

johertrich (changes requested):
Looks good to me. I have just a minor comment regarding the docs.

deepinv/datasets/patch_dataset.py
apolychronou: Changed that!

johertrich (approved):
Many thanks, looks fine to me.

tachella (approved):

PR #309, opened 2024-09-17
Unify metrics

### LINKED ISSUE #299: Consolidate metrics
- Consolidate metrics, which are currently spread across `utils.metric` and `loss.metric`.
- Put functions e.g. normalisation, mse into a `functional` module. 
- Let the metric classes be able to be used as losses (i.e. `l(x, x_net, *args, **kwargs)` or as metrics (`l(estimate, target)`)
- Consolidate normalisations before metrics - all metrics should be able to have normalisation e.g. L2 norm, L1 norm, min-max, mean-std (standardise) - `Metric` baseclass??

Andrewwango:
Closes #299. Unify metrics in one submodule. Note there are a lot of changed files from metrics being moved from `utils` into `loss.metric`.

- Add `Metric` baseclass that all metrics should inherit from
- All metrics can perform preprocessing (complex abs, norm) and postprocessing (training, reduce)
- Metrics can be used either as metrics, or as losses (e.g. replacing `torch.nn.MSELoss()`)
- Unit tests for metrics
- Add L1L2 metric


deepinv/loss/metric/metric.py
```diff
@@ -0,0 +1,158 @@
...
+        import pyiqa
+
+        return pyiqa
+    except ImportError:
+        raise ImportError(
+            "Metric not available. Please install the pyiqa package with `pip install pyiqa`."
+        )
+
+
+class Metric(Loss):
```
tachella, line 27: Do we want to inherit from Loss?

I am afraid that we end up with some circular patterns, eg EILoss using a metric

deepinv/loss/metric/metric.py
```diff
@@ -0,0 +1,158 @@
...
+    r"""
+    Base class for metrics.
+
+    See docs for ``forward()`` below for more details.
+
+    To create a new metric, inherit from this class and override the function :meth:`deepinv.loss.metric.Metric.metric`.
+
+    :param bool complex_abs: perform complex magnitude before passing data to metric function. If ``True``,
+        the data must either be of complex dtype or have size 2 in the channel dimension (usually the second dimension after batch).
+    :param bool train_loss: use metric as a training loss, by returning one minus the metric. If lower is better, does nothing.
```
tachella, line 37: I'm not sure if we should assume that metrics are always between 0 and 1, with 1 being bad.
I would remove this option.

deepinv/loss/metric/perceptual.py
```diff
@@ -0,0 +1,77 @@
...
+    Computes the perceptual similarity between two images, based on a pre-trained deep neural network.
+    Uses implementation from `pyiqa <https://pypi.org/project/pyiqa/>`_.
+
+    See docs for ``forward()`` below for more details.
+
+    :Example:
+
+    >>> from deepinv.utils.demo import get_image_url, load_url_image
+    >>> from deepinv.loss.metric import LPIPS
+    >>> ();m = LPIPS();() # doctest: +ELLIPSIS
```
tachella, line 17: what is this line doing?

deepinv/loss/scheduler.py
```diff
@@ -89,8 +89,9 @@ class RandomLossScheduler(BaseLossScheduler):
     :Example:
 
     >>> import torch
-    >>> from deepinv.loss import RandomLossScheduler, SupLoss, SSIM
-    >>> l = RandomLossScheduler(SupLoss(), SSIM()) # Choose randomly between Sup and SSIM
+    >>> from deepinv.loss import RandomLossScheduler, SupLoss
+    >>> from deepinv.loss.metric import SSIM
+    >>> l = RandomLossScheduler(SupLoss(), SSIM(train_loss=True)) # Choose randomly between Sup and SSIM
```
tachella, line 94: see issue with train_loss option above - I'm okay with having a special kwarg for SSIM

deepinv/tests/test_metric.py
```diff
@@ -0,0 +1,152 @@
...
+        get_image_url("celeba_example.jpg"), img_size=128, resize_mode="resize"
+    )
+
+    if complex_abs:
+        x = x[:, :2, ...]
+
+    x_hat = GaussianNoise(sigma=0.1, rng=rng)(x)
+
+    # Test metric worse when image worse
+    if train_loss:  # All metrics lower = better
```
tachella, line 61: some metrics can have higher=better, right?

deepinv/training/trainer.py
```diff
@@ -137,7 +138,7 @@ def get_samples_offline(self, iterators, g):
     :param None, deepinv.physics.generator.PhysicsGenerator physics_generator: Optional physics generator for generating
         the physics operators. If not None, the physics operators are randomly sampled at each iteration using the generator.
         Should be used in conjunction with ``online_measurements=True``. Also see ``loop_physics_generator``.
-    :param deepinv.loss.Loss, list[deepinv.loss.Loss] metrics: Metric or list of metrics used for evaluating the model.
+    :param deepinv.loss.metric.Metric, list[deepinv.loss.metric.Metric] metrics: Metric or list of metrics used for evaluating the model.
```
tachella, line 141: can we have deepinv.loss.Metric? otherwise I feel is a bit verbose

docs/source/deepinv.loss.rst
```diff
@@ -93,19 +93,58 @@ Training is implemented using :class:`deepinv.training.AdversarialTrainer` which
 
 Metrics
 --------
-Metrics are generally used to evaluate the performance of a model. Some of them can be used as training losses as well.
+Metrics are generally used to evaluate the performance of a model.
+
+Metrics inherit from the :class:`deepinv.loss.metric.Metric` baseclass, and take either ``x_net, x``
```
tachella, line 98: deepinv.loss.Metric

docs/source/deepinv.loss.rst
```diff
@@ -93,19 +93,58 @@ Training is implemented using :class:`deepinv.training.AdversarialTrainer` which
...
    :template: myclass_template.rst
    :nosignatures:
 
-        deepinv.loss.MSE
-        deepinv.loss.NMSE
-        deepinv.loss.PSNR
-        deepinv.loss.SSIM
-        deepinv.loss.LPIPS
-        deepinv.loss.NIQE
+        deepinv.loss.metric.Metric
```
tachella, line 139: not sure this renaming is that useful - what is the idea behind?

tachella: I worry that it is not backwards compatible

examples/plug-and-play/demo_vanilla_PnP.py
```diff
@@ -134,8 +134,8 @@
 )  # reconstruction with PnP algorithm
 
 # compute PSNR
-print(f"Linear reconstruction PSNR: {dinv.utils.metric.cal_psnr(x, x_lin):.2f} dB")
-print(f"PnP reconstruction PSNR: {dinv.utils.metric.cal_psnr(x, x_model):.2f} dB")
+print(f"Linear reconstruction PSNR: {dinv.metric.PSNR()(x, x_lin).item():.2f} dB")
```
tachella, line 137: shouldn't be dinv.loss.metric.PSNR?

tachella (commented):
I have some questions regarding the refactoring:

1. Do we want `Metric` to inherit from `Loss`? I'm worried there is some circularity here.
2. I'm not sure it is a good idea to refactor all metrics to `deepinv.loss.metric`. This is not backwards compatible and I'm not sure if it clarifies things much. If go down the non-backwards compatible path, it could directly be `deepinv.metric`.
3. I'm not sure regarding whether all metrics should verify lower = better, eg SSIM and MSE are opposite.

deepinv/loss/metric/perceptual.py
Andrewwango: It's because pyiqa is annoyingly written (they use conditional print statements in their code instead of warnings) + doctest is a bit fiddly so this is the only hacky way to capture the print output

deepinv/tests/test_metric.py
Andrewwango: i've added a comment to explain this in the test

deepinv/loss/metric/metric.py
Andrewwango: I've now changed it such that the train_loss option is only visible in the docs for SSIM, however I have kept it as an option for all metrics, so that it's very easy to implement for future metrics which would also require it. I have generalised it by adding a invert_metric method which can be overridden (e.g. for SSIM, it should be 1-m). I've made this clear in the docstrings.

deepinv/loss/metric/metric.py
Andrewwango: Metric now no longer inherits from loss

deepinv/training/trainer.py
Andrewwango: You can either import from `deepinv.loss.Metric`, `deepinv.metric.Metric`, or `deepinv.loss.metric.Metric`. These are all tested in the `test_metrics` tests. This means in most cases it remains backwards compatible.

docs/source/deepinv.loss.rst
Andrewwango: See above that you can import either `deepinv.metric.Metric`, `deepinv.loss.Metric`or `deepinv.loss.metric.Metric`. I've added a line explaining this. However annoyingly it doesn't create the link if I shorten it, so it will have to be verbose

docs/source/deepinv.loss.rst
Andrewwango: see above

docs/source/deepinv.loss.rst
Andrewwango: renaming them also emphasises that these are metrics, not normal losses, i.e. you can't use physics and model with them

examples/plug-and-play/demo_vanilla_PnP.py
Andrewwango: you can import directly with dinv.metric.PSNR, see above

Andrewwango:
Thanks for your review @tachella . Re. your comments:
1. I have removed Metric inheriting from Loss, and now Metric expects `(x_net, x=None, y=None)`
2. All metrics can be imported from `deepinv.loss.metric`, `deepinv.metric`, or `deepinv.loss`, so we will have backwards compatibility. For the old functions in `deepinv.utils.metric`, I have kept them and added deprecation errors.
3. `lower_better` is an attribute that should be defined in all implementations of metrics, this is what they do in `pyiqa` too. When `lower_better=True` (most of the metrics), it doesn't affect anything. However when `lower_better=False`, as is the case in SSIM, the inverse is taken when `train_loss=True` is passed. I have made this clearer in the docs + docstrings.

tachella (changes requested):
Many thanks for the refactoring. 

-  now that metric is a separate class from loss, I would move the metric code to a new folder, outside loss.

- some docstrings refer to the forward method but there is no specific forward method for those classes. I would incorporate a short description of the metric instead.

deepinv/models/dip.py
```diff
@@ -1,7 +1,7 @@
 import torch
 import torch.nn as nn
 import numpy as np
-from deepinv.loss import MCLoss
+from deepinv.loss.mc import MCLoss
```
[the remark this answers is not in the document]
Andrewwango: fixed

deepinv/loss/metric/distortion.py
```diff
@@ -0,0 +1,401 @@
...
+
+class QNR(Metric):
+    r"""
+    Quality with No Reference (QNR) metric for pansharpening.
+
+    QNR was proposed in Alparone et al., "Multispectral and Panchromatic Data Fusion Assessment Without Reference".
+
+    Note we don't use the torchmetrics implementation.
+
+    See docs for ``forward()`` below for more details.
```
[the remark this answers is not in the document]
Andrewwango: done

deepinv/loss/metric/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .metric import Metric
```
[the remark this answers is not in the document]
Andrewwango: done

deepinv/loss/metric/distortion.py
```diff
@@ -0,0 +1,401 @@
...
+
+    def metric(self, x_net, x, *args, **kwargs):
+        return cal_mae(x_net, x)
+
+
+class MSE(Metric):
+    r"""
+    Mean Squared Error metric.
+
+    See docs for ``forward()`` below for more details.
```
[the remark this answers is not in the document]
Andrewwango: done

deepinv/loss/metric/distortion.py
```diff
@@ -0,0 +1,401 @@
...
+        return cal_mse(x_net, x)
+
+
+class NMSE(MSE):
+    r"""
+    Normalised Mean Squared Error metric.
+
+    Normalises MSE by the L2 norm of the ground truth ``x``.
+
+    See docs for ``forward()`` below for more details.
```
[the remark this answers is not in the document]
Andrewwango: done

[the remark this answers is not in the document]
Andrewwango: done

deepinv/utils/plotting.py
```diff
@@ -211,7 +212,8 @@ def plot(
 
     :param list[torch.Tensor], torch.Tensor img_list: list of images to plot or single image.
     :param list[str] titles: list of titles for each image, has to be same length as img_list.
-    :param None, str, Path save_dir: path to save the plot.
+    :param None, str, Path save_fn: path to save the plot as a single image.
```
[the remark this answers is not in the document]
Andrewwango: I think it's useful, I've added an explanation in the docs

tachella (approved):

PR #310, opened 2024-09-23
[WIP] Champagne algorithms

Seon82:
Hello ! I'd like to contribute a series of variants of the Champagne algorithm to Deepinv.
The code is still quite rough, but I'd like to get some feedback before polishing it any further. I mostly want to check whether there's interest, and whether I'm getting the interface right and plugging into the right parts of deepinv.

### Background on the algorithm

Champagne is an algorithm that belongs to Sparse Bayesian Learning, and is notably used in MEG/EGG for source localization, where it shows very good performance.

In MEG source localization, we're solving $y = A x$, where $y$ contains sensor readings over time and is of shape `nb_sensors x time_steps`. The goal is to determine the activated sources withing the brain producing the sensor readings, by determining the `nb_sources x time_steps` values in $x$.  We have `nb_sources >> nb_sensors`.

The Champagne class of algorithms is a type-II Bayesian estimator that jointly determines the variance and the expectation for each source in $x$, while ensuring the sources are sparse.
It should also be noted that the algorithm makes assumptions on the prior and data fidelity terms, and doesn't allow for any flexibility there.

For more details on the background, [*Unification of sparse Bayesian learning algorithms for electromagnetic brain imaging with the majorization minimization framework*](https://www.sciencedirect.com/science/article/pii/S1053811921005851) contains a pretty good state-of-the art, and is what I've used as a basis for this implementation.

## Implementation questions
* The code is quite close to what's already implemented in `OptimIterator`, but with no well-defined f and g step, so I've subclassed `BaseOptim` directly. Does that look right to you, or should I be using something else ?
* I end up accessing the raw matrix `A` directly, by extracting it from the physics object using `A = physics.A(identity)`. How bad is it? It's very useful performance-wise, as the code uses quite a few multiplications by a (very large) diagonal matrix, and those can be done much faster with broadcasting to avoid ever having to build the diagonal matrix directly. I could perhaps add a warning if the physics object does not belong to `LinearPhysics` to ensure that approach is sound.
 


Andrewwango:
Hi @Seon82, thanks for the PR, will there be an implementation of the source localisation physics too?

Seon82:
Thanks for your comment! The forward operator in MEG/EEG is linear, so using `LinearPhysics` should work fine. I could create a subclass with the same behavior for better discoverability, but I don't think it's necessary at this point

Andrewwango:
> Thanks for your comment! The forward operator in MEG/EEG is linear, so using `LinearPhysics` should work fine. I could create a subclass with the same behavior for better discoverability, but I don't think it's necessary at this point

Thanks - but where is `A` defined then? E.g. if I want to simulate a dataset

Seon82:
I see what you mean. You’d typically use dedicated software/libraries like MNE or FieldTrip to generate $A$ for a given head geometry and sensor positions, then export it to use within Deepinv. 

That being said, we could consider adding a custom physics class with some level of integration with specialized libraries (most likely [MNE](https://mne.tools/stable/index.html)) to streamline the process.

Andrewwango:
Thanks for the clarity. I think it would be worth having at least a dummy implementation of A or a simple integration with MNE, which will have several benefits:
- Show that it's possible (e.g. integration in the [RI example](https://deepinv.github.io/deepinv/auto_examples/advanced/demo_ri_basic.html) with torchkbnufft)
- Easier to start experimenting
- To write unit tests for the Champagne algorithm

Any comments @tachella ?

PR #315, opened 2024-10-01
Generate dataset generator batch

apolychronou:
when sigma_generator not None and batch_size != 1 in generate_dataset there was a dimension mismatch because x was taken from the class and not the dataloader


Andrewwango:
Hi @apolychronou , thanks for flagging this issue.

Yes, you are correct, batch_size should be 1 where you highlighted.

I've just spotted a bigger error, `physics_generator` isn't actually being used in the data generation loops. I've raised that here: #319

Andrewwango (approved):

apolychronou:
Hi @Andrewwango ,
well it takes a step for each Physics operator.
I guess it should take a step for each batch

PR #324, opened 2024-10-20
generate dataset save params + fix bug

### LINKED ISSUE #201: `generate_dataset` save `physics_generator` params
In #183 `generate_dataset` can use physics that varies sample to sample, by using a `physics_generator`. Then it saves the generated samples and the `physics` state dict. 

It should also save the per-sample params (e.g. blur kernel) created using `params = physics_generator.step()` for each sample. This would be useful in scenarios where you want to:

- jointly estimate the params (e.g. blur kernel, or subsampling mask) in a blind deblurring problem, and evaluate the learnt kernels
- assume knowledge of the blur kernel during training, e.g. supervised learning.
- perform measurement consistency/data fidelity

### LINKED ISSUE #318: FEAT add `block=False` parameter to plot functions
`plot_*` functions in `deepinv.utils.plotting` use `plt.show()` which halts execution of the code until the matplotlib window is closed.

Would it make sense to use plt.show(block=False) instead in those?

### LINKED ISSUE #319: Generate dataset not stepping physics generator
It looks like `physics_generator` is not actually being used in the data generation loops in generate_dataset, it looks like an error from #183 : 

https://github.com/deepinv/deepinv/blob/main/deepinv/datasets/datagenerator.py#L206

Inside the loop (for both train_dataset and test_dataset), new physics generator params should be created each iteration.

Note this would require retraining examples which required this functionality (i.e. `physics_generator is not None`), such as the adversarial training example.

Note this could also be incorporated with fixing #201

### LINKED ISSUE #323: device inconsistent
Should we explicitly specify generator.device in /deepinv/transform/scale.py? an error has just been detected: 

/usr/local/lib/python3.10/dist-packages/deepinv/transform/scale.py](http://generator.dev/usr/local/lib/python3.10/dist-packages/deepinv/transform/scale.py) in sample_from(values, shape, dtype, device, generator)
     11     N = torch.tensor(len(values), device=device, dtype=dtype)
     12     indices = torch.floor(
---> 13         N * torch.rand(shape, device=device, dtype=dtype, generator=generator)
     14     ).to(torch.long)
     15     return values[indices]
RuntimeError: Expected a 'cuda' device type for generator but found 'cpu'

Andrewwango:
Fixes #319: fix error where generate dataset not stepping physics generator params

Closes #201:  now generate dataset can save and load physics generator params for better reproducibility. (+ lots of tests)

Fixes #323 

re-train adversarial training example as it was affected by #319

Add small explanation to `plot` to close #318

### DIFF
--- a/deepinv/datasets/datagenerator.py
+++ b/deepinv/datasets/datagenerator.py
@@ -17,12 +17,16 @@ if TYPE_CHECKING:
 
 class HDF5Dataset(data.Dataset):
     r"""
-    DeepInverse HDF5 dataset with signal/measurement pairs.
+    DeepInverse HDF5 dataset with signal/measurement pairs ``(x, y)``.
 
     If there is no training ground truth (i.e. ``x_train``) in the dataset file,
     the dataset returns the measurement again as the signal.
 
-    Optionally also return physics generator params as a dict per sample, if one was used during data generation.
+    Optionally also return physics generator params as a dict per sample ``(x, y, params)``, if one was used during data generation.
+
+    ..note::
+
+        We support all dtypes supported by ``h5py`` including complex numbers, which will be stored as complex dtype.
 
     :param str path: Path to the folder containing the dataset (one or multiple HDF5 files).
     :param bool train: Set to ``True`` for training and ``False`` for testing.
@@ -119,6 +123,12 @@ def generate_dataset(
     The generated dataset contains a train and test splits.
 
     Optionally, if random physics generator is used to generate data, also save physics generator params.
+    This is useful e.g. if you are performing a parameter estimation task and want to evaluate the learnt parameters,
+    or for measurement consistency/data fidelity, and require knowledge of the params when constructing the loss.
+
+    ..note::
+
+        We support all dtypes supported by ``h5py`` including complex numbers, which will be stored as complex dtype.
 
     :param torch.data.Dataset train_dataset: base dataset (e.g., MNIST, CelebA, etc.)
         with images used for generating associated measurements
--- a/deepinv/datasets/datagenerator.py
+++ b/deepinv/datasets/datagenerator.py
@@ -1,75 +1,124 @@
+from __future__ import annotations
+from typing import TYPE_CHECKING, Union, Callable, Tuple
+
 from tqdm import tqdm
 import os
 import h5py
 import torch
-from torch.utils.data import DataLoader, Subset
+from torch import Tensor
+from torch.utils.data import DataLoader, Subset, Dataset
 from torch.utils import data
+
+if TYPE_CHECKING:
+    from deepinv.physics import Physics
+    from deepinv.physics.generator import PhysicsGenerator
+    from deepinv.transform import Transform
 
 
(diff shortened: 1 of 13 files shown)


deepinv/datasets/datagenerator.py
```diff
@@ -1,76 +1,125 @@
...
+
+    Optionally also return physics generator params as a dict per sample, if one was used during data generation.
+
     :param str path: Path to the folder containing the dataset (one or multiple HDF5 files).
     :param bool train: Set to ``True`` for training and ``False`` for testing.
-    :param transform: A torchvision transform to apply to the data.
+    :param Transform, Callable transform: A deepinv or torchvision transform to apply to the data.
+    :param bool load_physics_generator_params: load physics generator params from dataset if they exist (e.g. if dataset created with :meth:`deepinv.datasets.generate_dataset`)
+    :param torch.dtype, str dtype: cast all real-valued data to this dtype.
+    :param torch.dtype, str complex_dtype: cast all complex-valued data to this dtype.
```
matthieutrs, line 32: Maybe we can be a bit more precise on the behaviour of such function: e.g. would imaginary part be ignored?

deepinv/loss/metric/perceptual.py
```diff
@@ -35,7 +35,7 @@ class LPIPS(Metric):
     def __init__(self, device="cpu", **kwargs):
         super().__init__(**kwargs)
         pyiqa = import_pyiqa()
-        self.lpips = pyiqa.create_metric("lpips").to(device)
+        self.lpips = pyiqa.create_metric("lpips", check_input_range=False).to(device)
```
matthieutrs, line 38: The pyiqa doc is not clear on that point but it seems that what you suggest here is departing from the default behaviour... How about adding `check_input_range` as an init optional parameter?

```python
     def __init__(self, device="cpu", check_input_range=False, **kwargs)
```

deepinv/loss/metric/perceptual.py
```diff
@@ -74,7 +74,7 @@ class NIQE(Metric):
     def __init__(self, device="cpu", **kwargs):
         super().__init__(**kwargs)
         pyiqa = import_pyiqa()
-        self.niqe = pyiqa.create_metric("niqe").to(device)
+        self.niqe = pyiqa.create_metric("niqe", check_input_range=False).to(device)
```
matthieutrs, line 77: Same comment here

deepinv/tests/test_loss_train.py
```diff
@@ -37,6 +43,129 @@ def test_generate_dataset(tmp_path, imsize, device):
...
+@pytest.mark.parametrize("bsize", [1, 4])
+@pytest.mark.parametrize(
+    "phys_gen",
+    [
+        "bernoulli_mask",
+        "sigma",
+        "diffraction",
+    ],
+)
+def test_generate_dataset_physics_generator(
```
matthieutrs, line 59: Really cool!

matthieutrs (commented):
Thanks for the many fixes! I just have minor comments

matthieutrs (approved):

PR #333, opened 2024-10-28
Doc clean up

tomMoral:
A few clean up for the doc:

- Add explicit deps when needed (wavelets, bm3d)
- Make sure all data downloaded by example are set in a single folder.
- Fix warning with deprecated imports+ un escaped `\`
- Change the order of the examples (start from basic, move advance at the end).
- Propose to change the landing page with a `User Guide` section that would  encompass all the API details.


docs/source/conf.py
```diff
@@ -35,9 +35,9 @@
 copybutton_exclude = ".linenos, .gp"
 
 intersphinx_mapping = {
-    "numpy": ("http://docs.scipy.org/doc/numpy/", None),
-    "torch": ("https://docs.pytorch.org/2.0/", None),
-    "python": ("https://docs.python.org/3.4", None),
+    "numpy": ("https://numpy.org/doc/stable/", None),
+    "torch": ("http://pytorch.org/docs/2.0/", None),
```
Andrewwango, line 39: Why `http` rather than `https`?

examples/basics/demo_custom_prior.py
```diff
@@ -17,7 +17,7 @@
 from deepinv.optim.optimizers import optim_builder
 from deepinv.training import test
 from torchvision import transforms
-from deepinv.utils.demo import load_dataset
+from deepinv.utils.demo import load_dataset, get_data_home
```
Andrewwango, line 20: why do we need to import get_data_home?

examples/basics/demo_transforms.py
```diff
@@ -121,7 +121,7 @@
 # Symmetrize function with respect to transform
 f_s = t.symmetrize(f, average=True)
 dinv.utils.plot(
-    [x, f(x), f_s(x)], titles=["Orig", "$f(x)$", "$\sum_i T_i^{-1}f(T_ix)$"]
+    [x, f(x), f_s(x)], titles=["Orig", "$f(x)$", "$\\sum_i T_i^{-1}f(T_ix)$"]
```
Andrewwango, line 124: Why is this needed? Seems like the Sigma looks fine on the [example](https://deepinv.github.io/deepinv/auto_examples/basics/demo_transforms.html)
(Secondly I still don't know why the example looks so ugly, any ideas @tomMoral ?)

Andrewwango:
I like the idea of a "Quick start" section that quickly explains the layout, but I'm not sure I'm a fan of "User Guide" hiding all the other sections on the homepage, it makes it slightly harder to find the API reference you want
image

docs/source/conf.py
tomMoral: probably a typo ^^
```suggestion
    "torch": ("https://pytorch.org/docs/2.0/", None),
```

examples/basics/demo_custom_prior.py
tomMoral: Good catch, I changed so many I got lost in the process I guess :)
```suggestion
from deepinv.utils.demo import load_dataset
```

examples/basics/demo_transforms.py
tomMoral: `\s` raises a warning when building the doc because it is not well encoded.
We can either use a raw string or escape the `\`. I choose to escape it but we could simply add `r` in from of the string.
As you prefer.

For the look of the example, I think the issue is that the text is too large, we should play with [`subpots_adjust`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.subplots_adjust.html#matplotlib.pyplot.subplots_adjust) to make enough room for the equation, or reduce the font size to make it fit.

PR #340, opened 2024-10-29
Sde solver

mh-nguyen712:


deepinv/sampling/edm/torch_utils/__init__.py
tachella, line 1: remove dummy file

deepinv/sampling/edm/__init__.py
```diff
@@ -0,0 +1,46 @@
+import torch
+from .dnnlib.util import open_url
+import pickle
+import torch.nn as nn
+import os
+import sys
+
+
+class ModelWrapper(nn.Module):
```
tachella, line 9: This is not very compatible with the other way of loading denoisers in the network.

I would prefer that we add fewer models, but in a more concise way than copy-pasting a large part of the edm library.

deepinv/sampling/edm/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+    def forward(self, x: torch.Tensor, t: float):
+        if isinstance(t, float):
+            t = torch.tensor([t] * x.size(0), device=x.device)
+        return self.edm_model.forward(x, sigma=t, class_labels=None)
+
+
+current_path = os.path.dirname(os.path.abspath(__file__))
+
+
+def load_model(model_name: str = "edm-afhqv2-64x64-uncond-ve.pkl") -> nn.Module:
```
tachella, line 23: are all these models sharing the same architecture?

If yes, I would only keep the architecture and select the correct weights using `pretrained` input (see for example how it is done in [DnCNN](https://deepinv.github.io/deepinv/api/stubs/deepinv.models.DnCNN.html#deepinv.models.DnCNN))

deepinv/sampling/edm/dnnlib/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
tachella, line 1: I would get rid of this if possible

tachella (changes requested):
Can we get rid of the edm library files? 
This forces us to follow their license, and I believe that we probably do not need all their functions.

Do we need their functions for something more than loading their networks?

PR #342, opened 2024-10-29
Add multiscale framework to physics

Nils-Laurent:
Not in a state to be merged yet, but gives visibility on the work. If you have any feedback, do not hesitate.


Andrewwango:
Should the downsample etc. functions be methods of `Physics`? It feels quite specialised, maybe it's better to put them in a separate class e.g. `MultiScaleBase` and when you want to use it you could do

```python
class MultiscaleInpainting(Inpainting, MultiScaleBase):
    pass
```

Andrewwango:
Marking as draft

tachella:
> Thank you for this interesting feedback @Andrewwango. I agree that having a separated multiscale framework could allow for more general algorithms, and that multiscale is not necessarily related to `Physics` class (e.g. an upscaling algorithm for handling non-uniform deformation with antialiasing as we discussed).
> 
> Maybe, as a first step, multiscale could be associated with `Physics` (with the currently proposed structure) and adapted later if a more relevant structure becomes easier to spot. I think about this strategy pro/cons as follows:
> 
> * '+' provides tools and interface for handling coarser scale in the context of inverse problems concisely (as shown in the example)
> * '-' will change the interface of physics : if we choose the proposed structure, it means that if we change `Physics` interface later, it might break users code
> 
> I think that if the interface on `Physics` makes sense, then we can accept the proposed structure (because if we implement in a separate class later, we can always use it in Physics afterward). How do you feel about it ? @Andrewwango, @tachella

I think that it makes sense to add it to Physics directly with a NoImplementedError by default, since it doesn't add overhead to the physics class, and doesn't require defining more intermediate classes.

Nils-Laurent:
@Andrewwango I like this idea. MultiscaleBase would then have multiscale related methods (so Physics would be unchanged) and the downsampling operator would be a member of MultiscaleBase. I have to think about how MultiscaleBase would use fine scale physics methods such as A_dagger (probably by sending "self" to __init__ of MultiscaleBase, which would be member, I will see if this is possible)

tachella:
> @Nils-Laurent @tachella agreed, it was more a conceptual point whether to_coarse should be that general (esp when lots of physics can't implement it) - just thinking ahead if we have many more functionalities like this it might crowd Physics ("cognitive overhead").
> 
> How about every time we want to let a new physics have multiscale functionality we let it derive from MultiScaleBase too? Eg
> 
> ```
> class Physics:
> ... unchanged 
> 
> class MultiscaleBase:
> ... as above 
> 
> class MRI(DecomposablePhysics):
> ... unchanged 
> 
> class Inpainting(DecompPhysics, MSBase):
> ... 
> ... def to_coarse(...):
> ...
> ```
> 
> This way the user doesn't need to implement their own class (unlike what I initially suggested) but it still separates the multiscale stuff into their own logical class

I think this solution is better, but I still don't agree with the idea that many operators will not be able to implement the multiscale approach. We could potentially provide a default implementation in physics of ``to_coarse`` that returns $A' = AS$ where $S$ is an upsampling operator, and this should work for any operator.

Nils-Laurent:
@tachella This is a good point. Does that mean that we would target `LinearPhysics` ?

Andrewwango:
> @tachella This is a good point. ~Does that mean that we would target `LinearPhysics` ?~ Nevermind, I agree it works in the general case

Cool - I didn't realise that there could be a default implementation - in which case that looks good to me. Looking forward to seeing what that would look like.

Nils-Laurent:
I have changed `Physics ` class to have a general `to_coarse` function.
I am unsure on one point, should there be a `initialize_downsampling_operator` ? (It has one usage in `downsample_signal` but it is mentioned in `to_coarse`, which cannot be used if there is no downsampling operator).

examples/plug-and-play/demo_PnP_multiscale.py
```diff
@@ -0,0 +1,154 @@
+"""
+Multiscale Plug-and-Play
+====================================================================================================
+
+This example describe the usage of multiple scales with Plug-and-Play schemes.
```
tachella, line 5: maybe add a more detailed description here

tachella (changes requested):
Hi Nils,

The PR is looking good, but I have a question/comment regarding how you construct the `to_coarse` operator. We should also add this method with the adjoint for the important case of LinearPhysics.

examples/plug-and-play/demo_PnP_multiscale.py
```diff
@@ -0,0 +1,163 @@
+"""
+Multiscale Plug-and-Play
```
tachella, line 2: I think we should wait until the paper is on arxiv to upload this example. I would nonetheless start merging the rest

examples/plug-and-play/demo_PnP_multiscale.py
```diff
@@ -0,0 +1,163 @@
...
+noise_level = 0.1
+noise_model = GaussianNoise(sigma=noise_level)
+physics = Inpainting(
+    tensor_size=set3c_img_shape, mask=0.5, noise_model=noise_model, device=device
+)
+
+prior = PnP(denoiser=DRUNet(pretrained="download", device=device))
+
+# set values of the PnP parameters
+max_iter = 200
```
tachella, line 74: Can you reduce this a bit? It will be a bit heavy to run every time the docs are built

examples/plug-and-play/demo_PnP_multiscale.py
```diff
@@ -0,0 +1,163 @@
...
+    metrics=[dinv.metric.PSNR()],
+    device=device,
+    online_measurements=True,
+    plot_images=True,
+    plot_convergence_metrics=True,
+    verbose=True,
+)
+
+# %%
+# Multiscale case : use a coarse setting to initialize the fine setting.
```
tachella, line 109: Remove space before ":"

examples/plug-and-play/demo_PnP_multiscale.py
tachella: Another option is to only change the initialisation, but then I wouldn't call it multilevel PnP

tachella (dismissed):

deepinv/physics/blur.py
```diff
@@ -282,6 +282,44 @@ def update_parameters(self, filter=None, **kwargs):
...
+
+        # downsample the blur filter
+        filt = torch.nn.functional.conv2d(pf, df, groups=pf.shape[1], padding="valid")
+        filt = filt[:, :, ::2, ::2]
+        filt = filt / torch.sum(filt) * torch.sum(in_filt)
+
+        return Blur(filter=filt, padding=self.padding, device=self.filter.device)
+
 
 class BlurFFT(DecomposablePhysics):
```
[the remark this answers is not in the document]
Nils-Laurent: You are right, I will define the code in a function and use it both for Blur and BlurFFT

examples/plug-and-play/demo_PnP_multiscale.py
Nils-Laurent: Once the rest is ready, we can merge it if that's fine and add the example later

examples/plug-and-play/demo_PnP_multiscale.py
Nils-Laurent: I took it down to 10 iterations which gives a good result

deepinv/physics/forward.py
```diff
@@ -531,6 +610,22 @@ def A_dagger(self, y, **kwargs):
...
+            raise ValueError(
+                "No downsampling operator, use initialize_downsampling_operator before."
+            )
+
+        ds = self.downsampling_operator
+
+        # factor**2 is only valid if x is an image
+        A_coarse = lambda x: self.A(ds.factor**2 * ds.A_adjoint(x))
+        A_coarse_adj = lambda y: ds.A(self.A_adjoint(y))
+        return LinearPhysics(A=A_coarse, A_adjoint=A_coarse_adj)
```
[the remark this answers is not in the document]
Nils-Laurent: Good point, I did not check it, and it does not pass the test. I will investigate

deepinv/tests/test_physics.py
```diff
@@ -1037,3 +1040,53 @@ def test_decolorize(srf, device, imsize):
...
+    r"""
+    Tests the similarity between a linear physics and its coarse version.
+
+    :param name: operator name (see find_operator)
+    :param device: (torch.device) cpu or cuda:x
+    """
+    physics, imsize, _, dtype = find_operator(name, device, sf=2)
+    if not isinstance(physics, dinv.physics.LinearPhysics):
+        pytest.skip("Skip " + name + " : not LinearPhysics")
+    if not len(imsize) == 3:
```
tachella, line 1072: Why are you only testing images with 3 channels?

jscanvic:
I believe `LinearPhysics.to_coarse` was mistakenly moved to `ComposedPhysics` during the last merge see:

https://github.com/deepinv/deepinv/blob/1e93d35067e8190612f0c0b6119116e7c9346a1b/deepinv/physics/forward.py#L822
https://github.com/deepinv/deepinv/blob/ec54d297e19cd0c134f6be31109df8747e4e5363/deepinv/physics/forward.py#L609

It appears to be what makes the tests fail too. I tried to fix it in 1cda9d1e8e96e1fb5f9c15e69fb76c3b0e751215 feel free to revert if it's not what's intended!

Nils-Laurent:
I have finally a new version of the PR that I think worth reviewing.
If any of you have time to look at it, thanks!

Tmodrzyk:
Hi @Nils-Laurent sorry we seem to have missed your update ! The multiscale physics is still very interesting for the lib so I'll try to have a look soon, don't hesitate to ping me or the other reviewers ! :smile:

Nils-Laurent:
> Hi @Nils-Laurent sorry we seem to have missed your update ! The multiscale physics is still very interesting for the lib so I'll try to have a look soon, don't hesitate to ping me or the other reviewers ! 😄

Hi! That's great, I have time in the incoming weeks to work on it.
I will check the merge conflicts and notify you and other reviewers once its done :)

Nils-Laurent:
Ok, so I think the PR is fine for being reviewed @tachella, @Tmodrzyk, @jscanvic , @Andrewwango. I will be glad to have your feedback on this work.

The last CI did fail `test_least_squares_implicit_backward[device0-BiCGStab-deblur_valid-2]`. I saw that happen for another PR (e.g. PR1295 Add RNG support to GammaNoise and FisherTippettNoise). Could this test randomly fail for accuracy issues in computations ?

@tachella, I checked the previous change requests, I think they are all either resolved or out of date (do not hesitate if it is not the case).

Tmodrzyk (dismissed):
Hi @Nils-Laurent, thank you again for this awesome PR ! 

There are a few high-level issues that should be addressed first:

- we now type basically everything, both in the functions declaration and in the docstring. The `wrappers.py` file lacks typing in several places.

- in the built demo file (you can check it here: https://github.com/deepinv/deepinv/actions/runs/30389846437/artifacts/8701226754) the PnP reconstruction seems barely above no-learning, is this expected ? Also in this cell, there is no explanation with regard to what we should or should not observe in the reconstructions. More explanations would be great. :D
Otherwise the multiscale reconstruction looks great.

- the demo uses the old way of constructing optimization algorithms. See the preferred syntax here https://deepinv.org/auto_examples/basics/demo_custom_optim.html#using-a-predefined-optimization-algorithm

deepinv/optim/data_fidelity.py
```diff
@@ -306,6 +308,28 @@ def prox(
         """
         return physics.prox_l2(x, y, self.norm * gamma)
 
+    def grad(self, x, y, physics, *args, **kwargs):
+        r"""
+        Calculates the gradient of the data fidelity term :math:`\datafidname` at :math:`x`.
+
+        The gradient is either computed using the chain rule, or using specific implementation of deepinv.physics.LinearPhysics.A_adjoint_A in the case of a LinearPhysics:
```
Andrewwango: small grammatical issue: surely the equation below refers to the chain rule, but this sentence makes it sound like the equation refers to the linear case. Maybe put the linear case instead below somewhere?

docs/source/api/deepinv.physics.rst
```diff
@@ -69,6 +69,10 @@ Operators
    deepinv.physics.Ptychography
    deepinv.physics.PtychographyLinearOperator
    deepinv.physics.Scattering
+   deepinv.physics.BlurMultiScaler
```
Andrewwango: do you need to add these to user guide too?

deepinv/physics/wrappers.py
```diff
@@ -37,17 +40,23 @@ def __init__(
...
         super().__init__(noise_model=physics.noise_model, device=device, **kwargs)
         self.base = physics
         self.factors = factors
         self.img_shape = img_shape
-        self.Upsamplings = [
-            Upsampling(img_size=img_shape, filter=filter, factor=factor, device=device)
-            for factor in factors
-        ]
+        self.Upsamplings = []
+        for factor in factors:
```
Andrewwango: can't we do this more compactly like above with 
```python
        self.Upsamplings = [
            Upsampling(img_size=img_shape, filter=filter if dtype is None else filter.to(dtype=dtype), factor=factor, device=device)
            for factor in factors
        ]

``` ?

File: deepinv/physics/wrappers.py [hunk 11]
```diff
@@ -204,6 +213,264 @@ def prox_l2(
...
+
+def to_multiscale(
+    physics: Physics,
+    img_shape: tuple,
+    dtype: type | None = None,
+    factors: tuple = (2, 4, 8),
+    device: str = "cpu",
+):
+    r"""
+    This function creates the proper MultiScaler object associated with the provided physics.
```
Andrewwango: since this is the top level function, this should be added to the user guide. Also, add more documentation to this docstring: what is the behaviour? i.e. blur, blurfft and inpainting get special treatment. Why do they need special treatment?

deepinv/physics/wrappers.py
```diff
@@ -204,6 +213,264 @@ def prox_l2(
...
+    See :class:`LinearPhysicsMultiScaler` for details.
+
+    :param physics: physics that should be converted to a MultiScaler
+    :param img_shape: shape of the image in the fine scale
+    :param torch.dtype, dtype: type to be associated with the signal
+    :param factors: downsampling factors used to get in coarser scales
+    :return: a MultiScaler version of the provided physics
+    """
+    if device is None:
+        device = physics.device
```
Andrewwango: not all physics have a device attribute - see #968 #989

deepinv/physics/wrappers.py
```diff
@@ -204,6 +213,264 @@ def prox_l2(
...
+        """
+        self.set_scale(scale)
+        physics = self.scaled_physics[self.scale - 1]
+        factor = self.factors[self.scale - 1]
+        return physics.A_adjoint_A(x) / factor**2
+
+
+def to_multiscale(
+    physics: Physics,
+    img_shape: tuple,
```
Andrewwango: we use img_size nomenclature rather than img_shape

deepinv/physics/wrappers.py
Andrewwango: also 2D only

deepinv/physics/wrappers.py
```diff
@@ -204,6 +213,264 @@ def prox_l2(
...
+        """
+        self.set_scale(scale)
+        if self.scale == 0:
+            return y
+        else:
+            return self.Upsamplings[self.scale - 1].A_adjoint(y)
+
+    def A_adjoint_A(self, x, scale=None, **kwargs):
+        r"""
+        Less computationnaly expensive version than parent class :class:`LinearPhysicsMultiScaler`
```
Andrewwango:
```suggestion
        Less computationally expensive version than parent class :class:`LinearPhysicsMultiScaler`
```

deepinv/physics/wrappers.py
Andrewwango: And if the user wants to speed up their own physics they can write their own wrapper class if the ywant

Andrewwango (changes requested):
super cool @Nils-Laurent and I'm happy to see this continued!

docs/source/api/deepinv.physics.rst
Nils-Laurent: No I agree, I think the top-level function is sufficient.

jscanvic:
Thanks for your review @Tmodrzyk @Andrewwango!

I'm helping Nils update the PR to the current standards

- The demo should be clearer, especially that the baseline PnP is expected to perform badly
- It should be clearer what coarser scale physics represent / how they're computed and why specific implementations for linear physics and blur and inpainting are provided
- I also added a virtual method for `downsample_measurement` to make it clearer that it's only implemented for certain operators as there's no easy way to compute it in general

With Nils' changes all points should be covered

@Nils-Laurent Don't hesitate to rewrite some of the stuff I added if it's incorrect / misleading!

Nils-Laurent:
Thank you all for your valuable feedback @Tmodrzyk, @Andrewwango, @jscanvic ,
and thank you very much @jscanvic for your help!

On my side, I think the PR looks good.

Maybe just a remark about the two multi-scale tests, i.e.
* test_multiscale_coarse_adjointness
* test_multiscale_A_adjoint_A

Their current implementation distinguish between two cases. Either `find_operator` gives sufficiently high resolution, and it is used as it is, otherwise, `find_operator` is called with a specific `imsize` setting the number of channels to `3`. When a new LinearPhysics is implemented, those tests could fail if the intended number of channel should not be set to `3`. Is that fine ?

Andrewwango (approved):
LGTM

examples/plug-and-play/demo_PnP_multiscale.py
```diff
@@ -0,0 +1,146 @@
...
+y = physics(x)
+dinv.utils.plot([x, y], ["Ground Truth", "Measurements"])
+
+# %%
+# Single-scale and multi-scale PnP models
+# ---------------------------------------
+# The multi-scale PnP model can be understood as a combination of multiple
+# single-scale PnP models operating at different scales. Here, we use two
+# scales: the fine scale corresponding to the scale of the original image, and
+# a coarse scale corresponding the fine scale downsampled by a factor of 2.
```
Tmodrzyk:
```suggestion
# a coarse scale corresponding to the fine scale downsampled by a factor of 2.
```

examples/plug-and-play/demo_PnP_multiscale.py
```diff
@@ -0,0 +1,146 @@
...
+    "plot_images": True,
+    "plot_convergence_metrics": True,
+    "verbose": True,
+}
+
+# Benchmark single-scale PnP
+dinv.test(model=model_fs, **test_kwargs)
+
+# Benchmark multi-scale PnP
+dinv.test(model=model_ms, **test_kwargs)
```
Tmodrzyk, line 146: These plots are awesome ! 
Nit: in the rendered doc, in the first plot the PSNR and residual subplots are slightly overlaping

Tmodrzyk (commented):
LGTM, nitpick on the demo

examples/plug-and-play/demo_PnP_multiscale.py
jscanvic: My bad good catch!

examples/plug-and-play/demo_PnP_multiscale.py
jscanvic: It appears it's a bug in `plot_curves` and it's not straightforward to fix - I opened #1308 so it can be fixed in another PR

jscanvic:
Thank you both! @Tmodrzyk it should be good now

jscanvic:
CI failures are unrelated - I'm merging

PR #357, opened 2024-11-17
Solve non-backward compatible changed in data-fidelity

samuro95:
Add the grad_d, prox_d and prox_d_conjugate function removed with the last reformat of the DataFidelity class? 


matthieutrs (approved):
Thanks for the fix! Just added some mini tests to check we do not run into the same issue in the future

PR #358, opened 2024-11-21
Modified JacobianSpectralNorm for batched processing and added 'reduction' parameter

Vladimirescu:
The `JacobianSpectralNorm` has been modified to compute spectral norms for each element in the batch, then perform a reduction over the 1D tensor containing the spectral norms. If `x` is a tensor with `x.dim() > 1` the first dimension is taken as the batch dimension. Otherwise, `x` is treated as a tensor with batch size 1.

### Observations:
- Some test failures regarding mismatches between tensor locations (`cpu` vs `cuda`) were observed. No failures related to the computations performed by `JacobianSpectralNorm` were observed
- The initial workings of `JacobianSpectralNorm` was to return the maximum spectral norm over that given batch. This is why parameter `reduction` is set to `max` by default.


matthieutrs:
Hi @Vladimirescu, thanks for the PR, looking good! Would it be possible for you to add some appropriate tests here?

https://github.com/deepinv/deepinv/blob/main/deepinv/tests/test_loss.py#L23C5-L23C34

(in particular checking that the reduction works well on some pre-defined examples)

Vladimirescu:
Hello @matthieutrs, sure! I'm on it.

deepinv/loss/regularisers.py
```diff
@@ -18,6 +18,7 @@ class JacobianSpectralNorm(Loss):
     :param float tol: tolerance for the convergence of the power method.
     :param bool eval_mode: set to `False` if one does not want to backpropagate through the spectral norm (default), set to `True` otherwise.
     :param bool verbose: whether to print computation details or not.
+    :param str reduction: one of ["mean", "sum", "max"], operation to be performed after all spectral norms have been computed. If None, a vector of length batch_size will be returned.
```
Andrewwango, line 21:
```suggestion
    :param str reduction: reduction in batch dimension. One of ["mean", "sum", "max"], operation to be performed after all spectral norms have been computed. If ``None``, a vector of length ``batch_size`` will be returned. Defaults to "max".
```

deepinv/loss/regularisers.py
```diff
@@ -18,6 +18,7 @@ class JacobianSpectralNorm(Loss):
     :param float tol: tolerance for the convergence of the power method.
     :param bool eval_mode: set to `False` if one does not want to backpropagate through the spectral norm (default), set to `True` otherwise.
```
Andrewwango, line 19:
```suggestion
    :param bool eval_mode: set to ``False`` if one does not want to backpropagate through the spectral norm (default), set to ``True`` otherwise.
```

deepinv/loss/regularisers.py
```diff
@@ -120,13 +159,14 @@ class FNEJacobianSpectralNorm(Loss):
...
```
Andrewwango, line 160:
```suggestion
    :param bool eval_mode: set to ``False`` if one does not want to backpropagate through the spectral norm (default), set to ``True`` otherwise.
```

deepinv/loss/regularisers.py
```diff
@@ -120,13 +159,14 @@ class FNEJacobianSpectralNorm(Loss):
...
     :param bool verbose: whether to print computation details or not.
+    :param str reduction: one of ["mean", "sum", "max"], operation to be performed after all spectral norms have been computed. If None, a vector of length batch_size will be returned.
```
Andrewwango, line 162:
```suggestion
    :param str reduction: reduction in batch dimension. One of ["mean", "sum", "max"], operation to be performed after all spectral norms have been computed. If ``None``, a vector of length ``batch_size`` will be returned. Defaults to "max".
```

deepinv/loss/regularisers.py
```diff
@@ -39,14 +40,36 @@ class JacobianSpectralNorm(Loss):
...
-    def __init__(self, max_iter=10, tol=1e-3, eval_mode=False, verbose=False):
+    def __init__(self, max_iter=10, tol=1e-3, eval_mode=False, verbose=False, reduction='max'):
         super(JacobianSpectralNorm, self).__init__()
         self.name = "jsn"
         self.max_iter = max_iter
         self.tol = tol
         self.eval = eval_mode
         self.verbose = verbose
 
+        self.reduction = lambda x: x
```
Andrewwango, line 51: I think add an 
```python
else:
    raise ValueError('reduction should be "mean", "sum", "max", "none" or None.')
```
Also sorry to be annoying, but please can you also add a "none" option? Just to be consistent with https://deepinv.github.io/deepinv/stubs/deepinv.loss.metric.Metric.html#deepinv.loss.metric.Metric

Andrewwango (commented):
Thanks for the PR! Minor docstrings changes

Vladimirescu:
The second test `Build docs / docs (pull_request) Failing after 7m` fails due to the fact the first test from here

https://github.com/deepinv/deepinv/blob/main/deepinv/tests/test_loss.py#L23C5-L23C34

uses an `x_detached` of simiar shape to `toymatrix`, which is `(50, 50)`. For the `toymatrix` to correspond to an operator with maximum spectral norm `50` it should be applied to a `x_detached` of shape `(50,)`. Right now, the first dimension of `x_detached` is considered as the batch size given the current implementation.

I think this initial test should be modified to make `x_detached` is a single `1D tensor` with random values. Also, I think another check should be added when running `JacobianSpectralNorm` to check if both `x` and `y` are single instances.

deepinv/tests/test_loss.py
```diff
@@ -35,18 +35,90 @@ def test_jacobian_spectral_values(toymatrix):
...
+    toy_operators = torch.Tensor([1, 2, 3, 4, 5])[:, None, None, None]
+
+    x_detached = torch.randn(B, C, H, W).requires_grad_()
+    out = toy_operators * x_detached
+
+    def model(x):
+        return toy_operators * x
+
+    # Nonec -> return all spectral norms
+    reg_l2 = JacobianSpectralNorm(
```
Andrewwango, line 61: I like the thorough tests, but perhaps this test could make better use of parametrized pytest? For example:
```python
@pytest.mark.parametrize("reg_type", ["jac", "fne"])
@pytest.mark.parametrize("reduction", ["none", "mean", "sum", "max"])
def test_jacobian_spectral_values(toymatrix, reg_type, reduction):
    ...
```
This way you will only probably need one assert statement, letting pytest do the combinations for you

Andrewwango (changes requested):

deepinv/tests/test_loss.py
matthieutrs: Oh yes much better! Actually not that easy to parametrize on the reg_type (in a succint way) since the two regs have different forwards. I've however parametrized the reduction part!

Andrewwango: nice

matthieutrs (approved):

PR #361, opened 2024-11-26
Sde minimal

mh-nguyen712:
Minimal implementation of SDE class and SDE solver for image generation. 


deepinv/sampling/sde.py
```diff
@@ -0,0 +1,408 @@
...
+        generating the random numbers.
+
+        .. note::
+           The :attr:`rng` attribute must be initialized for this method to work properly.
+        """
+        self.rng_manual_seed(seed)
+        return torch.empty_like(input).normal_(generator=self.rng)
+
+
+class DiffusionSDE(nn.Module):
```
tachella, line 130: This shouldn't inherit from BaseSDE?

deepinv/sampling/sde.py
```diff
@@ -0,0 +1,408 @@
...
+        most computation cost is from evaluating the `denoiser`, which will be always computed in `float32`.
+    :param torch.device device: device on which the computation is performed.
+    """
+
+    def __init__(
+        self,
+        drift: Callable = lambda x, t: -x,  # Default to Ornstein-Uhlenbeck process
+        diffusion: Callable = lambda t: math.sqrt(2.0),
+        use_backward_ode=False,
+        denoiser: nn.Module = None,
```
tachella, line 178: when we extend the class to inverse problems, here we should add a NoisyDataFidelity class input

deepinv/sampling/sde.py
```diff
@@ -0,0 +1,408 @@
...
+        **kwargs,
+    ) -> SDEOutput:
+        r"""
+        Sample the backward-SDE.
+
+        :param Union[Tensor, Tuple] x_init: Initial condition of the backward-SDE.
+            If it is a :meth:`torch.Tensor`, `x_init` should follow the distribution of :math:`p_T`, which is usually :math:`\mathcal{N}(0, \sigma_{\mathrm{max}}^2 \mathrm{Id})`.
+            If it is a tuple, it should be of the form `(B, C, H, W)`. A sample from the distribution :math:`p_T` will be generated automatically.
+
+        :param torch.Tensor timesteps: The time steps at which to discretize the backward-SDE, should be of shape `(n_steps,)`.
```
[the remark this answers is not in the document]
tachella: I think this parameter should go to the class constructor

deepinv/sampling/sde.py
```diff
@@ -0,0 +1,408 @@
...
+        :param str method: The method to discretize the backward-SDE, can be one of the methods available in :meth:`deepinv.sampling.sde_solver`.
```
tachella, line 246: I would move the method parameter to the class constructor and pass a Class, ie EulerSolver instead of a string

tachella (changes requested):
This PR represents a great upgrade to the library. I have a few different points that we should address before merging:

#### Classes & Organization

- I think it might be better to separate the SDE from ODE solutions in the `DiffusionSDE` class,  making the `DiffusionSDE` inherit from `BaseSDE`. Moreover, I would make `BaseSDE` inherit from `Reconstructor` and force the forward method to follow the template `forward(y, physics, **kwargs)`. In summary, I would make the following inheritance tree `torch.nn.Module -> deepinv.models.Reconstructor -> deepinv.sampling.BaseSDE -> deepinv.sampling.DiffusionSDE`

- Given the previous point, I would keep the forward method of the class for reconstruction, and move the unconditional generation to a `sample_prior` (or similar name) method.

- I would avoid passing `str` to the SDE classes (e.g. `"euler"`) and instead directly pass the class `EulerSolver`. This will avoid problems in the future if the solvers have specific parameters that need to be set.

- We would have to make separate ODE classes (although this can be left for another PR).

#### Documentation: 

- The new models need to be added to the API/User Guide. We should also ideally briefly describe their architecture on top of referring to the paper where they were proposed.
- The video plotter should also be added to the API/User Guide.
- Many docstrings use a single accent to represent `code` which is the standard way for markdown but does not work with restructuredText (on which sphinx works). We should use double accents, otherwise the text appears in italics instead of code mode.
- There are a few warnings in the **build_docs** Github action (see [here](https://github.com/deepinv/deepinv/actions/runs/12200795201/job/34037741662)) due to some formatting errors. We need to be careful to avoid these errors. You can find these warnings when building the docs with `make html-fast` in your machine.


ps. We can have a zoom call to discuss any of these points in more detail next week.

deepinv/sampling/sde.py
```diff
@@ -0,0 +1,408 @@
...
+from torch import Tensor
+import torch.nn as nn
+from typing import Callable, Union, Tuple, List, Any
+import numpy as np
+from numpy import ndarray
+import warnings
+from .sde_solver import select_solver, SDEOutput
+
+
+class BaseSDE(nn.Module):
```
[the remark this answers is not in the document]
mh-nguyen712: This is only for unconditional generation and don't see a direct relation to the reconstruction or optim. I think we can keep is at an independent class, and inherit from it together with deepinv.models.Reconstructor or BaseOptim for the conditional generation (for inverse problem, e.g. CondSDE).

deepinv/sampling/sde.py
```diff
@@ -0,0 +1,408 @@
...
+        method: str = "euler",
+        seed: int = None,
+        **kwargs,
+    ) -> SDEOutput:
+        r"""
+        Solve the SDE with the given timesteps.
+
+        :param torch.Tensor x_init: initial value.
+        :param timesteps: time steps at which to discretize the SDE, of shape `(n_steps,)`.
+        :param str method: method for solving the SDE. One of the methods available in :meth:`deepinv.sampling.sde_solver`.
```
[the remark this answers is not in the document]
mh-nguyen712: thanks!

deepinv/sampling/sde.py
mh-nguyen712: This class has a forward and a backward SDE, meanwhile the BaseSDE class has only one process (either forward or backward). We can make this class inherit from the BaseSDE and do the same operations only for the backward SDE, but I think we can keep this independent of the BaseSDE class. To be discussed more in details.

deepinv/sampling/sde.py
mh-nguyen712: Yes you're reason. I think that we have 2 options:
- add the `Score` input argument instead, which will be the unconditional score `\nabla \log p_t(x_t)` or the conditional score `\nabla \log p_t(x_t \vert y)`, but then we should define these scores properly outside the `DiffusionSDE`, which can be annoying when switching between the time and the noise level (which is dependent to the underlying SDE) (which is done by the function `sigma_t`).
- add an extra input argument `NoisyDataFidelity` when creating the SDE for the conditional sampling, but it is not used here. But I think that we have no problem when inheriting the ConditionalDiffusionSDE from the DiffusionSDE with an extra input argument. 
I think I prefer the option 2.

deepinv/sampling/sde.py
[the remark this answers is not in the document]
mh-nguyen712: We can sample from the same SDE with different timesteps. It provides more flexibility.

deepinv/sampling/sde.py
mh-nguyen712: It could be a good idea but it might be a little harder to use. We can make this similar to `https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html` or `https://github.com/google-research/torchsde`. In the future, when we add more solvers, we put other solver-related arguments to, for example, `**options`. I prefer to keep it as simple as possible, but I'am open to considering changes if necessary.

deepinv/sampling/sde_solver.py
```diff
@@ -0,0 +1,206 @@
...
+from numpy import ndarray
+import numpy as np
+
+
+class SDEOutput(dict):
+    r"""
+    A container for storing the output of an SDE solver, that behaves like a `dict` but allows access with the attribute syntax.
+    """
+
+    def __init__(self, sample: Tensor, trajectory: Tensor, timesteps: Tensor, nfe: int):
```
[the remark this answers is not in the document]
mh-nguyen712: added

docs/source/user_guide/reconstruction/sampling.rst
```diff
@@ -7,16 +7,89 @@ This package contains posterior sampling algorithms.
...
+.. list-table:: SDE/ODE solvers
+   :header-rows: 1
+
+   * - **Method**
+     - **Description**
+  
+   * - :class:`deepinv.sampling.sde_solver.BaseSDESolver`
+     - Base class of the solvers.
+
+   * - :class:`deepinv.sampling.sde_solver.EulerSolver`
```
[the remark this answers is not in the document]
mh-nguyen712: I added a wikipedia link to each method.

examples/sampling/demo_sde_image_generation.py
```diff
@@ -0,0 +1,316 @@
...
+# --------------------------------------
+#
+# Sampling is performed by solving the reverse-time SDE. To do so, we generate a reverse-time trajectory.
+
+num_steps = 20
+timesteps = np.linspace(0, 1, num_steps)[::-1]
+
+# The solution is obtained by calling the SDE object with a desired solver (here, Euler).
+
+solution = sde(
```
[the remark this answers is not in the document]
mh-nguyen712: by full_trajectory we mean that we track the SDE trajectory as every time step and return it in the solution.

deepinv/models/edm/guided_diffusion.py
```diff
@@ -0,0 +1,218 @@
...
+    Linear,
+    UNetBlock,
+    Conv2d,
+    GroupNorm,
+)
+
+
+class ADMUNet(torch.nn.Module):
+    r"""
+    Re-implementation of the architecture from the paper: Diffusion Models Beat GANS on Image Synthesis (https://arxiv.org/abs/2105.05233).
```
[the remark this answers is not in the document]
mh-nguyen712: Added

deepinv/sampling/sde_solver.py
```diff
@@ -0,0 +1,206 @@
...
+
+        return (
+            x0
+            + 0.5 * (drift_0 + drift_1) * dt
+            + 0.5 * (diffusion_0 + diffusion_1) * dW,
+            2,
+        )
+
+
+def select_solver(name: str = "euler") -> BaseSDESolver:
```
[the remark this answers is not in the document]
mh-nguyen712: Similar to this: https://github.com/deepinv/deepinv/pull/361#discussion_r1873812588, I think we can discuss more on this later.

deepinv/sampling/sde.py
tachella: yes, I think it is better to be more specific and ask directly for a `NoisyDataFidelity`

deepinv/sampling/sde.py
tachella: I think it is not optimal to have multiple options that are dependent on your specific solver.

We have currently the same problem with `deepinv.optim.optim_builder` which asks for a dictionary `params_algo` whose keys can change according to the chosen algorithm - some users told me that they found this not very intuitive.

mh-nguyen712: Agree! I'll try to make the necessary changes asap.

mh-nguyen712: I just made the necessary changes

examples/sampling/demo_diffusion_sde.py
```diff
@@ -0,0 +1,467 @@
...
+
+# %%
+# When the data fidelity is given, together with the measurements and the physics, this class can be used to perform posterior sampling for inverse problems.
+# For example, consider the inpainting problem, where we have a noisy image and we want to recover the original image.
+
+x = solution.sample
+physics = dinv.physics.Inpainting(tensor_size=x.shape[1:], mask=0.5, device=device)
+y = physics(x)
+
+posterior = PosteriorDiffusion(
```
tachella, line 417: model = PosteriorDiffusion(...)

examples/sampling/demo_diffusion_sde.py
```diff
@@ -0,0 +1,467 @@
...
+    data_fidelity=DPSDataFidelity(denoiser=denoiser),
+    unconditional_sde=sde,
+    dtype=dtype,
+    device=device,
+)
+
+# To perform posterior sampling, we need to provide the measurements, the physics and the solver.
+posterior_sample = posterior.forward(
```
tachella, line 425: posterior_sample = model(y, ...)

examples/sampling/demo_diffusion_sde.py
```diff
@@ -0,0 +1,478 @@
...
+
+# %%
+#  When the data fidelity is not given, the posterior diffusion is equivalent to the unconditional diffusion.
+model = PosteriorDiffusion(
+    data_fidelity=Zero(),
+    unconditional_sde=sde,
+    dtype=dtype,
+    device=device,
+)
+sample = model.forward(
```
tachella, line 403:
```suggestion
sample = model(
```

examples/sampling/demo_diffusion_sde.py
```diff
@@ -0,0 +1,478 @@
+r"""
+Image generation and Posterior Sampling with Stochastic Differential Equation Modeling
+====================================================================================================
+
+This code shows you how to use
+:meth:`deepinv.sampling.DiffusionSDE` to generate from a pre-trained denoiser and
+:meth:`deepinv.sampling.PosteriorDiffusion` to perform posterior sampling.
```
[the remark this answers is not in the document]
mh-nguyen712: Thanks for your clarifications

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -0,0 +1,148 @@
...
+
+                \nabla_x \log p(y|x) = \left(\operatorname{Id}+\nabla_x D(x)\right)^\top A^\top \left(y-\forw{D(x)}\right)
+
+        .. note::
+            The preconditioning term is computed with autodiff.
+
+        :param torch.Tensor x: Current iterate.
+        :param torch.Tensor y: Input data.
+        :param physics: physics model
+        :param float sigma: Standard deviation of the noise. (unused)
```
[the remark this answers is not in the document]
mh-nguyen712: In fact it is used (line 117), it was a docstring mistake.

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -0,0 +1,148 @@
...
+        return self.d(physics.A(x), y)
+
+
+class DPSDataFidelity(NoisyDataFidelity):
+    r"""
+    The DPS data-fidelity term.
+
+    This corresponds to the :math:`p(y|x)` prior as proposed in `Diffusion Probabilistic Models <https://arxiv.org/abs/2209.14687>`_.
+
+    :param denoiser: Denoiser network. # TODO: type?
```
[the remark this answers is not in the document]
mh-nguyen712: Thanks!

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -0,0 +1,148 @@
...
+from deepinv.optim.data_fidelity import L2
+
+# This file implements the p(y|x) terms as proposed in the `review paper <https://arxiv.org/pdf/2410.00083>`_ by Daras et al.
+
+
+class NoisyDataFidelity(L2):
+    r"""
+    Preconditioned data fidelity term for noisy data :math:`\datafid{x}{y}=\distance{\forw{x'}}{y'}`.
+
+    This is a base class for the conditional classes for approximating :math:`\log p(y|x)` used in diffusion
```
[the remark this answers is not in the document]
mh-nguyen712: Yes only x is perturbed. I think we left :math:`x_t` for generality, as it is defined by the underlying SDE, not necessary :math:`x + \epsilon` (the OU-process with zero drift).
I corrected some docstring mistakes.

deepinv/sampling/diffusion_sde.py
```diff
@@ -0,0 +1,438 @@
...
+        *args,
+        **kwargs,
+    ):
+        def backward_drift(x, t, *args, **kwargs):
+            return -forward_drift(x, t) + ((1 + alpha) / 2) * forward_diffusion(
+                t
+            ) ** 2 * self.score(x, t, *args, **kwargs)
+
+        def backward_diffusion(t):
+            return np.sqrt(alpha) * forward_diffusion(t)
```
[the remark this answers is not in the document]
mh-nguyen712: alpha can be a float, which means that torch.sqrt(alpha) will not work. I changed it to alpha ** 0.5 so that it can work whenever it is a float or a tensor

deepinv/models/precond.py
```diff
@@ -0,0 +1,85 @@
+import torch
+
+# ----------------------------------------------------------------------------
+# Improved preconditioning proposed in the paper "Elucidating the Design
+# Space of Diffusion-Based Generative Models" (EDM).
+
+
+class EDMPrecond(torch.nn.Module):
```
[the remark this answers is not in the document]
mh-nguyen712: Obviously. Thanks for your feedback!

examples/sampling/demo_diffusion_sde.py
```diff
@@ -0,0 +1,478 @@
...
+# %%
+# Below, we show the trajectory and the sample generated by the VE-SDE with the DRUNet denoiser.
+
+# sphinx_gallery_start_ignore
+# cleanup
+import os
+import shutil
+from pathlib import Path
+
+try:
```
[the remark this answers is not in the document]
mh-nguyen712: Because when the domo is launched, it will create some images and GIFs. We just moved them to the auto-examples folder

mh-nguyen712:
> The PR is getting there, I think we can still improve some bits.
> 
> Apart from minor docstring fixes, my main concerns is:
> 
> Can we remove one layer of abstraction, and remove unconditional sdes? I would have as an input to PosteriorDiffusion the denoiser, data fidelity and the sde separately, and only do unconditional generation by setting the data fidelity to zero. Moreover, can we pass the choice of solver at the constructor of the class instead of the forward? So the diffusion could be created as
> 
> ```
> model = PosteriorDiffusion(
>     data_fidelity=Zero(),
>     denoiser=DRUNet(),
>     sde=VarianceExplodingDiffusion(sigma_min,sigma_max, alpha),
>     solver=EulerSolver(timesteps=100)
>     dtype=dtype,
>     device=device,
> )
> 
> x = model(y, physics=None) # uncond generation
> 
> x, trajectory = model(y, physics=None, get_trajectory=True) # uncond generation with trajectory
> ```
> 
> Other detail:
> 
> * We shouldn't use :meth: but :class: or :func: instead (otherwise we have a bug with the pydata template)
> * We use `sigma `instead of `noise_level `for `Denoisers`.

Many thanks for your review. I have made your requested changes and correct some minor docstring errors. Hope that this PR is getting better.

tachella:
many thanks for all the changes!

I still have one important point: Do you think we can add the denoiser to `PosteriorDiffusion` instead of adding it to the sde? I would also not use the sde on its own (as it is currently done in the example), always use it together with `PosteriorDiffusion`

I think that this change will make the creation of the diffusion method more direct, where both the prior and the data fidelity are passed to the main class `PosteriorDiffusion`

mh-nguyen712:
> many thanks for all the changes!
> 
> 
> 
> I still have one important point: Do you think we can add the denoiser to `PosteriorDiffusion` instead of adding it to the sde? I would also not use the sde on its own (as it is currently done in the example), always use it together with `PosteriorDiffusion`
> 
> 
> 
> I think that this change will make the creation of the diffusion method more direct, where both the prior and the data fidelity are passed to the main class `PosteriorDiffusion`

Agree with you on the uses of SDE in the example. However, I think that passing the sde to the PosteriorDiffusion allows user to sample the posterior with different underlying SDE. Otherwise we can pass a string indicating the sde and construct it under the hood. What do you think?

tachella:
> > many thanks for all the changes!
> > I still have one important point: Do you think we can add the denoiser to `PosteriorDiffusion` instead of adding it to the sde? I would also not use the sde on its own (as it is currently done in the example), always use it together with `PosteriorDiffusion`
> > I think that this change will make the creation of the diffusion method more direct, where both the prior and the data fidelity are passed to the main class `PosteriorDiffusion`
> 
> Agree with you on the uses of SDE in the example. However, I think that passing the sde to the PosteriorDiffusion allows user to sample the posterior with different underlying SDE. Otherwise we can pass a string indicating the sde and construct it under the hood. What do you think?

I think it is a good idea to pass the SDE, I would just pass the denoiser to `PosteriorDiffusion` instead of the SDE, ie define the SDE without a denoiser, and get the score from `PosteriorDiffusion` directly.

mh-nguyen712:
> > > many thanks for all the changes!
> > > I still have one important point: Do you think we can add the denoiser to `PosteriorDiffusion` instead of adding it to the sde? I would also not use the sde on its own (as it is currently done in the example), always use it together with `PosteriorDiffusion`
> > > I think that this change will make the creation of the diffusion method more direct, where both the prior and the data fidelity are passed to the main class `PosteriorDiffusion`
> > 
> > 
> > Agree with you on the uses of SDE in the example. However, I think that passing the sde to the PosteriorDiffusion allows user to sample the posterior with different underlying SDE. Otherwise we can pass a string indicating the sde and construct it under the hood. What do you think?
> 
> I think it is a good idea to pass the SDE, I would just pass the denoiser to `PosteriorDiffusion` instead of the SDE, ie define the SDE without a denoiser, and get the score from `PosteriorDiffusion` directly.

I made some changes on the `PosteriorDiffusion`. It now takes an SDE as input, but without the need of a denoiser inside the SDE class. However, I keep the optional denoiser parameter in the SDE, making the implementation more easier (since sometimes the unconditional score depends both on the denoiser and the sde). 

From the point of view of users, this changes nothing, we will recommend to use `PosteriorDiffusion` for sampling.

examples/sampling/demo_diffusion_sde.py
```diff
@@ -0,0 +1,539 @@
...
+)
+# Here, we plot the original image, the measurement and the posterior sample
+dinv.utils.plot(
+    [x, y, x_hat],
+    show=True,
+    titles=["Original", "Measurement", "Posterior sample DiffUNet"],
+    save_fn="posterior_sample_DiffUNet.png",
+    figsize=(figsize * 3, figsize),
+)
+# sphinx_gallery_start_ignore
```
tachella, line 398: This bits look a bit hardcoded, can't we just remove the cleanup?

In the rest of the examples we don't do clean up

examples/sampling/demo_diffusion_sde.py
```diff
@@ -0,0 +1,539 @@
...
+    dtype=dtype,
+)
+
+# %%
+# Reverse-time SDE as generative process
+# --------------------------------------
+#
+# Sampling is performed by solving the reverse-time SDE. To do so, we generate a reverse-time trajectory.
+
+sample_seed_1, trajectory_seed_1 = sde.sample(
```
tachella, line 496: I would remove this possibility in my opinion, I think giving multiple ways of doing the same thing can be confusing for users

examples/sampling/demo_diffusion_sde.py
```diff
@@ -0,0 +1,539 @@
...
+    solver=solver,
+    dtype=dtype,
+    device=device,
+)
+
+# To perform posterior sampling, we need to provide the measurements, the physics and the solver.
+x_hat, trajectory = model(
+    y,
+    physics,
+    x_init=(1, 3, 64, 64),
```
tachella, line 317: Is this needed? I think we can guess the size from physics.A_adjoint(y)

examples/sampling/demo_diffusion_sde.py
```diff
@@ -0,0 +1,539 @@
...
+# EDM framework: https://arxiv.org/pdf/2206.00364 .
+# The network architecture is from Song et al: https://arxiv.org/abs/2011.13456 .
+unet = NCSNpp.from_pretrained("edm-ffhq64-uncond-ve")
+denoiser = EDMPrecond(model=unet).to(device)
+
+# The solution is obtained by calling the SDE object with a desired solver (here, Euler).
+# The reproducibility of the SDE Solver class can be controlled by providing the pseudo-random number generator.
+num_steps = 150
+rng = torch.Generator(device).manual_seed(42)
+timesteps = np.linspace(0.001, 1, num_steps)[::-1]
```
tachella, line 75: Can we pass this in torch instead of numpy?

examples/sampling/demo_diffusion_sde.py
```diff
@@ -0,0 +1,539 @@
...
+    DPSDataFidelity,
+    EulerSolver,
+    VarianceExplodingDiffusion,
+)
+from deepinv.optim import ZeroFidelity
+
+# In this example, we use the pre-trained FFHQ-64 model from the
+# EDM framework: https://arxiv.org/pdf/2206.00364 .
+# The network architecture is from Song et al: https://arxiv.org/abs/2011.13456 .
+unet = NCSNpp.from_pretrained("edm-ffhq64-uncond-ve")
```
tachella, line 68: I would do like with the rest of the denoisers of the library 


```suggestion
unet = NCSNpp(pretrained="edm-ffhq64-uncond-ve")
```

examples/sampling/demo_diffusion_sde.py
mh-nguyen712: It will work for both torch tensor and numpy array. I'll change it to torch by default.

examples/sampling/demo_diffusion_sde.py
mh-nguyen712: It was a trick of Mathieu so that we can render the gif in the docs nicely. These lines of code won't be shown up in the documentation and the generated jupyter notebook.

examples/sampling/demo_diffusion_sde.py
mh-nguyen712: Yes this is a good remark, Thanks

examples/sampling/demo_diffusion_sde.py
mh-nguyen712: Yes updated it

examples/sampling/demo_diffusion_sde.py
```diff
@@ -0,0 +1,420 @@
...
+#
+# Let us import the necessary modules, define the denoiser and the SDE.
+#
+# In this example, we use the Variance-Exploding SDE, whose forward process is defined as:
+#
+# .. math::
+#     d\, x_t = \sigma(t) d\, w_t \quad \mbox{where } \sigma(t) = \sigma_{\mathrm{min}}\left( \frac{\sigma_{\mathrm{max}}}{\sigma_{\mathrm{min}}}\right)^t
+
+import torch
+import numpy as np
```
tachella, line 48: Do we still need numpy?

docs/source/user_guide/reconstruction/sampling.rst
```diff
@@ -7,16 +7,117 @@ This package contains posterior sampling algorithms.
...
+    d\, x_t = f(x_t, t) d\,t + g(t) d\, w_t.
+
+Let :math:`p_t` denote the distribution of the random vector :math:`x_t`.
+The reverse-time SDE is defined as follows, running backward in time:
+
+.. math::
+
+    d\, x_{t} = \left( f(x_t, t) - \frac{1 + \alpha}{2} g(t)^2 \nabla \log p_t(x_t) \right) d\,t + g(t) \sqrt{\alpha} d\, w_{t}.
+
+The scalar :math:`\alpha \in [0,1]` weighting the diffusion term. :math:`\alpha = 0` corresponds to the ODE sampling and :math:`\alpha > 0` corresponds to the SDE sampling.
```
tachella, line 37:
```suggestion
The scalar :math:`\alpha \in [0,1]` weighting the diffusion term. :math:`\alpha = 0` corresponds to the ordinary differential equation (ODE) sampling and :math:`\alpha > 0` corresponds to the SDE sampling.
```

docs/source/user_guide/reconstruction/sampling.rst
```diff
@@ -7,16 +7,117 @@ This package contains posterior sampling algorithms.
...
+then
+
+.. math::
+
+    \nabla \log p_t(x_t) = \frac{\left(\mu_t D_{\sigma_t}(x_t) -  x_t \right)}{\sigma_t^2}.
+
+Starting from a random point following the end-point distribution :math:`p_T` of the forward process, 
+solving the reverse-time SDE gives us a sample of the data distribution :math:`p_0`.
+
+When turn out to Posterior Sampling, we need simply to repace the (unconditional) score function :math:`\nabla_{x_t} \log p_t(x_t)` by the conditional score function :math:`\nabla_{x_t} \log p_t(x_t|y)`. It can be decomposed using the Bayes' rule:
```
tachella, line 56:
```suggestion
In the case of posterior sampling, we need simply to repace the (unconditional) score function :math:`\nabla_{x_t} \log p_t(x_t)` by the conditional score function :math:`\nabla_{x_t} \log p_t(x_t|y)`. The conditional score can be decomposed using the Bayes' rule:
```

docs/source/user_guide/reconstruction/sampling.rst
```diff
@@ -7,16 +7,117 @@ This package contains posterior sampling algorithms.
...
+     - Base class for defining a SDE with a drift term and a diffusion coefficient
+
+   * - :class:`deepinv.sampling.DiffusionSDE`
+     - Define automatically the reverse-time SDE from a forward SDE and a denoiser. 
+
+   * - :class:`deepinv.sampling.VarianceExplodingDiffusion`
+     - The Variance-Exploding SDE, an instance of :meth:`deepinv.sampling.DiffusionSDE`
+
+   * - :class:`deepinv.sampling.PosteriorDiffusion`
+     - The Diffusion SDE class for Posterior Sampling, an subclass of :class:`deepinv.models.base.Reconstructor`
```
tachella, line 80:
```suggestion
     - The Diffusion SDE class for posterior sampling, an subclass of :class:`deepinv.models.base.Reconstructor`
```

docs/source/user_guide/reconstruction/sampling.rst
```diff
@@ -7,16 +7,117 @@ This package contains posterior sampling algorithms.
...
+
+
+
 .. _diffusion:
 
-Diffusion
----------
+Diffusion models for posterior sampling
+---------------------------------------
 We provide various sota diffusion methods for sampling from the posterior distribution.
```
tachella, line 121: We also provide custom implementations of some popular diffusion methods for posterior sampling.

docs/source/user_guide/reconstruction/sampling.rst
```diff
@@ -7,16 +7,117 @@ This package contains posterior sampling algorithms.
...
 Diffusion methods produce a sample from the posterior ``x`` given a
```
tachella, line 122: I would remove these lines I think as they're a bit redundant, and go directly to the list

docs/source/user_guide/reconstruction/sampling.rst
```diff
@@ -7,16 +7,117 @@ This package contains posterior sampling algorithms.
...
+   * - :class:`deepinv.sampling.HeunSolver`
+     - `Second order Heun solver. <https://en.wikipedia.org/wiki/Heun%27s_method>`_
+
+
+
 .. _diffusion:
 
-Diffusion
----------
+Diffusion models for posterior sampling
```
tachella, line 119: Custom diffusion posterior samplers


I would also put it as a subsection instead of a whole different section

docs/source/api/deepinv.sampling.rst
```diff
@@ -4,8 +4,30 @@ deepinv.sampling
...
+    deepinv.sampling.PosteriorDiffusion
+    deepinv.sampling.NoisyDataFidelity
+    deepinv.sampling.DPSDataFidelity
+    deepinv.sampling.BaseSDESolver
+    deepinv.sampling.EulerSolver
+    deepinv.sampling.HeunSolver
+    deepinv.sampling.SDEOutput
+
+
+Diffusion models for posterior sampling
```
tachella, line 28: Custom diffusion posterior samplers

deepinv/tests/test_sampling.py
```diff
@@ -186,3 +185,107 @@ def test_algo_inpaint(name_algo, device):
...
+        solver=solvers[0],
+        rescale=rescales[0],
+        dtype=torch.float64,
+        device=device,
+    )
+    x = sample_2
+    physics = dinv.physics.Inpainting(tensor_size=x.shape[1:], mask=0.5, device=device)
+    y = physics(x)
+
+    x_hat = posterior.forward(
```
tachella, line 284:
```suggestion
    x_hat = posterior(
```

deepinv/optim/data_fidelity.py
```diff
@@ -515,3 +515,44 @@ def __init__(self, N0=1024.0, mu=1 / 50.0):
     prox = data_fidelity.prox(
         x, y, physics, gamma=1.0
     )  # print prox_fA gives [0.6000, 3.6000]
+
+
+class ZeroFidelity(DataFidelity):
```
tachella, line 520: We should test this class

deepinv/models/ncsnpp.py
```diff
@@ -0,0 +1,312 @@
...
+from torch.nn import Linear, GroupNorm
+from .utils import get_weights_url
+
+
+class NCSNpp(Denoiser):
+    r"""Re-implementation of the DDPM++ and NCSN++ architectures from the paper: `Score-Based Generative Modeling through Stochastic Differential Equations <https://arxiv.org/abs/2011.13456>`_.
+
+    Equivalent to the original implementation by Song et al., available at `the official implementation <https://github.com/yang-song/score_sde_pytorch>`_.
+
+    The architecture consists of a series of convolution layer, down-sampling residual blocks and up-sampling residual blocks with skip-connections of scale :math:`\sqrt{0.5}`.
```
tachella, line 21:
```suggestion
    The architecture consists of a series of convolution layers, down-sampling residual blocks and up-sampling residual blocks with skip-connections of scale :math:`\sqrt{0.5}`.
```

examples/sampling/demo_diffusion_sde.py
```diff
@@ -0,0 +1,420 @@
...
+     d\, x_t = \left( f(x_t, t) - \frac{1 + \alpha}{2} g(t)^2 \nabla_{x_t} \log p_t(x_t | y) \right) d\,t + g(t) \sqrt{\alpha} d\, w_{t}
+
+where :math:`f` is the drift term, :math:`g` is the diffusion coefficient and :math:`w` is the standard Brownian motion. The drift term and the diffusion coefficient are defined by the underlying (unconditional) forward-time SDE `sde`. The (conditional) score function :math:`\nabla_{x_t} \log p_t(x_t | y)` can be decomposed using the Bayes' rule:
+
+.. math::
+     \nabla_{x_t} \log p_t(x_t | y) = \nabla_{x_t} \log p_t(x_t) + \nabla_{x_t} \log p_t(y | x_t).
+
+The first term is the score function of the unconditional SDE, which is typically approximated by a MMSE denoiser (`denoiser`) using the well-known Tweedie's formula, while the
+second term is approximated by the (noisy) data-fidelity term (`data_fidelity`).
+We implement various data-fidelity terms in :class:`deepinv.sampling.NoisyDataFidelity`.
```
tachella, line 34: I would point here to the docs section with the list of noisy data fidelities instead of the base class

tachella (changes requested):

tachella:
Last series of comments from my side (all quite minor), otherwise I think we can merge!

Thanks for this amazing work

deepinv/optim/data_fidelity.py
mh-nguyen712: Yes I'll add test function. Good remark!

examples/sampling/demo_diffusion_sde.py
mh-nguyen712: Updated!

examples/sampling/demo_diffusion_sde.py
```diff
@@ -0,0 +1,420 @@
...
+timesteps = torch.linspace(1, 0.001, 200)
+solver = EulerSolver(timesteps=timesteps, rng=rng)
+
+denoiser = dinv.models.DRUNet(pretrained="download").to(device)
+
+sde = VarianceExplodingDiffusion(
+    sigma_max=sigma_max, sigma_min=sigma_min, alpha=0.1, device=device, dtype=dtype
+)
+model = PosteriorDiffusion(
+    data_fidelity=dinv.optim.L2(sigma=0.6),
```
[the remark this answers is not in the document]
mh-nguyen712: I justed took a random value. I'll set leave the sigma value to the default value

mh-nguyen712:
> Last series of comments from my side (all quite minor), otherwise I think we can merge!
> 
> Thanks for this amazing work

Thanks again for your review! I think it's ready to merge!

tachella (approved):

PR #375, opened 2024-12-12
Reformat the way optimisation algorithms are defined 

samuro95:
New way to create an instance of the BaseOptim, BaseUnfolded or BaseDEQ class. Instead of using the optim_builder, we can now directly call the name of the algorithm as class name, for instance model = GradientDescent(...).  

- The classes for each algorithm are defined in optimizers.py. All the algo parameters are now explicit variables. Possibility to set them as unfolded or DEQ algorithms directly. For example ProximalGradientDescent(stepsize = 1., lambda_reg = 1.,..., unfold = True, ...)
- Change the test_optim and test_infolded accordingly. Also corrected some problems in the tests and added some Unfolded and DEQ tests. Added tests for Mirror Descent.
- Fixed bugs in Proximal Mirror Descent


examples/unfolded/None/convergence_metrics/curves.png
tachella, line 1: I think this shouldn't be in the PR?

pyproject.toml
```diff
@@ -64,6 +64,7 @@ doc = [
     "sphinx_copybutton",
     "sphinx_autoapi",
     "sphinx-design",
+    "sphinxcontrib-bibtex",
```
tachella, line 67: this shouldn't be in the bib PR?

tachella (changes requested):
Many thanks for the refactoring!

I have a couple of questions/comments:

- what happens with unrolled? I think it would be nice to add a `train` option to each of the classes to distinguish PnP and unrolled
- I would remove the `params_algo` input, and instead set specific input params for each iterator, ie. 

``` 
model = dinv.optim.ProximalGradientDescent(prior=prior, data_fidelity=data_fidelity, stepsize=1.0, g_param=0.1, train=False)
```
 
- I think we need to add the docstrings for each class, eg. `ProximalGradientDescent` (copying mostly what is explained in the iterator docs), but also specifying each algo's parameter ie `stepsize`, `g_param` etc (see point above)

PR #379, opened 2024-12-20
Docs hyperlinks fixes

tachella:
There are many hyperlinks in the current doc that do not work in the library, mostly due to bad usage of refs.

- added the `nitpicky = True` option to the sphinx configuration file `conf.py` to raise warnings each time there is a broken link in the doc. I found out a huge number of errors which I've fixed one by one.

- added a `-W` option to the sphinx builders in the Github action to convert any warnings into errors, in order to avoid these problems piling up in the future.

- Set single backquotes `` `code here` ``as default for code mode instead of italics, as many contributors prefer this way since it matches the markdown default. Double backquotes will keep working as well for code mode.

- Added to automatic typing for the docs in the `conf.py` file

```
# for python3 type hints
autodoc_typehints = "description"
autodoc_typehints_description_target = "documented"
# to handle functions as default input arguments
autodoc_preserve_defaults = True
```

Now it is possible to define the types following the python 3 standard:
```
def foo(a : torch.Tensor) -> torch.Tensor:
    """
    Function foo

    :param a: input tensor. 
    """
```
which will automatically add `torch.Tensor` to the input parameter and return type documentation, i.e. it is not mandatory to write `:param torch.Tensor a: input tensor.` (but it is still possible to do so).

There is however a catch.
If we write
```
from torch import Tensor
def foo(a : Tensor) -> Tensor:
    """
    Function foo

    :param a: input tensor. 
    """
```
This will also work, but the documentation will say that the type is `Tensor` and not `torch.Tensor` which is a bit less clear in my opinion.


Andrewwango:
@tachella 
> Set single backquotes...

Is this backward compatible with people still doing double backtick?

> Added to automatic typing

Is it still nice and clean when you have complex typing e.g. `x: Union[torch.Tensor, deepinv.utils.TensorList]`?
How about the [new notation from python 3.10](https://peps.python.org/pep-0604/) of `x: Tensor | TensorList`?

tachella:
> @tachella
> 
> > Set single backquotes...
> 
> Is this backward compatible with people still doing double backtick?

Yes, both will work

> 
> > Added to automatic typing
> 
> Is it still nice and clean when you have complex typing e.g. `x: Union[torch.Tensor, deepinv.utils.TensorList]`? How about the [new notation from python 3.10](https://peps.python.org/pep-0604/) of `x: Tensor | TensorList`?

Yes, both work, with the same flaws as stated above (ie only showing `Tensor` and not `torch.Tensor`

deepinv/datasets/fastmri.py
```diff
@@ -45,27 +45,27 @@ class FastMRISliceDataset(torch.utils.data.Dataset):
...
     :param bool load_metadata_from_cache: Whether to load dataset metadata from cache.
     :param bool save_metadata_to_cache: Whether to cache dataset metadata.
-    :param Union[str, Path] metadata_cache_file: A file used to cache dataset
+    :param Union[str, pathlib.Path] metadata_cache_file: A file used to cache dataset
         information for faster load times.
-    :param callable, optional sample_filter: A callable object that takes a
-        :meth:`SliceSampleFileIdentifier` as input and returns a boolean indicating
-        whether the sample should be included in the dataset.
-    :param float, optional sample_rate: A float between 0 and 1. This controls what
+    :param Callable sample_filter: (optional) A callable object that takes a
```
Andrewwango, line 55: I still don't fully understand what you mean by the links not working. For example I have just regenerated your docs - is this the intended output? E.g. Union and Callable aren't linked.
image

tachella: so `Union` is not generally shown as a link, and `Callable` is the only other example that doesn't show a link without flagging an error (not sure why).

deepinv/loss/adversarial/base.py
```diff
@@ -119,10 +124,10 @@ def __init__(
...
         r"""Typical adversarial loss in GAN discriminators.
 
-        :param Tensor real: image labelled as real, typically one originating from training set
-        :param Tensor fake: image labelled as fake, typically a reconstructed image
-        :param nn.Module D: discriminator/critic/classifier model. If None, then D passed from __init__ used. Defaults to None.
-        :return Tensor: discriminator adversarial loss
+        :param torch.Tensor real: image labelled as real, typically one originating from training set
+        :param torch.Tensor fake: image labelled as fake, typically a reconstructed image
+        :param torch.nn.Module D: discriminator/critic/classifier model. If None, then D passed from __init__ used. Defaults to None.
+        :return: (class:`torch.Tensor`) discriminator adversarial loss
```
Andrewwango, line 130:
```suggestion
        :return: (:class:`torch.Tensor`) discriminator adversarial loss
```

deepinv/training/trainer.py
```diff
@@ -586,7 +586,7 @@ def step(self, epoch, progress_bar, train=True, last_batch=False):
         It performs the forward pass, the backward pass, and the evaluation at each iteration.
 
         :param int epoch: Current epoch.
-        :param tqdm progress_bar: Progress bar.
+        :param progress_bar: tqdm progress bar.
```
Andrewwango, line 589: note that here, no type appears in the docs anymore for progress bar

deepinv/loss/measplit.py
```diff
@@ -154,9 +159,7 @@ def forward(self, x_net, y, physics, model, **kwargs):
 
         return loss_ms / mask2.mean()  # normalize loss
 
-    def adapt_model(
-        self, model: torch.nn.Module, eval_n_samples=None
-    ) -> SplittingModel:
+    def adapt_model(self, model: torch.nn.Module, eval_n_samples=None):
```
Andrewwango, line 162: Why remove the return typing for this method?

deepinv/loss/measplit.py
```diff
@@ -196,122 +199,137 @@ def adapt_model(
...
-        self.pixelwise = pixelwise
-
-    @staticmethod
-    def split(mask, y, physics):
-        return SplittingLoss.split(mask, y, physics)
-
-    def forward(
-        self, y: torch.Tensor, physics: Physics, update_parameters: bool = False
-    ):
+    class SplittingModel(Reconstructor):
```
Andrewwango, line 202: what's the purpose of including SplittingModel inside SplittingLoss?

tachella: To have it appearing in the docs, before it was not the case.

Andrewwango (changes requested):
Thanks for the hard work, please see comments above.

deepinv/loss/measplit.py
tachella: you are right, I undid the changes

deepinv/loss/adversarial/base.py
tachella: fixed

PR #382, opened 2025-01-02
API make the device argument more consistent for denoisers

### LINKED ISSUE #381: API inconsistent setting of device for various models
It is unclear what is the correct way to set the device for the various models, particularly for the denoisers.

The API specified in the base class `Denoiser/Reconstructor` states it should be passed at initialization.
However, the argument is not always present in the child class and is often not passed to the base class constructor.
It makes sense as calling `to(device)` in the `__init__` method of `Denoiser` before adding submodules in the subclass would have no effect.

There are various ways to make the API more constistent:
- we impose that all denoisers take in the `device` argument and call the following snippet at the end if it makes sense.
  ```python
  if device is not None:
      self.to(device)
  ```
- we do some black magic to try to make this automatic (with a wrapper class), but it feels like an overkill.
- we remove this argument and suggest using the `to` method to move the class to the expected device.

WDYT is the right behavior? I think the first option is the easiest one and can be easily tested so I would go for this one.

tomMoral:
This PR intends to make the `device` argument of the `Denoiser` more consistent.
If you are okay with the idea that all denoisers should take in the `device` argument, we can also add some unit tests to make sure this is the case.

Fixes #381

### DIFF
--- a/deepinv/models/base.py
+++ b/deepinv/models/base.py
@@ -19,10 +19,6 @@ class Denoiser(torch.nn.Module):
 
     """
 
-    def __init__(self, device="cpu"):
-        super().__init__()
-        self.to(device)
-
     def forward(self, x, sigma, **kwargs):
         r"""
         Applies denoiser :math:`\denoiser{x}{\sigma}`.
@@ -46,10 +42,6 @@ class Reconstructor(torch.nn.Module):
     The base class inherits from :class:`torch.nn.Module`.
 
     """
-
-    def __init__(self, device="cpu"):
-        super().__init__()
-        self.to(device)
 
     def forward(self, y, physics, **kwargs):
         r"""
--- a/deepinv/models/diffunet.py
+++ b/deepinv/models/diffunet.py
@@ -45,6 +45,7 @@ class DiffUNet(Denoiser):
         (only available for 3 input and output channels).
         Finally, ``pretrained`` can also be set as a path to the user's own pretrained weights.
         See :ref:`pretrained-weights <pretrained-weights>` for more details.
+    :param NoneType, torch.device device: Instruct our module to be either on cpu or on gpu. Default to ``None``, which suggests working on cpu.
     """
 
     def __init__(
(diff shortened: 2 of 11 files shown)


deepinv/models/base.py
```diff
@@ -19,10 +19,6 @@ class Denoiser(torch.nn.Module):
 
     """
 
-    def __init__(self, device="cpu"):
```
tachella, line 22: shouldn't we leave some abstract definition of the init so people implementing new denoisers follow the signature?

deepinv/models/diffunet.py
```diff
@@ -45,6 +45,7 @@ class DiffUNet(Denoiser):
         (only available for 3 input and output channels).
         Finally, ``pretrained`` can also be set as a path to the user's own pretrained weights.
         See :ref:`pretrained-weights <pretrained-weights>` for more details.
+    :param NoneType, torch.device device: Instruct our module to be either on cpu or on gpu. Default to ``None``, which suggests working on cpu.
```
Andrewwango, line 48:
```suggestion
    :param torch.device, str, None device: Instruct our module to be either on cpu or on gpu. Default to ``None``, which suggests working on cpu.
```

deepinv/models/base.py
tomMoral: I tried with abstract method but it does not seem to work....

The core issue here is that we need to call the `__init__` method of `torch.nn.Module` first and then call at the  end `self.to(device)` to move all parameters on the GPU.
When sub-classing, with the abstract method, we force the implementation of `__init__`, but when calling `super().__init__()` in the new class, it is hard to bypass the `denoiser.__init__` call.
The easier solution would be to not have the `__init__`, but just a template in the doc, but maybe it is not satisfactory.
Not sure of what's the best solution here.


```python
from abc import ABC, abstractmethod

class A:
    def __init__(self):
        print("init A")

class B(A, ABC):
    @abstractmethod
    def __init__(self):
        print("init B")

class C(B):
    def __init__(self):
        super().__init__()
        print("init C")
C()
```
You can see here that we are not bypassing `B` so if we put `self.to(device)` in it, we will call the method before creating the parameters.
```
init B
init C
```

tomMoral:
Closing this as this seems to have been fixed in #530

PR #341, opened 2024-10-29
Integrate flow-based model with conditioning on LinearPhysics class of inverse problem 

MaximeSongIdris:
### Checks to be done before submitting your PR
- [ ] `python3 -m pytest tests/` runs successfully.
- [ ] `black .` runs successfully.
- [ ] `make html` runs successfully (in the `docs/` directory).
- [ ] Updated docstrings related to the changes (as applicable).
- [ ] Added an entry to the [CHANGELOG.rst](../CHANGELOG.rst).

PR #384, opened 2025-01-09
Fixes

matthieutrs:
#341 is moving to draft for some of it functionalities are not expected to be ready soon but some bug fixes found along the way are still relevant


deepinv/physics/mri.py
```diff
@@ -399,6 +409,18 @@ def A(self, x, mask=None, coil_maps=None, **kwargs):
...
+        r"""
+        Incorporates noise into the measurements :math:`\tilde{y} = N(y)` and takes the mask into account.
+
+        :param torch.Tensor x:  clean measurements
+        :param None, float noise_level: optional noise level parameter
+        :return: noisy measurements
+
+        """
+
+        return self.mask[:, :, None] * self.noise_model(x, **kwargs)
```
tachella, line 422:
```suggestion
        return self.mask[..., None] * self.noise_model(x, **kwargs)
```

deepinv/physics/noise.py
```diff
@@ -117,6 +144,73 @@ def __init__(self, sigma=0.1, rng: torch.Generator = None):
         super().__init__(rng=rng)
         self.update_parameters(sigma=sigma)
 
+    def __add__(self, other):
+        r"""
+        Sum of 2 gaussian noises via + operator.
+
+        :math:`N(x) = \sqrt{N_1(x)^2 + N_2(x)^2}`
```
tachella, line 151:
```suggestion
        :math:`\sigma = \sqrt{\sigma_1^2 + \sigma_2^2}`
```

deepinv/physics/noise.py
```diff
@@ -117,6 +144,73 @@ def __init__(self, sigma=0.1, rng: torch.Generator = None):
...
+
+        :param deepinv.physics.GaussianNoise other: Gaussian noise :math:`N_2`
```
tachella, line 153:
```suggestion
        :param deepinv.physics.GaussianNoise other: Gaussian with standard deviation :math:`\sigma`
```

deepinv/physics/noise.py
```diff
@@ -117,6 +144,73 @@ def __init__(self, sigma=0.1, rng: torch.Generator = None):
...
+        r"""
+        Element-wise multiplication of a GaussianNoise via * operator.
+
+        0) If `other` is a :class:`NoiseModel`, then applies the NoiseModel multiplication.
+
+        1) If `other` is a :class:`float`, then the standard deviation of the GaussianNoise is multiplied by `other`.
+
+            | :math:`x=[x_1, ..., x_b]` a batch of images.
+            | :math:`\lambda` a float.
+            | :math:`[\lambda \times gaussian](x) = [\lambda \times gaussian(x_1), ..., \lambda \times gaussian(x_b)]`
```
tachella, line 172:
```suggestion
            | :math:`\sigma = [\lambda \times \sigma_1, ..., \lambda \times \sigma_b]`
```

deepinv/physics/noise.py
```diff
@@ -117,6 +144,73 @@ def __init__(self, sigma=0.1, rng: torch.Generator = None):
...
+
+        2) If `other` is a :class:`torch.Tensor`, then the standard deviation of the GaussianNoise is multiplied by `other`.
+
+            | :math:`x=[x_1, ..., x_b]` a batch of images.
+            | :math:`other=[[[[\lambda_1]]], ..., [[[\lambda_b]]]]` a batch of scaling factors.
+            | :math:`[other \times gaussian](x) = [\lambda_1 \times gaussian(x_1), ..., \lambda_b \times gaussian(x_b)]`
```
tachella, line 178:
```suggestion
           | :math:`\sigma = [\lambda \times \sigma_1, ..., \lambda \times \sigma_b]`
```

deepinv/physics/noise.py
```diff
@@ -129,12 +223,11 @@ def forward(self, x, sigma=None, seed=None, **kwargs):
...
         self.update_parameters(sigma=sigma)
-
-        if isinstance(sigma, torch.Tensor):
-            sigma = sigma[(...,) + (None,) * (x.dim() - 1)]
+        if isinstance(self.sigma, torch.Tensor):
+            sigma = self.sigma[(...,) + (None,) * (x.dim() - 1)]
         else:
             sigma = self.sigma
-        return x + self.randn_like(x, seed=seed) * sigma
+        return x + self.randn_like(x, seed=seed) * sigma.to(x.device)
```
tachella, line 230: I hope this doesn't slow down when sigma is already in the right device?

deepinv/training/adversarial.py
tachella, line 1: @Andrewwango, do you approve these changes?

deepinv/training/trainer.py
```diff
@@ -706,8 +737,13 @@ def plot(self, epoch, physics, x, y, x_net, train=True):
             else:
                 x_nl = None
 
+            if show_y:
```
tachella, line 740: this should be removed I think, since it is possible to show y, by choosing the `no_learning_method='y'` in the trainer

tachella (changes requested):
many thanks for the fixes! 

please see a few comments/corrections below

deepinv/physics/noise.py
matthieutrs: I did not notice a slowdown but I can benchmark more seriously

deepinv/training/trainer.py
matthieutrs: I am not sure - showing y and x_nl are two different things that may be used simultaneously, no? In my case I've used this as debugging, to check that in null-space of y in case of MRI there was no noise, while willing to check the adjoint at the same time.

tachella: I would remove it if you don't mind because it is a little hacky at the moment, otherwise we should add it to the parameters of the trainer and explain its usage

deepinv/physics/noise.py
tachella: ok

pyproject.toml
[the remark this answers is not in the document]
matthieutrs: due to a bug in setuptools at the install coming from one of the subpackages from pyqia, I'm trying to fix

deepinv/models/diffunet.py
```diff
@@ -292,12 +292,70 @@ def forward(self, x, t, y=None, type_t="noise_level"):
...
+            else:
+                raise ValueError('type_t must be either "timestep" or "noise_level"')
+            return out[..., pad[-2] :, pad[-4] :]
         else:
-            raise ValueError('type_t must be either "timestep" or "noise_level"')
+            return self.patch_forward(x, t, y=y, type_t=type_t, patch_size=512)
+
+    def patch_forward(self, x, t, y=None, type_t="noise_level", patch_size=512):
+        """
+        Splits an image tensor into patches, applies the model to each patch, and reconstructs the full image.
```
tachella, line 310:
```suggestion
        Splits an image tensor into patches (without overlapping), applies the model to each patch, and reconstructs the full image.
```

deepinv/physics/generator/downsampling.py
```diff
@@ -0,0 +1,103 @@
+import torch
+from typing import List
+from deepinv.physics.generator import PhysicsGenerator
+from deepinv.physics.blur import gaussian_blur, bilinear_filter, bicubic_filter
+
+
+class DownsamplingGenerator(PhysicsGenerator):
```
tachella, line 7: we should add this new class to the API and user guide

deepinv/physics/generator/downsampling.py
```diff
@@ -0,0 +1,103 @@
...
+    r"""
+    Random downsampling generator.
+
+    Generates random downsampling factors and filters.
```
tachella, line 11: should we add a ref to the Downsampling class?

deepinv/physics/generator/noise.py
```diff
@@ -56,3 +56,36 @@ def step(self, batch_size=1, seed: int = None, **kwargs):
             + self.sigma_min
         )
         return {"sigma": sigma}
+
+
+class GainGenerator(PhysicsGenerator):
```
tachella, line 61:
```suggestion
class GainGenerator(PhysicsGenerator):
   r"""

    Generator for the noise level :math:`\gamma` in the :class:`Poisson noise model <deepinv.physics.PoissonNoise>`.

    The gain is sampled uniformly from the interval :math:`[\gamma_\text{min}, \gamma_\text{max}]`.

    :param float gain_min: minimum noise level
    :param float gain_max: maximum noise level
    :param torch.Generator rng: random number generator
    :param str device: device where the tensor is stored
    :param torch.dtype dtype: data type of the generated tensor.

    |sep|

    :Examples:

    >>> from deepinv.physics.generator import SigmaGenerator
    >>> generator = SigmaGenerator()
    >>> sigma_dict = generator.step(seed=0) # dict_keys(['sigma'])
    >>> print(sigma_dict['sigma'])
    tensor([0.2532])

    """
```

deepinv/physics/remote_sensing.py
```diff
@@ -33,6 +33,8 @@ class Pansharpen(StackedLinearPhysics):
     :param str padding: options are ``'valid'``, ``'circular'``, ``'replicate'`` and ``'reflect'``.
         If ``padding='valid'`` the blurred output is smaller than the image (no padding)
         otherwise the blurred output has the same size as the image.
+    :param bool normalize: if ``True``, normalize the downsampling operator.
```
tachella, line 36:
```suggestion
    :param bool normalize: if ``True``, normalize the downsampling operator to have unit norm.
```

deepinv/training/adversarial.py
```diff
@@ -148,6 +148,7 @@ class AdversarialTrainer(Trainer):
     losses_d: Union[Loss, List[Loss]] = None
     D: Module = None
     step_ratio_D: int = 1
+    global_optimizer_step: bool = False
```
tachella, line 151: we should add this param to the docs

tachella: ie add this 

    :param bool global_optimizer_step: If ``True``, the optimizer step is performed once on all datasets. If ``False``, the optimizer step is performed on each dataset separately.

deepinv/physics/generator/noise.py
tachella, line 61: we should also add this to the user guide and API

tachella (commented):
some last points to consider before merging :)

deepinv/physics/generator/downsampling.py
matthieutrs: done

deepinv/training/adversarial.py
matthieutrs: sorry this got scrapped in a merge with main, adding it again

deepinv/physics/generator/noise.py
matthieutrs: done

deepinv/tests/test_physics.py
```diff
@@ -672,6 +677,60 @@ def test_MRI(mri, mri_img_size, device, rng):
...
+    if rng.device != device:
+        rng = torch.Generator(device=device).manual_seed(0)
+    x, y = (
+        torch.rand(mri_img_size, generator=rng, device=device) + 1,
+        torch.rand(mri_img_size, generator=rng, device=device) + 1,
+    )
+
+    coil_maps_kwarg = {}
+
+    if mri is MRI:
```
Andrewwango, line 700:
```suggestion
    if isinstance(mri, MRI):
```

deepinv/tests/test_physics.py
```diff
@@ -672,6 +677,60 @@ def test_MRI(mri, mri_img_size, device, rng):
...
+        x = x[:, :, 0, :, :]
+        y = y[:, :, 0, :, :]
+    elif mri is MultiCoilMRI:
```
Andrewwango, line 703:
```suggestion
    elif isinstance(mri, MultiCoilMRI):
```

deepinv/tests/test_physics.py
```diff
@@ -672,6 +677,60 @@ def test_MRI(mri, mri_img_size, device, rng):
...
+        # y treat T as coil dim for tests
+        x = x[:, :, 0, :, :]
+        coil_maps_kwarg = {"coil_maps": T}
+
+    for mask_size in [(H, W), (T, H, W), (C, T, H, W), (B, C, T, H, W)]:
+        # Remove time dim for static MRI
+        _mask_size = mask_size if mri is DynamicMRI else mask_size[:-3] + mask_size[-2:]
+
+        mask, mask2 = (
```
Andrewwango, line 712: What is mask2 for?

deepinv/physics/mri.py
```diff
@@ -403,6 +413,22 @@ def A(self, x, mask=None, coil_maps=None, **kwargs):
...
+        r"""
+        Incorporates noise into the measurements :math:`\tilde{y} = N(y)` and takes the mask into account.
+
+        :param torch.Tensor x:  clean measurements
+        :param None, float noise_level: optional noise level parameter
+        :return: noisy measurements
+
+        """
+        mask = self.mask
+        if mask.dim() == x.dim() - 1:
```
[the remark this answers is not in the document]
Andrewwango: and `self.check_mask` should always be called in `A` anyway - so this should be redundant anyway

deepinv/physics/mri.py
```diff
@@ -280,6 +280,16 @@ def A_adjoint(
...
         return x  # (B,C,...,H,W) where C=1 if mag else 2
 
+    def noise(self, x, **kwargs):
+        r"""
+        Incorporates noise into the measurements :math:`\tilde{y} = N(y)`
+
+        :param torch.Tensor x:  clean measurements
+        :return torch.Tensor: noisy measurements
+        """
+        noise = self.U(self.noise_model(x, **kwargs) * self.mask)
```
Andrewwango, line 290: I would be tempted to generalise this to
```python
def noise(self, x, **kwargs):
    return super().noise(x, **kwargs) * self.mask
```
And then `MultiCoilMRI` would have the same implementation?

deepinv/physics/mri.py
```diff
@@ -280,6 +280,16 @@ def A_adjoint(
             x = self.crop(x, crop=crop)
         return x  # (B,C,...,H,W) where C=1 if mag else 2
 
+    def noise(self, x, **kwargs):
```
Andrewwango, line 283: Finally, do you think it's worth having a class parameter e.g. `mask_noise`?

Otherwise, the change might affect lots of users who use noisy MRI. We could set the default to False but raise a warning that noise is being added where it shouldn't.

deepinv/training/trainer.py
```diff
@@ -95,7 +95,8 @@ class Trainer:
         Should be used in conjunction with ``online_measurements=True``, no effect when ``online_measurements=False``. Also see ``loop_random_online_physics``.
     :param bool loop_random_online_physics: if True, resets the physics generator **and** noise model back to its initial state at the beginning of each epoch,
         so that the same measurements are generated each epoch. Requires `shuffle=False` in dataloaders. If False, generates new physics every epoch.
-        Used in conjunction with ``physics_generator`` and ``online_measurements=True``, no effect when ``online_measurements=False``.
```
Andrewwango, line 98: why remove this?

deepinv/training/trainer.py
```diff
@@ -523,24 +528,27 @@ def compute_loss(self, physics, x, y, train=True, epoch: int = None):
...
 
-            meters = self.logs_total_loss_train if train else self.logs_total_loss_eval
-            meters.update(loss_total.item())
-            logs[f"TotalLoss"] = meters.avg
+                meters = (
+                    self.logs_total_loss_train if train else self.logs_total_loss_eval
+                )
+                meters.update(loss_total.item())
+                logs[f"TotalLoss"] = meters.avg
+        else:  # question: what do we want to do at test time?
```
Andrewwango, line 538:
```suggestion
        else:  # TODO question: what do we want to do at test time?
```

deepinv/training/trainer.py
```diff
@@ -621,7 +629,15 @@ def no_learning_inference(self, y, physics):
...
 
-    def step(self, epoch, progress_bar, train_ite=None, train=True, last_batch=False):
+    def step(
+        self,
+        epoch,
+        progress_bar,
+        train_ite=None,
+        train=True,
+        last_batch=False,
+        global_optimizer_step=True,
```
Andrewwango, line 639: Why add an extra param when we can take `self.global...` anyway?

deepinv/training/trainer.py
```diff
@@ -95,7 +95,8 @@ class Trainer:
         Should be used in conjunction with ``online_measurements=True``, no effect when ``online_measurements=False``. Also see ``loop_random_online_physics``.
     :param bool loop_random_online_physics: if True, resets the physics generator **and** noise model back to its initial state at the beginning of each epoch,
         so that the same measurements are generated each epoch. Requires `shuffle=False` in dataloaders. If False, generates new physics every epoch.
-        Used in conjunction with ``physics_generator`` and ``online_measurements=True``, no effect when ``online_measurements=False``.
+        Used in conjunction with ``physics_generator``.
+    :param bool global_optimizer_step: If ``True``, the optimizer step is performed once on all datasets. If ``False``, the optimizer step is performed on each dataset separately.
```
Andrewwango, line 99: Please can we add a test for this? (well, mostly for when `global_optimizer_step=False` that the original behaviour is recovered, and that global_optimizer_step does nothing when there is only one dataset

deepinv/physics/mri.py
matthieutrs: This would also extend to BlurFFT but I don't know if this is a desired behaviour (I would say no). @tachella what do you think? Should this be the default behaviour?

deepinv/tests/test_physics.py
matthieutrs: actually this was correct since mri is not instantiated at this stage, so `isinstance` does not work (whereas `is` does).

deepinv/physics/mri.py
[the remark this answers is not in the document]
matthieutrs: Thanks for the comment! I've cleaned it. Actually, this only concerns MultiCoilMRI which is not using self.check_mask, I've followed your implementation of the forward.

deepinv/physics/mri.py
tachella: it's true that it won't be backwards compatible, but at the same time, adding noise to the nullspace is not realistic either, so I'm okay to not give the option

deepinv/physics/generator/inpainting.py
```diff
@@ -44,6 +44,9 @@ def __init__(
         tensor_size: Tuple[int],
         split_ratio: float,
         pixelwise: bool = True,
+        random_split_ratio: bool = False,
```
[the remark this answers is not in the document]
matthieutrs: agreed, I've removed!

deepinv/training/trainer.py
matthieutrs: well spotted, that was a mistake, thanks!

deepinv/physics/mri.py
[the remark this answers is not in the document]
Andrewwango: Btw `MultiCoilMRI.A` does use `self.check_mask`, as it calls its `update_parameters`.

deepinv/tests/test_physics.py
Andrewwango: my bad!

deepinv/physics/mri.py
Andrewwango: Why would this affect BlurFFT?

deepinv/physics/blur.py
```diff
@@ -183,6 +160,47 @@ def splits(a, sf):
         else:
             return LinearPhysics.prox_l2(self, z, y, gamma, **kwargs)
 
+    def update_parameters(self, filter=None, factor=None, **kwargs):
+        r"""
+        Updates the current filter.
+
+        :param torch.Tensor filter: New filter to be applied to the input image.
```
Andrewwango, line 167: add factor docstring

deepinv/physics/forward.py
```diff
@@ -425,6 +424,15 @@ def __mul__(self, other):
             tol=self.tol,
         )
 
+    def update_parameters(self, **kwargs):
```
Andrewwango, line 427: 1. Remove update_parameters from decomposable physics as it will inherit this
2. Should we move `update_parameters `to base `Physics`? Since it's very general, and StackedPhysics already implements it. Currently Physics has no `update_parameters` so `StackedPhysics.update_parameters` might raise an error. If no, then I still think we should have a `def update_parameters: pass` in `Physics`

deepinv/physics/forward.py
```diff
@@ -425,6 +424,15 @@ def __mul__(self, other):
...
+        r"""
+        Updates the singular values of the operator.
```
Andrewwango, line 429:
```suggestion
        Updates the parameters of the operator.
```

deepinv/physics/generator/inpainting.py
```diff
@@ -44,6 +44,8 @@ def __init__(
         tensor_size: Tuple[int],
         split_ratio: float,
         pixelwise: bool = True,
+        min_split_ratio: float = 0.0,
```
Andrewwango, line 47: Wait, we should remove min_split_ratio and max_split_ratio too , they are unneeded too

deepinv/physics/generator/noise.py
```diff
@@ -56,3 +56,60 @@ def step(self, batch_size=1, seed: int = None, **kwargs):
...
+    :param float gain_max: maximum noise level
+    :param torch.Generator rng: random number generator
+    :param str device: device where the tensor is stored
+    :param torch.dtype dtype: data type of the generated tensor.
+
+    |sep|
+
+    :Examples:
+
+    >>> from deepinv.physics.generator import SigmaGenerator
```
Andrewwango, line 78:
```suggestion
    >>> from deepinv.physics.generator import GainGenerator
```

deepinv/physics/generator/noise.py
```diff
@@ -56,3 +56,60 @@ def step(self, batch_size=1, seed: int = None, **kwargs):
...
+    >>> generator = GainGenerator()
+    >>> sigma_dict = generator.step(seed=0) # dict_keys(['sigma'])
```
Andrewwango, line 80:
```suggestion
    >>> params = generator.step(seed=0) # dict_keys(['gain'])
```

deepinv/physics/generator/noise.py
```diff
@@ -56,3 +56,60 @@ def step(self, batch_size=1, seed: int = None, **kwargs):
...
+    >>> print(sigma_dict['gain'])
```
Andrewwango, line 81:
```suggestion
    >>> print(params['gain'])
```

Andrewwango (changes requested):
Looks good, only file I didn't review was DiffUNet because I'm not really sure on it

Todo: changelog

I like the new generators!

deepinv/physics/forward.py
matthieutrs: agreed

deepinv/physics/generator/inpainting.py
matthieutrs: Ok, this was inherited from problematic earlier merges. I think this is a nice feature and should be added back. The idea is to sample masks with split ratios that may differ within batch.

deepinv/physics/noise.py
```diff
@@ -117,6 +144,73 @@ def __init__(self, sigma=0.1, rng: torch.Generator = None):
         super().__init__(rng=rng)
         self.update_parameters(sigma=sigma)
 
+    def __add__(self, other):
```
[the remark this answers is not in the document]
matthieutrs: TODO

deepinv/physics/noise.py
```diff
@@ -129,12 +223,11 @@ def forward(self, x, sigma=None, seed=None, **kwargs):
         :returns: noisy measurements
         """
         self.update_parameters(sigma=sigma)
-
-        if isinstance(sigma, torch.Tensor):
-            sigma = sigma[(...,) + (None,) * (x.dim() - 1)]
+        if isinstance(self.sigma, torch.Tensor):
```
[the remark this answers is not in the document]
matthieutrs: sigma could be a float as well, e.g. `self.sigma = 0.1`

deepinv/physics/noise.py
```diff
@@ -356,8 +456,24 @@ def forward(self, x, gain=None, sigma=None, seed: int = None, **kwargs):
         """
         self.update_parameters(gain=gain, sigma=sigma)
         self.rng_manual_seed(seed)
-        y = torch.poisson(x / self.gain, generator=self.rng) * self.gain
-        y += self.randn_like(x) * self.sigma
+
+        if isinstance(self.gain, torch.Tensor):
```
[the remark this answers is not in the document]
matthieutrs: this would not work if `self.gain` is not already on gpu and the `gain` variable in `update_parameters(gain=gain)` is `None`

deepinv/sampling/diffusion.py
```diff
@@ -152,17 +153,17 @@ def forward(self, y, physics: deepinv.physics.DecomposablePhysics, seed=None):
...
+            nsr[case] = sigma_noise / (mask[case] + self.eps)
 
             # iteration 1
             # compute init noise
             mean = torch.zeros_like(y_bar)
             std = torch.ones_like(y_bar) * self.sigmas[0]
             mean[case] = y_bar[case]
             std[case] = (self.sigmas[0] ** 2 - nsr[case].pow(2)).sqrt()
-            x_bar = mean + std * torch.randn_like(y_bar)
+            x_bar = mean + std * torch.randn_like(y_bar) / np.sqrt(2.0)
```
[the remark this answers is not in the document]
matthieutrs: This algo was not working properly so I fixed it there. It was obvious to me back then but I forgot why, I'll double check. Nevertheless I've added a test in the sampling functions!

deepinv/training/trainer.py
```diff
@@ -523,24 +528,27 @@ def compute_loss(self, physics, x, y, train=True, epoch: int = None):
...
 
-            meters = self.logs_total_loss_train if train else self.logs_total_loss_eval
-            meters.update(loss_total.item())
-            logs[f"TotalLoss"] = meters.avg
+                meters = (
+                    self.logs_total_loss_train if train else self.logs_total_loss_eval
+                )
+                meters.update(loss_total.item())
+                logs[f"TotalLoss"] = meters.avg
+        else:  # TODO question: what do we want to do at test time?
```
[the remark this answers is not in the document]
matthieutrs: No, but I think we can leave as is today.

deepinv/physics/noise.py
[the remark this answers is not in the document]
Andrewwango: Do you think `.expand_as` is a clearer implementation than `[(...,) + (None,) * (x.dim() - 1)]`?

Andrewwango (changes requested):
Thanks Matthieu, just some final couple of qs!

deepinv/training/adversarial.py
```diff
@@ -186,12 +190,14 @@ def compute_loss(self, physics, x, y, train=True, epoch: int = None):
...
         :param bool train: If ``True``, the model is trained, otherwise it is evaluated.
         :param int epoch: current epoch.
+        :param bool step: If ``True``, perform an optimization step on all datasets before optimizer step.
         :returns: (tuple) The network reconstruction x_net (for plotting and computing metrics) and
             the logs (for printing the training progress).
         """
         logs = {}
 
-        self.optimizer.G.zero_grad()
+        if train and step:  # remove gradient
```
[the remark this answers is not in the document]
matthieutrs: I agree, I have disabled the multiple dataset option in the adversarial case.

deepinv/physics/noise.py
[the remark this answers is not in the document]
matthieutrs: Done!

deepinv/physics/noise.py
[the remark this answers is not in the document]
matthieutrs: I would keep it as is for now because of other places in the code following this, but it may be good to move to expand_as in the future

PR #386, opened 2025-01-16
Mini fixes + test for trainer when looping physics parameters

Andrewwango:
Now all eventualities for using trainer with various physics options are tested (across `test_loss_train` and `test_trainer`):
- Offline physics
- Offline physics with physics_generator (via `HDFDataset.load_physics_generator_params`)
- Online physics
- Online physics with physics_generator (completely random)
- Online physics with physics_generator and `loop_physics_generator_params=True`

Other changes
- Minor urgent fix to MO-EI loss
- Trainer load model smarter


deepinv/datasets/fastmri.py
```diff
@@ -380,14 +380,12 @@ def save_simple_dataset(
 
         :Example:
 
-        ::
```
tachella, line 383: without this :: it will not display as code

tachella (changes requested):
the PR looks good, there's only a small issue with the docstring of fastMRI (double check that it the code block is displaying correctly)

deepinv/datasets/fastmri.py
Andrewwango: No it is fine as in the following line I have the `::`, and it renders correctly as 
image

tachella (approved):

PR #395, opened 2025-02-10
WIP ENH add PnPFlow model

annegnx:
This is a model from https://arxiv.org/abs/2410.02423.
I have added two components:
- `FlowUNet` in `models/flowunet.py` (unet + checkpoints on CelebA) to model the velocity of the flow matching. 
- `PnPFlow` in `sampling/flow.py`: Image restoration method with flow matching
There is a quick test `debug.py`to test the method whilethis PR is in draft mode. 

@matthieutrs : 
- should we add utils to train the model, or a trained model on a specific dataset (we have checkpoints on celeba and afhq-cat) is enough ?
- currently the checkpoints are on a gdrive, that requires `gdown`to download. do you have a preferred way of storing and downloading checkpoints?
- do we need to do something to highlight the performance of the model on some problem ?


matthieutrs:
Hey thanks for the PR, that's a great start! Few quick comments before I have the time to properly have a look:
- I think you can remove the debug.py file
- It would be super cool to have an example file, it's quite easy to build (see for instance [this one](https://github.com/deepinv/deepinv/blob/main/examples/sampling/demo_diffpir.py))! This will showcase the performance of your algo on some problem

For now let's focus on the test, maybe we can add training later (I think this will require a bit more work because we did not put our training codes in a well organised repo yet but that's definitely something we should consider). Regarding the checkpoints I can upload it on HF with our models.

deepinv/sampling/pnpflow.py
```diff
@@ -0,0 +1,99 @@
...
+        self.verbose = verbose
+        self.device = device
+
+    def denoiser(self, x, t):
+        return x + (1 - t.view(-1, 1, 1, 1)) * self.model(x, t)
+
+    def interpolation_step(self, x, t):
+        return t * x + torch.randn_like(x) * (1 - t)
+
+    def forward(self, y, physics: deepinv.physics.DecomposablePhysics, seed=None):
```
tachella, line 79: why DecomposablePhysics? Do you need an SVD decomposition?

If you don't need an explicit SVD it should be LinearPhysics.

examples/sampling/demo_pnpflow.py
```diff
@@ -0,0 +1,158 @@
...
+    tensor_size=x.shape[1:],
+    noise_model=dinv.physics.GaussianNoise(sigma=sigma_noise),
+    device=device,
+)
+y = physics(2 * x - 1)
+
+# %%
+# Run the PnP method
+# ------------------
+x_hat = pnpflow.forward(y, physics)
```
tachella, line 101:
```suggestion
x_hat = pnpflow(y, physics)
```

deepinv/models/flowunet.py
```diff
@@ -0,0 +1,523 @@
+# Unet copied from "A Variational Perspective on Diffusion-Based Generative Models and Score Matching", NeurIPS 2021, Huang et al
```
mathurinm, line 1: @annegnx @SegoleneMartin THis should rather go in the docstring of the FlowUnet class IMO

deepinv/models/flowunet.py
```diff
@@ -0,0 +1,523 @@
...
+        input_channels,
+        input_height,
+        ch=32,
+        output_channels=None,
+        ch_mult=(1, 2, 4, 8),
+        num_res_blocks=6,
+        attn_resolutions=(16, 8),
+        dropout=0.0,
+        resamp_with_conv=True,
+        act=Swish(),
```
mathurinm, line 50: @matthieutrs how do you feel wrt having mutable parameters as defaults? SHould we do act=None and then `self.act = Swish() if act is None else act`

tachella (changes requested):
- change forward of model to `forward(x, sigma, **kwargs)` to make it work as denoiser by default
- keep another method for the velocity, e.g. `forward_velocity(x, sigma, **kwargs)`
- add tests for the denoiser alone, and the whole method (see where DPIR and denoisers are tested)
- in the demo put math details on the implementation of the PnP flow algo, maybe explaining step by step?
- move flow model to `.optim` instead of `.sampling`
- add denoiser & Flow to User Guide `docs/source/user_guide` (not only API `docs/source/api`)

deepinv/models/flowunet.py
```diff
@@ -0,0 +1,547 @@
...
+                ckpt = torch.load('./model_final_celeba.pt',
+                                  map_location=torch.device(self.device))
+            else:
+                ckpt = torch.load(
+                    "./model_final_celeba.pt", map_location=torch.device(self.device)
+                )
+                # ckpt = torch.load(
+                #     pretrained, map_location=lambda storage, loc: storage)
+
+            self.load_state_dict(ckpt, strict=True)
```
matthieutrs, line 216:
```suggestion
        if pretrained is not None:
            if pretrained == "download":
                if dataset_name == 'cat':
                    name = "ot_fm_cat.pt"
                elif dataset_name == 'celeba':
                    name = "ot_fm_celeba.pt"
                url = get_weights_url(model_name="drunet", file_name=name)
                ckpt_otfm = torch.hub.load_state_dict_from_url(
                    url, map_location=lambda storage, loc: storage, file_name=name
                )
            else:
                ckpt_otfm = torch.load(
                    pretrained, map_location=lambda storage, loc: storage
                )

            self.load_state_dict(ckpt_otfm, strict=True)
```

Something like this should now work (as done in [here](https://github.com/deepinv/deepinv/blob/main/deepinv/models/drunet.py#L135C1-L150C59))

annegnx:
Hey @matthieutrs  I think we have done the requested changes, let us know what you think! Thanks !

PR #410, opened 2025-02-21
VORTEX Loss

Andrewwango:

### DIFF
--- a/deepinv/loss/augmentation.py
+++ b/deepinv/loss/augmentation.py
@@ -1,0 +1,135 @@
+from __future__ import annotations
+
+from typing import Union
+
+import torch
+import torch.nn as nn
+from torch import Tensor
+
+from deepinv.loss.loss import Loss
+from deepinv.loss.metric.metric import Metric
+from deepinv.physics.mri import MRI
+from deepinv.transform.base import Transform
+from deepinv.transform.rotate import Rotate
+from deepinv.transform.shift import Shift
+from deepinv.transform.augmentation import RandomPhaseError, RandomNoise
+
+
+class VORTEXLoss(Loss):
+    r"""Measurement data augmentation loss.
+
+    Performs data augmentation in measurement domain as proposed by
+    `VORTEX: Physics-Driven Data Augmentations Using Consistency Training for Robust Accelerated MRI Reconstruction <https://arxiv.org/abs/2111.02549>`_.
+
+    The loss is defined as follows:
+
+    :math:`\mathcal{L}(T_e\inverse{y,A},\inverse{T_i y,A T_e^{-1}})`
+
+    where :math:`T_1` is a random :class:`deepinv.transform.Transform` defined in k-space and
+    :math:`T_2` is a random :class:`deepinv.transform.Transform` defined in image space.
+    By default, for :math:`T_1` we add random noise :class:`deepinv.transform.RandomNoise` and random phase error :class:`deepinv.transform.RandomPhaseError`.
+    By default, for :math:`T_2` we use random shift :class:`deepinv.transform.Shift` and random rotates :class:`deepinv.transform.Rotate`.
+
+    .. note::
+
+        See :ref:`transform` for a guide on all available transforms, and how to compose them. For example, you can easily
+        compose further transforms such as  ``Rotate(rng=rng, multiples=90) | Scale(factors=[0.75, 1.25], rng=rng) | Reflect(rng=rng)``.
+
+    .. note::
+
+        For now, this loss is only available for MRI problems, but it is easily generalisable to other problems.
+
+    :param deepinv.transform.Transform T_1: k-space transform.
+    :param deepinv.transform.Transform T_2: image transform.
+    :param deepinv.loss.metric.Metric, torch.nn.Module metric: metric for calculating loss.
+    :param bool no_grad: if ``True``, only propagate gradients through augmented branch as per original paper,
+        if ``False``, propagate through both branches.
+    :param torch.Generator rng: torch random number generator to pass to transforms.
+    """
+
+    def __init__(
+        self,
+        T_1: Transform = None,
+        T_2: Transform = None,
+        metric: Union[Metric, nn.Module] = torch.nn.MSELoss(),
+        no_grad: bool = True,
+        rng: torch.Generator = None,
+        *args,
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+        self.metric = metric
+        self.T_1 = (
+            T_1
+            if T_1 is not None
+            else RandomPhaseError(scale=0.1, rng=rng) * RandomNoise(rng=rng)
+        )
+        self.T_2 = (
+            T_2
+            if T_2 is not None
+            else Shift(shift_max=0.1, rng=rng) | Rotate(rng=rng, limits=15)
+        )
+        self.no_grad = no_grad
+
+    class TransformedMRI(MRI):
+        """Pre-multiply physics with transform.
+
+        :param deepinv.physics.MRI: original MRI physics
+        :param deepinv.transform.Transform: transform object
+        :param dict transform_params: fixed parameters for deterministic transform.
+        """
+
+        def __init__(
+            self,
+            physics: MRI,
+            transform: Transform,
+            transform_params: dict,
+            *args,
+            **kwargs,
+        ):
+            super().__init__(
+                *args,
+                img_size=physics.img_size,
+                mask=physics.mask,
+                device=physics.device,
+                three_d=physics.three_d,
+                **kwargs,
+            )
+            self.transform = transform
+            self.transform_params = transform_params
+
+        def A(self, x, *args, **kwargs):
+            return super().A(
+                self.transform.inverse(x, **self.transform_params), *args, **kwargs
+            )
+
+        def A_adjoint(self, y, *args, **kwargs):
+            return self.transform(
+                super().A_adjoint(y, *args, **kwargs), **self.transform_params
+            )
+
+    def forward(self, x_net: Tensor, y: Tensor, physics: MRI, model, **kwargs):
+        r"""
+        VORTEX loss forward pass.
+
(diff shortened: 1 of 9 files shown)


deepinv/loss/augmentation.py
```diff
@@ -0,0 +1,138 @@
...
+from deepinv.physics.mri import MRI
+from deepinv.physics.forward import LinearPhysics
+from deepinv.physics.inpainting import Inpainting
+from deepinv.transform.base import Transform
+from deepinv.transform.rotate import Rotate
+from deepinv.transform.shift import Shift
+from deepinv.transform.augmentation import RandomPhaseError, RandomNoise
+
+
+class VORTEXLoss(Loss):
```
tachella, line 20: I would call this DACLoss or DataAugmentationConsistencyLoss, since the idea is much more general than the VORTEX paper. We could anyway mention VORTEX in the docstring.

deepinv/loss/augmentation.py
```diff
@@ -0,0 +1,138 @@
...
+    r"""Measurement data augmentation loss.
+
+    Performs data augmentation in measurement domain as proposed by
+    `VORTEX: Physics-Driven Data Augmentations Using Consistency Training for Robust Accelerated MRI Reconstruction <https://arxiv.org/abs/2111.02549>`_.
+
+    The loss is defined as follows:
+
+    :math:`\mathcal{L}(T_e\inverse{y,A},\inverse{T_i y,A T_e^{-1}})`
+
+    where :math:`T_1` is a random :class:`deepinv.transform.Transform` defined in k-space and
```
tachella, line 30:
```suggestion
    where :math:`T_e` is a random :class:`deepinv.transform.Transform` defined in image space (that the model is supposed to be equivariant to) and
```

deepinv/loss/augmentation.py
```diff
@@ -0,0 +1,138 @@
...
+    :math:`T_2` is a random :class:`deepinv.transform.Transform` defined in image space.
```
tachella, line 31:
```suggestion
    :math:`T_i` is a random :class:`deepinv.transform.Transform` defined in measurement space (that the model is supposed to be invariant to).
```

deepinv/loss/augmentation.py
```diff
@@ -0,0 +1,138 @@
...
+    By default, for :math:`T_1` we add random noise :class:`deepinv.transform.RandomNoise` and random phase error :class:`deepinv.transform.RandomPhaseError`.
```
tachella, line 32:
```suggestion
    By default, for :math:`T_i` we add random noise :class:`deepinv.transform.RandomNoise`.
```

I would remove the default random phase error default since it only holds for MRI

deepinv/loss/augmentation.py
```diff
@@ -0,0 +1,138 @@
...
+    By default, for :math:`T_2` we use random shift :class:`deepinv.transform.Shift` and random rotates :class:`deepinv.transform.Rotate`.
```
tachella, line 33:
```suggestion
    By default, for :math:`T_e` we use random shift :class:`deepinv.transform.Shift` and random rotates :class:`deepinv.transform.Rotate`.
```

deepinv/loss/augmentation.py
```diff
@@ -0,0 +1,138 @@
...
+    .. note::
+
+        See :ref:`transform` for a guide on all available transforms, and how to compose them. For example, you can easily
+        compose further transforms such as  ``Rotate(rng=rng, multiples=90) | Scale(factors=[0.75, 1.25], rng=rng) | Reflect(rng=rng)``.
+
+    .. note::
+
+        For now, this loss is only available for MRI and Inpainting problems, but it is easily generalisable to other problems.
+
+    :param deepinv.transform.Transform T_1: k-space transform.
```
tachella, line 44:
```suggestion
    :param deepinv.transform.Transform T_i: measurement domain transform for which the model should be invariant.
```

deepinv/loss/augmentation.py
```diff
@@ -0,0 +1,138 @@
...
+    :param deepinv.transform.Transform T_2: image transform.
```
tachella, line 45:
```suggestion
    :param deepinv.transform.Transform T_e: image domain transform for which the model should be equivariant.
```

deepinv/loss/augmentation.py
```diff
@@ -0,0 +1,138 @@
...
+            )
+
+        def A_adjoint(self, y, *args, **kwargs):
+            return self.transform(
+                super().A_adjoint(y, *args, **kwargs), **self.transform_params
+            )
+
+    def forward(self, x_net: Tensor, y: Tensor, physics: MRI, model, **kwargs):
+        r"""
+        VORTEX loss forward pass.
```
tachella, line 116: change name

deepinv/loss/augmentation.py
```diff
@@ -0,0 +1,138 @@
...
+        :return: (:class:`torch.Tensor`) loss, the tensor size might be (1,) or (batch size,).
+        """
+        if self.no_grad:
+            x_net = x_net.detach()
+
+        # Sample image transform
+        e_params = self.T_2.get_params(x_net)
+
+        # Augment input
+        x_aug = self.T_2(physics.A_adjoint(self.T_1(y)), **e_params)
```
tachella, line 132: I'm not sure why you are adding T_2 here

tachella (changes requested):
thanks for the PR

- I think we should fix the naming of the transforms (the docs mix T_1, T_2 with T_i T_e), keeping only one option
- I would change the name of the loss since this technique is common beyond vortex (it is called data augmentation consistency)
- I would remove the MRI-specific default transforms to make it more generally applicable
- I have a doubt regading the implementation

deepinv/loss/augmentation.py
Andrewwango: This is from the [paper](https://arxiv.org/pdf/2111.02549) (Fig. 1)?

deepinv/loss/augmentation.py
Andrewwango: done

deepinv/loss/augmentation.py
Andrewwango: done

deepinv/loss/augmentation.py
```diff
@@ -0,0 +1,110 @@
...
+        # Transform physics
+        physics2 = deepcopy(physics)
+        A, A_adjoint, A_dagger = physics2.A, physics2.A_adjoint, physics2.A_dagger
+        physics2.A = lambda x, *args, **kwargs: A(
+            self.T_e.inverse(x, **e_params), *args, **kwargs
+        )
+        physics2.A_adjoint = lambda y, *args, **kwargs: self.T_e(
+            A_adjoint(y, *args, **kwargs), **e_params
+        )
+        physics2.A_dagger = lambda y, *args, **kwargs: self.T_e(
```
tachella, line 103: this only works for linear physics

I would prefer to compose physics as is done [here](https://github.com/deepinv/deepinv/blob/8189cbc8a75888fce1759dbe2564ecb2601f6c60/deepinv/transform/base.py#L297), where we first define a linear physics associated with the transform and then we just concatenate with the base physics.

deepinv/transform/augmentation.py
```diff
@@ -0,0 +1,114 @@
...
+        return torch.cat([n(y) * mask for n in noise_model])
+
+    def inverse(self, *args, **kwargs):
+        raise ValueError("Noise transform is not invertible.")
+
+
+class RandomPhaseError(Transform):
+    """Random phase error transform.
+
+    This transform adds a phase error to k-space using:
```
tachella, line 68:
```suggestion
    This transform is specific to MRI problems, and adds a phase error to k-space using:
```

deepinv/transform/augmentation.py
```diff
@@ -0,0 +1,114 @@
...
+
+    :math:`Ty=\exp(-i\phi_k)y` where :math:`\phi_k=\pi\alpha s_e` if :math:`k` is an even index,
+    or :math:`\phi_k=\pi\alpha s_o` if odd, and where :math:`\alpha` is a scale parameter,
+    and :math:`s_o,s_e\sim U(-1,1)`.
+
+    This transform is reproducible: for given param dict `se, so`, the transform is deterministic.
+
+    :param int, tuple[int, int] scale: scale parameter or range to pick randomly.
```
tachella, line 76:
```suggestion
    :param int, tuple[int, int] scale: scale parameters :math:`s_e` and :math:`s_o` or range to pick randomly.
```

tachella (changes requested):

deepinv/loss/augmentation.py
```diff
@@ -0,0 +1,110 @@
...
+        T_e: Transform = None,
+        metric: Union[Metric, nn.Module] = torch.nn.MSELoss(),
+        no_grad: bool = True,
+        rng: torch.Generator = None,
+        *args,
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+        self.metric = metric
+        self.T_i = T_i if T_i is not None else RandomNoise(rng=rng)
```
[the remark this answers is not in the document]
Andrewwango: done

deepinv/transform/augmentation.py
Andrewwango: done

deepinv/loss/augmentation.py
Andrewwango: I agree, that it would be nice to do:
```python
        # Transform physics
        physics2 = physics * LinearPhysics(
            A=lambda x: self.T_e.inverse(x, **e_params),
            A_adjoint=lambda x: self.T_e(x, **e_params),
        )
```
But I just tried this, and the results are very different. Something strange is going on.

I would propose leaving it as is for now, and when your equivaraint PR is ready for review, I can try changing this again.

Andrewwango: Added issue here https://github.com/deepinv/deepinv/issues/489

tachella (approved):

PR #414, opened 2025-02-25
Offline physics generator params in trainer

Andrewwango:
When training using physics generator and offline measurements, the physics params are loaded in for training (as expected) but the noise parameters (i.e. for random sigma noise) aren't in general!

This was because trainer was calling `physics.update_parameters` instead of `physics.update` so noise params (e.g. sigma) wasn't getting loaded in. Incidentally for some physics e.g. `Blur.update_parameters` also updates noise params but this isn't the case for Inpainting.

In this PR I add a test for this and modify trainer to call `physics.update` instead.

TODO
- [x] test noise params don’t update in A/Adjoint but params do
- [x] remove all user refs to update_parameters
- [x] change docs/examples to not reference update_parameters for users
- [x] make clear how to do unsup learning with offline/online measurements
- [x] discourage unsup learning with noisy online measurements
- [x] reproducible noise physics with online measurements in self-sup context?


tachella (approved):
many thanks for the bug fixes

PR #419, opened 2025-03-12
Simplified Jacobian free backprop implementation in DEQ

DavyL:
Hi,

I simplified the method for Jacobian free backprop by avoiding the backward_hook by returning the variable (with attached gradient of a single pass on the fixed_point_iterator) earlier.
Let me know what you think of it,
(I also added a test for Jacobian Free Backpropagation in test_unfolded)
Thanks,
Leo


deepinv/unfolded/deep_equilibrium.py
```diff
@@ -88,8 +87,13 @@ def forward(self, y, physics, x_gt=None, compute_metrics=False, **kwargs):
             X, cur_data_fidelity, cur_prior, cur_params, y, physics, **kwargs
         )["est"][0]
 
-        # Then we perform automatic differentiation.
-        # First another iteration for jacobian computation via automatic differentiation.
+        # If performing Jacobian Free Backpropagation, return x only with gradients from last pass
+        if self.jacobian_free is True:
```
samuro95, line 91: instead of moving the return there, what to you think about adding a if statement to the additional forward pass to avoid it if we set Jacobian_free to True ?

samuro95 (commented):
Thanks for the nice update !

PR #422, opened 2025-03-13
[Suggestion] Separable prior

DavyL:
Hi,  

Here is some code to extend already defined priors into **separable priors**, which can be written as:  

f(x) = \sum_i g(x_i).

The idea is that if a prior \( g \) is already defined, then calling `SeparablePrior(somePrior, ...)` creates a new prior that is separable along a specific axis.  
For example, if a prior is designed for grayscale images but we want to apply it independently to each channel of an image, we can define the separable axis as `"1"`.  

One key advantage is that the **proximal operator** of a separable function is simply the concatenation of the prox operators applied along the separable dimension.  

---

### 🔹 Remarks:  
- It might be useful to implement a `SeparablePotential` (or `StackedPotential`?) extending `Potential` instead of `Prior`, particularly since the gradient of a separable function is the sum of all gradients.  
- I have included an implementation for `dinv.utils.TensorList`, where the separable axis is defined over elements of the list.  
  - This is useful, for instance, when working with wavelet coefficients structured as:   [c_1, ..., c_J , d_J]
      
    where elements are tensors of different shapes but may share a similar regularization term \( ||\cdot||_1 \).  
- To allow for different regularizations on each element in the sum:  
  
  f(x) = \sum_i \lambda_i g(x_i),
    
  I introduced `separable_weights`, which must be passed as arguments.  
- Since weights must be non-negative, I parameterized them using an exponential function:  
  
  \lambda_i = \exp(w_i).
  
  This log-parametrization is not standard so give me your opinion on whether this approach makes sense.  
- A challenge when introducing weights is that the size of the `separable_axis` must be known in advance.  
  - A more flexible approach could be defining a `SeparablePrior` that takes a **list of Priors** as input:  
    ```python
    sepPrior = SeparablePrior([somePrior_1, ..., somePrior_J])
    ```
    This would allow defining functions of the form:  f(x) = \sum_j g(x_j),
    
    and weights could be directly embedded in each individual `somePrior_j`.  
- **Currently, there is no test for `ListSeparablePrior`**, as the existing implementation of `BaseOptim` algorithms does not seem particularly `TensorList`-friendly.  

---

### 🔹 Next Steps  
Let me know if you think it would be useful to add **separable functions** to the library. If so, I can write an example file.  

Before completing the PR, I believe the following questions should be addressed:  

1. **Should we extend `Prior` or `Potential`?**  
2. **Should weights be included in the definition, or should we pass a list of Priors in the `__init__` method?**  
3. **Should the implementation support both `torch.Tensor` and `dinv.utils.TensorList`?**  

Thanks, everyone, for this great library!  

**Leo**  


deepinv/optim/prior.py
```diff
@@ -665,3 +666,159 @@ def prox(self, x, *args, gamma=1.0, **kwargs):
...
+        is computed, and then the resulting slices are concatenated back along the separable_axis.
+
+        :param x: Input tensor.
+        :param gamma: A step-size parameter (on :math:`\tau f`).
+        :return torch.Tensor: :math:`\operatorname{prox}_{\tau f}(x)` of the same shape as :math:`x` after applying the proximity operator.
+        """
+        eseparable_weights = torch.exp(self.separable_weights)
+        prox_slices = []
+        # torch.unbind splits the tensor along the separable_axis without squeezing that axis.
+        for coord, sliced_x in enumerate(torch.unbind(x, dim=self.separable_axis)):
```
gRox167, line 748: `torch.split` splits the tensor without squeezing axis. 
`torch.split(x, 1, dim=self.separable_axis)` will work here.

deepinv/optim/prior.py
```diff
@@ -665,3 +666,159 @@ def prox(self, x, *args, gamma=1.0, **kwargs):
...
+            # Compute the prox for the current slice.
+            prox_slice = self.prior.prox(
+                sliced_x, *args, gamma=gamma * eseparable_weights[coord],  **kwargs
+            )
+            # unsqueeze to restore the separable axis for later concatenation.
+            prox_slices.append(prox_slice.unsqueeze(self.separable_axis))
```
gRox167, line 754: Also if we split the tensor without squeezing dim, we don't need to unsqueeze here.

deepinv/optim/prior.py
```diff
@@ -665,3 +666,159 @@ def prox(self, x, *args, gamma=1.0, **kwargs):
...
+    def __init__(self, prior, separable_axis, separable_weights, *args, **kwargs):
+        """
+        :param dinv.optim.Prior prior: a Prior defining the function :math:`g`
+        :param int separable_axis: index of the axis over which the prior is separable.
+        :param torch.Tensor separable_weights: a tensor of weights (in log-domain) for each slice along the separable axis.
+        """
+        super().__init__(*args, **kwargs)
+        self.prior = prior
+        self.separable_axis = separable_axis
+        self.separable_weights = separable_weights
```
gRox167, line 701: maybe also add a attribute for explicit_prior 
```python
        self.explicit_prior = (
            prior[0].explicit_prior if isinstance(prior, list) else prior.explicit_prior
        )
```

deepinv/optim/prior.py
```diff
@@ -665,3 +666,159 @@ def prox(self, x, *args, gamma=1.0, **kwargs):
...
+
+        :param torch.Tensor x: Input tensor.
+        :return torch.Tensor : value of :math:`f(x)` for each batch
+        """
+        # Exponentiate the (log-)weights.
+        eseparable_weights = torch.exp(self.separable_weights)
+        # Initialize f_total with zeros. Assumes the first dimension is the batch dimension.
+        f_total = torch.zeros(x.shape[0], device=x.device, dtype=x.dtype)
+        # Loop over the separable axis
+        for coord in range(x.shape[self.separable_axis]):
```
gRox167, line 719: same here, use torch.split

gRox167:
Just find out if we apply a single prior multiple times, some prior like TV, it's denoiser have `restart` attribute, which will save x2 and u2 from previous calculation. This will lead to `.prox()`'s output be identical for all separated slices.

jscanvic:
Thanks Leo for the suggestion and for implementing it!

I'm taking over this PR for the last few things that remain to be done given Leo won't have the time to do it in the coming weeks and it would be a shame to waste the work he's already done and miss on this new feature!

Also, thanks @gRox167 for reviewing the PR and for your insightful suggestions - I made the changes you suggested.

**Weights reparametrization**

In the original implementation, the weights were reparametrized using an exponential parametrization. The goal was to make it impossible to have non-positive weights but the choice of parametrization is arbitrary and it is overall too confusing for the benefits it brings in my opinion. I'd be happy to consider other solutions but for now I chose to get rid of it altogether and to simply rely on runtime failures if negative weights happen to pose a problem.

**ListSeparablePrior**

The optimization module of the library doesn't currently support TensorList even though as Leo suggests it can be a good way to represent wavelet representations and pyramid image representations. I expect it to be a lot of work to add support for it so I chose to leave out ListSeparablePrior for now. I'd be happy to have it added back in the future - of course! I also opened an issue for this: #638

**Changes**

* Fix squeezing-related bugs and switch to the more natural `torch.split` instead of `torch.unbind` as suggested by @gRox167
* Specify `Prior.explicit_prior` as suggested by @gRox167 
* Tidy the code (use shorter, more standard variable names, add intermediate variables)
* Remove the exponential reparametrization of the weights
* Get rid of ListSeparablePrior

jscanvic:
PS @gRox167 could you elaborate on your last comment? (the one about `restart`) I understand this is a bug with certain priors but it's not specific to this specific PR right? It'd be great if you could open a separate issue about it!

PR #423, opened 2025-03-14
CLN warn user about unused parameters in Physics

tomMoral:
Using the lib, I tried to use `Denoising(sigma=0.01)` which gave me unexpected results.
Upon investigating, I was surprised to see that `sigma` is actually ignored here (which makes sense).

This PR adds a warning to inform the user that the parameter is ignored and move up the `kwargs` to the base `Physics` class (as discussed IRL with @matthieutrs).
Another solution would be to remove the `kwargs` to raise an error in case a parameter is passed but irrelevant but that might break some code.

### DIFF
--- a/deepinv/physics/forward.py
+++ b/deepinv/physics/forward.py
@@ -1,4 +1,5 @@
 from typing import Union
+import warnings
 
 import torch
 from torch import Tensor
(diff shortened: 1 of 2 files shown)


tachella:
thanks for raising this!

I see the danger of `kwargs`, and your example is very relevant, but I'm afraid that we'll end up cluttering the output with warnings even in cases where the usage is correct.

tomMoral:
Which correct usage would you expect?

A possibility would also be to have a mechanism to ignore this warning explicitly.

Andrewwango (approved):
Looks good, thanks @tomMoral - before merging please can you check any warnings that may have arisen in the workflows since your change

PR #425, opened 2025-03-17
Add MINRES for solving linear systems

johertrich:
We add the minimal residual method (https://en.wikipedia.org/wiki/Minimal_residual_method) for solving linear systems (having mainly hypergradient computations for bilevel methods in mind, but it can also be used for solving least-squares problems).

The implementation is based on https://github.com/cornellius-gp/linear_operator. However, we simplified and modified this implementation a bit such that it can handle tensors with general shapes (in particular the image format B x C x H x W), complex linear systems and allows backprop (even though it is probably not a good idea to backprop through an iterative solver of a linear system).

Additionally, we added a unit-test for testing the linear system solvers for random matrices. As a side note: the bicgstab implementation doesn't work for complex or non-hermite matrices.

@Zeppo1994

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,7 @@ Current
 
 New Features
 ^^^^^^^^^^^^
+- Add minres least squared solver (:gh:`` by `Sebastian Neumayer`_ and `Johannes Hertrich`_)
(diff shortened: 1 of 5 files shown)


tachella (approved):
thanks for the great PR, it is good to go in my opinion

PR #428, opened 2025-03-19
Debugged MINRES convergence flag

MohammadSadeghSalehi:
Fix: Define flag in minres function

The flag variable in the minres function was not defined, which caused a local variable accessed before definition error when the function reached the maximum number of iterations without convergence.

Changes Made:
•	Added flag = True at the beginning of the function definition to ensure flag is always defined.

Impact:
•	Prevents runtime errors when the solver does not converge within the allowed iterations.


johertrich:
Hi Sadegh,

I just came across this pull request by random. I agree that there is a bug in the minres implementation since I forgot to initialize the convergence flag. However, I think the flag should not be a (keyword) argument of the function. The verbose argument already controls whether the convergence status will be printed or not (and I think this is the standard naming).

So short: I would suggest to initialize flag=True outside the function header but before the loop.

tachella:
Thanks for the fix!
I will close the PR, since the issue was fixed in #458, defining the flag inside the function instead of using a parameter

Issue #436, opened 2025-04-03
No early stopping in the deepinv.Trainer


Melvin-klein:
It would be great to support early stopping in the `deepinv.Trainer`.

tachella:
hey Melvin, many thanks for this suggestion!

I'm not sure I fully understand your point, how would the trainer be early-stopped?

tachella:
what do you mean? we already have a max number of training steps

Andrewwango:
Maybe do you mean to stop training when (`train` or `test`) `loss` or `metric` reaches `tol` instead of when `epoch` reaches `epochs`?

Melvin-klein:
Thanks for your answers ! I'm new in the ML field so I hope what I'm saying makes sense : I was thinking about stopping the training based on performance on a validation set. For example, if the performance measured on the validation set hasn't decreased for `n` epochs, we stop the training and keep the model learned `n` epochs before stopping.

tachella:
Thanks for the clarification, we are currently working on this!

Andrewwango:
Closed by #437

Issue #398, opened 2025-02-12
Trainer save best model


Andrewwango:
Currently, `deepinv.Trainer` only has the option to save the model every n epochs. It would be good to have functionality, e.g. `ckp_interval='best'`, where the model gets saved every time if the eval metric is greater than the current best.

alexmehta:
Would a good solution be to implement ability to add hooks/callbacks when training like keras lightning and other training libraries do? Would allow for saving best model but also might make other features such as early stopping, wandb/tensorboard integration more clean/extensible.

Andrewwango:
Thanks for the ideas @alexmehta !

I think an eventual solution that might be cleaner and more scalable than what we have now could be achieved with hooks/callbacks. Currently our framework is to provide modular methods with one unified set of class params that are passed around.

I am still unsure whether callbacks are appropriate right now, so for now we have added this feature in #437 in our existing framework, but I will add your comment to discussions as I think it will be likely that we will need your advice soon again!

Issue #390, opened 2025-01-24
Missing scheduler.state_dict() when `Trainer.save_model`


MaximeSongIdris:
When saving training to resume a training later on, if we are using a scheduler, we need to save its state_dict :

```
        torch.save({
            'epoch': epoch,
            'model_state_dict': model.state_dict(),
            'optimizer_state_dict': optimizer.state_dict(),
            'scheduler_state_dict': scheduler.state_dict()
        }, 'checkpoint.pth')
```

MaximeSongIdris:
When loading from checkpoint:
https://github.com/deepinv/deepinv/blob/main/deepinv/training/trainer.py#L270

we need to pay attention to torch.load(), because some training may happen on a device "cuda:1" and not available when resuming training -> to make things proper, we can load everything on cpu first (instanciating model on cpu, loading checkpoint on cpu, loading weight on cpu, then moving model to gpu

MaximeSongIdris:
https://deepinv.github.io/deepinv/_modules/deepinv/training/trainer.html#Trainer
When loading from ckpt_pretrained, `epoch_start = checkpoint["epoch"]` in `load_model` is wrong, as the model is already trained at this epoch.
It should be: `epoch_start = checkpoint["epoch"] + 1`

Andrewwango:
For anyone working on this issue, please note the latest modifications in `load_model` in #386 #394

Andrewwango:
Closed by #437

PR #437, opened 2025-04-09
Trainer updates

tachella:
Trainer fixes

- adds early stopping #436 
- adds optional max batch iterations
- clarifies data formats for dataloaders + allow nan, y
- metric history fixes
- fixed tqdm eval (and made it green :P)
- saves best model #398 
- save and load scheduler #390 

### DIFF
--- a/deepinv/utils/plotting.py
+++ b/deepinv/utils/plotting.py
@@ -110,7 +110,7 @@ def prepare_images(x=None, y=None, x_net=None, x_nl=None, rescale_mode="min_max"
             out = preprocess_img(img, rescale_mode=rescale_mode)
             vis_array.append(out)
         vis_array = torch.cat(vis_array)
-        grid_image = make_grid(vis_array, nrow=x.shape[0])
+        grid_image = make_grid(vis_array, nrow=x_net.shape[0])
 
     for k in range(len(imgs)):
         imgs[k] = preprocess_img(imgs[k], rescale_mode=rescale_mode)


deepinv/training/trainer.py
```diff
@@ -786,11 +815,11 @@ def save_model(self, epoch, eval_metrics=None, state={}):
                 state["eval_metrics"] = eval_metrics
             if self.wandb_vis:
                 state["wandb_id"] = wandb.run.id
+
+            name = "best_model.pth.tar" if best_model else f"ckp_{epoch}.pth.tar"
```
Andrewwango, line 819:
```suggestion
            name = "ckp_best.pth.tar" if best_model else f"ckp_{epoch}.pth.tar"
```

deepinv/training/trainer.py
```diff
@@ -39,9 +39,17 @@ class Trainer:
...
+        If a validation dataloader `eval_dataloader` is provided, the trainer will also **save the best model** according to the
+        first metric in the list, using the following format:
+        ``save_path/yyyy-mm-dd_hh-mm-ss/best_model.pth.tar``.
+
+    Assuming that `x` is the ground-truth reference and `y` is the measurement and `params` is a dict of :ref:`physics parameters <physics_generators>`,
+    the **dataloaders** should return one of the following options:
+
+    1. `(x, y)` or `(x, y, params)`, which requires `online_measurements=False` (default) otherwise `y` will be ignored and new measurements will be generated online.
+    2. `(x)` or `(x, params)`, which requires `online_measurements=True` for generating measurements in an online manner as `y=physics(x)`. Otherwise, first generate a dataset of `(x,y)` with :class:`deepinv.datasets.generate_dataset` and then use option 1 above.
+    3. If you have a dataset of measurements only `(y)` or `(y, params)` you should modify it such that it returns `(torch.nan, y)` or `(torch.nan, y, params)`  and use options 1 or 2 above with `online_measurements=False`.
```
Andrewwango, line 52: Can we add a safety net where losses are called `l(x_net, x, y, physics, model, ...)` or metrics are called `m(x_net, x)` and `x` is `nan` then `None` is passed instead, so that if the user is using a full-reference metric, the metric throws a runtime error

deepinv/utils/plotting.py
```diff
@@ -82,9 +82,13 @@ def prepare_images(x, y, x_net, x_nl=None, rescale_mode="min_max"):
     :returns: The images, the titles, the grid image, and the caption.
     """
     with torch.no_grad():
-        imgs = [x]
-        titles = ["Ground truth"]
-        caption = "From left to right: Ground truth, "
+        imgs = []
+        titles = []
+        caption = "From left to right: "
+        if not torch.any(torch.isnan(x)):
```
Andrewwango, line 88: I think in general we should still allow nans in data when plotting. If we add logic in trainer to convert nans to Nones, then we can just detect Nones here?

deepinv/training/trainer.py
Andrewwango, line 52: How can option 3 use option 2? Maybe leave out the last bit of this sentence and just say 
```suggestion
    3. If you have a dataset of measurements only `(y)` or `(y, params)` you should modify it such that it returns `(torch.nan, y)` or `(torch.nan, y, params)`. Set `online_measurements=False`.
```

Can you also add a warning that if x is ALL nan, i.e. we have option 3, then online_measurements must be false?

deepinv/tests/test_trainer.py
```diff
@@ -317,3 +317,78 @@ def test_trainer_test_metrics(device, rng):
...
+        epochs=1,
+        physics=physics,
+        train_dataloader=dataloader,
+        optimizer=optimizer,
+    )
+    # Check that the model is trained without errors
+    trainer.train()
+
+
+def test_early_stop(dummy_dataset, imsize, device, dummy_model):
```
Andrewwango, line 365: Can we also add a test for stopping when max batch iters is reached (and also, whilst we're at it, when epochs is reached)

deepinv/training/trainer.py
tachella: done

deepinv/training/trainer.py
tachella: agreed, I now change to None in get_samples

deepinv/utils/plotting.py
tachella: same as above

deepinv/training/trainer.py
tachella: done

deepinv/tests/test_trainer.py
tachella: done

deepinv/training/trainer.py
```diff
@@ -59,85 +56,150 @@ class Trainer:
...
+
     :param None, torch.utils.data.DataLoader, list[torch.utils.data.DataLoader] eval_dataloader: Evaluation data loader(s)
-        should provide a signal x or a tuple of (x, y) signal/measurement pairs or a tuple `(x, y, params)`
-        where `params` is a dict of physics generator parameters to be loaded into the physics each iteration.
-    :param None, torch.optim.lr_scheduler.LRScheduler scheduler: Torch scheduler for changing the learning rate across iterations.
-    :param bool online_measurements: Generate the measurements in an online manner at each iteration by calling
-        ``physics(x)``. This results in a wider range of measurements if the physics' parameters, such as
-        parameters of the forward operator or noise realizations, can change between each sample;
-        the measurements are loaded from the training dataset.
+        should provide a signal `x` or a tuple of `(x, y)` signal/measurement pairs or a tuple `(x, y, params)`
```
Andrewwango, line 103: Similarly here remove description and refer to list of options above on how to pass dataloaders

deepinv/training/trainer.py
```diff
@@ -566,6 +644,9 @@ def compute_metrics(
                     x_net=x_net,
                     x=x,
                     epoch=epoch,
+                    model=self.model,
```
Andrewwango, line 647: Why have you added these three lines? I remember us discussing this when I created the `Metric` interface, and you said that we shouldn't allow metrics to take `y`, `model` or `physics`

In fact, this is why I had to caveat QNR, if you wanted to use QNR in trainer you would've had to override `compute_metrics`
https://deepinv.github.io/deepinv/api/stubs/deepinv.loss.metric.QNR.html#deepinv.loss.metric.QNR.metric

deepinv/training/trainer.py
```diff
@@ -814,6 +895,61 @@ def reset_metrics(self):
         for l in self.logs_metrics_eval:
             l.reset()
 
+    def save_best_model(self, epoch, train_ite, metric_history, metrics, **kwargs):
```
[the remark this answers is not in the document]
Andrewwango: Same with `metrics`

deepinv/training/trainer.py
```diff
@@ -814,6 +895,61 @@ def reset_metrics(self):
...
+            curr_metric = history[-1]
+            if (lower_better and curr_metric <= best_metric) or (
+                not lower_better and curr_metric >= best_metric
+            ):
+                # Saving the model
+                self.save_model("ckp_best.pth.tar", epoch)
+                if self.verbose:
+                    print(f"Best model saved at epoch {epoch + 1}")
+
+    def stop_criterium(self, epoch, train_ite, metric_history, metrics, **kwargs):
```
Andrewwango, line 925:
```suggestion
    def stop_criterion(self, epoch, train_ite, metric_history, metrics, **kwargs):
```

deepinv/training/trainer.py
```diff
@@ -814,6 +895,61 @@ def reset_metrics(self):
...
+        r"""
+        Stop criterium for early stopping.
```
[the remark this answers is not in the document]
Andrewwango: By default, stop based when first eval metric reaches xxx. Override this method to early stop on a custom condition.

deepinv/training/trainer.py
```diff
@@ -814,6 +895,61 @@ def reset_metrics(self):
...
+        history = metric_history[metrics[k].__class__.__name__]
+        if hasattr(metrics[k], "lower_better"):
+            lower_better = metrics[k].lower_better
+        else:  # assume is a loss which is always lower is better
+            lower_better = True
+
+        best_metric = min(history) if lower_better else max(history)
+        best_epoch = history.index(best_metric)
+
+        if self.verbose:
```
Andrewwango, line 945: Shouldn't this only be printed if the return is True?

deepinv/training/trainer.py
```diff
@@ -814,6 +895,61 @@ def reset_metrics(self):
...
+            print(
+                "Early stopping triggered as validation metrics have not improved in "
+                "the last 5 epochs, disable it with early_stop=False"
+            )
+
+        return epoch > 5 + best_epoch
```
Andrewwango, line 951: Doesn't this condition assume that the the stable metric is the best metric? This is not necessarily appropriate for all experiments. Maybe we should just do when the latest metric has been stable (ignoring what is the "best" metric)?

deepinv/training/trainer.py
```diff
@@ -909,21 +1048,58 @@ def train(
...
+            if stop_flag:
+                break
 
         if self.wandb_vis:
             wandb.save("model.h5")
             wandb.finish()
 
+        # load best model
+        if self.save_path is not None:
+            path = os.path.join(Path(self.save_path), Path("ckp_best.pth.tar"))
```
Andrewwango, line 1094: Path / Path instead of os

deepinv/training/trainer.py
```diff
@@ -909,21 +1048,58 @@ def train(
...
+
+            if stop_flag:
+                break
 
         if self.wandb_vis:
             wandb.save("model.h5")
             wandb.finish()
 
+        # load best model
+        if self.save_path is not None:
```
Andrewwango, line 1093: This feels quite dangerous - if you have a bad best criterion, then this could potentially discard lots of precious training?

I would rather remove this logic. We could add a parameter to Trainer.test e.g. `ckpt` that modifies `self.ckpt_pretrained` which is used in `load_model` or something.

(note: we have `load_model` which handles model loading logic)

Andrewwango (changes requested):
Looks good, thanks!

Some other todos

- [ ] Trainer.test docstring dataloader should refer to the options on how we pass data (options 1,2,3) in the main docstring
- [ ] Trainer.test docstring what happens when save_path is none - what is the default (does it depend on what `save_path` or `plot_images` was in Trainer init)?

deepinv/utils/plotting.py
```diff
@@ -75,20 +75,25 @@ def prepare_images(x, y, x_net, x_nl=None, rescale_mode="min_max"):
...
     with torch.no_grad():
-        imgs = [x]
-        titles = ["Ground truth"]
-        caption = "From left to right: Ground truth, "
-        if y.shape == x.shape:
+        imgs = []
+        titles = []
+        caption = "From left to right: "
+        if len(y.shape) == 4:
             imgs.append(y)
```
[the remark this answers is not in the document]
tachella: because our deepinv's `plot` function can handle different sizes, whereas wandb can't

I think it's nice to see the measurements when we can, even if they are smaller (ie super res or blur without padding)

deepinv/training/trainer.py
```diff
@@ -59,85 +56,150 @@ class Trainer:
...
     :param torch.nn.Module model: Reconstruction network, which can be PnP, unrolled, artifact removal
         or any other custom reconstruction network.
     :param deepinv.physics.Physics, list[deepinv.physics.Physics] physics: Forward operator(s) used by the reconstruction network.
-    :param int epochs: Number of training epochs. Default is 100.
-    :param torch.optim.Optimizer optimizer: Torch optimizer for training the network.
     :param torch.utils.data.DataLoader, list[torch.utils.data.DataLoader] train_dataloader: Train data loader(s) should provide a
         a signal `x` or a tuple of `(x, y)` signal/measurement pairs or a tuple `(x, y, params)`
-        where `params` is a dict of physics generator parameters to be loaded into the physics each iteration.
+        where `params` is a dict of physics generator parameters to be loaded into the physics each iteration. Default is ``None``.
+    :param bool online_measurements: Generate the measurements in an online manner at each iteration by calling
```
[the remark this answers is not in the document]
tachella: I agree, we can add add this comment

deepinv/training/trainer.py
```diff
@@ -39,15 +39,12 @@ class Trainer:
...
+    Assuming that `x` is the ground-truth reference and `y` is the measurement and `params` is a dict of :ref:`physics parameters <physics_generators>`,
+    the **dataloaders** should return one of the following options:
 
-    .. note::
-
-        The training code can synchronize with `Weights & Biases <https://wandb.ai/site>`_ for logging and visualization
-        by setting ``wandb_vis=True``. The user can also customize the wandb setup by providing
-        a dictionary with the setup for wandb.
+    1. `(x, y)` or `(x, y, params)`, which requires `online_measurements=False` (default) otherwise `y` will be ignored and new measurements will be generated online.
+    2. `(x)` or `(x, params)`, which requires `online_measurements=True` for generating measurements in an online manner as `y=physics(x)`. Otherwise, first generate a dataset of `(x,y)` with :class:`deepinv.datasets.generate_dataset` and then use option 1 above.
```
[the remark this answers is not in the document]
tachella: I added option 2 which is the best in my opinion

deepinv/training/trainer.py
```diff
@@ -59,85 +56,150 @@ class Trainer:
...
-        **Caveat**: this requires ``shuffle=False`` in your dataloaders.
+        The training code can synchronize with `Weights & Biases <https://wandb.ai/site>`_ for logging and visualization
+        by setting ``wandb_vis=True``. The user can also customize the wandb setup by providing
+        a dictionary with the setup for wandb.
 
-        An alternative, safer solution is to generate and save params offline using :func:`deepinv.datasets.generate_dataset`.
-        The params dict will then be automatically updated every time data is loaded.
+    :Basics:
 
     :param torch.nn.Module model: Reconstruction network, which can be PnP, unrolled, artifact removal
```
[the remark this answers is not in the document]
tachella: done

deepinv/training/trainer.py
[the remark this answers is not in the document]
tachella: done

deepinv/training/trainer.py
```diff
@@ -59,85 +56,150 @@ class Trainer:
...
+    :Optimization:
+
+    :param None, torch.optim.Optimizer optimizer: Torch optimizer for training the network. Default is ``None``.
+    :param int epochs: Number of training epochs.
+        Default is 100. The trainer will perform gradient steps equal to the `min(epochs*n_batches, max_batch_steps)`.
+    :param int max_batch_steps: Number of gradient steps per iteration.
+        Default is 1e10. The trainer will perform gradient steps equal to the `min(epochs*n_batches, max_batch_steps)`.
+    :param None, torch.optim.lr_scheduler.LRScheduler scheduler: Torch scheduler for changing the learning rate across iterations. Default is ``None``.
+    :param bool early_stop: If ``True``, the training stops when the evaluation loss is not improving. Default is ``False``.
+        The user can modify the strategy for saving the best model by overriding the :func:`deepinv.Trainer.stop_criterium` method.
```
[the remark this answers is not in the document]
tachella: haha yes, I was wondering about this, I tried to make up a latin word it seems

deepinv/training/trainer.py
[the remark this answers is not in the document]
tachella: I was thinking of callbacks, it is much easier for the user to change this function without requiring to understand all the properties of the class, ie just using the inputs

deepinv/training/trainer.py
```diff
@@ -814,6 +895,61 @@ def reset_metrics(self):
...
+
+        :param int epoch: Current epoch.
+        :param int train_ite: Current gradient iteration, gives a more fine-grained control than epochs.
+        :param dict metric_history: Dictionary containing the metrics history, with the metric name as key.
+        :param list metrics: List of metrics used for evaluation.
+        """
+        if train_ite > 1:  # do at least one gradient step
+            k = 0  # index of the first metric
+            history = metric_history[metrics[k].__class__.__name__]
+            if hasattr(metrics[k], "lower_better"):
```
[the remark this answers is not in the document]
tachella: nice trick, didn't know this

deepinv/training/trainer.py
tachella: what is currently done is checking if the first metric hasn't improved in the last 5 epochs, I'm not sure I understand your question

deepinv/training/trainer.py
tachella: Ok I removed it, but I thought it would be nice to automate picking the best model for users.

deepinv/training/trainer.py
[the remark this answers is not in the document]
Andrewwango: Ok, I've just fixed the implementation, can you check that it is ok and add a test (should be able to add it to `test_trainer`/`test_trainer_physics_generator_params`)

deepinv/training/trainer.py
Andrewwango: Sorry I misunderstood the code!

Andrewwango (changes requested):

Andrewwango (approved):
lgtm, thanks

PR #444, opened 2025-04-14
paper : add optim

samuro95:
Add details on optimization


tachella (changes requested):
minor latex detail, otherwise I think its great!

PR #454, opened 2025-04-19
ENSURE loss

Andrewwango:
ENSURE loss from https://arxiv.org/abs/2010.10631

### DIFF
--- a/docs/source/api/deepinv.loss.rst
+++ b/docs/source/api/deepinv.loss.rst
@@ -53,6 +53,7 @@ Self-Supervised Learning
     deepinv.loss.TVLoss
     deepinv.loss.R2RLoss
     deepinv.loss.ScoreLoss
+    deepinv.loss.ENSURELoss
 
 Adversarial Learning
 --------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,8 @@ Current
 New Features
 ^^^^^^^^^^^^
 
+- Add ENSURE loss (:gh:`430` by `Andrew Wang`_)
+
 Changed
 ^^^^^^^
 
--- a/deepinv/loss/__init__.py
+++ b/deepinv/loss/__init__.py
@@ -5,7 +5,7 @@ from .sup import SupLoss
 from .score import ScoreLoss
 from .tv import TVLoss
 from .r2r import R2RLoss
-from .sure import SureGaussianLoss, SurePoissonLoss, SurePGLoss
+from .sure import SureGaussianLoss, SurePoissonLoss, SurePGLoss, ENSURELoss
 from .regularisers import JacobianSpectralNorm, FNEJacobianSpectralNorm
 from .measplit import (
     SplittingLoss,
--- a/deepinv/loss/sure.py
+++ b/deepinv/loss/sure.py
@@ -1,9 +1,17 @@
-import torch
-import torch.nn as nn
-import numpy as np
-import deepinv.physics
-from deepinv.loss.loss import Loss
+from __future__ import annotations
+from typing import TYPE_CHECKING
 
+import numpy as np
+from torch import Tensor
+import torch
+
+from deepinv.loss.loss import Loss
+from deepinv.physics.mri import MRI, MRIMixin
+
+if TYPE_CHECKING:
+    from deepinv.physics.generator.base import PhysicsGenerator
+    from deepinv.models.base import Reconstructor
+    from deepinv.physics.forward import Physics
 
 def hutch_div(y, physics, f, mc_iter=1, rng=None):
     r"""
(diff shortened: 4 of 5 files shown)


tachella (commented):
I feel that this loss is a too specific version of the SURE loss, where the metric B is set using an expectation over the physics generator.

In my opinion this would clutter a bit the library and confuse practitioners, so I'm not sure about adding it.

why not include it in an example?

tachella:
The loss is too close to SURE to motivate a new class in my opinion.


I would nonetheless maybe keep a function for computing the metric based on a physics generator, e.g. as done in

```
d = physics_generator.step(batch_size=2000)["mask"].mean(0, keepdim=True)
self.dsqrti = 1 / d.sqrt()
```

but otherwise I would suggest to close the PR

deepinv/loss/mri/sure.py
```diff
@@ -0,0 +1,93 @@
...
+
+    .. math::
+
+        y \sim\mathcal{N}(u,\sigma^2 I) \quad \text{with}\quad u= A(x).
+
+    The loss is computed as
+
+    .. math::
+
+        \frac{1}{m}\|W^{-1}P(A^{\dagger}y - \inverse{y})\|_2^2 +\frac{2\sigma^2}{m\tau}b^{\top} \left(\inverse{A^{\top}y+\tau b_i} -
```
tachella, line 31: why \inverse{A^{\top}y} ? can't this loss work for a general network R(y,A)?

If that's the case, we should specify that it only works for Artifact Removal networks?

deepinv/loss/mri/sure.py
```diff
@@ -0,0 +1,93 @@
...
+        r"""
+        Monte-Carlo estimation for the divergence of f(x).
+
+        :param torch.Tensor x_net: Reconstructions.
+        :param torch.Tensor y: Measurements.
+        :param deepinv.physics.Physics physics: Forward operator associated with the measurements.
+        :param torch.nn.Module f: Reconstruction network.
+        """
+        b = torch.empty_like(y).normal_(generator=self.rng)
+        x2 = f(physics.A(physics.A_adjoint(y) + b * self.tau), physics)
```
tachella, line 75: this doesn't exactly match the math definition above

tachella (changes requested):

deepinv/loss/mri/sure.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def __init__(
+        self,
+        sigma: float,
+        physics_generator: PhysicsGenerator,
+        tau: float = 0.01,
+        rng: torch.Generator = None,
+    ):
+        super().__init__(sigma=sigma, tau=tau, rng=rng)
+        d = physics_generator.step(batch_size=2000)["mask"].mean(0, keepdim=True)
```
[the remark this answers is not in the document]
Andrewwango: Yes potentially, and I raise this issue here: https://github.com/deepinv/deepinv/issues/486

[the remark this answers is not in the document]
Andrewwango: But for now it should be ok, and I propose a `physics_generator.average`

deepinv/loss/mri/sure.py
```diff
@@ -0,0 +1,93 @@
...
+
+    The loss is computed as
+
+    .. math::
+
+        \frac{1}{m}\|W^{-1}P(A^{\dagger}y - \inverse{y})\|_2^2 +\frac{2\sigma^2}{m\tau}b^{\top} \left(\inverse{A^{\top}y+\tau b_i} -
+        \inverse{A^{\top}y}\right)
+
+    where :math:`R` is the trainable network, :math:`A` is the forward operator,
+    :math:`y` is the noisy measurement vector of size :math:`m`, :math:`W^{-1}P` is a weighted projection operator determined by the set of measurement operators,
```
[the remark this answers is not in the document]
Andrewwango: done

deepinv/loss/mri/sure.py
```diff
@@ -0,0 +1,93 @@
...
+    from deepinv.physics.generator.base import PhysicsGenerator
+    from deepinv.models.base import Reconstructor
+    from deepinv.physics.forward import Physics
+
+
+class ENSURELoss(SureGaussianLoss):
+    r"""
+    ENSURE loss for image reconstruction in Gaussian noise.
+
+    The loss function extends :class:`deepinv.loss.SureGaussianLoss` and is designed for the following noise model:
```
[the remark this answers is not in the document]
Andrewwango: done

deepinv/loss/mri/sure.py
```diff
@@ -0,0 +1,93 @@
...
+
+    .. math::
```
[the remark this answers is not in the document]
Andrewwango: done

deepinv/loss/mri/sure.py
Andrewwango: They state this is the case in their eq. 15
image
This is valid for ArtifactRemoval, unrolled etc.

Andrewwango: See below comment on how we generalise this

deepinv/loss/mri/sure.py
Andrewwango: Firstly, we note that ENSURE only applies for networks $R(\cdot)=r(A^\top\cdot)$ such that $R: u\rightarrow x,u=A^\top y$ as noted in the above comment. Assume that the user is using a valid model.

To be compatible with our model framework $model: y\rightarrow x$, we use the fact that $A^\top A A^\top = A^\top$ and add the `physics.A`.

Andrewwango: Would you agree that this is a sufficiently general way of doing this, as long as we assume that the model that the user passes in is indeed in the form $r(A^\top\cdot)$?

Andrewwango: Would you agree that this is a sufficiently general way of doing this, as long as we assume that the model that the user passes in is indeed in the form $r(A^\top\cdot)$?

tachella: this also assumes that A^{t}AA^{t} = A^t which is only true for operators with binary eigenvalues

deepinv/loss/mri/sure.py
tachella: they require in the proof that $\hat{\rho}$ (the reconstruction net) is independent of $s$ (aka $A_s$)
image

Andrewwango: I've addressed this in the docs and by adding an artifactremoval switch

deepinv/loss/mri/sure.py
Andrewwango: I've addressed this in the docs and by adding an artifactremoval switch

deepinv/loss/mri/sure.py
```diff
@@ -0,0 +1,108 @@
...
+    :math:`b\sim\mathcal{N}(0,I)`, :math:`\tau\geq 0` is a hyperparameter controlling the
+    Monte Carlo approximation of the divergence, and :math:`\Beta=W^{-1}P`
+    where :math:`W` is a weighting determined by the set of measurement operators and :math:`P` is the projection operator onto the range space of :math:`\A^\top`.
+
+    The ENSURE loss was proposed in `Aggarwal et al <https://arxiv.org/abs/2010.10631>`_ for MRI.
+
+    .. warning::
+
+        This loss was originally proposed only to be used with :class:`artifact removal models <deepinv.models.ArtifactRemoval>` which can be written in the form :math:`\inverse{\cdot}=r(A^\top\cdot)`.
+        If an artifact removal model is not used, then we input to network instead :math:`A A^\top\cdot` (assuming that :math:`A A^\top=I`).
```
tachella, line 48: sorry I'm still bugging on this, but I would evaluate a general model as `model(y,physics)` instead of adding :math:`A A^\top\cdot`

deepinv/loss/mri/sure.py
```diff
@@ -0,0 +1,108 @@
...
+
+    .. math::
+
+        \frac{1}{m}\|\Beta(A^{\dagger}y - \inverse{y})\|_2^2 +\frac{2\sigma^2}{m\tau}b^{\top} \left(\inverse{A^{\top}y+\tau b_i} -
+        \inverse{A^{\top}y}\right)
+
+    where :math:`R` is the trainable network (which takes :math:`A^\top y` as input),
+    :math:`A` is the forward operator,
+    :math:`y` is the noisy measurement vector of size :math:`m`,
+    :math:`b\sim\mathcal{N}(0,I)`, :math:`\tau\geq 0` is a hyperparameter controlling the
```
tachella, line 39: I would be more specific on how W is computed.

I think this only works for forward operators with an SVD decomposition, ie LinearPhysics, otherwise it is not feasible to compute W. We could clarify this too.

tachella (changes requested):

deepinv/loss/mri/sure.py
```diff
@@ -0,0 +1,109 @@
...
+        :param torch.Tensor x_net: Reconstructions.
+        :param torch.Tensor y: Measurements.
+        :param deepinv.physics.Physics physics: Forward operator associated with the measurements.
+        :param torch.nn.Module f: Reconstruction network.
+        """
+        b = torch.empty_like(y).normal_(generator=self.rng)
+        u = physics.A_adjoint(y) + b * self.tau
+
+        if isinstance(f, ArtifactRemoval):
+            x2 = f.backbone_inference(u, physics, None)
```
tachella: Sorry for going in circles, but I don't get why we need to do this special treatment here. Why not defining u=y + tau * b and then applying model(y, physics)?

Andrewwango: Because according to the paper, the model input is $A^\top y + b\tau $, not $A^\top(y+b\tau)$?

Andrewwango: I wanted to replicate the formulae from the paper rather than trying generalise

Andrewwango: i.e. figure 1b) of https://arxiv.org/abs/2010.10631

tachella: this is too hacky for me, I think it defeats the point of the library to have very hand-crafted solutions.

It will only work when `y` has the same size as `A_adjoint(y)`, and it is very weird to add `b` noise in the nullspace of `A`

I would go for the same divergence computation as in the SureGaussianLoss

Andrewwango: done

docs/source/api/deepinv.loss.rst
```diff
@@ -67,6 +67,17 @@ Specialized self-supervised losses for MRI
...
+--------------------------
+.. userguide:: mri-losses
+
+.. autosummary::
+   :toctree: stubs
+   :template: myclass_template.rst
+   :nosignatures:
+
+    deepinv.loss.mri.ENSURELoss
+
```
tachella, line 70: Phase2Phase and Artifact2Artifact WeightedSplittingLoss RobustSplittingLoss shouldn't go here too?

deepinv/loss/mri/sure.py
```diff
@@ -0,0 +1,101 @@
...
+    ) -> Tensor:
+        r"""
+        Monte-Carlo estimation for the divergence of f(x).
+
+        :param torch.Tensor x_net: Reconstructions.
+        :param torch.Tensor y: Measurements.
+        :param deepinv.physics.Physics physics: Forward operator associated with the measurements.
+        :param torch.nn.Module f: Reconstruction network.
+        """
+        return mc_div(physics(x_net), y, f, physics, tau=self.tau, rng=self.rng)
```
tachella:
```suggestion
        return mc_div(physics.A(x_net), y, f, physics, tau=self.tau, rng=self.rng)
```

deepinv/loss/mri/sure.py
```diff
@@ -0,0 +1,100 @@
...
+
+    where :math:`R` is the trainable network (which takes :math:`A^\top y` as input),
+    :math:`A` is the forward operator,
+    :math:`y` is the noisy measurement vector of size :math:`m`,
+    :math:`b\sim\mathcal{N}(0,I)`, :math:`\tau\geq 0` is a hyperparameter controlling the
+    Monte Carlo approximation of the divergence, and :math:`\Beta=W^{-1}P`
+    where :math:`P` is the projection operator onto the range space of :math:`\A^\top`
+    and :math:`W` is a weighting determined by the set of measurement operators where :math:`W^2=\mathbb{E}\left[P\right]`.
+
+    The ENSURE loss was proposed in `Aggarwal et al <https://arxiv.org/abs/2010.10631>`_ for MRI.
```
tachella: use footcite for the citation

tachella (changes requested):
final thing to fix

deepinv/loss/mri/sure.py
Andrewwango: done

tachella:
There's some black issues

Issue #383, opened 2025-01-02
API BlurFFT operator not compatible with torch.nn.Module API


tomMoral:
This script raises an error on the second call to `phy(x)` as the attribute `self.angle` is not moved to `cuda` while the other parameters are:

```python
from deepinv.physics import BlurFFT
import torch


h = torch.ones((1, 1, 5, 5))
phy = BlurFFT((15, 15), h)

x = torch.ones((1, 1, 15, 15))
phy(x)

x = x.cuda()
phy = phy.cuda()
phy(x)
```

A possible solution is to use this patch:
```diff
diff --git a/deepinv/physics/blur.py b/deepinv/physics/blur.py
index b489aeee..836368aa 100644
--- a/deepinv/physics/blur.py
+++ b/deepinv/physics/blur.py
@@ -375,7 +375,9 @@ class BlurFFT(DecomposablePhysics):
 
             mask = filter_fft_2d(filter, self.img_size).to(self.device)
             self.angle = torch.angle(mask)
-            self.angle = torch.exp(-1.0j * self.angle).to(self.device)
+            self.angle = torch.nn.Parameter(
+                torch.exp(-1.0j * self.angle).to(self.device), requires_grad=False
+            )
             mask = torch.abs(mask).unsqueeze(-1)
             mask = torch.cat([mask, mask], dim=-1)
             self.mask = torch.nn.Parameter(mask, requires_grad=False)
```

But this seems to go against the current behavior, where the `device` argument is strictly enforced (thus requiring to know it in advance).
I would favor making the device parameter optional (with `device=None`), putting all the important quantities as  `Parameter` and using a final call to `self.to(self.device)` if it is explicitly set. This way explicit device is enforced, while this allows to use the `cuda/to` API from the `torch.nn.Module`.

Issue #460, opened 2025-04-24
Moving physics between devices


jscanvic:
The base class `deepinv.physics.Physics` inherits from the class `torch.nn.Module` and as such it could support the method `Module.to` which is meant to render the module compatible with inputs of a certain dtype and device. Even though `Module.to` works fairly automatically by traversing the tree of parameters and submodules, the way certain operators are currently implemented in the library makes it insufficient.

I suggest to make it standard for physics operators to be movable between devices using `Physics.to`.

**Reproducing**

```python
import deepinv as dinv
import torch

physics = dinv.physics.Tomography(angles=50, img_width=256, parallel_computation=True, device="cpu")
x = torch.randn(1, 1, 256, 256, device="cpu")

x = x.to("cuda:0")
physics.to("cuda:0")

# Raises an exception.
x_hat = physics.A_dagger(physics.A(x))
```

**Expected behavior**

No exception should be raised.

**Implementation**

From what I can tell, it should be possible to enable that by following a few guidelines

* Devices and dtypes that are passed in to physics operators through their constructor should only be used for initialization purposes, and more precisely, they should not be stored as attributes.
* Tensors stored as attributes should always be registered as parameters, either implicitely or explicitely. By doing so, they will be automatically moved to the right device when calling `Physics.to`.
* Callable class instances stores as attributes should be registered as submodules, either implicitely or explicitely. For instance, `Tomography.iradon` is rightly registered as a submodule, although implicitely, as `IRadon` is a subclass of `Module`.
* Tensors created in forward calls should always use the same device and dtype as either input tensors or stored parameters depending on the situation. They should crucially not used a value stored as an attribute that would not be automatically updated by `Physics.to`.

(Many classes in the code base currently follow the (anti-)pattern of storing as attributes dtypes and devices passed in at construction.)

By applying these guidelines to the class `IRadon` used by `Tomography`, I managed to make the example work fairly easily.

tachella:
I fully agree with this proposal - we should work on this

mh-nguyen712:
Totally agree! I already addressed this issue in my recent PR #462

jscanvic:
I figured an automated fuzzer might come in handy to find device moving related bugs.

It hooks up to the unit tests to get physics instances and input tensors and checks if, after moving the physics and inputs to some device, applying the operator succeeds as expected.

@mh-nguyen712 You might find it helpful to check if #462 completely solves this issue!

[...]

Issue #430, opened 2025-03-20
Bug with physics Downsampling A_adjoint function


xuwayyy:
When call A_adjoint function in Downsampling class with a str parameter 'filter', the function does not convert it to a torch.nn.Parameter like in __init__.py
![Image](https://github.com/user-attachments/assets/51e62504-7587-4bb0-826b-cf35e1f94572)

Issue #412, opened 2025-02-21
`to_nn_parameter` should be `update_parameters`


Andrewwango:
The `to_nn_parameter` function should be a method of the noise model baseclass i.e. `NoiseModel.update_parameters(**kwargs)` and then `GaussianNoise` etc. call `super().update_parameters(sigma=sigma)`. This is inline with other `Physics` e.g. `MRI`.

https://github.com/deepinv/deepinv/blob/568e51f2f4f64e224ee9ae54af13009303717228/deepinv/physics/noise.py#L494

Andrewwango:
Related to #414

PR #462, opened 2025-04-26
Make physics classes differentiable

### LINKED ISSUE #438: Physics parameters are non-differentiable
The current `update_parameters()` of `Physics` and `NoiseModel`  does not allow one to differentiate the output of the physics w.r.t to the parameter. For example, the following code snippet won't work:
```
physics = dinv.physics.Blur(device='cpu')
filter = torch.rand((1, 1, 5, 5))
filter /= filter.sum(dim=(-2,-1), keepdim=True)

with torch.enable_grad():
    filter.requires_grad_(True)
    y_cur = physics.A(x, filter=filter)
    loss = torch.nn.functional.mse_loss(y_cur, y)
    
grad = torch.autograd.grad(loss, filter)[0]
```

The reason is inside the `update_parameters()`, we forced the `filter` to be a `nn.Parameter` with `requires_grad = False`.

A simple work-around could be:
```
  def update_parameters(self, **kwargs):
      if kwargs:
          for key, value in kwargs.items():
              if (
                  value is not None
                  and hasattr(self, key)
                  and isinstance(value, torch.Tensor)
              ):
                  self.register_buffer(key, value)
```
The `register_buffer` acts similar to `nn.Parameter`, except that `filter` won't be show up in `physics.parameters()`, but I think it's not very important. The `filter` will be in the `state_dict` . Methods like `physics.to(...)` or `physics.cuda()`  will work properly.

This is related to #412  #387 #414 #383

I can create a PR soon for adressing this issue.

mh-nguyen712:
Close #438 and related to #430 #414 #412 #387 #383 #460 
- Make physics classes differentiable, with an example and test
- Make physics device consistent: `.to()`, `.cuda()`, `.cpu()` work properly 
- `physics.update_parameters()` properly handles submodules like `noise_model` and pseudo-random number generator `rng`
- Some minor fixes in code and docs.
  

deepinv/tests/test_physics.py
```diff
@@ -1240,5 +1185,121 @@ def test_unmixing(device):
...
+        y2 = physics(x)
+        assert y2.device == cuda
+
+        # Denoising add random noise in each forward call
+        if not isinstance(physics, dinv.physics.Denoising):
+            if isinstance(y2, TensorList):
+                for y11, y22 in zip(y1, y2):
+                    torch.allclose(y11.to(cuda), y22, rtol=1e-3)
+            else:
+                assert torch.allclose(y1.to(cuda), y2, rtol=1e-3)
```
jscanvic, line 1305: I think it tests some of the operators (those in OPERATORS) but not those in NONLINEAR_OPERATORS, PHASE_RETRIEVAL_OPERATORS, and NOISES. It might be worth testing them as well!

mh-nguyen712: Many thanks! Added!

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,268 @@
...
+    y_hat = to_optimize_physics.A(x)
+    loss = loss_fn(y_hat, y)
+    loss.backward()
+
+    # a gradient step
+    optimizer.step()
+
+    # projection step, when doing additional steps, it's important to change only
+    # the tensor data to avoid breaking the gradient computation
+    to_optimize_physics.filter.data = projection_simplex_sort(
```
tachella, line 244: I'm not a big fan of the name `to_optimize_physics`, can we just write `physics`?

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,268 @@
...
+kernel_init[..., 5:-5, 5:-5] = 1.0
+kernel_init = projection_simplex_sort(kernel_init)
+
+kernel_hat = kernel_init.clone()
+optimizer = torch.optim.Adam([kernel_hat], lr=0.1)
+
+# We will alternate a gradient step and a projection step
+losses = []
+n_iter = 200
+for i in tqdm(range(n_iter)):
```
tachella, line 171: the learning rate doesn't seem to have any effect in this method, could you check that it works for you locally?

Otherwise I would remove it, since it is quite similar to the last option, which I find more elegant.

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,268 @@
...
+# %%
+#
+# Run the algorithm
+#
+kernel_init = torch.zeros_like(true_kernel)
+kernel_init[..., 5:-5, 5:-5] = 1.0
+kernel_init = projection_simplex_sort(kernel_init)
+n_iter = 1000
+stepsize = 0.7
+kernel_hat, losses = projected_gradient_descent(
```
tachella, line 132: I would write the for loop here, instead of using a function, to match the other options below where you wrote an explicit for loop

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,268 @@
...
+    return w.reshape(shape)
+
+
+# Now we can define the projected gradient descent steps
+def projected_gradient_descent(physics, x, y, kernel_init, n_iter=100, stepsize=0.01):
+    kernel_hat = kernel_init.clone()
+    losses = []
+    for i in tqdm(range(n_iter)):
+        # gradient step
+        grad = gradient(physics, x, y, kernel_hat)
```
tachella, line 113: I would explicit this grad step here, ie copy the contents of the `gradient` function here

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,268 @@
...
+#
+
+
+def loss_fn(physics, x, y, filter):
+    y_hat = physics.A(x, filter=filter)
+    loss = (y_hat - y).pow(2).mean()
+    return 0.5 * loss
+
+
+def gradient(physics, x, y, filter):
```
tachella, line 82: see comment below - I would move this to the for loop instead of using a dedicated function

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,268 @@
...
+    filter = filter.clone().requires_grad_(True)
+    with torch.enable_grad():
+        y_hat = physics.A(x, filter=filter)
+        loss = torch.nn.functional.mse_loss(y_hat, y)
```
tachella, line 86: here I would suggest to use a data fidelity instead, ie define `data_fidelity = dinv.optim.L2()`
```suggestion
        physics.update(filter=filter)
        loss = data_fidelity(y=y, x=x, physics=physics) / y.numel()
```

Normalize by `y.numel()` if you want to have the same loss value as the current one

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,268 @@
...
+        loss.backward()
+    return filter.grad
+
+
+@torch.no_grad
```
tachella, line 91: for some reason I'm getting an error with this decorator, can we move it to `with torch.no_grad():` inside the function?

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,268 @@
...
+    losses = []
+    for i in tqdm(range(n_iter)):
+        # gradient step
+        grad = gradient(physics, x, y, kernel_hat)
+        kernel_hat = kernel_hat - stepsize * grad
+        # projection step
+        kernel_hat = projection_simplex_sort(kernel_hat)
+        # loss
+        with torch.no_grad():
+            losses.append((physics.A(x, filter=kernel_hat) - y).pow(2).sum().item())
```
tachella, line 119: if you move the gradient computation in the loop, you can change this line to avoid double computation
```suggestion
            losses.append(loss.item())
```

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+def loss_fn(physics, x, y, filter):
+    y_hat = physics.A(x, filter=filter)
+    loss = (y_hat - y).pow(2).mean()
+    return 0.5 * loss
+
+
+def gradient(physics, x, y, filter):
+    filter = filter.clone().requires_grad_(True)
```
tachella, line 83: I don't fully understand if this hack of cloning the filter to compute gradients is needed?

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,268 @@
...
+
+y = physics(x, filter=true_kernel)
+
+dinv.utils.plot([x, y, true_kernel], titles=["Sharp", "Blurry", "True kernel"])
+
+# %%
+# Define an optimization algorithm
+# --------------------------------
+#
+# The convolution kernel lives in the simplex: positive entries summing to 1.
```
tachella, line 69:
```suggestion
# The convolution kernel lives in the simplex, ie the kernel must have positive entries summing to 1.
```

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,268 @@
...
+# We can use one of the most basic optimization algorithm -- Projected Gradient Descent.
```
tachella, line 70:
```suggestion
# We can use a simple optimization algorithm - Projected Gradient Descent - to enforce this constraint.
```

docs/source/user_guide/physics/intro.rst
```diff
@@ -93,6 +93,7 @@ imaging system optimization, etc. The following example shows how operators and
    >>> dict_params = {'filter': theta, 'dummy': None}
    >>> y = physics(x, **dict_params) # # we define the blur by passing in the dictionary
 
+One can also optimize the parameter :math:`\theta`, as show in this example: :ref:`sphx_glr_auto_examples_advanced_demo_optimizing_physics_parameter.py`
```
tachella, line 96: can you include a little code snippet here showing the differentiation?

deepinv/tests/test_physics.py
```diff
@@ -677,57 +708,6 @@ def test_MRI(mri, mri_img_size, device, rng):
             assert xrss.shape == (x.shape[0], 1, *x.shape[2:])  # B,1,H,W
 
 
-@pytest.mark.parametrize("mri", [MRI, DynamicMRI, MultiCoilMRI])
```
tachella, line 680: why removing this test?

deepinv/tests/test_physics.py
```diff
@@ -1051,38 +1030,6 @@ def test_downsampling_adjointness(device):
                     assert torch.abs(Axy - Atyx) < 1e-3
 
 
-def test_prox_l2_downsampling(device):
```
tachella, line 1054: why removing this test?

deepinv/tests/test_physics.py
```diff
@@ -1240,5 +1187,156 @@ def test_unmixing(device):
...
+        # Move to GPU if cuda is available
+        if torch.cuda.is_available():
+            torch.manual_seed(11)
+            cuda = torch.device("cuda:0")
+            physics = physics.to(cuda)
+            x = x.to(cuda)
+            y2 = physics(x)
+            assert y2.device == cuda
+
+            # Denoising add random noise in each forward call
```
tachella, line 1336:
```suggestion
            # skip Denoising that adds random noise in each forward call
```

docs/source/user_guide/physics/intro.rst
```diff
@@ -65,7 +65,7 @@ More details can be found in the doc of each class.
 Parameter-dependent operators
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Many (linear or non-linear) operators depend on (optional) parameters :math:`\theta` that describe the imaging system, ie
+Many (linear or non-linear) operators depend on (optional) parameters :math:`\theta` that describe the imaging system, i.e
```
tachella, line 68: I would keep `ie `(british english) or go for `i.e.` (american english)

deepinv/tests/test_generators.py
```diff
@@ -228,16 +228,17 @@ def test_downsampling_generator(num_channels, device, dtype):
         filter="bicubic",
         factor=4,
     )
-    list_filters = ["bilinear", "bicubic"]
-    list_factors = [2, 4]
+    # unused?
```
tachella, line 231: we can remove them if they are not being used

deepinv/physics/noise.py
```diff
@@ -90,6 +97,82 @@ def randn_like(self, input: torch.Tensor, seed: int = None):
...
+        device = self._get_device_from_args(*args, **kwargs)
+        if device is not None and self.rng is not None:
+            state = self.rng.get_state()
+            # Move the generator to the specified device
+            self.rng = torch.Generator(device=device)
+            try:
+                self.rng.set_state(state)
+            except RuntimeError:
+                warnings.warn(
+                    "Moving the random number generator between CPU/GPU is not possible. Re-initializing the generator."
```
tachella, line 138:
```suggestion
                    "Moving the random number generator between CPU/GPU is not possible. Reinitialize the generator on the correct device."
```

how often does this happen?

CHANGELOG.rst
```diff
@@ -45,6 +45,11 @@ New Features
 - Online training with noisy physics now can repeat the same noise each epoch (:gh:`414` by `Andrew Wang`_)
 - Trainer test can return unaggregated metrics (:gh:`420` by `Andrew Wang`_)
 - MoDL model (:gh:`435` by `Andrew Wang`_)
+- Add ComposedLinearPhysics (used my __mul__ method) (:gh:`462` by `Minh Hai Nguyen`_)
```
tachella, line 48:
```suggestion
- Add ComposedLinearPhysics (via __mul__ method) (:gh:`462` by `Minh Hai Nguyen`_)
```

tachella (changes requested):
thanks for this PR, it is fixing lots of things.

Please see some comments to address below.

I think we should also add a test checking that loading & saving physics is working fine, something like

```
physics = dinv.physics.Inpainting(tensor_size=(1, 32, 32))
mask1 = physics.mask
torch.save("my_physics.pt", physics.state_dict())

# generate new physics
physics2 = dinv.physics.Inpainting(tensor_size=(1, 32, 32))

assert physics2.mask != mask1

# load saved physics
physics2.load_state_dict(torch.load("my_physics.pt"))
assert physics2.mask == mask1
```
(ideally generalizing this for all operators)

deepinv/physics/forward.py
```diff
@@ -616,6 +614,41 @@ def A_dagger(
         )
 
 
+class ComposedLinearPhysics(LinearPhysics):
```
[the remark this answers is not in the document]
mh-nguyen712: Thanks Julian. I think that your implementation is good. I'll send your a message for merging this

deepinv/physics/noise.py
mh-nguyen712: I only happens if one initialize the rng on CPU and send to GPU

deepinv/tests/test_generators.py
mh-nguyen712: done

deepinv/tests/test_physics.py
mh-nguyen712: Sorry for this in fact I'm removing nothing here. I was comment out other tests when testing on my local to accelerate the tests and forgot to uncomment them before pushing so the git keeps tracking theses changes. But they are identical.

examples/advanced/demo_optimizing_physics_parameter.py
mh-nguyen712: Thanks for your detailed reviews Julian!

examples/advanced/demo_optimizing_physics_parameter.py
mh-nguyen712: I forgot the parentheses

examples/advanced/demo_optimizing_physics_parameter.py
mh-nguyen712: good idea, thanks

examples/advanced/demo_optimizing_physics_parameter.py
mh-nguyen712: Sure. I changed it to physics

examples/advanced/demo_optimizing_physics_parameter.py
mh-nguyen712: It do have an effect the learning rate on my local. It's true that this method is very similar to the last option, I just want to show that there are different options to optimize the parameter

deepinv/tests/test_physics.py
mh-nguyen712: Added a gain. Really sorry about this

examples/advanced/demo_optimizing_physics_parameter.py
tachella: ok, we can keep it then

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,258 @@
...
+
+# We will alternate a gradient step and a projection step
+losses = []
+n_iter = 200
+for i in tqdm(range(n_iter)):
+    optimizer.zero_grad()
+    # compute the gradient, this will directly change the gradient of `kernel_hat`
+    with torch.enable_grad():
+        kernel_hat.requires_grad_(True)
+        physics.update_parameters(filter=kernel_hat)
```
tachella, line 163: we should use 
```suggestion
        physics.update(filter=kernel_hat)
```

(update_parameters shouldn't be called explicitly, as it is handled internally by `update`)

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,258 @@
...
+n_iter = 1000
+stepsize = 0.7
+
+kernel_hat = kernel_init
+losses = []
+for i in tqdm(range(n_iter)):
+    # compute the gradient
+    with torch.enable_grad():
+        kernel_hat.requires_grad_(True)
+        physics.update_parameters(filter=kernel_hat)
```
tachella, line 111: same here
```suggestion
        physics.update(filter=kernel_hat)
```

examples/advanced/demo_optimizing_physics_parameter.py
mh-nguyen712: I'm not fully understand the logic of `update` and `update_parameters`, can you clarify? Currently I suppose that they do the same job.

deepinv/physics/forward.py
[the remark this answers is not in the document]
mh-nguyen712: Done

examples/advanced/demo_optimizing_physics_parameter.py
tachella: physics.update() calls internally the update_parameters() of both the operator **and the noise**.

So if you do physics.update(sigma=.1, filter=filter) this will call


physics.update_parametera(filter=filter)
physics.noise_model.update_parameters(sigma=sigma)

tachella (approved):
looks good to me, thanks Hai

deepinv/physics/noise.py
```diff
@@ -70,7 +71,12 @@ def reset_rng(self):
         r"""
         Reset the random number generator to its initial state.
         """
-        self.rng.set_state(self.initial_random_state)
+        if self.rng is not None:
+            self.rng.set_state(self.initial_random_state)
+        else:
+            warnings.warn(
+                "Cannot reset state for random number generator because it is not initialized. This is ignore."
```
matthieutrs, line 78:
```suggestion
                "Cannot reset state for random number generator because it was not initialized. This is ignored."
```

deepinv/physics/noise.py
```diff
@@ -90,6 +96,82 @@ def randn_like(self, input: torch.Tensor, seed: int = None):
         self.rng_manual_seed(seed)
         return torch.empty_like(input).normal_(generator=self.rng)
 
+    def update_parameters(self, **kwargs):
+        r"""
+
```
matthieutrs, line 101:
```suggestion
```

deepinv/tests/test_physics.py
```diff
@@ -7,6 +7,10 @@
 import deepinv as dinv
 from deepinv.optim.data_fidelity import L2
 from deepinv.physics.mri import MRI, MRIMixin, DynamicMRI, MultiCoilMRI, SequentialMRI
+import copy
+from deepinv.utils import TensorList
+import os
+import shutil
```
matthieutrs, line 13: maybe move core libraries (copy, os, shutil) on top

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,258 @@
...
+#
+# We can plot the loss to make sure that it decreases
+#
+plt.figure()
+plt.plot(range(n_iter), losses)
+plt.title("Loss evolution")
+plt.yscale("log")
+plt.xlabel("Iteration")
+plt.tight_layout()
+plt.show()
```
matthieutrs, line 191: Using `plt.semilogy` can make you gain 1 line

```suggestion
plt.figure()
plt.semilogy(range(n_iter), losses)
plt.title("Loss evolution")
plt.xlabel("Iteration")
plt.tight_layout()
plt.show()
```

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,258 @@
...
+    optimizer.zero_grad()
+    y_hat = physics.A(x)
+    loss = loss_fn(y_hat, y)
+    loss.backward()
+
+    # a gradient step
+    optimizer.step()
+
+    # projection step, when doing additional steps, it's important to change only
+    # the tensor data to avoid breaking the gradient computation
```
matthieutrs, line 233:
```suggestion
    # projection step. 
    # Note: when doing additional steps, it's important to change only
    # the tensor data to avoid breaking the gradient computation
```

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,258 @@
...
+)
+
+# %%
+#
+# We can plot the loss to make sure that it decreases
+#
+plt.figure()
+plt.plot(range(n_iter), losses)
+plt.title("Loss evolution")
+plt.yscale("log")
```
matthieutrs, line 253: same as above, you can use plt.semilogy

matthieutrs (changes requested):
Thanks @mh-nguyen712 for the great PR! See my comments below.

IMO one thing that is still missing in the PR is a test on `ComposedLinearPhysics`.

Is it possible to add a toy test checking that the behaviour of the class is correct?

For instance with 2 mask physics, checking that the kernel of the ComposedPhysics is equivalent to a product of the masks, and check the adjoint too. Similar, I would do a test on a product of Inpainting and Blur. Would that be possible?

deepinv/physics/forward.py
```diff
@@ -616,6 +603,112 @@ def A_dagger(
...
+    def __init__(self, physics_list: list[Physics], **kwargs):
+        super().__init__(physics_list, **kwargs)
+
+    def A_adjoint(self, y: Tensor, **kwargs) -> Tensor:
+        r"""
+        Computes adjoint of composed operator
+
+        .. math::
+
+            x = A_1^{\top}(A_2^{\top}(\dots(N_k^{\top}(y))))
```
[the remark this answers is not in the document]
mh-nguyen712: Currently only the noise of the last physics is used. I'm not if it's the good behavior or we should take all noise model into account.

deepinv/physics/unmixing.py
```diff
@@ -66,8 +63,8 @@ def __init__(
 
         self.E, self.C = M.shape
 
-        self.update_parameters(M=M, **kwargs)
-        self.M_pinv = torch.linalg.pinv(self.M)
+        self.register_buffer("M", M.to(device))
+        self.register_buffer("M_pinv", torch.linalg.pinv(self.M.to(device)))
```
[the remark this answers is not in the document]
mh-nguyen712: good remark, thanks Mathieu!

mh-nguyen712:
Thanks @matthieutrs  for your review! For the test, currently there are already 2 tests of the composition (`composition1` and `composition2` in the `find_operator` of test_physics, between Downsampling and BlurFFT). I can add a more explitit test for mask operator like Inpainting

deepinv/physics/forward.py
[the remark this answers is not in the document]
tachella: I think this is a reasonable behaviour for most practical applications

deepinv/physics/forward.py
```diff
@@ -616,6 +603,112 @@ def A_dagger(
...
+
+class ComposedLinearPhysics(ComposedPhysics, LinearPhysics):
+    r"""
+    Composes multiple linear physics operators into a single operator.
+
+    The measurements produced by the resulting model are defined as
+
+    .. math::
+
+        Ax = N_k(A_k\dotsA_1(x))
```
tachella, line 686:
```suggestion
        Ax = N_k(A_k\dotsA_2(A_1(x)))
```

deepinv/physics/forward.py
```diff
@@ -63,13 +64,8 @@ def __mul__(self, other):  #  physics3 = physics1 \circ physics2
...
 
         """
-        A = lambda x: self.A(other.A(x))  # (A' = A_1 A_2)
-        noise = self.noise_model
-        sensor = self.sensor_model
-        return Physics(
-            A=A,
-            noise_model=noise,
-            sensor_model=sensor,
+        return ComposedPhysics(
```
tachella, line 67: I think we need to add the `compose` function below to handle combinations of linear + linear = linear, non_linear + linear = non_linear, and non_linear + non_linear = non_linear
```suggestion
        return compose([other, self])
```
with 
```
def compose(*physics: Union[Physics, LinearPhysics]):
    r"""
    Composes multiple forward operators :math:`A = A_1\circ A_2\circ \dots \circ A_n`.

    The measurements produced by the resulting model are :class:`deepinv.utils.TensorList` objects, where
    each entry corresponds to the measurements of the corresponding operator.

    :param deepinv.physics.Physics physics: Physics operators :math:`A_i` to be composed.
    """
    if all(isinstance(phys, LinearPhysics) for phys in physics):
        return ComposedLinearPhysics(physics)
    else:
        return ComposedPhysics(physics)
```

deepinv/physics/forward.py
```diff
@@ -409,25 +405,16 @@ def A_adjoint_A(self, x, **kwargs):
...
-            self.A_adjoint(x, **kwargs), **kwargs
-        )
-        noise = self.noise_model
-        sensor = self.sensor_model
-        return LinearPhysics(
-            A=A,
-            A_adjoint=A_adjoint,
-            noise_model=noise,
-            sensor_model=sensor,
+        return ComposedLinearPhysics(
```
tachella, line 416: same here
```suggestion
        return compose([other, self])
```

CHANGELOG.rst
```diff
@@ -63,6 +63,11 @@ New Features
 - Trainer test can return unaggregated metrics (:gh:`420` by `Andrew Wang`_)
 - MoDL model (:gh:`435` by `Andrew Wang`_)
 - Add conversion to Hounsfield Units (HUs) for LIDC IDRI (:gh:`459` by `Jérémy Scanvic`_)
+- Add ComposedLinearPhysics (via __mul__ method) (:gh:`462` by `Minh Hai Nguyen`_)
```
tachella, line 66:
```suggestion
- Add ComposedLinearPhysics (via __mul__ method) (:gh:`462` by `Minh Hai Nguyen`_ and `Julian Tachella`_ )
```

tachella (changes requested):
I think we need to add the auxiliary `compose` function to handle combination of linear and non-linear operators (see comments below)

deepinv/physics/forward.py
mh-nguyen712: Good point. Many thanks Julian!

deepinv/physics/forward.py
```diff
@@ -616,6 +595,127 @@ def A_dagger(
...
 
+class ComposedPhysics(Physics):
+    r"""
+    Composes multiple physics operators into a single operator.
+
+    The measurements produced by the resulting model are defined as
+
+    .. math::
+
+        A(x) = N_k(A_k(\dots(A_1(x))))
```
tachella, line 606:
```suggestion
        A(x) = N_k(A_k(\dots A_2(A_1(x))))
```

I would add this to make clear that only the last noise is applied

tachella (approved):
lgtm, many thanks Hai

Andrewwango:
Todo
- [x] add a test for the minimum reproducible example of the composed physics error in #489

deepinv/utils/tensorlist.py
```diff
@@ -235,6 +235,10 @@ def squeeze(self, dim=None):
     def unsqueeze(self, dim=None):
         return TensorList([xi.unsqueeze(dim=dim) for xi in self.x])
 
+    @property
+    def device(self):
+        return self.x[0].device
```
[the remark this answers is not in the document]
mh-nguyen712: They should be in the same device (as implemented in `.to()`, `.cuda()`, ...) and I think a property should not modify the object so we can raise a warning if the tensors are not in the same device

deepinv/tests/test_physics.py
```diff
@@ -1279,5 +1316,257 @@ def test_unmixing(device):
...
+
+        # Check two physics have the same output
+        x = torch.randn(imsize, device=device, dtype=dtype).unsqueeze(0)
+        assert torch.allclose(physics(x), new_physics(x))
+
+        # Remove the cache dir
+        shutil.rmtree(cache_dir)
+
+
+def test_composed_linear_physics(device):
```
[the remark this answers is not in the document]
mh-nguyen712: I added a test for checking the expected behavior. We can discuss further in the issue [#489](https://github.com/deepinv/deepinv/issues/489)

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,255 @@
+r"""
+Optimizing your operator with Deepinv Physics
```
Andrewwango: You could promote this more clearly in the title or summary, e.g. solving blind inverse problems with DeepInverse / estimating physics parameters

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,255 @@
...
+========================================================
+
+This demo shows you how to use
+:class:`deepinv.physics.Physics` together with automatic differentiation to optimize your operator.
+
+Consider the forward model
+
+.. math::
+    y = N(A(x, \theta))
```
Andrewwango:
```suggestion
    y = \noise{\forw{x, \theta}}
```

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,255 @@
...
+
+where :math:`N` is the noise model, :math:`A(\cdot, \theta)` is the forward operator, with the parameter :math:`\theta` (e.g., the filter in :class:`deepinv.physics.Blur`).
```
Andrewwango:
```suggestion
where :math:`N` is the noise model, :math:`\forw{\cdot, \theta}` is the forward operator, and the goal is to learn the parameter :math:`\theta` (e.g., the filter in :class:`deepinv.physics.Blur`).
```

examples/advanced/demo_optimizing_physics_parameter.py
Andrewwango: i.e. "optimizing physics parameter" might suggest you're making the parameter better, whereas you're just trying to learn the true parameter

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,255 @@
...
+
+In a typical blind inverse problem, given a measurement :math:`y`, we would like to recover both the underlying image :math:`x` and the operator parameter :math:`\theta`,
+resulting in a highly ill-posed inverse problem.
+
```
Andrewwango, line 17: before going into the specific details of the experiment, explain how deepinverse facilitates solving these types of problems. i.e. we provide end-to-end differentiable physics where physics params can easily be added as  variables

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,255 @@
...
+In this example, we only focus on a much more simpler problem: given the measurement :math:`y` and the ground truth :math:`x`, find the parameter :math:`\theta`.
+This can be reformulated as the following optimization problem:
+
+.. math::
+    \min_{\theta} \frac{1}{2} \|A(x, \theta) - y \|^2
```
Andrewwango:
```suggestion
    \min_{\theta} \frac{1}{2} \|\forw{x, \theta} - y \|^2
```

examples/advanced/demo_optimizing_physics_parameter.py
Andrewwango: that is the purpose of this demo! not showing a basic first order optimisaation

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,255 @@
...
+
+# %%
+# Import required packages
+#
+import deepinv as dinv
+import torch
+from tqdm import tqdm
+import matplotlib.pyplot as plt
+
+device = "cuda" if torch.cuda.is_available() else "cpu"
```
Andrewwango: get freer gpu?

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,255 @@
...
+
+dinv.utils.plot([x, y, true_kernel], titles=["Sharp", "Blurry", "True kernel"])
+
+# %%
+# Define an optimization algorithm
+# --------------------------------
+#
+# The convolution kernel lives in the simplex, ie the kernel must have positive entries summing to 1.
+# We can use a simple optimization algorithm - Projected Gradient Descent - to enforce this constraint.
+# The following function allows one to compute the orthogonal projection onto the simplex, by a sorting algorithm (Reference: Large-scale Multiclass Support Vector Machine Training via Euclidean Projection onto the Simplex
```
Andrewwango: URL

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,255 @@
...
+plt.yscale("log")
+plt.xlabel("Iteration")
+plt.tight_layout()
+plt.show()
+
+# %%
+# Combine with arbitrary optimizer
+# --------------------------------
+#
+# Pytorch provides a wide range of optimizer for training neural networks.
```
Andrewwango:
```suggestion
# Pytorch provides a wide range of optimizers for training neural networks.
```

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,255 @@
...
+In a typical blind inverse problem, given a measurement :math:`y`, we would like to recover both the underlying image :math:`x` and the operator parameter :math:`\theta`,
+resulting in a highly ill-posed inverse problem.
+
+In this example, we only focus on a much more simpler problem: given the measurement :math:`y` and the ground truth :math:`x`, find the parameter :math:`\theta`.
+This can be reformulated as the following optimization problem:
+
+.. math::
+    \min_{\theta} \frac{1}{2} \|A(x, \theta) - y \|^2
+
+This problem can be addressed by first-order optimization if we can compute the gradient of the above function with respect to :math:`\theta`.
```
Andrewwango, line 24: I would add somewhere here: we show 3 example methods on how to use differentiable parameters to estimate parameters: manual optimisation, torch optimisation...

deepinv/tests/test_physics.py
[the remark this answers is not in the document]
Andrewwango: ok

deepinv/tests/test_physics.py
```diff
@@ -1279,5 +1316,277 @@ def test_unmixing(device):
...
```
Andrewwango, line 1541: is there a test for composed physics differentiability?

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,255 @@
...
+#
+# The convolution kernel lives in the simplex, ie the kernel must have positive entries summing to 1.
+# We can use a simple optimization algorithm - Projected Gradient Descent - to enforce this constraint.
+# The following function allows one to compute the orthogonal projection onto the simplex, by a sorting algorithm (Reference: Large-scale Multiclass Support Vector Machine Training via Euclidean Projection onto the Simplex
+# Mathieu Blondel, Akinori Fujino, and Naonori Ueda)
+#
+
+
+@torch.no_grad()
+def projection_simplex_sort(v):
```
Andrewwango: even though it's a helper function, I still think having type annotations and docstrings are useful

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,255 @@
...
+plt.tight_layout()
+plt.show()
+
+
+# %%
+#
+# Optimizing the physics as a usual neural network
+# ------------------------------------------------
+#
+# Below we show another way to optimize the parameter of the physics, as we usually do for neural networks
```
Andrewwango, line 206: Would you say this is the most convenient way for most users? Even though it's not the most elucidating (i.e. doesn't show how the gradient step works) compared to the previous examples, I think a novice user would want to see this first as this is how they might be using it for deep learning

examples/advanced/demo_optimizing_physics_parameter.py
```diff
@@ -0,0 +1,255 @@
...
+    filter=kernel_init.clone().requires_grad_(True), device=device
+)
+
+# Set up the optimizer by giving the parameter to an optimizer
+# Try to change your favorite optimizer
+optimizer = torch.optim.AdamW([physics.filter], lr=0.1)
+
+
+# Try to change another loss function
+# loss_fn = torch.nn.MSELoss()
```
Andrewwango, line 223: why commented?

docs/source/user_guide/physics/physics.rst
```diff
@@ -11,6 +11,7 @@ Operators
 Operators describe the forward model :math:`z = A(x,\theta)`, where
 :math:`x` is the input image and :math:`\theta` are the parameters of the operator.
 The parameters :math:`\theta` can be sampled using random generators, which are available for some specific classes.
+Using automatic differentiation, we can compute derivatives w.r.t to both the input :math:`x` or the parameters :math:`\theta`.
```
Andrewwango, line 14: clarify why this might be useful for users: e.g. for blind inverse problems, parameter estimation problems...

docs/source/user_guide/physics/intro.rst
```diff
@@ -94,6 +94,25 @@ imaging system optimization, etc. The following example shows how operators and
...
+
+.. doctest::
+    
+	>>> import torch
+	>>> from deepinv.physics import Blur
+	>>> x = torch.rand((1, 1, 16, 16))
+	>>> theta = torch.ones((1, 1, 2, 2)) / 4 # a basic 2x2 averaging filter
+	>>> physics = Blur(filter=theta, padding='circular') # we instantiate a blur operator with its convolution filter
+	>>> y = physics(x)
+	>>> theta_2 = torch.ones((1, 1, 3, 3)) / 9
```
Andrewwango: explain why and how this is different from theta

deepinv/tests/test_physics.py
```diff
@@ -676,9 +716,7 @@ def test_MRI(mri, mri_img_size, device, rng):
         y1 = physics(x)
         if isinstance(physics, MultiCoilMRI):
             y1 = y1[:, :, 0]  # check 0th coil
-        # check shapes
-        print(y1.shape, mask.shape, physics.mask.shape)
-        assert torch.all((y1 == 0) == (physics.mask == 0))
+        assert torch.all((y1 == 0) == (mask == 0))
```
Andrewwango: why this change? the original tests a) physics mask gets updated correctly in init and b) it masks the data correctly.

deepinv/tests/test_physics.py
```diff
@@ -1041,7 +1077,8 @@ def test_tomography(device):
                     )
                     y = physics.A(r)
                     error = (physics.A_dagger(y) - r).flatten().mean().abs()
-                    assert error < 0.2
+                    epsilon = 0.2 if device == "cpu" else 0.3  # Relax a bit of GPU
```
Andrewwango, line 1076:
```suggestion
                    epsilon = 0.2 if device == "cpu" else 0.3  # Relax a bit if GPU
```

docs/source/user_guide/physics/physics.rst
Andrewwango: shouldn't there also be a doc somewhere advertising that physics (physics, linear physics) are composable (although not decomposable physics), which is very useful? And they're also useful for not just composing with other physics, but the general case where you want to add a function to A and A_adjoint

deepinv/physics/forward.py
```diff
@@ -63,16 +64,7 @@ def __mul__(self, other):  #  physics3 = physics1 \circ physics2
...
-        noise = self.noise_model
-        sensor = self.sensor_model
-        return Physics(
-            A=A,
-            noise_model=noise,
-            sensor_model=sensor,
-            max_iter=self.max_iter,
-            tol=self.tol,
-        )
+        return compose(other, self, max_iter=self.max_iter, tol=self.tol)
```
Andrewwango, line 74: `if isinstance(other, DecomposablePhysics)` then raise a warning that the resultant physics won't return the decomposability (i.e. the closed form dagger and closed form prox)?

if you agree with this, then I think this should also be caveated in the docs too because it could lead to hard to diagnose problems as reported in #489

Andrewwango, line 74: I think we should also add a warning here or in the docs that whenever you compose, you lose the physics attributes. Let's say a user wants to add a premultiplier to their inpainting physics so they do `physics_new = physics * LinearPhysics(A=lambda x: x*3)`. I think it would be reasonable that the user would expect `torch.all(physics_new.mask == physics.mask) == True` . Feel free to disagree!

Andrewwango: See review summary!

Andrewwango: See review summary!

deepinv/tests/test_physics.py
```diff
@@ -1279,5 +1316,277 @@ def test_unmixing(device):
...
+            physics, imsize, dtype = finder(name)
+            break
+        except Exception:
+            continue
+    else:
+        raise ValueError(f"Could not find an operator for {name}")
+
+    # The current radio physics depends on torchkbnufft, which seems to be not compatible.
+    if "radio" in name:
+        return 1
```
Andrewwango: pytest skip

deepinv/physics/forward.py
```diff
@@ -616,6 +595,127 @@ def A_dagger(
...
+    .. math::
+
+        N(A(x)) = N_k(A_k(\dots A_2(A_1(x))))
+
+    where :math:`A_i` is the ith physics operator and :math:`N_k` is the noise of the last operator.
+
+    :param list[deepinv.physics.Physics] physics_list: list of physics operators to compose.
+    """
+
+    def __init__(self, physics_list: List[Physics], device=None, **kwargs):
```
Andrewwango: It could be nice to do `__init__(self, *physics: Physics, device=None)` but I don't mind too much

deepinv/physics/forward.py
```diff
@@ -616,6 +595,127 @@ def A_dagger(
...
+
+class ComposedLinearPhysics(ComposedPhysics, LinearPhysics):
+    r"""
+    Composes multiple linear physics operators into a single operator.
+
+    The measurements produced by the resulting model are defined as
+
+    .. math::
+
+        N(A(x)) = N_k(A_k \dots A_2(A_1(x)))
```
Andrewwango: \forw and \noise

deepinv/physics/forward.py
```diff
@@ -616,6 +595,127 @@ def A_dagger(
...
 
+class ComposedPhysics(Physics):
+    r"""
+    Composes multiple physics operators into a single operator.
+
+    The measurements produced by the resulting model are defined as
+
+    .. math::
+
+        N(A(x)) = N_k(A_k(\dots A_2(A_1(x))))
```
Andrewwango: \forw and \noise

deepinv/physics/forward.py
```diff
@@ -616,6 +595,127 @@ def A_dagger(
...
+
+        :param torch.Tensor y: measurements
+        :return: signal/image
+        """
+        for physics in reversed(self.physics_list):
+            y = physics.A_adjoint(y, **kwargs)
+        return y
+
+
+def compose(*physics: Union[Physics, LinearPhysics], **kwargs):
```
Andrewwango, line 711: why can't this be a (static) method of Physics that LinearPhysics inherits?

deepinv/physics/forward.py
```diff
@@ -1061,12 +1161,3 @@ def A_adjoint(self, y: TensorList, **kwargs) -> torch.Tensor:
                 for i, physics in enumerate(self.physics_list)
             ]
         )
-
-    def update_parameters(self, **kwargs):
```
Andrewwango, line 1086: why is this removed?

deepinv/physics/blur.py
```diff
@@ -65,14 +64,15 @@ def __init__(
...
-        self.factor = factor
         assert isinstance(factor, int), "downsampling factor should be an integer"
-        # assert len(img_size) == 3, "img_size should be a tuple of length 3, C x H x W"
+
         self.imsize = img_size
         self.padding = padding
         self.device = device
-        self.filter = filter
+
+        self.register_buffer("filter", None)
```
Andrewwango, line 73: Sorry but why do we not also register factor?

deepinv/physics/blur.py
```diff
@@ -167,41 +166,34 @@ def update_parameters(self, filter=None, factor=None, **kwargs):
         :param int factor: New downsampling factor to be applied to the input image.
         """
         if factor is not None:
+            assert isinstance(factor, int), "downsampling factor should be an integer"
```
Andrewwango: raise ValueError instead of assert

deepinv/physics/blur.py
```diff
@@ -167,41 +166,34 @@ def update_parameters(self, filter=None, factor=None, **kwargs):
...
+                filter_fft_2d(self.filter, self.imsize, real_fft=False).to(self.device),
             )
-            self.Fhc = torch.nn.Parameter(torch.conj(self.Fh), requires_grad=False)
-            self.Fh2 = torch.nn.Parameter(self.Fhc * self.Fh, requires_grad=False)
-            self.Fhc = torch.nn.Parameter(self.Fhc, requires_grad=False)
-            self.Fh2 = torch.nn.Parameter(self.Fh2, requires_grad=False)
+            self.register_buffer("Fhc", torch.conj(self.Fh))
+            self.register_buffer("Fh2", self.Fhc * self.Fh)
+
+        super().update_parameters(**kwargs)
```
Andrewwango, line 200: nice

deepinv/physics/cassi.py
```diff
@@ -93,11 +93,12 @@ def __init__(
...
             ).step()["mask"]
 
-        self.update_parameters(mask=mask, **kwargs)
-
         # In SS-CASSI, masking happens on the padded image after shearing
         if self.mode == "ss":
-            self.mask = self.pad(self.mask)
+            mask = self.pad(mask)
+
+        self.register_buffer("mask", mask)
```
Andrewwango, line 100: I'm not sure I understand. Why does this init not need update_parameters anymore?

docs/source/user_guide/physics/physics.rst
Andrewwango: we should add the composing physics to this doc here https://deepinv.github.io/deepinv/user_guide/physics/intro.html#physics-combining

deepinv/physics/inpainting.py
```diff
@@ -117,7 +117,8 @@ def __init__(
             mask = mask.unsqueeze(0)
 
         self.tensor_size = tensor_size
-        self.update_parameters(mask=mask)
```
Andrewwango, line 122: So after your changes, is `update_parameters` used anywhere anymore? Has it become redundant? Why are we changing these to `register_buffer` when `update_parameters` just calls `register_buffer` anyway? Sorry, I think I'm missing something!

Andrewwango: Since `register_buffer` might be new to some developers, maybe it's worth adding in the base physics docstring OR the "Defining new operator" docs, that new physics should use `register_buffer`, and users should use `physics.update` to update parameters (is this true, e.g. when we use `update` in the trainer?)

deepinv/tests/test_physics.py
```diff
@@ -152,6 +166,7 @@ def find_operator(name, device):
             n_coils
         )  # B,N,H,W where N is coil dimension
         p = MultiCoilMRI(coil_maps=maps, img_size=img_size, device=device)
+        params = ["mask"]
```
Andrewwango: should we also test the `coil_maps` param?

deepinv/physics/noise.py
```diff
@@ -90,6 +95,81 @@ def randn_like(self, input: torch.Tensor, seed: int = None):
...
+        if kwargs:
+            for key, value in kwargs.items():
+                if (
+                    value is not None
+                    and hasattr(self, key)
+                    and isinstance(value, (torch.Tensor, float))
+                ):
+                    self.register_buffer(key, self._float_to_tensor(value))
+
+    def _float_to_tensor(self, value):
```
Andrewwango, line 113: docstrings and typing for all these methods please

deepinv/tests/test_physics.py
```diff
@@ -1279,5 +1316,277 @@ def test_unmixing(device):
...
+        if torch.cuda.is_available():
+            torch.manual_seed(11)
+            cuda = torch.device("cuda:0")
+            physics = physics.to(cuda)
+            x = x.to(cuda)
+            y2 = physics.A(x)
+            assert y2.device == cuda
+
+            # skip denoising that adds random noise in each forward call
+            if not isinstance(physics, dinv.physics.Denoising):
```
Andrewwango, line 1463: Does it not even work with denoising with a rng set in the noise model?

deepinv/tests/test_physics.py
```diff
@@ -1279,5 +1316,277 @@ def test_unmixing(device):
...
+                if isinstance(y2, TensorList):
+                    for y11, y22 in zip(y1, y2):
+                        assert torch.linalg.norm((y11.to(cuda) - y22).ravel()) < 1e-3
```
Andrewwango: 1e-3 feels quite generous - are most physics approaching this bound?

examples/advanced/demo_optimizing_physics_parameter.py
tachella: I would avoid using expressions like 'with deepinverse' in the examples. In my mind, it is obvious that all examples are running with deepinverse

deepinv/physics/inpainting.py
tachella: I agree with Andrew, seems an important point to show in the demo how to use it

tachella:
> 2. After you compose a DecomposablePhysics with either something simple or another DecomposablePhysics, you lose the benefits of the decomposable physics. We can discuss this point further in [Physics mul transform might be wrong? #489](https://github.com/deepinv/deepinv/issues/489)
> 

Just adding my response over here too: 

I could imagine some cases where it is possible to easily obtain an SVD decomposition when composing with a transform (eg `Inpainting `+ circular `Shift`), but these should be handled on a per-case basis (as far as I know there is no automatic way of doing this), and this is probably out of the scope of this PR.

deepinv/physics/blur.py
mh-nguyen712: It's only an integer not a tensor so we don't need to register.

deepinv/physics/cassi.py
mh-nguyen712: `mask` is the only parameter we have here so calling `super().update_parameters()` will do nothing, unless there is something in **kwargs. I just translated what have been done before to the buffers.

deepinv/physics/forward.py
mh-nguyen712: Good point!

deepinv/physics/forward.py
mh-nguyen712: I want to make it simple in this PR. Maybe we could make a PR dedicated to the composition of physics, as it seems to be a complicated problem.

deepinv/physics/noise.py
mh-nguyen712: These are private methods so I think the docstring is not necessary (PEP8). The method name itself already explained what it does I think. Correct me if I'm wrong.

deepinv/tests/test_physics.py
mh-nguyen712: It's the same code. You would like to suggest something different?

deepinv/tests/test_physics.py
mh-nguyen712: Sorry I don't remember why, this could be my mistake. I reverted to the original test.

examples/advanced/demo_optimizing_physics_parameter.py
mh-nguyen712: Just want to illustrate that we can use different losses, I dont want to run all the examples. I can remove it if it makes more sense

examples/advanced/demo_optimizing_physics_parameter.py
mh-nguyen712: I agree that this is very convenient but I think I prefer starting by the explicit gradient steps. The example is short enough so I think users can go to this part very rapidly.

deepinv/physics/forward.py
mh-nguyen712: It will be automatically inherited from StackedPhysics, no need to duplicate the implementation

deepinv/physics/inpainting.py
mh-nguyen712: Sorry for this confusion. The `update_parameter` can still be used anywhere (except in this __init__, it should be only used at the end of __init__). The problem is that, in `update_parameter`, we only update tensor attributes of the class, so in the __init__ we need to define which are the tensor attributes (which should be registered as buffers). 
Then in update_parameters, it will call `register_buffer` to all tensor attributes. In summary, the only change is instead of doing `self.attribute = tensor`, it becomes `self.register_buffer(attribute, tensor)`, the logic of `update` and `update_parameters` keep unchanged.
I'll clarify this in the docs

deepinv/tests/test_physics.py
mh-nguyen712: Yes because pytorch generate different values on CPU and GPU, even the same seed is used.

deepinv/tests/test_physics.py
mh-nguyen712: I changed to 1e-5, it passed on my local. Let's see what we have on the github CI

deepinv/tests/test_physics.py
mh-nguyen712: They are tested in the `test_operators_differentiability` for all physics in `OPERATORS`, which includes 2 compositions: `composition` and `composition2`

docs/source/user_guide/physics/physics.rst
mh-nguyen712: That's a good point! But could we leave it for another PR dedicated to the ComposedPhysics? At this stage, the behavior of ComposedPhysics is not clear. The current implementation is minimalist, avoiding possible bugs from using lambda. I think we should discuss and work more on that, which is out of the scope of this PR.

mh-nguyen712:
Thanks @Andrewwango  for your detailed reviews. I addressed most of your comments. The main remaining issue is related to the behavior of the ComposedPhysics, which is not clear how we should handle it and it's out of scope of this PR. I think we could discuss further and make a dedicated PR for this.

docs/source/user_guide/physics/physics.rst
Andrewwango: @tachella what do you think? I think if we implement a functionality in DeepInverse it should be well documented on why and how it works

deepinv/physics/forward.py
Andrewwango: Ok, this isn't an important comment anyway

docs/source/user_guide/physics/defining.rst
```diff
@@ -18,6 +18,10 @@ a :class:`deepinv.physics.Physics.A` method that computes the forward operator.
     you can define the adjoint automatically using autograd with :class:`deepinv.physics.adjoint_function`.
     Note however that coding a closed form adjoint is generally more efficient.
 
+.. tip::
+
+    To make the new physics compatible with other torch functionalities, physics parameters (tensor attributes) should be registered as module buffers by using `self.register_buffer(param_name, param_tensor)`. This ensures methods like `.to(), .cuda()` work properly, allowing one to train a model using DDP.
```
Andrewwango:
```suggestion
    To make the new physics compatible with other torch functionalities, all physics parameters (i.e. attributes of type :class:`torch.Tensor`) should be registered as `module buffers <https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_buffer>` by using `self.register_buffer(param_name, param_tensor)`. This ensures methods like `.to(), .cuda()` work properly, allowing one to train a model using Distributed Data Parallel. 
```

deepinv/physics/inpainting.py
Andrewwango: Thanks!

deepinv/physics/noise.py
Andrewwango: It would be still good to have at least a [comment inside each function describing what they do](https://peps.python.org/pep-0008/#:~:text=Docstrings%20are%20not%20necessary%20for%20non%2Dpublic%20methods%2C%20but%20you%20should%20have%20a%20comment%20that%20describes%20what%20the%20method%20does.)

Andrewwango (changes requested):
Thanks for addressing the comments! I have left one small comment here + left one comment from the previous review unresolved. 

- [ ] Please can you link clearly which existing issues #430 #414 #412 #387 #383 #460 will be closed by this PR?

Once these have been addressed it should be good to go!

I've added all longer concerns about composablephysics to a new issue #533.

docs/source/user_guide/physics/physics.rst
Andrewwango: I have moved over the discussion to a new issue #533 for future work.

deepinv/physics/forward.py
Andrewwango: I have moved over the issue to a new issue #533 for future work.

Andrewwango: I have moved over the issue to a new issue #533 for future work.

mh-nguyen712:
This will close #383 #412 #430 #438 #460 #387

Andrewwango (approved):
Thank you!

PR #463, opened 2025-04-27
FIX doctest run in documentation workflow

tomMoral:
- Add a `IGNORE_OUTPUT` primitive to skip check in doctest when the output depends on the context
- Move doctest check to only `rst` files, using generated stubs for each classes.
- Add cache to `pip` dependencies
- Add concurrency to cancel some runs when pushing to the same PR
- Pin torch to version before `2.7` as it seems the latest release broke something for the homography.


tomMoral:
:tada:

matthieutrs (approved):

PR #464, opened 2025-04-29
DOC add doc artifact GH action

mathurinm:
Create yaml file for integration in circleci. This will allow to view docs ran in PR pushes.


matthieutrs (approved):
Thanks for this, super useful!

PR #468, opened 2025-04-30
Clean up file system pollution in get_freer_gpu

### LINKED ISSUE #465: File system pollution in get_freer_gpu
Currently, the function `deepinv.utils.get_freer_gpu` uses a shell pipeline to extract the memory usage of available GPUs. The result of the pipeline is written to a file named "tmp" in the current working directory, which is then read using Python's  `open` function. This file is not subsequently removed and thus pollutes the file system. There is also no fail-safe preventing unwarranted file overwriting, which might result in data losses.

**Current implementation**

https://github.com/deepinv/deepinv/blob/a9f113a7a8c3054129ff681a559a6950c9aba268/deepinv/utils/nn.py#L18-L23

**Expected behavior**

The function should not write anything to the file system.

**Fix**

Instead of redirecting the standard output to a file, then reading it, it should not be redirected in the first place and the standard output should be read directly, avoiding any write to the file system. The typical way to do so in Python would be to use `subprocess.run` instead of `os.system` as it allows to run a command (possibly in a shell), to wait for its completion, and to get its standard output as a string directly.

jscanvic:
Fixes #465

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,7 @@ Changed
 Fixed
 ^^^^^
 - Fix latex rendering in `deepinv.utils.plotting.config_matplotlib`  (:gh:`452` by `Romain Vo`_)
+- Get rid of unnecessary file system writes in `get_freer_gpu` (:gh:`468` by `Jérémy Scanvic`_)
 
 
 v0.3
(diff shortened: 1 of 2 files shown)


mh-nguyen712:
Tested it and it looks good for me! I suggest that you can add a test function (that something I think that we need to improve in each PR to make the library more stable, and pass the code coverage check).   
A possible way is to use a `unittest.mock.patch` to mock to output of `subprocess.run` to simulate different situations.

tachella:
lgtm, thanks @jscanvic!

PR #474, opened 2025-05-03
Tomography operator with astra-toolbox

romainvo:
I have implemented a `TomographyWithAstra` operator using the https://astra-toolbox.com/.  It uses the GPULink functionality of astra to share the underlying CUDA memory of torch.Tensor and arrays used by `astra`. By wrapping astra, the operator supports 2d geometries (parallel and fanbeam) as well as 3d (parallel and conebeam). 

Astra can be installed easily with `conda install -c astra-toolbox -c nvidia astra-toolbox`.

The contribution consists in two main classes:
- a `deepinv.physics.functional.astra.XrayTransform` object, which wraps the astra-computation, both forward and adjoint computation, with torch.Tensor as inputs.
- a `deepinv.physics.TomographyWithAstra` LinearPhysics operator, which instanciates an XrayTransform as attribute (similar to how the Radon module is used). The calls to XrayTransform are wrapped in a custom `torch.autograd.Function` (AutogradTransform in astra.py) to make the computation of `A()` and `A_adjoint()` differentiable with torch.

The operator is more memory efficient, though slower for batched computations in parallel 2d. This is because ASTRA does not handle batched computations, therefore the custom autograd.Function processes batch elements sequentially. It is a priori, not easily _vmappable_ (https://pytorch.org/docs/stable/notes/extending.func.html) because the XrayTransform needs to access pointers of batch elements, which is not possible for Tensors produced during vmap.

I have also adapted the following examples, to check for an `astra-toolbox` install and adapt the CT operator used automatically.
- `examples/plug-and-play/demo_vanilla_PnP.py`
- `examples/patch-priors/demo_patch_priors_CT.py`

I haven't touched the learned primal-dual example which runs significantly slower while using the `astra` operator.

Other changes:
- I have simplified and adapted the `AbstractFiler` module in `deepinv.physics.functional.radon.py` to handle sinogram data for both the `Tomography` operator and the `TomographyWithAstra` operator.

Test:
- I have implemented an independant test for the TomographyWithAstra operator, which is definitely not going to pass the adjointness and pseudoinverse tests of linear OPERATORS. Not sure if this what is intended ? They seem to pass, and skipped by `pytest` if `astra` is not installed.


deepinv/tests/test_physics.py
```diff
@@ -1009,6 +1019,86 @@ def test_tomography(device):
...
+            At_y_x = torch.sum(At_y * x).item()
+
+            relative_error = abs(Ax_y - At_y_x) / At_y_x
+            assert relative_error < 0.01  # at least 99% adjoint
+
+            ## --- Test pseudoinverse ---
+            r = physics.A_adjoint(physics.A(x))
+            y = physics.A(r)
+            error = torch.linalg.norm(physics.A_dagger(y) - r) / torch.linalg.norm(r)
+            assert error < 0.2
```
matthieutrs, line 1064: This seems like a quite high threshold given that you (correctly) test on the range of A... Do you think it's possible to reduce this?

examples/patch-priors/demo_patch_priors_CT.py
```diff
@@ -45,6 +45,20 @@
...
+
+try:
+    import astra
+
+    USE_ASTRA = True
+    from deepinv.physics import TomographyWithAstra
+
+    print(
+        "Detected `astra-toolbox` package, the example will use the TomographyWithAstra operator"
+    )
```
matthieutrs, line 57: I am not sure I would put this here. My argument would be that some people would copy-paste this, think "uh-oh I need astra" eventhough they just wanted to run the toy tomography example.
Maybe instead I would add a new section in the examples "Working with external libraries" and try to put hands on experiments with quite technical libraries.
What do you think?

examples/plug-and-play/demo_vanilla_PnP.py
```diff
@@ -16,6 +16,17 @@
 from deepinv.utils.demo import load_url_image, get_image_url
 from deepinv.utils.plotting import plot, plot_curves
 
+try:
+    import astra
+
+    USE_ASTRA = True
+    print(
+        "Detected `astra-toolbox` package, the example will use the TomographyWithAstra operator"
+    )
```
matthieutrs, line 25: Same comment as in the other example

examples/patch-priors/demo_patch_priors_CT.py
Andrewwango: I agree, and it would be a great example to have to show DeepInverse's integratibility with other libraries, setting the precedent for future integrations (e.g. sigpy, mri-nufft...)

deepinv/tests/test_physics.py
```diff
@@ -162,6 +162,16 @@ def find_operator(name, device):
         p = dinv.physics.Tomography(
             img_width=img_size[-1], angles=img_size[-1], device=device
         )
+    elif name == "TomographyWithAstra":
```
Andrewwango, line 165: Perhaps if adjointness etc. tests don't apply to this, it might be misleading to include it in these tests (only to be skipped anyway)

deepinv/tests/test_physics.py
```diff
@@ -1009,6 +1019,86 @@ def test_tomography(device):
...
+    """
+
+    pytest.importorskip(
+        "astra",
+        reason="This test requires astra-toolbox. It should be "
+        "installed with `conda install -c astra-toolbox -c nvidia astra-toolbox`",
+    )
+
+    ## Test 2d transforms
+    for geometry_type in ["parallel", "fanbeam"]:
```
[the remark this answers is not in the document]
Andrewwango: and/or for normalize

deepinv/physics/functional/astra.py
```diff
@@ -0,0 +1,545 @@
...
+        To handle standard PyTorch pipelines, :class:`XrayTransform` is instanciated inside a :class:`deepinv.physics.TomographyWithAstra` operator.
+
+    :param dict[str, Any] projection_geometry: Dictionnary containing the parameters of the projection geometry. It is passed to the ``astra.create_projector()`` function to instanciate the projector.
+    :param dict[str, Any] object_geometry:  Dictionnary containing the parameters of the object geometry. It is passed to the ``astra.create_projector()`` function to instanciate the projector.
+    :param bool is_2d: Specifies if the geometry is flat (2d) or describe a real 3d reconstruction setup.
+    """
+
+    def __init__(
+        self,
+        projection_geometry: dict[str, Any],
```
Andrewwango, line 36: Very minor detail, prefer `typing.Dict` over `dict`

deepinv/physics/functional/astra.py
```diff
@@ -0,0 +1,545 @@
...
+        else:
+            return 1.0
+
+    def __call__(
+        self, x: torch.Tensor, out: Optional[torch.Tensor] = None
+    ) -> torch.Tensor:
+        r"""Forward projection.
+
+        :param torch.Tensor x: Tensor of shape [1,H,W] in 2d, or [D,H,W] in 3d.
+        :param torch.Tensor | None out: To avoid unecessary copies, provide tensor of shape [...,A,N]
```
Andrewwango, line 154:
```suggestion
        :param torch.Tensor, None out: To avoid unecessary copies, provide tensor of shape [...,A,N]
```

Andrewwango (commented):
Thanks for the PR and it's really cool to provide this functionality! Some intermediate minor comments:

deepinv/tests/test_physics.py
romainvo: Although we test on the range of A, I guess because the adjoint is not matched we still get a somewhat high error. 

I have fixed the detector and angular parameters of the fan/cone setup to get a nice geometry with a proper field of view. At such low-resolution, the error does not seem to go below 0.05 for parallel, though it does decrease with higher resolution.. For fan/cone beam the error is a little higher (0.1/0.15), which is expected.

examples/patch-priors/demo_patch_priors_CT.py
romainvo: I also agree that it could be misleading ! I have added an "external-libraries" section with a low-dose CT example with astra. Tell me if the naming of the section is appropriate or if you had something else in mind ?

deepinv/tests/test_physics.py
romainvo: I have removed the TomographyWithAstra from find_operator

deepinv/physics/functional/astra.py
romainvo: I think typing.Dict is deprecated in favor of builtins.dict since version 3.9 (https://peps.python.org/pep-0585/) , though I can switch to it for consistency if you prefer ?

romainvo:
Hi everyone. I've tried to resolve the last checks failing. The doc won't build for some files because TomographyWithAstra only supports CUDA operations, this concerns:
- the code snippets in the docstring
- and the example in `external-libraries/demo_astra_tomography.py`

Is there a quick way to skip those checks if a CUDA device is not available ? 

Another point concerns the supported platform, at the moment, ASTRA only provides a pip package for linux. To get support for windows, you need to install it via their conda channel. I've restricted the dependency install to Linux platform for the moment.

Andrewwango:
One solution to the lack of CPU tests that we discussed with @tachella is to mock the Astra library calls, like @jscanvic does to create dataset tests when we can't actually download the datasets in #490. Do you think this could be possible?

The philosophy here is that as Astra is an external library, our tests should not have to cover its functionality, only our own wrapper. Then it is also ok that we don't have tests for adjointness etc, only that it runs ok and outputs correct shapes etc. Would you agree with this idea?

romainvo:
@Andrewwango I've patched the calls to the astra kernel with dummy functions in a new test, it should be good, tell me what you think. I put it in a `test_external_libraries.py`.

deepinv/tests/test_external_libraries.py
```diff
@@ -0,0 +1,111 @@
...
+            (True, "parallel", True),
+            (True, "fanbeam", False),
+            (True, "fanbeam", True),
+            (False, "parallel", False),
+            (False, "parallel", True),
+            (False, "conebeam", False),
+            (False, "conebeam", True),
+        ],
+    )
+    @mock.patch.object(
```
Andrewwango, line 28: I like this test! But do you think it would make more sense to not repeat code? One solution would be to merge it with the other tomography test and instead have an environment variable DEEPINV_MOCK_EXTERNAL or something like we have for mocking datasets in #490, and if it's true we run the mocked test

deepinv/physics/tomography.py
```diff
@@ -185,3 +201,355 @@ def A_adjoint(self, y, **kwargs):
...
+            object_geometry=self.object_geometry,
+            projection_geometry=self.projection_geometry,
+            is_2d=self.is_2d,
+        )
+
+        self.filter = RampFilter(dtype=torch.float32, device=self.device)
+
+        if normalize:
+            self.operator_norm = self.compute_norm(
+                torch.randn(self.img_shape, device=self.device)[None, None]
```
Andrewwango, line 439: Set a rng here to ensure reproducibility?

deepinv/physics/tomography.py
```diff
@@ -185,3 +201,355 @@ def A_adjoint(self, y, **kwargs):
...
+    It is more memory efficient than the :class:`deepinv.physics.Tomography` operator and support 3d geometries.
+    See documentation of :class:`deepinv.physics.functional.XrayTransform` for more
+    information on the ``astra`` wrapper.
+
+    Mathematically, it is described as a ray transform
+    :math:`R` which linearly integrates an object :math:`x` along straight
+    lines
+
+    .. math::
+        y = Rx
```
Andrewwango, line 217: Is R standard? Just because our deepinv convention is R=\inverse{} ie the reconstruction operator, and A=\forw{} ie the forward operator

deepinv/physics/tomography.py
```diff
@@ -1,10 +1,26 @@
...
+)
+from warnings import warn
+
+try:
+    import astra
+except:
+    astra = ImportError("The astra-toolbox package is not installed.")
+
 
 PI = 4 * torch.ones(1).atan()
```
Andrewwango, line 25: Consistent with the use of `from math import pi as PI` elsewhere?

deepinv/physics/functional/astra.py
Andrewwango: My bad! I'm clearly still stuck in 3.8...

Andrewwango (changes requested):
Looks good so far!

examples/external-libraries/_demo_astra_tomography.py
```diff
@@ -0,0 +1,163 @@
...
+prior = dinv.optim.prior.TVPrior(n_it_max=20)
+
+# Logging parameters
+verbose = True
+plot_convergence_metrics = (
+    True  # compute performance and convergence metrics along the algorithm.
+)
+
+# Algorithm parameters
+SCALING = 1 / physics.compute_norm(torch.rand_like(test_imgs)).item()
```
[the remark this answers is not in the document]
tachella: can you please remove the all caps from both the `demo_vanilla_PnP `example and this example?

examples/external-libraries/_demo_astra_tomography.py
```diff
@@ -0,0 +1,163 @@
...
+
+# Select the data fidelity term
+data_fidelity = L2()
+
+# Specify the prior (we redefine it with a smaller number of iteration for faster computation)
+prior = dinv.optim.prior.TVPrior(n_it_max=20)
+
+# Logging parameters
+verbose = True
+plot_convergence_metrics = (
```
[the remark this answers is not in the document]
tachella: I think you can remove the parentheses and just write
```suggestion
plot_convergence_metrics = True
```

examples/external-libraries/README.rst
```diff
@@ -0,0 +1,2 @@
+External libraries
```
tachella: where is this going?

deepinv/physics/tomography.py
```diff
@@ -179,9 +192,365 @@ def A_adjoint(self, y, **kwargs):
...
+    .. math::
+        y = Ax
+
+    where :math:`y` is the set of line integrals, called sinogram in 2D, or
+    radiographs in 3D. An object is typically scanned using a surrounding circular
+    trajectory. Given different acquisition systems, the lines along which
+    the integrals are computed follow different geometries:
+
+    * parallel. (2d and 3d)
+        Per view, all rays intersecting the object are parallel. In 2d, all rays live on the same plane, perpendicular
```
tachella: can you change all the '3d' -> '3D' and '2d' to '2D' for consistency?

deepinv/physics/tomography.py
```diff
@@ -179,9 +192,365 @@ def A_adjoint(self, y, **kwargs):
...
+    .. warning::
+
+        Due to computational efficiency reasons, the projector and backprojector
+        implemented in ``astra`` are not matched. The projector is typically ray-driven,
+        while the backprojector is pixel-driven. The adjoint of the forward Ray Transform
+        is approximated by rescaling the backprojector.
+
+    .. warning::
+
+        The :class:`deepinv.physics.functional.XrayTransform` used in :class:`TomographyWithAstra` sequentially processes batch elements, which can make the 2d parallel beam operator significantly slower than the its native torch counterpart with :class:`deepinv.physics.Tomography` (though still more memory-efficient).
```
tachella:
```suggestion
        The :class:`deepinv.physics.functional.XrayTransform` used in :class:`deepinv.physics.TomographyWithAstra` sequentially processes batch elements, which can make the 2d parallel beam operator significantly slower than the native torch counterpart with :class:`deepinv.physics.Tomography` (though still more memory-efficient).
```

deepinv/physics/tomography.py
```diff
@@ -179,9 +192,365 @@ def A_adjoint(self, y, **kwargs):
...
+
+    :param tuple[int, ...] img_size: Shape of the object grid, either a 2 or 3-element tuple, for respectively 2d or 3d.
```
tachella: what are the ... ? Does this display nicely in sphinx?

deepinv/physics/tomography.py
```diff
@@ -179,9 +192,365 @@ def A_adjoint(self, y, **kwargs):
...
+    :param int num_angles: Number of angular positions sampled uniformly in ``angular_range``. (default: 180)
+    :param int | tuple[int, ...], None num_detectors: In 2d, specify an integer for a single line of detector cells. In 3d, specify a 2-element tuple for (row,col) shape of the detector. (default: None)
+    :param tuple[float, float] angular_range: Angular range, defaults to (0,``torch.pi``).
+    :param float | tuple[float, float] detector_spacing: In 2d the width of a detector cell. In 3d a 2-element tuple specifying the (vertical, horizontal) dimensions of a detector cell. (default: 1.0)
+    :param tuple[float, ...] object_spacing: In 2d, the (x,y) dimensions of a pixel in the reconstructed image. In 3d, the (x,y,z) dimensions of a voxel. (default: ``(1.,1.)``)
+    :param tuple[float, ...], None aabb: Axis-aligned bounding-box of the reconstruction area [min_x, max_x, min_y, max_y, ...]. Optional argument, if specified, overrides argument ``object_spacing``. (default: None)
```
tachella: should we give a more meaningful name to this variable, eg`bounding_box`?

tachella (changes requested):
The PR looks very nice!

some minor comments

deepinv/physics/tomography.py
romainvo: It's the syntax for variable-length tuple of homogeneous type (see https://docs.python.org/3.9/library/typing.html). It's useful here to specify 2D and 3D use-cases where the tuple can be of size 2 or 3 dependingly, same for the bounding box arguments for exemple. 

typing-sphinx-display

deepinv/physics/tomography.py
romainvo: I think `aabb` is standard in computer graphics, but `bounding_box` works also fine !

examples/external-libraries/README.rst
romainvo: What do you mean ? I think the file is needed by sphinx to display the `examples/external-libraries` and parse the examples under this section (``docs/source/conf.py``).

tachella: ah yes, I didn't realise! i'm not sure if we want to create a new section.. should we add it in the advanced examples?

deepinv/physics/tomography.py
tachella: ok, thanks!

deepinv/physics/tomography.py
```diff
@@ -179,9 +192,368 @@ def A_adjoint(self, y, **kwargs):
...
+    It is more memory efficient than the :class:`deepinv.physics.Tomography` operator and support 3D geometries.
+    See documentation of :class:`deepinv.physics.functional.XrayTransform` for more
+    information on the ``astra`` wrapper.
+
+    Mathematically, it is described as a ray transform
+    :math:`A` which linearly integrates an object :math:`x` along straight
+    lines
+
+    .. math::
+        y = Ax
```
```suggestion
        y = \forw{x}
```

deepinv/physics/tomography.py
```diff
@@ -1,20 +1,33 @@
...
+    astra = ImportError("The astra-toolbox package is not installed.")
 
 
 class Tomography(LinearPhysics):
     r"""
     (Computed) Tomography operator.
 
     The Radon transform is the integral transform which takes a square image :math:`x` defined on the plane to a function
-    :math:`y=Rx` defined on the (two-dimensional) space of lines in the plane, whose value at a particular line is equal
+    :math:`y=Ax` defined on the (two-dimensional) space of lines in the plane, whose value at a particular line is equal
```
```suggestion
    :math:`y=\forw{x}` defined on the (two-dimensional) space of lines in the plane, whose value at a particular line is equal
```

examples/external-libraries/README.rst
Andrewwango: This is ok to keep in external librariers.

Andrewwango (changes requested):
Very minor, but apart from these LGTM and happy to approve once everything resolved

Andrewwango (approved):

Andrewwango:
@romainvo Please can you check out the bugs in the workflows?

tachella (approved):

PR #455, opened 2025-04-19
Plot inset fix

Andrewwango:
Close #451

### Checks to be done before submitting your PR
- [x] `python3 -m pytest tests/` runs successfully.
- [x] `black .` runs successfully.
- [x] `make html` runs successfully (in the `docs/` directory).
- [x] Updated docstrings related to the changes (as applicable).
- [x] Added an entry to the [CHANGELOG.rst](../CHANGELOG.rst).

PR #478, opened 2025-05-04
fix display demos

matthieutrs:
Since a recent upgrade, `plot()` function stopped displaying in sphinx.


.circleci/config.yml
```diff
@@ -44,13 +44,15 @@ jobs:
...
 
-        # Build docs
+        # Build docs (fast for now, too slow for circleci in non-fast mode)
         - run:
             name: make html
             no_output_timeout: 120m
             command: |
               cd docs;
-              make html;
+              make SPHINXOPTS=-v html-fast;
```
tachella, line 53: let's go back to make html and merge :)

tachella:
I think there is a clash with #455 - it is worth checking if both PRs are proposing the same solution

matthieutrs:
@tachella yes this is proposing the same solution as #455 , I've added the contribution of @Andrewwango in the CHANGELOG (although for some reason his commits wouldn't appear)

.circleci/config.yml
matthieutrs: I cannot for now - we need to upgrade our circleci with more compute bc atm this is too heavy (memory and compute-wise) for base plan

tachella: ah shit, ok - we can check if we can pay this later, and keep the fast for the moment which is already quite useful for checking formatting errors

tachella (approved):
lgtm

PR #416, opened 2025-03-02
Better FastMRI, better splitting and weighted SSDU

g-keying:
SSDU:
- Weighted-SSDU loss from [Millard & Chiew](https://arxiv.org/abs/2205.10278) - theoretical basis for SSDU based on Noisier2Noise
- Robust SSDU loss from [Millard & Chiew 2024](https://arxiv.org/abs/2210.01696)
- Generalised multiplicative splitting mask generator
- Polynomial variable-density acceleration mask generator
- Add more measplit tests

MRI:
- Estimate coil maps using ESPIRiT via sigpy

FastMRI:
- FastMRITransform better loads raw data by estimating coil maps and generating masks

Other:
- Add more metric standardisation
- Add HaarPSI metric from [this paper](https://arxiv.org/abs/2410.24098), code copied and adapted from https://github.com/ideal-iqa/haarpsi-pytorch, fixing [an issue](https://github.com/ideal-iqa/haarpsi-pytorch/issues/1)

### Checks to be done before submitting your PR
- [x] `python3 -m pytest tests/` runs successfully.
- [x] `black .` runs successfully.
- [x] `make html` runs successfully (in the `docs/` directory).
- [x] Updated docstrings related to the changes (as applicable).
- [x] Added an entry to the [CHANGELOG.rst](../CHANGELOG.rst).

PR #488, opened 2025-05-12
add PhysicsGenerator.average

### LINKED ISSUE #486: Physics generator averaging better memory handling
In #416 and #454, physics generator averaging is proposed to calculate the mask mean:

```python
(in deepinv.physics.generator.base.PhysicsGenerator)

    def average(self, n: int = 2000) -> Dict:
        """Calculate average of physics generator.
        :param int n: number of samples to average over, defaults to 2000
        """
        params = self.step(batch_size=n)
        return {k: v.mean(0, keepdim=True) for (k, v) in params.items()}
```

This risks going out of memory. Therefore, a better mini batched version should be propsoed.

jscanvic:
Fixes #486 

### DIFF
--- a/deepinv/physics/generator/base.py
+++ b/deepinv/physics/generator/base.py
@@ -135,6 +135,36 @@ class PhysicsGenerator(nn.Module):
 
         return PhysicsGenerator(step=step)
 
+    def average(self, n: int = 2000, batch_size: int = 1) -> dict:
+        """Calculate average of physics generator.
+        :param int n: number of samples to average over, defaults to 2000
+        :param int n: number of samples to compute in parallel, higher means faster but more costly memory-wise, defaults to 1
+        :returns: A dictionary with the new parameters, that is ``{param_name: param_value}``.
+        """
+        assert n > 0, "n must be positive"
+        assert batch_size >= 1, "batch_size must be positive"
+        params_sum = None
+        keys = None
+        n_processed = 0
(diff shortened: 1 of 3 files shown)


jscanvic:
The implementations in #416 and #454 don't support cases like `ProductConvolutionBlurGenerator` because not all of the parameters are tensor-valued, one of them (padding) is string-valued.

I'm not sure how we want to take care of these cases. The options I see are: 1) let them fail, 2) make it so non-numeric parameters are supported as long as they always have the same value when calling `PhysicsGenerator.step`.

deepinv/physics/generator/base.py
```diff
@@ -135,6 +135,36 @@ def step(batch_size: int = 1, seed: int = None, **kwargs):
...
+        :param int n: number of samples to average over, defaults to 2000
+        :param int n: number of samples to compute in parallel, higher means faster but more costly memory-wise, defaults to 1
+        :returns: A dictionary with the new parameters, that is ``{param_name: param_value}``.
+        """
+        assert n > 0, "n must be positive"
+        assert batch_size >= 1, "batch_size must be positive"
+        params_sum = None
+        keys = None
+        n_processed = 0
+        while n_processed != n:
```
Andrewwango, line 149: Is this safe or should it be > n?

Andrewwango (approved):
thanks Jérémy!

PR #501, opened 2025-05-24
Upgrade deprecated types

### LINKED ISSUE #500: Consistent typing
Would be good to have consistent type annotations: typing in docstrings should be reflected in the annotations, and be brought up to date with latest pep, eg Dict->dict

jscanvic:
Fixes #500

From what I witnessed, black does not flag use of deprecated `typing.T` types, e.g., `typing.Dict` instead of `dict`, but [ruff](https://docs.astral.sh/ruff/) does with the rule [UP006](https://docs.astral.sh/ruff/rules/non-pep585-annotation/).

On a tangential note, maybe it'll be worth integrating a stronger linter like ruff into the CI pipeline at some point. We probably don't want to enforce all of the default rules to avoid making contributing overly painful, but for specific things I believe it might be worth having.

**Find offending lines**

```bash
ruff check --target-version "py39" --select "UP006"
```

(See ruff.txt for the output)

**Fix**

Apparently, ruff is able to fix the code automatically but it requires using the option `--unsafe-fixes` so I'm not sure if it can be trusted. This is what I've used for this PR.

```bash
ruff check --unsafe-fixes --fix --target-version "py39" --select "UP006"
```

**ruff.txt** ([download](https://github.com/user-attachments/files/20427012/ruff.txt), excerpt below)

```
deepinv/datasets/cmrxrecon.py:108:19: UP006 Use `tuple` instead of `Tuple` for type annotation
    |
106 |         mask_generator: Optional[BaseMaskGenerator] = None,
107 |         transform: Optional[Callable] = None,
108 |         pad_size: Tuple[int, int] = (512, 256),
    |                   ^^^^^ UP006
109 |         noise_model: NoiseModel = None,
110 |     ):
    |
    = help: Replace with `tuple`

deepinv/datasets/cmrxrecon.py:171:10: UP006 Use `dict` instead of `Dict` for type annotation
    |
169 |     def _retrieve_metadata(
170 |         self, fname: Union[str, Path, os.PathLike]
171 |     ) -> Dict[str, Any]:
    |          ^^^^ UP006
172 |         """Open file and retrieve metadata
    |
    = help: Replace with `dict`

(Truncated)

Found 90 errors.
No fixes available (90 hidden fixes can be enabled with the `--unsafe-fixes` option).
```


jscanvic:
Thanks @Andrewwango!

I don't know much about mypy but it seems like a neat piece of software. It'd be interesting to experiment with it and see how we could use it for the library!

Andrewwango (approved):
Thanks for doing this!

So does your ruff command work?

`ruff check --target-version "py39" --select "UP006"`

And the the set of errors it brought up is exactly the set of errors you fix here?

If yes, then let's include it in CI!

I have moved the mypy suggestion to #536.

jscanvic:
It does most of the work the only thing it does not catch is types in docstrings so I handled them manually. It's added!

PR #508, opened 2025-06-03
Fix R2R results

bemc22:
minor changes

* register `normalize` from `PoissonNoise` paramter to be saved in `state_dict` 
* adjust `learning rate` in demo and uncomment loading weights


tachella:
Some additional points to fix: 

The note here is not showing well (sphinx formatting issue)
image

I would explain here that we start from a pretrained model to speed up training, with a similar explanation to that in the other self-supervised examples

image

bemc22:
> Some additional points to fix:
> 
> The note here is not showing well (sphinx formatting issue) image
> 
> I would explain here that we start from a pretrained model to speed up training, with a similar explanation to that in the other self-supervised examples
> 
> image

thanks, fixed in [be0197c]

examples/self-supervised-learning/demo_r2r_denoising.py
```diff
@@ -122,27 +123,27 @@
 #
 
 epochs = 1  # choose training epochs
-learning_rate = 1e-3
-batch_size = 32 if torch.cuda.is_available() else 1
+learning_rate = 1e-9
```
tachella: this looks super small - is this expected?

bemc22: yes! for the poisson case, since we are loading pre-trained weights, the idea it's to not modify them

tachella: I would still choose a learning rate that modifies them, does the performance get much worse?
An option to avoid overfitting is to split the training set and use the R2R loss in the validation set too (ie self-supervised)

tachella: @bemc22 ?

bemc22: Sorry, yes agree! I will fix it as you suggest

examples/self-supervised-learning/demo_r2r_denoising.py
```diff
@@ -153,9 +154,11 @@
 verbose = True  # print training information
 wandb_vis = False  # plot curves and images in Weight&Bias
 
+# self-supervised training!
 train_dataloader = DataLoader(
-    train_dataset, batch_size=batch_size, num_workers=num_workers, shuffle=False
+    test_dataset, batch_size=batch_size, num_workers=num_workers, shuffle=False
```
tachella: Why do you train with test?

bemc22: Sorry, I misunderstood your last comment. Since MNIST does not have a validation set, I used the training set again. The current restoration quality is 20.5 dB compared to 12 dB without learning.

bemc22:
The build documentation failed due to multiple warnings of the type:

`WARNING: Git clone too shallow [git.too_shallow]`

I resolved this issue by updating `documentation.yaml` to load the complete Git history instead of a shallow clone:

```
  with:
    fetch-depth: 0
```

A drawback of this solution is a slight increase in runtime, from 53 minutes to 54 minutes.

I am not certain whether this is the optimal solution @tachella .

jscanvic:
@bemc22  Good catch! I faced the same issue in #555 and managed to identify the root cause:

We use `sphinx_sitemap` in the documentation pipeline but we haven't pinned down its version so 2 days ago, when it was updated to v2.7.0, it started being used in our CI pipeline. The thing is it now uses timestamps from git commits to populate the `<lastmod>` tag (a neat feature!) which is prone to failure when using shallow clones. It's a known issue and there are suggested fixes:

https://github.com/mgeier/sphinx-last-updated-by-git/

For now I suggest to temporarily pin down `sphinx_sitemap` to v2.6.0 as I do in feecb82f75494ae76445f670abd24fe2de807c29 and figure out later which solution is the best (disabling the new feature, using a regular clone, silencing those specific warnings, ...).

Related: https://github.com/jdillard/sphinx-sitemap/pull/95

bemc22:
> @bemc22  Good catch! I faced the same issue in #555 and managed to identify the root cause:
> 
> We use `sphinx_sitemap` in the documentation pipeline but we haven't pinned down its version so 2 days ago, when it was updated to v2.7.0, it started being used in our CI pipeline. The thing is it now uses timestamps from git commits to populate the `<lastmod>` tag (a neat feature!) which is prone to failure when using shallow clones. It's a known issue and there are suggested fixes:
> 
> https://github.com/mgeier/sphinx-last-updated-by-git/
> 
> For now I suggest to temporarily pin down `sphinx_sitemap` to v2.6.0 as I do in feecb82f75494ae76445f670abd24fe2de807c29 and figure out later which solution is the best (disabling the new feature, using a regular clone, silencing those specific warnings, ...).
> 
> Related: https://github.com/jdillard/sphinx-sitemap/pull/95

Thanks! I will update this PR accordingly

deepinv/physics/noise.py
```diff
@@ -491,7 +491,8 @@ def __init__(
         self, gain=1.0, normalize=True, clip_positive=False, rng: torch.Generator = None
     ):
         super().__init__(rng=rng)
-        self.normalize = normalize
+        self.register_buffer("normalize", torch.tensor(normalize, dtype=torch.bool))
+        self.update_parameters(gain=gain)
```
tachella: is this `update_parameters` necessary? the rest of the noise models do not update inside the init?

bemc22: You are correct; I seem have an outdated version. Currently in the main branch, it is saved as `self.register_buffer`, and the other noise models follow the same approach for their parameters. I will update it accordingly.

```python
        self.register_buffer(
            "gain", self._float_to_tensor(gain).to(getattr(rng, "device", "cpu"))
        )
```
source: https://github.com/deepinv/deepinv/blame/3abc8e607c8cd74299531b3e89bd4090316fd5e9/deepinv/physics/noise.py#L496-L498

PR #522, opened 2025-06-07
Speed up the test suite by parallelizing the slowest outlier

jscanvic:
The test suite takes about 30 minutes to run and about 20 of them are spent on the single test `test_sde` in `deepinv.tests.test_sampling`.

I'm not sure why so much time is spent on this single test but we should be able to get a 10 minutes (30%) speed up by letting it run in parallel of the other tests.

As a matter of comparison, the 2nd slowest test is `test_priors_algo` in `deepinv.tests.test_optim` and it only takes 3 minutes to finish off.

**How**

The test suite currently runs sequentially, one test at a time, but it is possible to parallelize it using `pytest-xdist`. Here, we run the tests on two separate workers, one for `test_sde` and one for the other tests which remain sequentially executed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,12 +48,10 @@ jobs:
 
       - name: Test with pytest and generate coverage report
         run: |
-          python3 -m pytest --cov=deepinv --junitxml=junit.xml -o junit_family=legacy -v
+          python3 -m pytest -n 2 --dist=loadgroup --cov=deepinv --junitxml=junit.xml -o junit_family=legacy -v
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v5
 
       - name: Upload test results to Codecov
         if: ${{ !cancelled() }}
         uses: codecov/test-results-action@v1
-
-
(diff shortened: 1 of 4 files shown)


Andrewwango (approved):
LGTM, thanks for doing this @jscanvic. Does this affect the time taken in GitHub actions  or just locally?

jscanvic:
It affects both but the actual speedup depends on the number of CPU cores. The ~15% speedup I get is for CI (GitHub machines have 4 cores). See:
https://github.com/deepinv/deepinv/actions/runs/15508239459/job/43665730679
https://github.com/deepinv/deepinv/actions/runs/15508239459/job/43665730678
https://github.com/deepinv/deepinv/actions/runs/15511955911/job/43674001011
https://github.com/deepinv/deepinv/actions/runs/15511955911/job/43674001016

Andrewwango:
> It affects both but the actual speedup depends on the number of CPU cores. The ~15% speedup I get is for CI (GitHub machines have 4 cores). See:

Amazing!

mh-nguyen712:
Bravo @jscanvic ! Thanks for speeding up my slow example in demo_sde.py

PR #524, opened 2025-06-09
RAM

matthieutrs:


tachella:
do you think you can make a demo file too? maybe with some finetuning at the end?

deepinv/models/ram.py
```diff
@@ -0,0 +1,1023 @@
...
+        self,
+        physics,
+        img_shape,
+        filter="sinc",
+        scales=[2, 4, 8],
+        device="cpu",
+        **kwargs,
+    ):
+        super().__init__(noise_model=physics.noise_model, **kwargs)
+        self.base = physics
```
tachella: should we estimate the norm of the operator and normalize?

deepinv/models/ram.py
```diff
@@ -0,0 +1,1023 @@
...
+        return super().A_adjoint(x, **kwargs)
+
+    def A_adjoint(self, y, **kwargs):
+        return super().A(y, **kwargs)
+
+    def prox_l2(self, z, y, gamma, **kwargs):
+        return super().prox_l2(z, y, gamma, **kwargs)
+
+
+class MultiScalePhysics(Physics):
```
tachella: are we adding this to the library separately from RAM?

docs/source/api/deepinv.models.rst
```diff
@@ -92,6 +92,7 @@ Artifact Removal
    :nosignatures:
 
    deepinv.models.ArtifactRemoval
+   deepinv.models.RAM
```
tachella, line 96: add to User Guide too

matthieutrs:
still todo:
- add example
- check docs

deepinv/physics/blur.py
```diff
@@ -204,6 +204,21 @@ def update_parameters(self, filter=None, factor=None, **kwargs):
             self.Fh2 = torch.nn.Parameter(self.Fh2, requires_grad=False)
 
 
+class Upsampling(Downsampling):
+    r"""
```
tachella, line 268: we should add some proper docs to this

deepinv/models/ram.py
```diff
@@ -0,0 +1,927 @@
...
+from deepinv.physics import Physics, LinearPhysics, Downsampling
+from deepinv.utils.tensorlist import TensorList
+
+from huggingface_hub import hf_hub_download
+
+cuda = True if torch.cuda.is_available() else False
+Tensor = torch.cuda.FloatTensor if cuda else torch.FloatTensor
+
+
+class RAM(nn.Module):
```
tachella:
```suggestion
class RAM(deepinv.models.Reconstructor, deepinv.models.Denoiser):
```

tachella: I think it could work for both denoiser and reconstructor?

examples/unfolded/demo_ram.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+    x_train = x[..., :64, :64]  # take a small patch of the image
+    y_train = physics_train(x_train)
+else:
+    physics_train = physics.clone()
+    x_train = x.clone()
+    y_train = y.clone()
+
+
+mc_loss = dinv.loss.SureGaussianLoss(physics.noise_model.sigma)
```
tachella: Move to Recorrupted2Recorrupted, which is a variant of SURE that works really well in low-photon Poisson
```suggestion
mc_loss = dinv.loss.R2RLoss()
```

examples/unfolded/demo_ram.py
```diff
@@ -0,0 +1,160 @@
...
+# We showcase this in the following, where the model is fine-tuned on the measurement vector `y` itself.
+# Here, since this example is run in a no-GPU environment, we will use a small patch of the image to speed up training,
+# but in practice, we can use the full image.
+from deepinv.datasets.utils import UnsupDataset
+
+# Define small physics in the case of no-GPU
+if not torch.cuda.is_available():
+    physics_train = dinv.physics.Demosaicing(
+        img_size=(3, 64, 64),
+        noise_model=dinv.physics.GaussianNoise(0.05),
```
tachella:
```suggestion
        noise_model=dinv.physics.PoissonNoise(0.1),
```

examples/unfolded/demo_ram.py
```diff
@@ -0,0 +1,160 @@
...
+
+t = dinv.transform.Shift(shift_max=0.4)
+eq_loss = dinv.loss.EILoss(t, weight=0.1)
+
+losses = [mc_loss, eq_loss]
+
+dataset = UnsupDataset(y_train)
+
+train_dataloader = torch.utils.data.DataLoader(dataset, batch_size=1, shuffle=True)
+
```
tachella, line 170:
```suggestion
y_val = physics(dinv.utils.load_example(another image??).to(device))
eval_dataloader = torch.utils.data.DataLoader(UnsupDataset(y_val), batch_size=1, shuffle=True)
```

examples/unfolded/demo_ram.py
```diff
@@ -0,0 +1,160 @@
...
+optimizer = torch.optim.Adam(model.parameters(), lr=5e-5)
+
+max_epochs = 20
+trainer = dinv.Trainer(
+    model=model,
+    physics=physics_train,
+    eval_interval=max_epochs,
+    ckp_interval=max_epochs - 1,
+    metrics=losses[0],
+    early_stop=True,
```
tachella, line 180: this should do the job of early stopping with the validation self-sup loss

examples/unfolded/demo_ram.py
```diff
@@ -0,0 +1,160 @@
...
+
+train_dataloader = torch.utils.data.DataLoader(dataset, batch_size=1, shuffle=True)
+
+optimizer = torch.optim.Adam(model.parameters(), lr=5e-5)
+
+max_epochs = 20
+trainer = dinv.Trainer(
+    model=model,
+    physics=physics_train,
+    eval_interval=max_epochs,
```
tachella: Make more evals to make sure that we stop well
```suggestion
    eval_interval=5,
```

docs/source/user_guide/reconstruction/artifact.rst
```diff
@@ -14,3 +14,24 @@ reconstruction network :class:`deepinv.models.Reconstructor` :math:`R` by doing
 - Pseudoinverse: :math:`\inversef{y}{A}=\phi(A^{\dagger}y)` with ``mode='pinv'``.
 - | Direct: :math:`\inversef{y}{A}=\phi(y)` with ``mode='direct'``.
   | This option serves as a wrapper to obtain a :class:`Reconstructor <deepinv.models.Reconstructor>`, and can be used to adapt a generic denoiser or image-to-image network into one that is specific to an inverse problem.
+
+
+Deep artfact removal networks
```
tachella:
```suggestion
General reconstruction models
```

deepinv/datasets/utils.py
```diff
@@ -180,3 +181,22 @@ def forward(self, x: Tensor):
         :param torch.Tensor x: image tensor of shape (..., H, W)
         """
         return stack([x, zeros_like(x)], dim=-3)
+
+
+class UnsupDataset(torch.utils.data.Dataset):
```
Andrewwango: Why is this in `dataset/utils` if it's only used in the example? You could just define this in the example.

If you think it's sufficiently useful for other cases, then please could you add it to docs (maybe reference it in the Trainer user guide / docstring) + test it

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+import deepinv as dinv
+from deepinv.physics import MultiScaleLinearPhysics, Pad
+from deepinv.utils.tensorlist import TensorList
+from .base import Reconstructor
```
Andrewwango:
```suggestion
from deepinv.models.base import Reconstructor
```

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+
+from huggingface_hub import hf_hub_download
+
+cuda = True if torch.cuda.is_available() else False
```
Andrewwango: What is this line + the line below for?

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+        self.up2 = upsample_convtranspose(nc[2], nc[1], bias=False, mode="2")
+        self.up1 = upsample_convtranspose(nc[1], nc[0], bias=False, mode="2")
+
+        self.m_tail = OutTail(nc[0], in_channels)
+
+        # load pretrained weights from hugging face
+        if pretrained:
+            self.load_state_dict(
+                torch.load(
+                    hf_hub_download(repo_id="mterris/ram", filename="ram.pth.tar"),
```
Andrewwango: Why not `torch.hub.load_state_dict_from_url` which we use in our other models? Won't huggingface_hub add another dependency

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+
+from huggingface_hub import hf_hub_download
+
+cuda = True if torch.cuda.is_available() else False
+Tensor = torch.cuda.FloatTensor if cuda else torch.FloatTensor
+
+
+class RAM(Reconstructor):
+    r"""
+    RAM model.
```
Andrewwango:
```suggestion
    Reconstruct Anything Model.
```

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+
+    This model (proposed in `this paper <https://arxiv.org/abs/2503.08915>`_) is a convolutional neural network (CNN)
+    designed for image reconstruction tasks.
+
+    :param in_channels: Number of input channels. If a list is provided, the model will have separate heads for each channel.
```
Andrewwango: typing for the docstrings please

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+            )
+
+        if device is not None:
+            self.to(device)
+
+    def constant2map(self, value, x):
+        r"""
+        Converts a constant value to a map of the same size as the input tensor x.
+
+        :params float value: constant value
```
Andrewwango:
```suggestion
        :param float value: constant value
```

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+        :params torch.Tensor x: input tensor
```
Andrewwango:
```suggestion
        :param torch.Tensor x: input tensor
```

examples/unfolded/demo_ram.py
```diff
@@ -0,0 +1,164 @@
+"""
+RAM model for solving inverse problems.
```
Andrewwango:
```suggestion
Reconstruct Anything Model for solving inverse problems.
```

docs/source/user_guide/reconstruction/weights.rst
```diff
@@ -48,4 +48,6 @@ associated reference and relevant details. All pretrained weights are hosted on
    * - :class:`deepinv.models.Restormer`
      - from `Restormer: Efficient Transformer for High-Resolution Image Restoration <https://arxiv.org/abs/2111.09881>`_. Pretrained parameters from `swz30 github <https://github.com/swz30/Restormer/tree/main>`_.
    * -
-     - Also available on the `deepinverse Restormer HugginfaceHub <https://huggingface.co/deepinv/Restormer/tree/main>`_.
\ No newline at end of file
+     - Also available on the `deepinverse Restormer HugginfaceHub <https://huggingface.co/deepinv/Restormer/tree/main>`_.
```
Andrewwango:
```suggestion
     - Also available on the `deepinverse Restormer HuggingfaceHub <https://huggingface.co/deepinv/Restormer/tree/main>`_.
```

deepinv/tests/test_models.py
```diff
@@ -649,6 +684,42 @@ def __len__(self):
...
+    # skip test if channel is 3 and physics_name is "MRI"
+    if channels == 3 and physics_name == "MRI":
+        pytest.skip(f"Skipping {model} with {physics_name} for 3 channel input.")
+
+    model = choose_restoration_model(model).to(device)
+    torch.manual_seed(0)
+
+    imsize = (
+        channels,
+        32,
```
Andrewwango: Can we put imsize as a fixture and test a variety of imsizes?

deepinv/tests/test_models.py
```diff
@@ -649,6 +684,42 @@ def __len__(self):
...
+    )  # minimal size for this model is (3, 64, 64), but we want to test non-square images
+    physics, imsize, _, dtype = find_operator(physics_name, device, imsize=imsize)
+
+    x = torch.randn(imsize, device=device, dtype=dtype, generator=rng).unsqueeze(0)
+    y = physics(x)
+
+    with torch.no_grad():
+        x_hat = model(y, physics)
+
+    assert x_hat.shape == x.shape
```
Andrewwango, line 931: Can we do a performance check like y = blur(x); assert psnr(model(y), x) > psnr(y, x)

deepinv/tests/test_models.py
```diff
@@ -94,6 +121,14 @@ def choose_denoiser(name, imsize):
     return out.eval()
 
 
+def choose_restoration_model(name):
```
Andrewwango: Could we integrate tests for the other reconstruction models into here? e.g. 
```python
deepinv.models.ArtifactRemoval
deepinv.models.MoDL
deepinv.models.VarNet
deepinv.models.PanNet
```

deepinv/physics/multiscale.py
```diff
@@ -0,0 +1,137 @@
...
+    def A_adjoint(self, y, scale=None, **kwargs):
+        self.set_scale(scale)
+        y = self.base.A_adjoint(y, **kwargs)
+        if self.scale == 0:
+            return y
+        else:
+            return self.Upsamplings[self.scale - 1].A_adjoint(y)
+
+
+class Pad(LinearPhysics):
```
Andrewwango: Can this not be achieved with the ComposedPhysics operators? e.g. 
```python
physics_padded = physics * LinearPhysics(A=lambda: unpad, A_adjoint=lambda: pad)
```

If there is a reason not to do that, then this physics should be put in the docs + tested. I would also maybe rename it something clearer like `PhysicsPadder`

deepinv/physics/multiscale.py
```diff
@@ -0,0 +1,137 @@
+import torch
+
+from deepinv.physics import Physics, LinearPhysics
+from deepinv.physics.blur import Upsampling
+
+
+class MultiScalePhysics(Physics):
```
Andrewwango: I feel like `MultiScalePhysics` sounds like a base class when it is a wrapper class. Maybe `PhysicsMultiScaler`?

Andrewwango: Can this not be achieved with the ComposedPhysics operators? e.g.

```python
physics_ms = physics * Physics(A=lambda x: upsample(x))
physics_ms = physics * LinearPhysics(A=lambda x: upsample(x), A_adjoint=lambda y: downsample(y))
```
These classes feel quite specific to RAM, so the helper methods `upsample(...)` etc. can be put in RAM, rather than having this class in physics.

If there is a reason not to do that, then this physics should be put in the docs + tested. 

Similar comment with Pad below.

deepinv/physics/blur.py
Andrewwango: and tests

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+
+        out = self.forward_unet(x_in, sigma=sigma, gamma=gamma, physics=physics, y=y)
+
+        out = physics.remove_pad(out)
+
+        return out
+
+
+### --------------- MODEL ---------------
+class BaseEncBlock(nn.Module):
```
Andrewwango, line 330: docstring

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+                x, physics=physics, y=y, img_channels=img_channels, scale=scale
+            )
+        return x
+
+
+def krylov_embeddings(y, p, factor, v=None, N=4, x_init=None):
+    r"""
+    Efficient Krylov subspace embedding computation with parallel processing.
+
+    :params torch.Tensor y: Input tensor.
```
Andrewwango: param instead of params

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+    :params p: An object with A and A_adjoint methods (linear operator).
```
Andrewwango: typing

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+        out = torch.cat([out, x_k], dim=1)
+
+    return out
+
+
+class MeasCondBlock(nn.Module):
+    r"""
+    Measurement conditioning block for the RAM model.
+
+    :param out_channels: Number of output channels.
```
Andrewwango: typing

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+            aux = self.nl_2(aux)
+            aux_0 = getattr(self, f"conv2{i}")(aux)
+            aux_1 = getattr(self, f"skipconv{i}")(x)
+            x = aux_0 + aux_1
+
+        return x
+
+
+# --------------------------------------------------------------------------------------
+class AffineConv2d(nn.Conv2d):
```
Andrewwango, line 900: Docstring for this class

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+                stride=self.stride,
+                dilation=self.dilation,
+                groups=self.groups,
+            )
+
+
+"""
+Functional blocks below
+
+Parts of code borrowed from
```
Andrewwango: I think for borrowing code it's worth properly mentioning them where appropriate in each function's docstring to make sure code is credited in the right place

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+        u = self.nl(u)
+        u_2 = self.conv2(u)
+        emb_grad = self.PhysicsBlock(
+            u, y, physics, img_channels=img_channels, scale=scale
+        )
+        u_1 = self.gain * emb_grad
+        return x + u_2 + u_1
+
+
+class InHead(torch.nn.Module):
```
Andrewwango, line 591: sorry to be a pain: docstrings and typing in these few classes below

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+        noise_level_map = self.constant2map(sigma, x)
+        gamma_map = self.constant2map(gamma, x)
+        return torch.cat((x, noise_level_map, gamma_map), 1)
+
+    def realign_input(self, x, physics, y):
+        r"""
+        Realign the input x based on the measurements y and the physics model.
+        Applies the proximity operator of the L2 norm with respect to the physics model.
+
+        :params torch.Tensor x: Input tensor
```
Andrewwango: params -> param

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+        gamma = gamma[(...,) + (None,) * (x.dim() - 1)]
+        model_input = physics.prox_l2(x, y, gamma=gamma * self.fact_realign)
+
+        return model_input
+
+    def forward_unet(self, x0, sigma=None, gamma=None, physics=None, y=None):
+        r"""
+        Forward pass of the UNet model.
+
+        :params torch.Tensor x0: init image
```
Andrewwango: param

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+        x = self.m_up1(x, physics=physics, y=y, img_channels=img_channels, scale=0)
+
+        x = self.m_tail(x + x1, img_channels)
+
+        return x
+
+    def forward(self, y=None, physics=None):
+        r"""
+        Reconstructs a signal estimate from measurements y
+        :param torch.tensor y: measurements
```
Andrewwango:
```suggestion
        :param torch.Tensor y: measurements
```

deepinv/models/ram.py
```diff
@@ -0,0 +1,928 @@
...
+        :param deepinv.physics.Physics physics: forward operator
```
Andrewwango, line 233: return value please in docstrings

Andrewwango (changes requested):
CHANGELOG please.

Thanks for integrating RAM into deepinv, here are some comments!

deepinv/datasets/utils.py
matthieutrs: defined in example

deepinv/models/ram.py
matthieutrs: removed

deepinv/models/ram.py
matthieutrs: good point

deepinv/tests/test_models.py
matthieutrs: I think this belongs to another PR since this would mean debugging potential suboptimalities of these models in this PR. Maybe the best is to open an issue asking for more thorough testing of those?

matthieutrs:
Thanks @Andrewwango! 

I'm halfway through your comments - I'll ping you when done. 

TODO: compare with MultiScalePhysics

deepinv/physics/multiscale.py
Andrewwango: I'm happy to leave this refactoring, I've mentioned this in https://github.com/deepinv/deepinv/issues/533#issuecomment-3036180494 to address in future. Please address the remaining suggestions in the above comment, and maybe bring Pad and MultiScalePhysics closer together (so they are both in form of:

```python
class Wrapper(LinearPhysics):
    def A(self, x):
        return super().A(hijack(x))
    def A_adjoint(self, y):
        return hijack_adjoint(super().A_adjoint(y))
```
)

deepinv/physics/multiscale.py
Andrewwango: see comment with Pad above

deepinv/tests/test_models.py
matthieutrs: I've split the test in image size test + performance test. I've also added the models you are suggesting, but saddly the tests fail for those. Skipping for now, but we should change this in the future (TODO).

docs/source/user_guide/reconstruction/artifact.rst
```diff
@@ -14,3 +14,24 @@ reconstruction network :class:`deepinv.models.Reconstructor` :math:`R` by doing
 - Pseudoinverse: :math:`\inversef{y}{A}=\phi(A^{\dagger}y)` with ``mode='pinv'``.
 - | Direct: :math:`\inversef{y}{A}=\phi(y)` with ``mode='direct'``.
   | This option serves as a wrapper to obtain a :class:`Reconstructor <deepinv.models.Reconstructor>`, and can be used to adapt a generic denoiser or image-to-image network into one that is specific to an inverse problem.
+
+
+General reconstruction models
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+We provide the following list of deep artifact removal architectures.
```
tachella:
```suggestion
We provide the following list of multiphysics reconstruction models.
```

docs/source/user_guide/reconstruction/artifact.rst
```diff
@@ -14,3 +14,24 @@ reconstruction network :class:`deepinv.models.Reconstructor` :math:`R` by doing
...
+See :ref:`pretrained-weights` for more information on pretrained denoisers.
+
+.. list-table:: Deep denoisers
```
tachella:
```suggestion
.. list-table:: Multiphysics reconstruction models
```

deepinv/models/ram.py
```diff
@@ -0,0 +1,1047 @@
...
+        x = self.m_up2(x, physics=physics, y=y, img_channels=img_channels, scale=1)
+
+        x = self.up1(x + x2)
+        x = self.m_up1(x, physics=physics, y=y, img_channels=img_channels, scale=0)
+
+        x = self.m_tail(x + x1, img_channels)
+
+        return x
+
+    def forward(self, y=None, physics=None):
```
tachella: I would allow users to use it as a Denoiser

docs/source/user_guide/reconstruction/artifact.rst
```diff
@@ -14,3 +14,24 @@ reconstruction network :class:`deepinv.models.Reconstructor` :math:`R` by doing
...
+   :widths: 15 25 15 15 10
+   :header-rows: 1
+
+   * - Model
```
tachella, line 32: I would also add RAM to the list of denoisers

deepinv/models/ram.py
```diff
@@ -0,0 +1,1056 @@
...
+        x = self.m_tail(x + x1, img_channels)
+
+        return x
+
+    def forward(self, y, physics=None, sigma=None, gain=None):
+        r"""
+        Reconstructs a signal estimate from measurements y
+
+        :param torch.Tensor y: measurements
+        :param deepinv.physics.Physics physics: forward operator
```
tachella, line 233: Add gain and sigma

deepinv/physics/multiscale.py
```diff
@@ -0,0 +1,146 @@
...
+    def A_adjoint(self, y, scale=None, **kwargs):
+        self.set_scale(scale)
+        y = self.base.A_adjoint(y, **kwargs)
+        if self.scale == 0:
+            return y
+        else:
+            return self.Upsamplings[self.scale - 1].A_adjoint(y)
+
+
+class PhysicsCropper(LinearPhysics):
```
tachella: I think we should call this Outpainting, and use compose with another physics instead of asking an input linear physics

Andrewwango: Julián, I think it's safer to do this explicitly for now, based on our past discussions on composed physics. We discussed this point with Matthieu and for now I have added this point to https://github.com/deepinv/deepinv/issues/533#issuecomment-3036180494 for us to address later.

examples/unfolded/demo_ram.py
```diff
@@ -0,0 +1,217 @@
+"""
+Reconstruct Anything Model (RAM) for solving inverse problems.
+====================================================================================================
+
+This example shows how to use the RAM model method to solve inverse problems. The RAM model, described in
```
Andrewwango: Mention foundation model somewhere here

examples/unfolded/demo_ram.py
```diff
@@ -0,0 +1,217 @@
...
+    [
+        "Original",
+        "Measurement\n PSNR = {:.2f}dB".format(in_psnr),
+        "Reconstruction\n PSNR = {:.2f}dB".format(out_psnr),
+    ],
+    figsize=(8, 3),
+)
+
+# %%
+# This model is not trained on all degradations, so it may not perform well on all inverse problems.
```
Andrewwango:
```suggestion
# This model is not trained on all degradations, so it may not perform well on all inverse problems out of the box.
```

examples/unfolded/demo_ram.py
```diff
@@ -0,0 +1,217 @@
...
+# For instance, it is not trained on image demosaicing. Applying it to a demosaicing problem will yield poor results,
```
Andrewwango:
```suggestion
# For instance, it is not trained on image demosaicing. Applying it to a demosaicing problem out-of-the-box will yield poor results,
```

examples/unfolded/demo_ram.py
```diff
@@ -0,0 +1,217 @@
...
+
+train_dataloader = torch.utils.data.DataLoader(dataset, batch_size=1, shuffle=True)
+
+# %%
+# In order to check the performance of the fine-tuned model, we will use a validation set.
+# We will use a small patch of another image. Note that this validation is also performed in an unsupervised manner,
+# so we will not use the ground truth validation image.
+y_val = physics_train(dinv.utils.load_example("leaves.png")[..., :64, :64].to(device))
+eval_dataloader = torch.utils.data.DataLoader(
+    UnsupDataset(y_val), batch_size=1, shuffle=True
```
Andrewwango: Remove `batch_size=1` and `shuffle=True` for simplicity

examples/unfolded/demo_ram.py
```diff
@@ -0,0 +1,217 @@
...
+mc_loss = dinv.loss.R2RLoss()
+
+t = dinv.transform.Shift(shift_max=0.4)
+eq_loss = dinv.loss.EILoss(t, weight=0.1)
+
+losses = [mc_loss, eq_loss]
+
+dataset = UnsupDataset(y_train)
+
+train_dataloader = torch.utils.data.DataLoader(dataset, batch_size=1, shuffle=True)
```
Andrewwango: Remove `batch_size=1` and `shuffle=True` for simplicity

examples/unfolded/demo_ram.py
```diff
@@ -0,0 +1,217 @@
...
+x_train = x[..., :64, :64]  # take a small patch of the image
+y_train = physics_train(x_train)
+
+
+mc_loss = dinv.loss.R2RLoss()
+
+t = dinv.transform.Shift(shift_max=0.4)
+eq_loss = dinv.loss.EILoss(t, weight=0.1)
+
+losses = [mc_loss, eq_loss]
```
Andrewwango: The loss variable names aren't immediately obvious, but it's not the focus of this example anyway. I'd just combine them to
```python
losses = [
    dinv.loss.R2RLoss(),
    dinv.loss.EILoss(dinv.transform.Shift(shift_max=0.4), weight=0.1)
]
```

deepinv/models/ram.py
```diff
@@ -0,0 +1,1056 @@
...
+            )
+
+        x_temp = physics.A_adjoint(y)
+        pad = (-x_temp.size(-2) % 8, -x_temp.size(-1) % 8)
+        physics = PhysicsCropper(physics, pad)
+
+        x_in = physics.A_adjoint(y)
+
+        sigma = (
+            physics.noise_model.sigma if hasattr(physics.noise_model, "sigma") else 1e-3
```
Andrewwango:
```suggestion
            getattr(physics.noise_model, "sigma", 1e-3)
```

deepinv/models/ram.py
```diff
@@ -0,0 +1,1056 @@
...
+        )
+        sigma = self._handle_sigma(max(sigma, self.sigma_threshold))
+
+        gain = (
+            physics.noise_model.gain if hasattr(physics.noise_model, "gain") else 1e-3
```
Andrewwango:
```suggestion
            getattr(physics.noise_model, "gain", 1e-3)
```

deepinv/models/ram.py
```diff
@@ -0,0 +1,1056 @@
...
+        x_temp = physics.A_adjoint(y)
+        pad = (-x_temp.size(-2) % 8, -x_temp.size(-1) % 8)
+        physics = PhysicsCropper(physics, pad)
+
+        x_in = physics.A_adjoint(y)
+
+        sigma = (
+            physics.noise_model.sigma if hasattr(physics.noise_model, "sigma") else 1e-3
+        )
+        sigma = self._handle_sigma(max(sigma, self.sigma_threshold))
```
Andrewwango: What is `_handle_sigma`? I can't find it anywhere

deepinv/models/ram.py
Andrewwango: The `1e-3` is redundant as it will always be set to `sigma_threshold` anyway, assuming that `sigma_threshold` is always greater than 1e-3. What do you think about changing this to 0 here so users aren't wondering why there is a 1e-3 constant here?

deepinv/models/ram.py
```diff
@@ -0,0 +1,1056 @@
...
+    :param bool pretrained: If True, the model will be initialized with pretrained weights.
+    :param float sigma_threshold: Threshold (minimum value) for the noise level. Default is 1e-3.
+    """
+
+    def __init__(
+        self,
+        in_channels=[1, 2, 3],
+        device=None,
+        pretrained=True,
+        sigma_threshold=1e-3,
```
Andrewwango: Why not `5e-3` as discussed offline?

deepinv/models/ram.py
```diff
@@ -0,0 +1,1056 @@
...
+
+        if device is not None:
+            self.to(device)
+
+    def constant2map(self, value, x):
+        r"""
+        Converts a constant value to a map of the same size as the input tensor x.
+
+        :param float value: constant value
+        :param torch.Tensor x: input tensor
```
Andrewwango, line 120: return doc + type please!

deepinv/models/ram.py
```diff
@@ -0,0 +1,1056 @@
...
+                    (x.size(0), 1, x.size(2), x.size(3)), device=x.device
+                ) * value[None, None, None, None].to(x.device)
+        else:
+            value_map = (
+                torch.ones((x.size(0), 1, x.size(2), x.size(3)), device=x.device)
+                * value
+            )
+        return value_map
+
+    def base_conditioning(self, x, sigma, gain):
```
Andrewwango, line 139: docstring here please

deepinv/models/ram.py
```diff
@@ -0,0 +1,1056 @@
...
+        return torch.cat((x, noise_level_map, gain_map), 1)
+
+    def realign_input(self, x, physics, y):
+        r"""
+        Realign the input x based on the measurements y and the physics model.
+        Applies the proximity operator of the L2 norm with respect to the physics model.
+
+        :param torch.Tensor x: Input tensor
+        :param deepinv.physics.Physics physics: Physics model
+        :param torch.Tensor y: Measurements
```
Andrewwango, line 159: return value + type here please

deepinv/models/ram.py
```diff
@@ -0,0 +1,1056 @@
...
+        :param torch.Tensor x: input tensor
+        """
+        if isinstance(value, torch.Tensor):
+            if value.ndim > 0:
+                value_map = value.view(x.size(0), 1, 1, 1)
+                value_map = value_map.expand(-1, 1, x.size(2), x.size(3))
+            else:
+                value_map = torch.ones(
+                    (x.size(0), 1, x.size(2), x.size(3)), device=x.device
+                ) * value[None, None, None, None].to(x.device)
```
Andrewwango: Am I right in thinking this is taken from original drunet code? This feels like it could all be simplified with `expand_as` or `einops.rearrange`

deepinv/models/ram.py
```diff
@@ -0,0 +1,1056 @@
...
+                device=y.device,
+            )
+
+        x_temp = physics.A_adjoint(y)
+        pad = (-x_temp.size(-2) % 8, -x_temp.size(-1) % 8)
+        physics = PhysicsCropper(physics, pad)
+
+        x_in = physics.A_adjoint(y)
+
+        sigma = (
```
Andrewwango: why does `realign_input` need all the `base.base...` to check for sigma, whereas here we're only looking at `physics.noise_model.sigma`?

deepinv/models/ram.py
```diff
@@ -0,0 +1,1056 @@
...
+
+    def forward(self, y, physics=None, sigma=None, gain=None):
+        r"""
+        Reconstructs a signal estimate from measurements y
+
+        :param torch.Tensor y: measurements
+        :param deepinv.physics.Physics physics: forward operator
+        :return: torch.Tensor: reconstructed signal estimate
+        """
+        assert (
```
Andrewwango: raise ValueError instead

deepinv/models/ram.py
```diff
@@ -0,0 +1,1056 @@
...
+                    out_channels,
+                    bias=bias,
+                    img_channels=img_channels,
+                    decode_upscale=decode_upscale,
+                )
+                for _ in range(nb)
+            ]
+        )
+
+    def forward(self, x, physics=None, y=None, img_channels=None, scale=0):
```
Andrewwango, line 367: docstring, mostly because there are new parameters like `scale`

deepinv/models/ram.py
```diff
@@ -0,0 +1,1056 @@
...
+    :param torch.Tensor v: Precomputed values to subtract from Krylov sequence. Defaults to None.
+    :param int N: Number of Krylov iterations. Defaults to 4.
+    :param torch.Tensor x_init: Initial guess. Defaults to None.
+    :return: torch.Tensor: a stacked tensor over the channel dimension containing Krylov embeddings.
+    """
+
+    if x_init is None:
+        x = p.A_adjoint(y)
+    else:
+        x = x_init.clone()  # Extract the first img_channels
```
Andrewwango:
```suggestion
        x = x_init.clone()
```

deepinv/tests/test_models.py
```diff
@@ -800,6 +851,121 @@ def __len__(self):
...
+        x_hat = model(y, physics)
+
+    assert x_hat.shape == x.shape
+
+
+@pytest.mark.parametrize("whsize", LIST_IMAGE_WHSIZE)
+@pytest.mark.parametrize("model_name", REST_MODEL_LIST)
+@pytest.mark.parametrize("physics_name", LINEAR_OPERATORS)
+@pytest.mark.parametrize("channels", CHANNELS)
+def test_restoration_model_performance(
```
Andrewwango: What makes this test fundamentally different from the previous test such that you can't merge them together, and have the performance check after the shapes test? There is a lot of repeated code

deepinv/tests/test_models.py
```diff
@@ -800,6 +851,121 @@ def __len__(self):
...
+
+@pytest.mark.parametrize("whsize", LIST_IMAGE_WHSIZE)
+@pytest.mark.parametrize("model_name", REST_MODEL_LIST)
+@pytest.mark.parametrize("physics_name", LINEAR_OPERATORS)
+@pytest.mark.parametrize("channels", CHANNELS)
+def test_restoration_model_sizes(
+    device, model_name, physics_name, channels, rng, whsize
+):
+
+    # skip test if channel is 1 and physics_name is in ["demosaicing", "MRI"]
```
Andrewwango: These comments are a bit unnecessary

deepinv/tests/test_models.py
```diff
@@ -800,6 +851,121 @@ def __len__(self):
...
+    y = physics(x)
+
+    with torch.no_grad():
+        x_hat = model(y, physics)
+
+    psnr_fn = PSNR(max_pixel=1)
+
+    if (
+        not (physics_name == "super_resolution_circular" and channels == 2)
+        and model_name == "ram"
```
Andrewwango, line 937: Don't need this if you already `pytest.skip` if not RAM above?

deepinv/tests/test_models.py
```diff
@@ -800,6 +851,121 @@ def __len__(self):
...
+    ):  # suboptimal performance in this case
+        psnr_in = psnr_fn(physics.A_dagger(y), x)
+        psnr_out = psnr_fn(x_hat, x)
+        assert torch.all(psnr_out > psnr_in)
```
Andrewwango, line 943: Nice!

deepinv/tests/test_models.py
Andrewwango: This should be addressed in #622

docs/source/api/deepinv.physics.rst
```diff
@@ -18,6 +18,9 @@ Base Classes
    deepinv.physics.DecomposablePhysics
    deepinv.physics.StackedPhysics
    deepinv.physics.StackedLinearPhysics
+   deepinv.physics.PhysicsMultiScaler
```
Andrewwango, line 21: Add `physics.Upsampling` to API too

docs/source/user_guide/physics/physics.rst
```diff
@@ -110,6 +110,28 @@ This is particular useful when dealing with blind inverse problems or parameter
        | :func:`generate_shifts <deepinv.physics.phase_retrieval.generate_shifts>`
 
 
+.. _wrapper_list:
+
```
Andrewwango, line 116: Upsampling physics to user guide too

deepinv/physics/multiscale.py
```diff
@@ -0,0 +1,146 @@
...
+    Cropping for linear physics operators.
+
+    Given a linear physics operator :math:`A`, this operator instantiates a new operator :math:`\tilde{A} = A \circ C` where :math:`C` is a cropping operator that crops the input tensor.
+    The adjoint operator is defined as :math:`\tilde{A}^{\top} = C^{\top} \circ A^{\top}` and :math:`C^{\top}` is a padding operator that pads the input tensor to the original size.
+
+    :param deepinv.physics.LinearPhysics physics: base linear physics operator.
+    :param tuple pad: padding to apply to the input tensor, e.g., (pad_height, pad_width).
+    """
+
+    def __init__(self, physics, crop):
```
Andrewwango: param says crop but doc says pad

deepinv/physics/multiscale.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def A(self, x):
+        return self.base.A(x[..., self.crop[0] :, self.crop[1] :])
+
+    def A_adjoint(self, y):
+        y = self.base.A_adjoint(y)
+        y = torch.nn.functional.pad(y, (self.crop[1], 0, self.crop[0], 0))
+        return y
+
+    def remove_pad(self, x):
```
Andrewwango: can `A` use `remove_pad`

deepinv/tests/test_physics.py
```diff
@@ -560,6 +619,114 @@ def test_operators_adjointness(name, device, rng):
...
+    """
+
+    list_ops = [
+        "down_resolution_circular",
+        "down_resolution_reflect",
+        "down_resolution_replicate",
+        "down_resolution_constant",
+    ]
+
+    for kernel in ["bilinear", "bicubic", "sinc", "gaussian"]:
```
Andrewwango: mark parametrize

deepinv/tests/test_physics.py
```diff
@@ -560,6 +619,114 @@ def test_operators_adjointness(name, device, rng):
...
+        for name in list_ops:
```
Andrewwango: mark parametrize

deepinv/physics/multiscale.py
```diff
@@ -0,0 +1,146 @@
+import torch
```
Andrewwango: Maybe call this file `wrapper`?

docs/source/user_guide/physics/physics.rst
```diff
@@ -110,6 +110,28 @@ This is particular useful when dealing with blind inverse problems or parameter
...
+~~~~~~~~~
+Wrappers are operators that can be used to adapt existing operators to a new problem.
+
+.. list-table:: Wrappers
+    :header-rows: 1
+
+    * - **Family**
+      - **Operators**
+
+    * - Multicale
```
Andrewwango:
```suggestion
    * - Multiscale
```

docs/source/user_guide/physics/physics.rst
```diff
@@ -110,6 +110,28 @@ This is particular useful when dealing with blind inverse problems or parameter
...
+      -
+         | :class:`deepinv.physics.PhysicsMultiScaler`
+         | :class:`deepinv.physics.LinearPhysicsMultiScaler`
+
+    * - Padding
```
Andrewwango:
```suggestion
    * - Padding/Cropping
```

deepinv/physics/multiscale.py
```diff
@@ -0,0 +1,146 @@
...
+    This class applies a physics model at a given scale
+    by upsampling the input signal before applying the base physics operator.
+
+    .. math::
+
+        A(x) = A_{base}(U_{scale}(x))
+
+    where :math:`U_{scale}` is the upsampling operator for the given scale and :math:`A_{base}` is the base physics operator.
+
+    By default, we assume that the factors for the different scales are [2, 4, 8]. The 0th scale corresponds to no upsampling,
```
Andrewwango: This paragraph is a bit confusing. If the 0th scale corresponds to no upsampling, why is `factors[0] == 2`?

Andrewwango (changes requested):
Great progress! some more changes (sorry...)

deepinv/models/ram.py
matthieutrs: It's in the base denoiser class, allowing to handle types / shapes of sigma!

deepinv/models/ram.py
matthieutrs: good point

deepinv/models/ram.py
matthieutrs: good point, I've also removed it as an init param

deepinv/tests/test_models.py
matthieutrs: Yes you are right. My main reason is that I wanted to skip tests for the models that are not tested rather than just writing "if" conditions for better tracking of undone tests. I still believe that it's best as I've proposed here, although when we do not have these skips we can merge the two tests.

deepinv/models/ram.py
```diff
@@ -0,0 +1,1090 @@
...
+            raise ValueError(
+                "Either physics, sigma or gain must be provided to the RAM model."
+            )
+
+        if physics is None:
+            gain = 1e-3 if gain is None else gain
+            sigma = self.sigma_threshold if sigma is None else sigma
+
+            physics = dinv.physics.Denoising(
+                noise_model=dinv.physics.PoissonGaussianNoise(sigma=sigma, gain=gain),
```
tachella, line 248: is this mapping sigma and gain to the right device?

deepinv/models/ram.py
```diff
@@ -0,0 +1,1090 @@
...
+        :param float, torch.Tensor gain: Poisson noise level
+        :return: torch.Tensor: reconstructed signal estimate
+        """
+        if physics is None and sigma is None and gain is None:
+            raise ValueError(
+                "Either physics, sigma or gain must be provided to the RAM model."
+            )
+
+        if physics is None:
+            gain = 1e-3 if gain is None else gain
```
tachella: this seems very hardcoded, shouldn't be proportional to the norm of y?

deepinv/physics/multiscale.py
```diff
@@ -0,0 +1,146 @@
...
+    :param tuple pad: padding to apply to the input tensor, e.g., (pad_height, pad_width).
+    """
+
+    def __init__(self, physics, crop):
+        super().__init__(noise_model=physics.noise_model)
+        self.base = physics
+        self.crop = crop
+
+    def A(self, x):
+        return self.base.A(x[..., self.crop[0] :, self.crop[1] :])
```
Andrewwango: This should use `self.remove_pad(x)`

deepinv/physics/multiscale.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def A_adjoint(self, y):
+        y = self.base.A_adjoint(y)
+        y = torch.nn.functional.pad(y, (self.crop[1], 0, self.crop[0], 0))
```
Andrewwango: Cleaner if you move the padding out to a function called `def pad(self, y)`

deepinv/models/ram.py
matthieutrs: this works on gpu on my side yes

examples/unfolded/demo_ram.py
```diff
@@ -0,0 +1,215 @@
...
+physics_train = dinv.physics.Demosaicing(
+    img_size=(3, 64, 64),
+    noise_model=dinv.physics.PoissonNoise(0.1, clip_positive=True),
+    device=device,
+)
+x_train = x[..., :64, :64]  # take a small patch of the image
+y_train = physics_train(x_train)
+
+
+mc_loss = dinv.loss.R2RLoss()
```
Andrewwango: remove this line

deepinv/models/ram.py
```diff
@@ -0,0 +1,1064 @@
...
+
+        if physics is None:
+            gain = 1e-3 * y.abs().max() if gain is None else gain
+            sigma = self.sigma_threshold if sigma is None else sigma
+
+            physics = dinv.physics.Denoising(
+                noise_model=dinv.physics.PoissonGaussianNoise(sigma=sigma, gain=gain),
+                device=y.device,
+            )
+        else:
```
Andrewwango: There is no fallback if `physics = None`, `sigma = gain = None`, `physics.noise_model` does not exist. I think this should be 
```suggestion
        elif hasattr(physics, "noise_model"):
            ...
        else:
            # some fallback e.g.use the same sigma and gain code from above when `physics is None`?
```

deepinv/models/ram.py
```diff
@@ -0,0 +1,1064 @@
...
+                "Either physics, sigma or gain must be provided to the RAM model."
+            )
+
+        if physics is None:
+            gain = 1e-3 * y.abs().max() if gain is None else gain
+            sigma = self.sigma_threshold if sigma is None else sigma
+
+            physics = dinv.physics.Denoising(
+                noise_model=dinv.physics.PoissonGaussianNoise(sigma=sigma, gain=gain),
+                device=y.device,
```
Andrewwango: physics doesn't have a device param, remove this

deepinv/models/ram.py
```diff
@@ -0,0 +1,1064 @@
...
+            )
+        else:
+            if hasattr(physics, "noise_model"):
+                sigma = getattr(physics.noise_model, "sigma", self.sigma_threshold)
```
Andrewwango: Questions:
1. What happens when `physics` doesn't have a noise model but `physics.base` has a `noise_model`?
2. Conceptually, what happens in e.g. Pansharpening when there are two sub-physics with different noise models?

Andrewwango (changes requested):
Thanks Matthieu! hopefully nearing the end, some comments and questions:

deepinv/models/ram.py
```diff
@@ -0,0 +1,1083 @@
...
+from deepinv.models.base import Reconstructor, Denoiser
+
+
+class RAM(Reconstructor, Denoiser):
+    r"""
+    Reconstruct Anything Model.
+
+    This model (proposed in `this paper <https://arxiv.org/abs/2503.08915>`_) is a convolutional neural network (CNN)
+    designed for image reconstruction tasks.
+
```
tachella, line 36: add a note saying that the operator should be normalized (ie norm =1 ) for best results

deepinv/models/ram.py
```diff
@@ -0,0 +1,1083 @@
...
+    def threshold_snr(self, val, y, physics, threshold=1e-2, eps=1e-6):
+        r"""
+        Performs the operation
+
+        .. math::
+            \text{threshold\_snr}(x) = \max(\frac{\text{val}}{\|A^\top y\|/\sqrt{m} + \epsilon}, \text{threshold}) * \|A^\top y\|/\sqrt{m}
+
+        """
+        Aty = physics.A_adjoint(y)
```
tachella: we can use `x_temp` to avoid a useless A_adjoint call

deepinv/models/ram.py
tachella, line 36: Explain that the model can be used both as denoiser `model(y,sigma)` or as reconstructor `model(y,physics)`

deepinv/models/ram.py
```diff
@@ -0,0 +1,1083 @@
...
+from deepinv.physics import LinearPhysicsMultiScaler, PhysicsCropper
+from deepinv.utils.tensorlist import TensorList
+from deepinv.models.base import Reconstructor, Denoiser
+
+
+class RAM(Reconstructor, Denoiser):
+    r"""
+    Reconstruct Anything Model.
+
+    This model (proposed in `this paper <https://arxiv.org/abs/2503.08915>`_) is a convolutional neural network (CNN)
```
tachella:
```suggestion
    Convolutional neural network model :footcite:t:`terris2025ram` that has been trained to work on a large variety of linear image reconstruction tasks (deblurring, inpainting, denoising, tomography, MRI, etc.).
    
    The model takes as input a physics operator (optionally set to the identity for denoising), a gaussian noise level (optionally set to 0), and a Poisson noise level (optionally set to 0).
    
```

tachella: also add `terris2025ram` arxiv ref to the refs.bib file

tachella (changes requested):
some more changes, mostly on docs and efficiency

deepinv/models/ram.py
tachella: and that can be checked using physics.compute_norm() function (add a link to the function)

Andrewwango (changes requested):
Adding a `Request changes` to fix performance tests discussed offline.

Andrewwango:
I've added the RAM example directly in [e89ffaa](https://github.com/deepinv/deepinv/pull/622/commits/e89ffaa1d8c97047f419f79971808052e3619ad7) in my PR #622 and made some changes, so I'm happy to merge this first

deepinv/models/ram.py
Andrewwango: done

deepinv/models/ram.py
Andrewwango: done

Andrewwango: done

deepinv/models/ram.py
Andrewwango: done

deepinv/models/ram.py
```diff
@@ -0,0 +1,1111 @@
...
+    def threshold_snr(self, Aty, val, threshold=1e-2, eps=1e-6):
+        r"""
+        Performs the operation
+
+        .. math::
+            \text{threshold\_snr}(x) = \max(\frac{\text{val}}{\|A^\top y\|/\sqrt{m} + \epsilon}, \text{threshold}) * \|A^\top y\|/\sqrt{m}
+
+        """
+        # num = Aty.pow(2).mean(dim=tuple(range(1, Aty.ndim))).sqrt()
+        num = torch.tensor(1.0)  # now assuming that range is in (0, 1)
```
Andrewwango:
```suggestion
        num = torch.tensor(1.0, device=Aty.device)  # now assuming that range is in (0, 1)
```

deepinv/models/ram.py
```diff
@@ -0,0 +1,1130 @@
...
+        :param float gamma: Poisson noise gain
+        :param deepinv.physics.Physics physics: physics measurement operator
+        :param torch.Tensor y: measurements
+        """
+        img_channels = x0.shape[1]
+        physics = LinearPhysicsMultiScaler(physics, x0.shape[-3:], device=x0.device)
+
+        if self.separate_head and img_channels not in self.in_channels:
+            raise ValueError(
+                f"Input image has {img_channels} channels, but the network only have heads for {self.in_channels} channels."
```
Andrewwango:
```suggestion
                f"Input image has {img_channels} channels, but the network only has heads for {self.in_channels} channels."
```

Issue #532, opened 2025-06-13
Physics clone


Andrewwango:
It would be good to have a `physics.clone` that deepcopies a physics object along with attributes (buffers, parameters), but references new memory. And a test to make sure these are copied into new memory.

PR #534, opened 2025-06-13
Add Physics.clone

jscanvic:
Implements #532

The class `Physics` is a subclass of the class `torch.nn.Module` and its instances might hold tensors as (possibly nested) parameters and buffers. The method `clone` should make a deep copy of the physics instance and reallocate all of the tensors at a separate memory location.

**Implementation**

It mostly suffices to use `copy.deepcopy` since `torch.Tensor` implements the special method `__deepcopy__` which internally calls `torch.Tensor.clone` [1].

The only difficulty is that physics instances might hold instances of `torch.Generator` which cannot be copied using `copy.deepcopy` [2]. For this reason, we do the copying of torch generators manually using the parameter `memo` of `copy.deepcopy`. The torch generators are located by traversing the object hierarchy of the physics instance.

[1] https://github.com/pytorch/pytorch/blob/b6add8c8ba927a9f1687134ec2734b02a6ccb3c0/torch/_tensor.py#L161C1-L161C42
[2] https://github.com/pytorch/pytorch/issues/43672


deepinv/physics/forward.py
```diff
@@ -244,6 +248,56 @@ def update(self, **kwargs):
         if hasattr(self.noise_model, "update_parameters"):
             self.noise_model.update_parameters(**kwargs)
 
+    # NOTE: Physics instances can hold instances of torch.Generator as
```
Andrewwango, line 250: Worth putting this inside the docstring?

deepinv/physics/forward.py
```diff
@@ -244,6 +248,56 @@ def update(self, **kwargs):
...
+    # (possibly nested) attributes and they cannot be copied using deepcopy
+    # natively. For this reason, we manually copy them beforehand and populate
+    # the copies using the memo parameter of deepcopy. For more details, see:
+    # https://github.com/pytorch/pytorch/issues/43672
+    # https://github.com/pytorch/pytorch/pull/49840
+    # https://discuss.pytorch.org/t/deepcopy-typeerror-cant-pickle-torch-c-generator-objects/104464
+    def clone(self):
+        r"""
+        Clone the forward operator.
```
Andrewwango: ... by performing deepcopy to copy all attributes to new memory.

deepinv/tests/test_physics.py
```diff
@@ -1317,3 +1317,40 @@ def test_adjoint_autograd(name, device):
...
+    assert parameter_names == parameter_names_clone, "Parameter names do not match."
+
+    for name in parameter_names.intersection(parameter_names_clone):
+        param = physics.get_parameter(name)
+        param_clone = physics_clone.get_parameter(name)
+
+        # Check that buffers have been reallocated somewhere else in the memory space
+        assert (
+            param.data_ptr() != param_clone.data_ptr()
+        ), f"Parameter {name} has not been cloned properly."
```
Andrewwango, line 1658: nice! can you also do simple checks:
- Before `physics.clone()`: `if hasattr(physics.mask): physics.mask += 7` ; Here: `if hasattr(physics.mask): assert physics.mask == physics_clone.mask`
- say modify `physics_clone.A` to `lambda *args: return "hi"` and check that `physics.A` doesn't return `"hi"`

deepinv/physics/forward.py
```diff
@@ -244,6 +248,56 @@ def update(self, **kwargs):
...
+        """
+        memo = {}
+
+        # Traverse the graph of nested attributes of the forward operator
+        traversal_queue = [self]
+        seen = set()
+        while traversal_queue:
+            # 1. Get the next node to process
+            node = traversal_queue.pop()
```
Andrewwango, line 272: I think I'm gonna have to trust you on this one... if it works it works!

deepinv/tests/test_physics.py
```diff
@@ -1317,3 +1317,40 @@ def test_adjoint_autograd(name, device):
...
+
```
Andrewwango, line 1659: Maybe add a test that the new `torch.Generator` has the same state as the old one?

Andrewwango (changes requested):
nice so far!

deepinv/tests/test_physics.py
```diff
@@ -1317,3 +1317,40 @@ def test_adjoint_autograd(name, device):
     delta_y = y.grad
     Az = physics.A(z)
     assert torch.allclose(delta_y, Az, rtol=1e-5)
+
+
+@pytest.mark.parametrize("name", OPERATORS)
+def test_clone(name, device):
+    physics, imsize, _, dtype = find_operator(name, device)
+
+    physics_clone = physics.clone()
```
Andrewwango, line 1640: also the simplest `assert physics.__name__ == physics_clone.__name__` or `isinstance(physics_clone, physics.__class__)`

deepinv/tests/test_physics.py
Andrewwango: and that they evolve together - i.e. they are not entangled in any way

Andrewwango (changes requested):
Sorry another comment!
Also please can you replace where we currently use deepcopy in the codebase?

e.g. https://github.com/search?q=repo%3Adeepinv%2Fdeepinv%20deepcopy&type=code

deepinv/physics/forward.py
jscanvic: It's very low-level stuff and I wish we could avoid it but I couldn't find a better way! deepcopy uses pickle under the hood but it doesn't expose it so we can't filter generators there directly. Instead it exposes the memory cache so that we can make our own copies but we still have to find the objects in the hierarchy. By hierarchy I mean both normal attributes & values for mappings used in e.g. submodules (saved as values in the dict Module._modules). I couldn't find a standard traversal function so I had to implement it myself and it should work as long as we don't put generators in very exotic places.

jscanvic:
Thanks for the review! Should be good now

deepinv/tests/test_physics.py
```diff
@@ -1317,3 +1317,99 @@ def test_adjoint_autograd(name, device):
...
+    # Save original values
+    saved_A = physics.A
+    physics.A = lambda *args, **kwargs: "hi"
+
+    x = torch.randn(imsize, device=device, dtype=dtype).unsqueeze(0)
+    assert physics.A(x) == "hi"
+    assert physics_clone.A(x) != "hi"
+
+    # Restore original values
+    physics.A = saved_A
```
mh-nguyen712, line 1757: Thanks for the hard works! Can you also test the forward of the copied physics, with `update_parameters` to check that the it also works? I think the best way it's to add a copied physics in the list OPERATOR so that all physics tests will also be applied on the copied one.

mh-nguyen712: Espectially the case where a buffer is required grad. Does the copied instance also requires grad? Do we need to do as in torch: `.detach().clone()`?

jscanvic:
Thanks the for review @mh-nguyen712! I'll make the changes asap.

I think deepcopy already does what we want wrt autograd. Here's my understanding of it:

* requires_grad is an object attribute so it gets copied with deepcopy
* Autograd works with functions that save a reference to their inputs (e.g., see [1]) so when deepcopy is used the original tensors remain in the direct acyclic graph (DAG) 
* In `__deepcopy__` [2] the original tensor is cloned in a `no_grad` context so they (the copies) end up detached automatically

[1] https://docs.pytorch.org/tutorials/beginner/examples_autograd/two_layer_net_custom_function.html
[2] https://github.com/pytorch/pytorch/blob/c83041cac2e9ca537b210b74876dbcee5e08ae9a/torch/_tensor.py#L145

jscanvic:
I've added checks for `Tensor.requires_grad` and for `Tensor.grad` - is that what you had in mind? @mh-nguyen712

deepinv/tests/test_physics.py
jscanvic: I've added `inpainting_clone` which does that!

mh-nguyen712:
Thanks for adding the tests @jscanvic . LGTM now!

mh-nguyen712 (approved):

jscanvic:
Thanks @mh-nguyen712! 

Some of the tests were not hit because no physics has parameters (at least not the ones that are tested) so I added a dummy parameter at the start to make them run.

mh-nguyen712:
Currently physics parameters are only registered as buffers, not parameters so they only appear in `physics.buffers()` or `physics.named_buffers()` so your check in `physics.parameters()` will not be hit. We prefer buffers over parameters because they can hold no_grad tensors. But as you tested both cases, it should be good I think.

deepinv/tests/test_physics.py
```diff
@@ -1616,3 +1621,227 @@ def test_adjoint_autograd(name, device):
...
+
+    # Add a dummy parameter used for further testing
+    dummy_tensor = torch.randn(
+        imsize,
+        device=device,
+        dtype=dtype,
+        generator=torch.Generator(device).manual_seed(0),
+    )
+    dummy_parameter = torch.nn.Parameter(dummy_tensor)
+    physics.register_parameter("dummy", dummy_parameter)
```
Andrewwango, line 1638: Does the copy stuff only work when the parameter is registered as a Parameter? How about just a class attribute like `physics.dummy = dummy_parameter`?

jscanvic: It works for regular attributes as well! See 8b3984b

Andrewwango (approved):
LGTM, thanks for your hard work and thorough tests

PR #550, opened 2025-06-17
Add an error on trainer directory collisions instead of overwriting files silently

### LINKED ISSUE #525: Trainer causes data losses
The trainer is designed so that the training data gets written to a separate directory for each run but two might actually end up sharing one, resulting in data losses due to overwrites.

**Cause**

The output directory is determined by the current timestamp which provides no uniqueness guarantee when training two models in parallel.

https://github.com/deepinv/deepinv/blob/9297caccd0f4f93d4d13b8af797dbb8909748c37/deepinv/training/trainer.py#L338-L340

Even though the bug might seem unlikely it did in fact hit me. Moreover, I suspect it is actually very likely to happen when starting multiple trainings a few seconds apart due to caching and resource sharing in high-performance computing systems.

**Fix**

An exception should at the very least be raised when trying to acquire ownership of an already-owned directory. In the current implementation, ownership acquisition is creating the not-already-created directory. The fix would amount to passing `exist_ok=False` to `os.makedirs` the first time it is called and `exist_ok=True` the subsequent times:

https://github.com/deepinv/deepinv/blob/9297caccd0f4f93d4d13b8af797dbb8909748c37/deepinv/training/trainer.py#L904-L904

This alone would get rid of the data losses but it would result in unnecessary crashes. A better solution would involve changing the name until ownership acquisition succeeds, e.g., by appending a count or a random string to the directory name.

jscanvic:
Fixes #525

This PR is a quick fix to prevent data losses due to overwriting until we decide how best to deal with collisions (let acquisition fail or add a retry strategy).

We let the training fail when the output directory already exists before training starts thereby preventing concurrent training writing (and overwriting) files in the same directory.

The other tests are currently prone to collisions and fail with this fix so I change them so they use unique directories.

**Reproducing**

```python
import deepinv as dinv
import torch
from pathlib import Path
from deepinv.utils.demo import load_dataset

from unittest.mock import patch
import math
import threading
import copy
import os

physics = dinv.physics.Inpainting(tensor_size=(3, 16, 16), mask=0.5, device="cpu")

train_dataset = dinv.datasets.CBSD68(
    "data/Set14",
    transform=transforms.Compose(
        [
            transforms.ToTensor(),
            transforms.Resize((16, 16)),
        ]
    ),
    download=True,
)
train_dataloader = DataLoader(train_dataset, batch_size=1, shuffle=False)

backbone = dinv.models.UNet(in_channels=3, out_channels=3, scales=2)

losses = [
    dinv.loss.SupLoss(),
]

optimizer = torch.optim.AdamW(model.parameters(), lr=1e-3)

# Increase the probability of collision by running the same code many times
for _ in range(100):
    paths = []

    def train_worker(run_idx: int):
        trainer = dinv.Trainer(
            model=copy.deepcopy(model),    # avoid parameter clashes
            device="cpu",
            save_path="ckps",
            verbose=True,
            show_progress_bar=False,
            physics=physics,
            epochs=2,
            losses=copy.deepcopy(losses),  # avoid shared state
            optimizer=copy.deepcopy(optimizer),
            online_measurements=True,
            check_grad=True,
        )

        trainer.train()
        paths.append(trainer.save_path)

    threads = []
    for i in range(2):                      # spin up exactly two threads
        t = threading.Thread(target=train_worker, args=(i,))
        t.start()
        threads.append(t)

    for t in threads:                       # wait for both to finish
        t.join()

    if len(paths) != len(set(paths)):
        raise RuntimeError(
            f"Collision detected: {len(paths)} paths, {len(set(paths))} unique paths"
        )

```

### DIFF
--- a/deepinv/tests/test_trainer.py
+++ b/deepinv/tests/test_trainer.py
@@ -17,11 +17,12 @@ import builtins
 import io
 import contextlib
 import re
+import tempfile
+import os
 
 from conftest import no_plot
 
 NO_LEARNING = ["A_dagger", "A_adjoint", "prox_l2", "y"]
-
 
 @pytest.fixture
 def imsize():
@@ -40,8 +41,15 @@ def physics(imsize, device):
     return dinv.physics.BlurFFT(img_size=imsize, filter=filter, device=device)
 
 
+@pytest.fixture
+def save_path():
+    # NOTE: The save path should be unique to avoid collision errors.
+    with tempfile.TemporaryDirectory() as dirname:
+        yield dirname
+
+
 @pytest.mark.parametrize("no_learning", NO_LEARNING)
-def test_nolearning(imsize, physics, model, no_learning, device):
+def test_nolearning(imsize, physics, model, no_learning, device, save_path):
     y = torch.ones((1,) + imsize, device=device)
     trainer = dinv.Trainer(
         model=model,
@@ -51,6 +59,7 @@ def test_nolearning(imsize, physics, model, no_learning, device):
         physics=physics,
         compare_no_learning=True,
         no_learning_method=no_learning,
+        save_path=save_path,
     )
     x_hat = trainer.no_learning_inference(y, physics)
     assert (physics.A(x_hat) - y).pow(2).mean() < 0.1
@@ -126,6 +135,7 @@ def test_get_samples(
     use_physics_generator,
     online_measurements,
     rng,
+    save_path,
 ):
     # Dummy constant GT dataset
     class DummyDataset(Dataset):
@@ -217,6 +227,7 @@ def test_get_samples(
             if online_measurements and physics_generator is not None
             else None
         ),
+        save_path=save_path,
     )
 
     iterator = iter(dataloader)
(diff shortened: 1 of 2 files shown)


deepinv/tests/test_trainer.py
```diff
@@ -39,8 +42,15 @@ def physics(imsize, device):
     return dinv.physics.BlurFFT(img_size=imsize, filter=filter, device=device)
 
 
+@pytest.fixture
+def save_path():
```
jscanvic: Use the more standard `tmpdir` fixture featured by pytest:

https://docs.pytest.org/en/latest/reference/reference.html#std-fixture-tmpdir

tachella (approved):
LGTM, thanks Jérémy

I think I never had this issue, since you have to train to models on the same second, but it is definitely nice to have safeguards

jscanvic:
For future reference, here's a workaround I'm using to prevent the trainer from crashing when a collision happens and to make it try a different directory instead: b0a05c0f425708a861dcae426b3a7a9c581909c9

Issue #487, opened 2025-05-12
Increasing test coverage


jscanvic:
The code coverage is currently at 78.35% [1] and while it is not that low, it is not that high either. We arguably don't want to chase a 100% code coverage because of diminishing returns, but there is a clear benefit to having more of the code base tested out (easier contributing & reviewing), and there might be low-hanging fruits. 

As of today, there are 15,674 tracked lines of code, 12,282 of which are covered, leaving 3,392 of them uncovered or only partially covered. This means that in order to attain a coverage of say 85%, we'd have to cover about 30% of remaining lines, or about 1,000 more lines.

I suggest we try to get there.

**How**

- [x] Testing dataset code by stubbing data #490

The `deepinv.datasets` module contains about 400 uncovered lines of code. One of the challenges with testing dataset code is that datasets can't generally be downloaded during CI as it takes time and space. As long as this is the case, we won't be able to meaningfully test the consistency between the actual data and our code, but we can still test the surrounding logic by stubbing in fake data

For instance, in `Kholer`, there is code responsible for computing the correspondence between (linear) indices, and filenames stored on disk. This is code worth testing. In `LidcIdriSliceDataset`, there is code responsible for making indexation deterministic, and there is also code responsible for processing DICOM files.

In general, it is also possible to test out that transforms passed in at construction are correctly applied when calling `__getitem__`.

A lot of the code is also responsible for running integrity checks, those are a bit more difficult to test but there should be ways to do it. Maybe by factoring out most of the code in a separate module responsible for integrity checks, then testing it independently and stubbing it on specific calls.

- [x] Improving coverage of the `deepinv.utils` module #555 

There are 445 uncovered lines of code in the `deepinv.utils` module. Some of the untested code deals with IO, e.g., in `deepinv.utils.demo` but even IO related functions contain some logic that can be tested out by stubbing the IO related functions. Other functions are purely computational and can easily be tested, e.g., code in `deepinv.utils.tensorlist` and `deepinv.utils.phantoms`. Deprecated functions in `deepinv.utils.metrics` can easily be tested out.

3. Other low-hanging fruits #623 

There are cases where code redundancy make for a large number of uncovered lines. This is the case for code in `deepinv.models.UNet` as the forward pass is implemented separately for scales 2, 3, 4 and 5, even though the logic is the same and could be factored in a single algorithm that would depend on the number of scales. In particular, scales 3 and 5 are not tested and amount for almost 50 uncovered lines of code.

[...]

PR #555, opened 2025-06-19
Increase test coverage of `deepinv.utils`

jscanvic:
Towards #487

### DIFF
--- a/deepinv/tests/test_utils.py
+++ b/deepinv/tests/test_utils.py
@@ -3,6 +3,8 @@ import torch
 import pytest
 from deepinv.utils.decorators import _deprecated_alias
 import warnings
+import numpy as np
+from contextlib import nullcontext
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


deepinv/tests/test_utils.py
```diff
@@ -75,14 +119,109 @@ def test_tensordict_append(tensorlist):
...
+
+@pytest.mark.parametrize("C", [1, 3])
+@pytest.mark.parametrize("n_images", range(1, 3))
+@pytest.mark.parametrize("save_plot", [False, True])
+@pytest.mark.parametrize("cbar", [False, True])
+@pytest.mark.parametrize("with_titles", [False, True])
+@pytest.mark.parametrize("dict_img_list", [False, True])
+@pytest.mark.parametrize("suptitle", [None, "dummy_title"])
+def test_plot(
+    tmpdir,
```
jscanvic: Use `tmp_path` instead of the deprecated `tmpdir`:
https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures

deepinv/utils/__init__.py
```diff
@@ -37,3 +37,4 @@
 )
 from .phantoms import RandomPhantomDataset, SheppLoganDataset
 from .patch_extractor import patch_extractor
+from .parameters import get_GSPnP_params
```
tachella, line 40: this was missing?

deepinv/tests/test_utils.py
```diff
@@ -3,6 +3,33 @@
...
+import unittest.mock as mock
+from unittest.mock import patch
+import subprocess
+import os
+import inspect
+import itertools
+import pathlib
+import torchvision
+import torchvision.transforms as transforms
+import PIL
```
tachella, line 20: this library is not in the dependencies of deepinv - is it really necessary?

jscanvic: You mean PIL? We're already using it extensively in `deepinv.datasets` so it shouldn't be a problem right? I don't think I've added any new third-party dependency

deepinv/utils/__init__.py
jscanvic: They're used in one of the demos with the full qualifier `deepinv.utils.parameters` instead of the simpler `deepinv.utils`. I made the change then reverted it (36c0709015caf47f9c1f82dc495a0a71d0c1bd0c) & forgot to get rid of the other one in `__init__py`. I can get rid of it if you think it doesn't make sense!

deepinv/tests/test_utils.py
Andrewwango: I think it is - PIL is a dependency of torchvision, which is a dependency of deepinv. https://pypistats.org/packages/torchvision

deepinv/tests/test_utils.py
```diff
@@ -3,6 +3,33 @@
...
+import io
+import copy
+
+
+@pytest.fixture
+def non_interactive_matplotlib():
```
Andrewwango: Can you unify this with `deepinv/tests/conftest.no_plot`?

deepinv/tests/test_utils.py
```diff
@@ -75,14 +120,156 @@ def test_tensordict_append(tensorlist):
...
-        deepinv.utils.plot({k: v for k, v in zip(titles, imgs)}, show=False)
+# The class TensorList features many utility methods that we do not test in
+# depth but verify that they do not raise any exception when called. To do
+# that, we get a tensor list instance, we iterate over its methods and try to call
+# them filling in the required parameters in the process. We use the name of
+# the parameter to determine what to pass in, e.g., if the parameter is named
+# shape we use the input tensor list shape as the value for that parameter. By
+# default we use the tensor list itself for every optional parameter.
+def test_tensorlist_methods(tensorlist):
+    tensorlist, _ = tensorlist
```
Andrewwango: What does this mean?

jscanvic: The code is a bit confusing because tensorlist is not a TensorList instance, it's a tuple of two TensorList instances. Anyway, I changed this snippet to make it more consistent with the rest of the tests!

deepinv/tests/test_utils.py
jscanvic: Done!

jscanvic:
Thanks for the review! Should be good now.

Issue #539, opened 2025-06-16
`ruff` linter


Andrewwango:
@jscanvic  has added a partial `ruff` CI rule in #501 that checks deprecated types.

It would be nice to have more `ruff` linting to check for other problems, bugs and inefficiencies. The rules should be set to:
- Be in accordance with current deepinv coding conventions
- Switch off formatter so we can keep existing black formatter
- Not apply any changes, just throw warnings
- Be mostly aimed at preventing bugs & measurable inefficiencies (instead of more vague criteria like code cleanness)

Note this should also be orthogonal to type checking in #536

jscanvic:
Agreed!

Linting arguably raises the bar for contributions and we want it to be worthwhile so I'd also add:
* Be mostly aimed at preventing bugs & measurable inefficiencies (instead of more vague criteria like code cleanness)

Andrewwango:
Removing good first issue as it probably isn't

Tmodrzyk:
I could add a few other cases to the linting and take care of the eventual manual fixes. 
Could you assign me to this issue @Andrewwango ?

Andrewwango:
Note #628 which is an issue evidencing why we need a linter

Andrewwango:
Update: help wanted and good first issue! See the draft PR #705

PR #557, opened 2025-06-20
Flag unused imports in CI

jscanvic:
Towards #539

Imports changes in PRs give valuable clues about implementation details right away but only when they're actually relied on. Unused imports additions and deletions convey no meaningful information and make reviewing more tedious. I suggest making CI fail when the code base contains unused imports so contributors have immediate feedback and can get rid of them before reviewing starts - hopefully making reviewing more efficient for everyone involved.

I do it in fedabd615fa9c1ae57d451d92403a703fcbaa145 by ticking [rule F401](https://docs.astral.sh/ruff/rules/unused-import/#unused-import-f401) flagging unused imports.

**Getting rid of unused imports**

There are currently 416 imports flagged as unused by ruff. I get rid of 60 of them in 8ff5657a299c818583a234e180d72937c9134d59 using:

```bash
ruff check --target-version "py39" --fix --select "F401"
``` 

It leads to minor issues with pytest fixtures and imports with side effects. I use `# noqa: F401` to turn off the rule for those.

Most (352) of the remaining ones are imports in `__init__.py` files that are not listed in `__all__`. I fix them (leaving 4 remaining) using the unsafe fix in fe5541e3ea4c505f0b1ff31f94c9acb1dd031785:

```bash
ruff check --target-version "py39" --preview --fix --unsafe-fixes --select "F401"
```

From what I can tell what it does is:
* Get rid of standard and third-party modules (so that we don't re-export them)
* Change first-party imports to be in the redundant form `from foo import bar as bar` (so that ruff knows it's intentionally exported)

I fix the last 4 errors in 8ab2ae5 manually
* By adding `train`, `test` and `Trainer` to `__all__` in the root `__init__.py` 
* By replacing the availability check for `astra` based on try/expect by a call to `importlib.util.find_spec`


deepinv/models/varnet.py
```diff
@@ -12,7 +12,7 @@
 from deepinv.physics.mri import MRIMixin, MRI, MultiCoilMRI
 
 if TYPE_CHECKING:
-    from deepinv.physics.forward import Physics
+    pass
```
Andrewwango: can you remove the TYPE_CHECKING and its import

Andrewwango (changes requested):
Thanks for making these changes @jscanvic ! Everything looks good, some comments:

- Please can you address the conflicts
- Question: doing `from foo import bar as bar` in the `__init__`s might be unnatural for some developers. Is this automatically done by ruff? If not, if a developer forgets to do this (i.e. doing `from foo import bar`) and ruff flags this, is it obvious to developers that this is the fix (`as bar`)?

jscanvic:
Thanks for the review @Andrewwango!

I agree those imports look very wacky but it's well-supported by ruff.

Namely, unsafe fixing does it automatically and it's suggested as a fix in the checker. For instance, if I get rid of one of them and run the checker it's exactly what it tells me to do:

```
$ git diff HEAD
diff --git a/deepinv/loss/__init__.py b/deepinv/loss/__init__.py
index 5acda128..f071a718 100644
--- a/deepinv/loss/__init__.py
+++ b/deepinv/loss/__init__.py
@@ -1,4 +1,4 @@
-from .mc import MCLoss as MCLoss
+from .mc import MCLoss
 from .ei import EILoss as EILoss
 from .moi import MOILoss as MOILoss, MOEILoss as MOEILoss
 from .sup import SupLoss as SupLoss

$ ruff check --target-version "py310" --select "F401"
deepinv/loss/__init__.py:1:17: F401 `.mc.MCLoss` imported but unused; consider removing, adding to `__all__`, or using a redundant alias
  |
1 | from .mc import MCLoss
  |                 ^^^^^^ F401
2 | from .ei import EILoss as EILoss
3 | from .moi import MOILoss as MOILoss, MOEILoss as MOEILoss
  |
  = help: Use an explicit re-export: `MCLoss as MCLoss`

Found 1 error.
```

I don't think it'd make contributing harder given 1) it only matters for `__init__.py` files which are not changed that often and 2) the explanation and the fix would be in the failure logs anyway.

jscanvic:
Actually I think there's a better solution - I couldn't make it work at first but I think I found a way.

I suggest to revert back to normal imports in `__init__.py` files and simply disable ruff's unused import rule for those.

The rule barely brings any value for those files anyway because we don't have clear module boundaries.

I should be able to make it work using this setting:

https://docs.astral.sh/ruff/settings/#lint_per-file-ignores

@Andrewwango What do you think?

Andrewwango:
Thanks for looking into it @jscanvic ! 

> Namely, unsafe fixing does it automatically and it's suggested as a fix in the checker

I agree, the checker message is very clear, so I think if we go for it, it wouldn't be a problem.

> Actually I think there's a better solution - I couldn't make it work at first but I think I found a way.

I don't mind this either, it would save some minor initial inconvenience for in making sure all PRs start using `x as x` but if it's lots more work to revert the `__init__` changes, then don't worry!

jscanvic:
I made the change - it should be good for merging now!

PR #561, opened 2025-06-24
Add new noise datafidelities for diffusion-based posterior sampling 

samuro95:
- Change the default VP parameters to the Song default ones for efficient VP 

TOCOMPLETE

### DIFF
--- a/deepinv/sampling/diffusion_sde.py
+++ b/deepinv/sampling/diffusion_sde.py
@@ -347,8 +347,8 @@ class VariancePreservingDiffusion(DiffusionSDE):
     def __init__(
         self,
         denoiser: Denoiser = None,
-        beta_min: float = 0.0001,
-        beta_max: float = 5.0,
+        beta_min: float = 0.1,
+        beta_max: float = 20,
         alpha: float = 1.0,
         solver: BaseSDESolver = None,
         dtype=torch.float64,


deepinv/models/ncsnpp.py
```diff
@@ -88,10 +90,23 @@ def __init__(
         pixel_std: float = 0.75,
         device=None,
     ):
+        assert model_name in ["ncsn", "ddpm"]
```
mh-nguyen712:
```suggestion
        assert model_name.lower() in ["ncsn", "ddpm"], "'model_name' should be 'ncsn' or 'ddpm'."
```

deepinv/sampling/diffusion_sde.py
```diff
@@ -10,24 +10,32 @@
...
 from deepinv.sampling.noisy_datafidelity import NoisyDataFidelity
 from copy import deepcopy
+from deepinv.sampling.utils import trapz_torch
 
 
 class _WrapperDenoiserMinusOneOne(nn.Module):
     """
-    A wrapper for denoisers trained on [0, 1] images to be used with [-1, 1] images.
-    This is important for diffusion sampling
+    A wrapper for denoisers trained on :math:`[x_{min}, x_{max}]` images to be used with math:`[-1, 1]` images, i.e. on diffusion sampling iterates.
```
mh-nguyen712:
```suggestion
    A wrapper for denoisers trained on :math:`[x_{\mathrm{min}}, x_{\mathrm{max}}]` images to be used with math:`[-1, 1]` images, i.e. on diffusion sampling iterates.
```

mh-nguyen712: Nice

deepinv/sampling/diffusion_sde.py
```diff
@@ -239,74 +247,124 @@ def scale_t(self, t: Tensor | float) -> Tensor:
...
-    where :math:`\alpha \in [0,1]` is a constant weighting the diffusion term.
-
-    This class is the reverse-time SDE of the VE-SDE, serving as the generation process.
-
-    :param deepinv.models.Denoiser denoiser: a denoiser used to provide an approximation of the score at time :math:`t` :math:`\nabla \log p_t(x_t)`.
-    :param float sigma_min: the minimum noise level.
-    :param float sigma_max: the maximum noise level.
-    :param float alpha: the weighting factor of the diffusion term.
-    :param deepinv.sampling.BaseSDESolver solver: the solver for solving the SDE.
+        d x_t = \frac{s'(t)}{s(t)} x_t dt + s(t) \sqrt{2 \sigma(t) \sigma'(t)} d w_t
```
mh-nguyen712, line 261: This is different than eq(6) in EDM paper? It's similar to the ODE formulation but I dont know if the SDE is ok

deepinv/sampling/diffusion_sde.py
```diff
@@ -239,74 +247,124 @@ def scale_t(self, t: Tensor | float) -> Tensor:
...
+                    t = self._handle_time_step(t)
+                    return torch.ones_like(t)
+
+                def scale_prime_t(t):
+                    t = self._handle_time_step(t)
+                    return torch.zeros_like(t)
+
+            else:
+                raise ValueError(
+                    "scale_t must be provided if variance_preserving or variance_exploding is False"
```
mh-nguyen712:
```suggestion
                    "'scale_t' must be provided if 'variance_preserving' and 'variance_exploding' is False"
```

deepinv/sampling/diffusion_sde.py
```diff
@@ -239,74 +247,124 @@ def scale_t(self, t: Tensor | float) -> Tensor:
...
+                )
+        self.scale_t = scale_t
+
+        if sigma_prime_t is None:
+
+            def sigma_prime_t(t):
+                t = torch.tensor(self._handle_time_step(t), requires_grad=True)
```
mh-nguyen712:
```suggestion
                t = self._handle_time_step(t).requires_grad_(True)
```

deepinv/sampling/diffusion_sde.py
```diff
@@ -239,74 +247,124 @@ def scale_t(self, t: Tensor | float) -> Tensor:
...
+                s = self.sigma_t(t)
+                s.backward()
+                return t.grad.item()
+
+        self.sigma_prime_t = sigma_prime_t
+
+        if scale_prime_t is None:
+
+            def scale_prime_t(t):
+                t = torch.tensor(self._handle_time_step(t), requires_grad=True)
```
mh-nguyen712:
```suggestion
                t =self._handle_time_step(t).requires_grad_(True)
```

deepinv/sampling/diffusion_sde.py
```diff
@@ -320,12 +378,23 @@ def forward_diffusion(t):
             *kwargs,
         )
 
-        self.sigma_min = sigma_min
-        self.sigma_max = sigma_max
+    def score(self, x: Tensor, t: Union[Tensor, float], *args, **kwargs) -> Tensor:
```
mh-nguyen712:
```suggestion
    def score(self, x: Tensor, t: Tensor | float, *args, **kwargs) -> Tensor:
```

deepinv/sampling/diffusion_sde.py
```diff
@@ -334,26 +403,249 @@ def sample_init(self, shape, rng: torch.Generator) -> Tensor:
...
+
+        if B_t is None:
+
+            def B_t(t: Union[Tensor, float], n_steps: int = 1000) -> Tensor:
+                t = self._handle_time_step(t)
+                return trapz_torch(
+                    beta_t, torch.tensor(0.0, device=t.device), t, n_steps
+                )
+
+        def scale_t(t: Union[Tensor, float]) -> Tensor:
```
mh-nguyen712:
```suggestion
        def scale_t(t: Tensor | float) -> Tensor:
```

deepinv/sampling/diffusion_sde.py
```diff
@@ -334,26 +403,249 @@ def sample_init(self, shape, rng: torch.Generator) -> Tensor:
...
+            return torch.exp(-B_t(t))
+
+        def scale_prime_t(t: Union[Tensor, float]) -> Tensor:
```
mh-nguyen712: same here

deepinv/sampling/diffusion_sde.py
```diff
@@ -334,26 +403,249 @@ def sample_init(self, shape, rng: torch.Generator) -> Tensor:
...
+        def sigma_t(t: Tensor | float) -> Tensor:
+            t = self._handle_time_step(t)
+            if variance_preserving:
+                return (1 / scale_t(t) ** 2 - 1) ** 0.5
+            else:
+
+                def integrand(s: torch.Tensor) -> torch.Tensor:
+                    return xi_t(s) / (scale_t(s) ** 2)
+
+                integral = trapz_torch(
```
mh-nguyen712, line 576: n_steps is not defined?

deepinv/sampling/diffusion_sde.py
samuro95: it is indeed the same with \beta(t) = \alpha s(t)^2 \sigma(t) \sigma'(t)

I added this in comments.

deepinv/models/ncsnpp.py
```diff
@@ -57,6 +58,7 @@ class NCSNpp(Denoiser):
 
     def __init__(
         self,
+        model_name: str = "ncsn",  # Model name, 'ncsn' or 'ddpm'.
```
mh-nguyen712: Do you think "ddpm" is confusing with the DDPM scheduler? And would change this to "model_type". 
Moreover, if we keep this parameter, we should remove all related params such as "embedding_type", "encoder_type", ... as they are defined implicitly from this parameter

deepinv/models/ncsnpp.py
```diff
@@ -222,11 +238,20 @@ def __init__(
                 )
 
         if pretrained is not None:
-            if (
-                pretrained.lower() == "edm-ffhq64-uncond-ve"
-                or pretrained.lower() == "download"
+            if pretrained.lower() == "edm-ffhq64-uncond-ve" or (
+                pretrained.lower() == "download" and model_name == "ncsn"
+            ):
+                name = "edm-ffhq-64x64-uncond-ve.pt"
```
mh-nguyen712: Did you change the checkpoint name on the huggingface hub?

deepinv/sampling/diffusion_sde.py
```diff
@@ -167,13 +175,20 @@ def __init__(
         *args,
         **kwargs,
     ):
+        if not isinstance(alpha, Callable):
+            alpha_value = alpha
+
+            def alpha(t: Tensor | float) -> Tensor:
+                t = self._handle_time_step(t)
+                return alpha_value * torch.ones_like(t)
```
mh-nguyen712:
```suggestion
            def alpha(t: Tensor | float) -> Tensor:
                return alpha_value
```

deepinv/sampling/diffusion_sde.py
```diff
@@ -134,11 +142,11 @@ class DiffusionSDE(BaseSDE):
     This class define the following reverse-time SDE:
 
     .. math::
-        d x_{t} = \left( f(x_t, t) - \frac{1 + \alpha}{2} g(t)^2 \nabla \log p_t(x_t) \right) dt + g(t) \sqrt{\alpha} d w_{t}.
+        d x_{t} = \left( f(x_t, t) - \frac{1 + \alpha(t)}{2} g(t)^2 \nabla \log p_t(x_t) \right) dt + g(t) \sqrt{\alpha(t)} d w_{t}.
 
     :param Callable drift: a time-dependent drift function :math:`f(x, t)` of the forward-time SDE.
     :param Callable diffusion: a time-dependent diffusion function :math:`g(t)` of the forward-time SDE.
-    :param Callable alpha: a scalar weighting the diffusion term. :math:`\alpha = 0` corresponds to the ODE sampling and :math:`\alpha > 0` corresponds to the SDE sampling.
+    :param Callable, float alpha: a (possibly time-dependent) scalar weighting the diffusion term. :math:`\alpha(t) = 0` corresponds to ODE sampling and :math:`\alpha(t) > 0` corresponds to SDE sampling.
```
mh-nguyen712: `alpha(t)` should be between 0 and 1 or anything?

mh-nguyen712:
```suggestion
    :param Callable, float alpha: a (possibly time-dependent) scalar weighting the diffusion term. A  constant function :math:`\alpha(t) = 0` corresponds to ODE sampling and :math:`\alpha(t) > 0` corresponds to SDE sampling.
```

deepinv/sampling/diffusion_sde.py
```diff
@@ -239,74 +254,124 @@ def scale_t(self, t: Tensor | float) -> Tensor:
...
+    Generative diffusion Stochastic Differential Equation.
 
-    This class implements the reverse-time SDE of the Variance-Exploding SDE (VE-SDE) :footcite:t:`song2020score`.
-
-    The forward-time SDE is defined as follows:
+    This class implements the diffusion generative SDE based on the formulation from :footcite:t:`karras2022elucidating` (with :math:`\beta(t) = \alpha(t) s(t)^2 \sigma(t) \sigma'(t)`):
 
     .. math::
-        d x_t = g(t) d w_t \quad \mbox{where } g(t) = \sigma_{\mathrm{min}} \left( \frac{\sigma_{\mathrm{max}}}{\sigma_{\mathrm{min}}} \right)^t
+        d x_t = \left(\frac{s'(t)}{s(t)} x_t - \frac{1 + \alpha(t)}{2} s(t)^2 \sigma(t) \sigma'(t) \nabla \log p_t(x_t) \right) dt + s(t) \sqrt{2 \sigma(t) \sigma'(t)} d w_t
```
mh-nguyen712: For me I still dont see the equivalence of this with eq(6) in EDM paper, even with your beta(t). It seems that you're missing a `sqrt(alpha(t))` in the diffusion term. For me, this SDE is the `DiffusionSDE` with `g(t) = s(t)\sqrt{\sigma(t) \sigma'(t)}` and `f(x, t) = \frac{s'(t)}{s(t)} x_t`
The eq. 6 in EDM paper looks strange to me as well because they don't have any `x` term in the drift.

deepinv/sampling/diffusion_sde.py
```diff
@@ -239,74 +254,124 @@ def scale_t(self, t: Tensor | float) -> Tensor:
...
         device=torch.device("cpu"),
         *args,
         **kwargs,
     ):
-        def forward_drift(x, t, *args, **kwargs):
-            r"""
-            The drift term of the forward VE-SDE is :math:`0`.
+        self.T = T
+        self.sigma_t = sigma_t
+        self.variance_preserving = variance_preserving
```
mh-nguyen712:
```suggestion
        assert variance_preserving and variance_exploding
        self.variance_preserving = variance_preserving
        self.variance_exploding = variance_exploding
```

deepinv/sampling/diffusion_sde.py
```diff
@@ -239,74 +254,124 @@ def scale_t(self, t: Tensor | float) -> Tensor:
...
-            :param torch.Tensor, float t: The current time
-            :return: The drift term, which is 0 for VE-SDE since it only has a diffusion term
-            :rtype: float
-            """
-            return 0.0
+            def sigma_prime_t(t):
+                t = self._handle_time_step(t).requires_grad_(True)
+                s = self.sigma_t(t)
+                s.backward()
+                return t.grad.item()
```
mh-nguyen712: Just to avoid error when running sampling in a `torch.no_grad()` context 
```suggestion
            def sigma_prime_t(t):
                with torch.enable_grad():
                    t = self._handle_time_step(t).requires_grad_(True)
                    s = self.sigma_t(t)
                    s.backward()
                    return t.grad.detach()
```

deepinv/sampling/diffusion_sde.py
```diff
@@ -239,74 +254,124 @@ def scale_t(self, t: Tensor | float) -> Tensor:
...
+
+        self.sigma_prime_t = sigma_prime_t
+
+        if scale_prime_t is None:
+
+            def scale_prime_t(t):
+                t = self._handle_time_step(t).requires_grad_(True)
+                s = self.scale_t(t)
+                s.backward()
+                return t.grad.item()
```
mh-nguyen712: same here
```suggestion
            def scale_prime_t(t):
                with torch.enable_grad():
                    t = self._handle_time_step(t).requires_grad_(True)
                    s = self.scale_t(t)
                    s.backward()
                    return t.grad.detach()
```

deepinv/sampling/diffusion_sde.py
```diff
@@ -239,74 +254,124 @@ def scale_t(self, t: Tensor | float) -> Tensor:
...
         *args,
         **kwargs,
     ):
-        def forward_drift(x, t, *args, **kwargs):
-            r"""
-            The drift term of the forward VE-SDE is :math:`0`.
+        self.T = T
+        self.sigma_t = sigma_t
+        self.variance_preserving = variance_preserving
+        if scale_t is None:
```
mh-nguyen712, line 333: Can we change this logic to favor the use of `variance_preserving` or `variance_exploding` ? See my previous comment

deepinv/sampling/diffusion_sde.py
```diff
@@ -334,96 +410,201 @@ def sample_init(self, shape, rng: torch.Generator) -> Tensor:
...
+    Compared to the EDM formulation in :class:`deepinv.sampling.EDMDiffusionSDE`, the scale :math:`s(t)` and noise :math:`\sigma(t)` schedulers are defined with respect to :math:`\beta(t)` and :math:`\xi(t)` as follows:
 
     .. math::
-        d x_t = -\frac{1}{2} \beta(t) x_t dt + \sqrt{\beta(t)} d w_t \quad \mbox{ where } \beta(t) = \beta_{\mathrm{min}}  + t \left( \beta_{\mathrm{max}} - \beta_{\mathrm{min}} \right)
+        s(t) = \exp\left(-\int_0^t \beta(s) ds\right), \quad \sigma(t) = \sqrt{2 \int_0^t \frac{\xi(s)}{s(s)^2} ds}.
 
-    The reverse-time SDE is defined as follows:
+    Common choices include the variance-preserving formulation :math:`\beta(t) = \xi(t)` and the variance-exploding formulation :math:`\beta(t) = 0`.
+    For choosing variance-preserving formulation, set `variance_preserving=True` and `beta_t` and `xi_t` will be automatically set to be the same function.
+    For choosing variance-exploding formulation, set `variance_exploding=True` and `beta_t` will be automatically set to `0`.
```
mh-nguyen712: I would add newline here for better visibility
```suggestion

    For choosing variance-preserving formulation, set `variance_preserving=True` and `beta_t` and `xi_t` will be automatically set to be the same function.

    For choosing variance-exploding formulation, set `variance_exploding=True` and `beta_t` will be automatically set to `0`.
```

deepinv/sampling/diffusion_sde.py
```diff
@@ -334,96 +410,201 @@ def sample_init(self, shape, rng: torch.Generator) -> Tensor:
...
+            t = self._handle_time_step(t)
+            return (b_prime_t(t) * a_t(t) - b_t(t) * a_prime_t(t)) / (a_t(t) ** 2)
 
         super().__init__(
-            forward_drift=forward_drift,
+            scale_t=scale_t,
+            scale_prime_t=scale_prime_t,
+            sigma_t=sigma_t,
+            sigma_prime_t=sigma_prime_t,
             forward_diffusion=forward_diffusion,
```
mh-nguyen712: `forward_diffusion` is not defined yet

deepinv/sampling/diffusion_sde.py
```diff
@@ -432,46 +613,113 @@ def forward_diffusion(t):
...
         )
 
-        self.beta_min = beta_min
-        self.beta_max = beta_max
-        self.beta_d = beta_max - beta_min
 
-    def sample_init(self, shape, rng: torch.Generator) -> Tensor:
-        r"""
-        Sample from the initial distribution of the reverse-time diffusion SDE, which is the standard Gaussian distribution.
+class VarianceExplodingDiffusion(EDMDiffusionSDE):
```
mh-nguyen712: It's fine but why we should not inherit VE from SongDiffusion?

examples/sampling/demo_diffusion_sde.py
```diff
@@ -266,11 +266,13 @@
 from deepinv.sampling import VariancePreservingDiffusion
 
 del trajectory
+# denoiser = NCSNpp(pretrained="download", model_name='ddpm').to(device)
```
mh-nguyen712:
```suggestion
```

examples/sampling/demo_diffusion_sde.py
```diff
@@ -266,11 +266,13 @@
...
 sde = VariancePreservingDiffusion(device=device, dtype=dtype)
 model = PosteriorDiffusion(
     data_fidelity=dps_fidelity,
     denoiser=denoiser,
     sde=sde,
+    alpha=0.0,
```
mh-nguyen712: there is no `alpha` parameter in PosteriorDiffusion. It should be set in the sde.

deepinv/sampling/utils.py
```diff
@@ -37,3 +38,30 @@ def refl_projbox(x, lower: Tensor, upper: Tensor) -> Tensor:
...
+    Differentiable trapezoidal integration of func(x) between a and b along the last dimension.
+
+    :param func: Callable that takes a Tensor of shape (..., N) and returns Tensor of same shape.
+    :param torch.Tensor a: Lower integration bound. Can be scalar or batched (...,).
+    :param torch.Tensor b: Upper integration bound. Must be broadcastable with a.
+    :param int n_steps: Number of discretization steps.
+    :return torch.Tensor: Integral of func from a to b. Shape (...), integration along the last dimension.
+    """
+    # Create uniform grid between a and b
+    shape = a.shape
```
mh-nguyen712:
```suggestion
```

deepinv/models/ncsnpp.py
samuro95: I think we should keep the name DDPM as it is the one known in the community. I update the other things.

deepinv/models/ncsnpp.py
samuro95: yes to differentiate the "edm" and "baseline" versions (even though baseline is not used here, could be useful to add it at some point)

deepinv/sampling/diffusion_sde.py
samuro95: good point but I wanted to output a Tensor so that alpha is always tensor

deepinv/sampling/diffusion_sde.py
samuro95: anything positive !

deepinv/sampling/diffusion_sde.py
samuro95: yes it was missing alpha(t) in the definition of the reverse SDE. This is exactly the reverse SDE you mention (as defined in the functions forward_drift and forward_diffusion of the class.

samuro95: (with a factor 2 in the sqrt of the g(t) )

deepinv/sampling/diffusion_sde.py
samuro95: why ?

deepinv/sampling/diffusion_sde.py
samuro95: I don't get your point on this

deepinv/sampling/diffusion_sde.py
samuro95: It makes more sense to be defined by EDM because it is defined via the choice of sigma_t and not the choice of xi_t or beta_t

samuro95: While the VariancePreservingDiffusion is defined via the choice of beta_t

deepinv/models/ncsnpp.py
```diff
@@ -17,14 +17,16 @@ class NCSNpp(Denoiser):
     r"""Implementation of the DDPM++ and NCSN++ architectures.
 
     Equivalent to the original implementation by :footcite:t:`song2020score`, available at `the official implementation <https://github.com/yang-song/score_sde_pytorch>`_.
-
+    The DDPM model has been trained with the VP-SDE from :footcite:t:`song2020score` whle the NCSN++ model has been trained with the VE-SDE.
```
mh-nguyen712:
```suggestion
    The DDPM model has been trained with the VP-SDE from :footcite:t:`song2020score` while the NCSN++ model has been trained with the VE-SDE.
```

deepinv/sampling/diffusion_sde.py
mh-nguyen712: I think for alpha, a scalar is fine (for the current use of alpha). When it's a tensor, we need to make sure that its shape is broadcastable with the image (eg in the backward_drift). Or we should make sure that alpha_t is always a tensor with the good shape (and check it when a custom alpha_t is given). So I think a scalar is a simple choice.

deepinv/sampling/diffusion_sde.py
mh-nguyen712: we cannot use variance_preserving and variance_exploding at the same time (the first assert). 
then just want to set both variance_exploding and variance_preserving as class attribute (you only set variance_preserving)

deepinv/sampling/diffusion_sde.py
```diff
@@ -239,74 +270,135 @@ def scale_t(self, t: Tensor | float) -> Tensor:
...
+    Generative diffusion Stochastic Differential Equation.
 
-    This class implements the reverse-time SDE of the Variance-Exploding SDE (VE-SDE) :footcite:t:`song2020score`.
-
-    The forward-time SDE is defined as follows:
+    This class implements the diffusion generative SDE based on the formulation from :footcite:t:`karras2022elucidating` (with :math:`\beta(t) = \alpha(t) s(t)^2 \sigma(t) \sigma'(t)`):
 
     .. math::
-        d x_t = g(t) d w_t \quad \mbox{where } g(t) = \sigma_{\mathrm{min}} \left( \frac{\sigma_{\mathrm{max}}}{\sigma_{\mathrm{min}}} \right)^t
+        d x_t = \left(\frac{s'(t)}{s(t)} x_t - \frac{1 + \alpha(t)}{2} s(t)^2 \sigma(t) \sigma'(t) \nabla \log p_t(x_t) \right) dt + s(t) \sqrt{2 \alpha(t) \sigma(t) \sigma'(t)} d w_t
```
mh-nguyen712: If we apply the reverse-time SDE formula in Y Song paper. I think there is a missing factor 2 in the term related to the score. (I think the correct one should be (1 + \alpha(t)) * ... instead of (1 + \alpha(t)) / 2 *)

deepinv/sampling/diffusion_sde.py
```diff
@@ -320,40 +412,295 @@ def forward_diffusion(t):
...
-        std = self.sigma_t(t)
-        denoised = self.denoiser(
-            x.to(torch.float32), self.sigma_t(t).to(torch.float32), *args, **kwargs
-        ).to(self.dtype)
-        score = (denoised - x.to(self.dtype)) / std.pow(2)
-        return score
+    This class implements the diffusion generative SDE based the formulation from :footcite:t:`song2020score`:
+
+    .. math::
+        d x_t = -\left(\frac{1}{2} \beta(t) x_t + \frac{1 + \alpha(t)}{2} \xi(t) \nabla \log p_t(x_t) \right) dt + \sqrt{\alpha(t) \xi(t)} d w_t
```
mh-nguyen712, line 483: I would use `g(t)` as the diffusion term instead of `\xi_t`

deepinv/sampling/diffusion_sde.py
```diff
@@ -320,40 +412,295 @@ def forward_diffusion(t):
...
+            dtype=dtype,
+            device=device,
+            *args,
+            *kwargs,
+        )
+
+
+class FlowMatching(EDMDiffusionSDE):
+    r"""
+    Generative Flow Matching process.
```
mh-nguyen712, line 605: Can we write the backward time SDE here? Similar to other classes

deepinv/sampling/diffusion_sde.py
```diff
@@ -320,40 +412,295 @@ def forward_diffusion(t):
...
+            T=0.99,
+            denoiser=denoiser,
+            solver=solver,
+            dtype=dtype,
+            device=device,
+            *args,
+            *kwargs,
+        )
+
+    def velocity(self, x, t, *args, **kwargs):
```
mh-nguyen712: docstring

docs/source/user_guide/reconstruction/sampling.rst
```diff
@@ -122,21 +122,30 @@ In the case of posterior sampling, we need simply to replace the (unconditional)
 by the conditional score function :math:`\nabla \log p_t(x_t|y)`. The conditional score can be decomposed using the Bayes' rule:
 
 .. math::
+    \begin{align}
```
mh-nguyen712, line 125: Do we need to update the `stochastic differential equations` section above. Eg make `alpha` becomes `\alpha(t)`... 

Also for the notation of a noise realization, here we use `\omega`, in the Diffusion classes we use `\epsilon`. I would use `\epsilon` everywhere

examples/sampling/demo_diffusion_sde.py
```diff
@@ -266,7 +266,7 @@
 from deepinv.sampling import VariancePreservingDiffusion
 
 del trajectory
-sde = VariancePreservingDiffusion(device=device, dtype=dtype)
+sde = VariancePreservingDiffusion(alpha=0.0, device=device, dtype=dtype)
```
mh-nguyen712: The generated sample with `alpha=0` is too smooth

mh-nguyen712: image

examples/sampling/demo_flow_matching.py
```diff
@@ -0,0 +1,312 @@
...
+    solver=solver,
+    dtype=dtype,
+    device=device,
+    verbose=True,
+)
+
+seed_1 = 1
+
+
+x_hat, trajectory = model(
```
mh-nguyen712, line 134: The result is weird...

examples/sampling/demo_flow_matching.py
```diff
@@ -0,0 +1,312 @@
...
+model = PosteriorDiffusion(
+    data_fidelity=dps_fidelity,
+    sde=sde,
+    solver=solver,
+    dtype=dtype,
+    device=device,
+    verbose=True,
+)
+
+x_hat, trajectory = model(
```
mh-nguyen712, line 183: This generates the following

examples/sampling/demo_flow_matching.py
mh-nguyen712: image

examples/sampling/demo_flow_matching.py
mh-nguyen712: image

deepinv/sampling/diffusion_sde.py
samuro95: good catch, thanks

deepinv/sampling/diffusion_sde.py
samuro95: good idea

deepinv/models/__init__.py
```diff
@@ -37,7 +37,9 @@
 from .precond import EDMPrecond
 from .ram import RAM
 from .client import Client
-from .wrapper import DiffusersDenoiserWrapper, ComplexDenoiserWrapper
+from .wrapper import DiffusersDenoiserWrapper
```
mh-nguyen712:
```suggestion
```
duplicate

deepinv/models/ncsnpp.py
```diff
@@ -16,17 +16,26 @@
...
     Each residual block has a self-attention mechanism with multiple channels per attention head.
     The noise level can be embedded using either Positional Embedding  or Fourier Embedding with optional augmentation linear layer.
 
+    :param str model_type: Model type, which defines the architecture and embedding types. Options are:
+
+        - `'ncsn'` for the NCSN++ architecture: the following arguments will be ignored and set to `embedding_type='fourier'`, `channel_mult_noise=2`, `encoder_type='residual'`, `decoder_type='standard'`, `resample_filter=[1,3,3,1]`.
+        - `'ddpm'` for the  DDPM++ architecture: the following arguments will be ignored and set to `embedding_type='positional'`, `channel_mult_noise=1`, `encoder_type='standard'`, `decoder_type='standard'`, `resample_filter=[1,1]`.
+
+    Default is `'ncsn'`.
+
```
mh-nguyen712: Remove blank line so that the generated docs do not break into 2 parts (see fig)

deepinv/models/ncsnpp.py
```diff
@@ -39,26 +48,31 @@ class NCSNpp(Denoiser):
...
-    :param list resample_filter: Resampling filter: [1,1] for DDPM++, [1,3,3,1] for NCSN++.
-    :param str, None pretrained: use a pretrained network. If ``pretrained=None``, the weights will be initialized at random
-        using Pytorch's default initialization. If ``pretrained='download'``, the weights will be downloaded from an
-        online repository (the default model trained on FFHQ at 64x64 resolution (`ffhq64-uncond-ve`) with default architecture).
-        Finally, ``pretrained`` can also be set as a path to the user's own pretrained weights.
-        In this case, the model is supposed to be trained on `[0,1]` pixels, if it was trained on `[-1, 1]` pixels, the user should set the attribute `_was_trained_on_minus_one_one` to `True` after loading the weights.
+    :param str encoder_type: Encoder architecture: `'standard'` for DDPM++, `'residual'` for NCSN++.
+    :param str decoder_type: Decoder architecture: `'standard'` for both DDPM++ and NCSN++.
+    :param list resample_filter: Resampling filter: `[1,1]` for DDPM++, `[1,3,3,1]` for NCSN++.
+    :param str, None pretrained: use a pretrained network.
```
mh-nguyen712: Maybe we can provide simple string to download more pretrained weights from huggingface hub (eg imagenet / afhq models).

deepinv/models/wrapper.py
```diff
@@ -1,9 +1,323 @@
...
 
-class DiffusersDenoiserWrapper(Denoiser):
+class ScoreModelWrapper(Denoiser):
+    r"""
+    Wraps a score model as a DeepInv Denoiser.
+
+    Given a noisy sample :math:`x_t = s_t(x_0 + \sigma_t \varepsilon)`, where :math:`\varepsilon \sim \mathcal{N}(0, I)`,
+    depending on the `prediction_type`, the input `score_model` is trained to predict, either:
+
+        * the noise :math:`\varepsilon` (`prediction_type = 'noise'`)
```
mh-nguyen712: I would use `epsilon` instead of `noise` here, as it's more standard (eg in diffusers)

deepinv/models/wrapper.py
```diff
@@ -1,9 +1,323 @@
...
+        * the velocity (or drift) of the corresponding ODE/SDE :math:`s_t (\varepsilon - \sigma_t * x_0)` as typically the case for flow-matching models (`prediction_type = 'velocity'`)
+
+    :param torch.nn.Module | Callable score_model: score model to be wrapped.
+    :param str prediction_type: type of prediction made by the score model.
+    :param bool clip_output: whether to clip the output to the model range. Default is `True`.
+    :param Callable | torch.Tensor sigma_schedule: continuous function or tensor (of shape `[N]` with `N` the number of time steps) defining the noise schedule :math:`\sigma_t`.
+    :param Callable | torch.Tensor scale_schedule: function or tensor (of shape `[N]` with `N` the number of time steps) defining the scaling schedule :math:`s_t`.
+    :param Callable sigma_inverse: analytic inverse of the `sigma_schedule`. If not provided, a numeric inversion is used.
+    :param bool variance_preserving: whether the schedule is variance-preserving. If `True`, the `scale_schedule` is computed from the `sigma_schedule`.
+    :param bool variance_exploding: whether the schedule is variance-exploding. If `True`, the `scale_schedule` is set to `1`.
```
mh-nguyen712: I would remove these 2 parameters and give 2 examples of `sigma_schedule` and `scale_schedule` corresponding to VE and VP, with small code snippet. I think it's a bit confusing to have all these 2 parameters (which only used for initializing the schedules)

deepinv/models/wrapper.py
```diff
@@ -1,9 +1,323 @@
...
+    :param float T: maximum time value for continuous schedules. Default is `1.0`.
+    :param bool takes_integer_time: whether the model takes integer time steps (in `[0, n_timesteps-1]`) as input. Default is `False`.
+    :param int n_timesteps: number of time steps for discrete schedules. Default is `1000`.
+    :param bool _was_trained_on_minus_one_one: whether the model was trained on images in `[-1, 1]` range (`True`) or `[0, 1]` range (`False`). Default is `True`.
```
mh-nguyen712, line 31: it's a bit weird to have private arg (`_was...`) here but I don't see how to remove it easily...

deepinv/models/wrapper.py
```diff
@@ -1,9 +1,323 @@
...
+        return a / b
+
+    def score(
+        self, x: torch.Tensor, t: float | torch.Tensor = None, *args, **kwargs
+    ) -> torch.Tensor:
+        r"""
+        Computes the score function :math:`\nabla_x \log p_t(x)`.
+
+        :param torch.Tensor x: input tensor of shape `[B, C, H, W]`.
+        :param torch.Tensor | float t: single time or tensor of shape `[B]` or `[]`.
```
mh-nguyen712:
```suggestion
        :param torch.Tensor | float t: single timestep or tensor of shape `[B]` or `[]`.
```

deepinv/models/wrapper.py
```diff
@@ -1,9 +1,323 @@
...
+        if isinstance(scale_schedule, torch.Tensor):
+            self.register_buffer("scale_schedule", scale_schedule)
+        else:
+            self.scale_schedule = scale_schedule
+
+        self.sigma_inverse = sigma_inverse
+        self.T = T
+        self.to(device)
+
+    def _handle_time_step(
```
mh-nguyen712: we can remove this function and use `self._handle_sigma`

deepinv/models/wrapper.py
```diff
@@ -1,9 +1,323 @@
...
+
+    def get_schedule_value(
+        self,
+        schedule: Callable | torch.Tensor,
+        t: torch.Tensor,
+        target_size: torch.Size = None,
+    ) -> torch.Tensor:
+        r"""
+        Get the value of a schedule (function or tensor) at given time steps.
+        :param Callable | torch.Tensor schedule: schedule function or tensor.
```
mh-nguyen712, line 172: add a newline here

deepinv/models/wrapper.py
mh-nguyen712: It's indeed `epsilon` in `forward`

deepinv/models/wrapper.py
```diff
@@ -1,9 +1,323 @@
...
+        x: torch.Tensor,
+        sigma: float | torch.Tensor = None,
+        input_in_minus_one_one: bool = False,
+        *args,
+        **kwargs,
+    ) -> torch.Tensor:
+        r"""
+        Applies denoiser :math:`\denoiser{x}{\sigma}`.
+        If `input_in_minus_one_one` is `False` (default value), the input `x` is expected to be in `[0, 1]` range (up to random noise) and the output is also in `[0, 1]` range.
+        Else, both input and output are expected in `[-1, 1]` range.
```
mh-nguyen712:
```suggestion
        Otherwise, both input and output are expected in `[-1, 1]` range.
```

deepinv/models/wrapper.py
```diff
@@ -48,60 +362,89 @@ def __init__(
...
-            diffs = (sigmas[None, :] - sigma[:, None]).abs()  # [B, T]
-            timestep = torch.argmin(diffs, dim=1)
-        return timestep
+        self.tokenizer = pipeline.tokenizer if hasattr(pipeline, "tokenizer") else None
+        self.text_encoder = (
+            pipeline.text_encoder if hasattr(pipeline, "text_encoder") else None
+        )
+        self.bert = pipeline.bert if hasattr(pipeline, "bert") else None
+        self.vae = pipeline.vae if hasattr(pipeline, "vae") else None
+        self.vqvae = pipeline.vqvae if hasattr(pipeline, "vqvae") else None
```
mh-nguyen712: maybe it's cool to have these attributes I'm not sure if we need them

deepinv/models/wrapper.py
```diff
@@ -48,60 +362,89 @@ def __init__(
...
-        ), "Currently, only DDPMScheduler is supported."
+        model = pipeline.unet
+        scheduler = pipeline.scheduler
+        prediction_type = getattr(scheduler.config, "prediction_type", "epsilon")
 
-        self.clip_output = clip_output
-
-        # Precompute sigma(t) over training timeline
-        ac = self.scheduler.alphas_cumprod
+        if isinstance(scheduler, (PNDMScheduler, DDPMScheduler, DDIMScheduler)):
```
mh-nguyen712, line 447: super cool that it now works with more Scheduler!

deepinv/models/wrapper.py
mh-nguyen712: So we could remove the note before (L330-331)

deepinv/sampling/diffusion_sde.py
```diff
@@ -8,26 +8,37 @@
 from deepinv.models.base import Reconstructor, Denoiser
 from deepinv.optim.data_fidelity import ZeroFidelity
 from deepinv.sampling.sde_solver import BaseSDESolver, SDEOutput
-from deepinv.sampling.noisy_datafidelity import NoisyDataFidelity
-from copy import deepcopy
+from deepinv.sampling.noisy_datafidelity import NoisyDataFidelity, DPSDataFidelity
+from deepinv.sampling.utils import trapz_torch
+import threading
+
+_print_lock = threading.Lock()
```
mh-nguyen712: it's debugging code? remove it

deepinv/sampling/diffusion_sde.py
```diff
@@ -8,26 +8,37 @@
...
+        self.xmax = xmax
 
     def forward(self, x: Tensor, sigma: Tensor, *args, **kwargs) -> Tensor:
-        # Scale from [-1, 1] to [0, 1]
-        x = (x + 1) / 2
-        denoised_01 = self.denoiser(x, sigma / 2, *args, **kwargs)
-        # Scale back to [-1, 1]
-        denoised = denoised_01 * 2 - 1
+        # Scale from [-1, 1] to [xmin, xmax], except if specified otherwise with the 'input_in_minus_one_one' argument in kwargs
+        if "input_in_minus_one_one" in kwargs and not kwargs["input_in_minus_one_one"]:
```
mh-nguyen712:
```suggestion
        if not kwargs.get("input_in_minus_one_one", True)
```
can be a bit simpler

deepinv/sampling/diffusion_sde.py
```diff
@@ -8,26 +8,37 @@
...
+            x = (x + 1) / 2 * (self.xmax - self.xmin) + self.xmin
+            sigma = sigma * (self.xmax - self.xmin) / 2
+        denoised = self.model(x, sigma, *args, **kwargs)
+        # Scale back to [-1, 1], except if specified otherwise with the 'input_in_minus_one_one' argument in kwargs
+        if "input_in_minus_one_one" in kwargs and not kwargs["input_in_minus_one_one"]:
```
mh-nguyen712:
```suggestion
        if not kwargs.get("input_in_minus_one_one", True)
```
same here, or even store it as a variable

deepinv/sampling/diffusion_sde.py
```diff
@@ -134,11 +165,11 @@ class DiffusionSDE(BaseSDE):
     This class define the following reverse-time SDE:
 
     .. math::
-        d x_{t} = \left( f(x_t, t) - \frac{1 + \alpha}{2} g(t)^2 \nabla \log p_t(x_t) \right) dt + g(t) \sqrt{\alpha} d w_{t}.
+        d x_{t} = \left( f(x_t, t) - \frac{1 + \alpha(t)}{2} g(t)^2 \nabla \log p_t(x_t) \right) dt + g(t) \sqrt{\alpha(t)} d w_{t}.
 
     :param Callable drift: a time-dependent drift function :math:`f(x, t)` of the forward-time SDE.
     :param Callable diffusion: a time-dependent diffusion function :math:`g(t)` of the forward-time SDE.
-    :param Callable alpha: a scalar weighting the diffusion term. :math:`\alpha = 0` corresponds to the ODE sampling and :math:`\alpha > 0` corresponds to the SDE sampling.
+    :param Callable, float alpha: a (possibly time-dependent) positive scalar weighting the diffusion term. A  constant function :math:`\alpha(t) = 0` corresponds to ODE sampling and :math:`\alpha(t) > 0` corresponds to SDE sampling.
```
mh-nguyen712: by `positively time-dependent` you meant it's increasing w.r.t `t`? otherwise I would remove `positively`

deepinv/sampling/diffusion_sde.py
mh-nguyen712: I think the factor 2 issue is not resolved yet?

deepinv/sampling/diffusion_sde.py
```diff
@@ -239,74 +274,134 @@ def scale_t(self, t: Tensor | float) -> Tensor:
...
+        This SDE must be solved by going reverse in time i.e. from :math:`t=T` to :math:`t=0`.
+
+    :param Callable sigma_t: a time-dependent noise level schedule.
+    :param Callable scale_t: a time-dependent scale schedule.
+        If not provided, it will be set to :math:`s(t) = \left(1 + \sigma(t)^2\right)^{-1/2}` if `variance_preserving=True`, or :math:`s(t) = 1` if `variance_exploding=True`.
+        If both `variance_preserving` and `variance_exploding` are `False`, `scale_t` must be provided. Default to `None`.
+    :param Callable sigma_prime_t: the derivative of `sigma_t`. If not provided, it will be computed using autograd. Default to `None`.
+    :param Callable scale_prime_t: the derivative of `scale_t`. If not provided, it will be computed using autograd. Default to `None`.
+    :param bool variance_preserving: whether to use the variance-preserving formulation, which imposes :math:`s(t) = \left(1 + \sigma(t)^2\right)^{-1/2}`. Default to `False`.
+    :param bool variance_exploding: whether to use the variance-exploding formulation, which imposes :math:`s(t) = 1`. Default to `False`.
```
mh-nguyen712: Similar here. I would remove these 2 parameters and give the corresponding schedules in the docs. I think it will simplify a bit. We can chose a default schedule, eg VP.

deepinv/sampling/diffusion_sde.py
```diff
@@ -320,110 +415,171 @@ def forward_diffusion(t):
...
+        self, x: Tensor, model_output: Tensor, sigma: Tensor, scale: Tensor
+    ) -> Tensor:
+        denoised = scale * model_output
+        score = (denoised - x.to(self.dtype)) / (scale * sigma).pow(2)
+        return score
 
     def sample_init(self, shape, rng: torch.Generator) -> Tensor:
         r"""
-        Sample from the initial distribution of the reverse-time diffusion SDE, which is a Gaussian with zero mean and covariance matrix :math:`\sigma_{max}^2 \operatorname{Id}`.
+        Sample from the initial distribution of the reverse-time diffusion SDE, which is a Gaussian with zero mean and covariance matrix :math:` s(t)^2 \sigma(T)^2 \operatorname{Id}`.
```
mh-nguyen712:
```suggestion
        Sample from the initial distribution of the reverse-time diffusion SDE, which is a Gaussian with zero mean and covariance matrix :math:` s(T)^2 \sigma(T)^2 \operatorname{Id}`, where :math:`T` is the endpoint of the forward SDE. 
```

deepinv/sampling/diffusion_sde.py
```diff
@@ -320,110 +415,171 @@ def forward_diffusion(t):
...
         self,
-        denoiser: Denoiser = None,
-        beta_min: float = 0.0001,
-        beta_max: float = 5.0,
-        alpha: float = 1.0,
+        beta_t: Callable = None,
+        B_t: Callable = None,
+        xi_t: Callable = None,
+        variance_preserving: bool = False,
+        variance_exploding: bool = False,
```
mh-nguyen712, line 528: similar discussion about these 2 params

deepinv/sampling/diffusion_sde.py
```diff
@@ -432,46 +588,218 @@ def forward_diffusion(t):
...
-        scale = self.scale_t(t)
+    .. math::
+        s(t) = a(t), \quad \sigma(t) = \frac{b(t)}{a(t)} .
 
-        denoised = scale * self.denoiser(
-            (x / scale).to(torch.float32),
-            sigma.to(torch.float32),
+    .. note::
+
+        This SDE must be solved going reverse in time i.e. from :math:`t=T` to :math:`t=0`.
```
mh-nguyen712: Maybe add a warning that here we implement x_0 as the data distribution. In standard flowing matching, x_0 usually denotes the noise and x_1 is the data.
I don't know if this creates confusion for people that already know flow matching.

mh-nguyen712 (changes requested):
Thanks for making these updates!

deepinv/models/wrapper.py
samuro95: I kept them for future integration of text-conditional generation, I agree to remove them for now

deepinv/sampling/diffusion_sde.py
samuro95: mmh, I believe they are quite useful, we can discuss it if you want

deepinv/models/wrapper.py
```diff
@@ -1,9 +1,323 @@
...
+
+    Given a noisy sample :math:`x_t = s_t(x_0 + \sigma_t \varepsilon)`, where :math:`\varepsilon \sim \mathcal{N}(0, I)`,
+    depending on the `prediction_type`, the input `score_model` is trained to predict, either:
+
+        * the noise :math:`\varepsilon` (`prediction_type = 'noise'`)
+
+        * the denoised sample :math:`x_0` (`prediction_type = 'denoised'`)
+
+        * the `v-prediction` :math:`s_t (\varepsilon - \sigma_t * x_0)` as proposed by :footcite:`salimans2022progressive` (`prediction_type = 'v_prediction'`)
+        * the velocity (or drift) of the corresponding ODE/SDE :math:`s_t (\varepsilon - \sigma_t * x_0)` as typically the case for flow-matching models (`prediction_type = 'velocity'`)
```
mh-nguyen712: This is not implemented

deepinv/sampling/diffusion_sde.py
```diff
@@ -239,74 +274,134 @@ def scale_t(self, t: Tensor | float) -> Tensor:
...
-    :param torch.device device: device on which the computation is performed.
+    :param torch.device device: device on which the computation is performed. Default to CPU.
     :param \*args: additional arguments for the :class:`deepinv.sampling.DiffusionSDE`.
     :param \*\*kwargs: additional keyword arguments for the :class:`deepinv.sampling.DiffusionSDE`.
-
     """
 
     def __init__(
         self,
+        sigma_t: Callable,
```
mh-nguyen712, line 304: we could change to `sigma_schedule` as in the Wrapper, or keep `sigma_t` here and change `sigma_schedule` to `sigma_t`.

deepinv/sampling/diffusion_sde.py
```diff
@@ -239,74 +274,134 @@ def scale_t(self, t: Tensor | float) -> Tensor:
...
+            else:
+                raise ValueError(
+                    "'scale_t' must be provided if 'variance_preserving' and 'variance_exploding' is False"
+                )
+        self.scale_t = scale_t
+
+        if sigma_prime_t is None:
+
+            def sigma_prime_t(t):
+                with torch.enable_grad():
```
mh-nguyen712, line 374: So by doing this, you're supposing that `t` is a scalar (not batched). We should clarify the input / output type of `sigma_t` and `scale_t` in the doc

deepinv/sampling/diffusion_sde.py
samuro95: I changed in the wrapper

Issue #556, opened 2025-06-20
Tests failing on GPU


jscanvic:
**Reproducing**

On a machine with a GPU:

```
$ CUDA_VISIBLE_DEVICES=-1 python -m pytest deepinv/tests/test_trainer.py
...
45 passed, 8 skipped, 42 warnings in 4.78s
```

```
$ python -m pytest deepinv/tests/test_trainer.py
...
24 failed, 21 passed, 8 skipped, 25 warnings in 4.13s
```

mh-nguyen712:
I also experienced this during development. There are also other tests that fail on a GPU machine. It would be great to correct all of them.
E.g. in `test_unfolded` ,  `test_DEQ` fails.

jscanvic:
@mh-nguyen712  Thanks for pointing this out!

Tmodrzyk:
Just pointing out that some examples in the doc do also fail when built on GPU, so if users try them on a GPU machine they'll probably end up with an error. 
For instance this one: [demo_diffpir.py](https://github.com/deepinv/deepinv/blob/main/examples/sampling/demo_diffpir.py)

PR #569, opened 2025-06-25
Fixing tests on GPU

jscanvic:
Towards #556 

**Testing command**

```bash
DEEPINV_MOCK_TESTS=True python -m pytest -v deepinv/tests 2>&1 > >(tee test.log)
```

**Results before fixes**

[full logs](https://github.com/user-attachments/files/20909794/test.log)

```
=484 failed, 701 passed, 99 skipped, 137 warnings, 363 errors in 559.19s (0:09:19) =
```


jscanvic:
We're almost there but there's a last bug I can't manage to fix - I'll leave it for another PR.

deepinv/loss/metric/distortion.py
```diff
@@ -680,16 +680,16 @@ def metric(self, x_net: Tensor = None, x: Tensor = None, *args, **kwargs) -> Ten
         coeffs_deg_y = self._haar_wavelet_decompose(deg_y, n_scales)
         if is_color_image:
             coefficients_ref_i = torch.abs(
-                self._convolve2d(ref_i, torch.ones((2, 2)) / 4.0)
+                self._convolve2d(ref_i, torch.ones((2, 2), device=ref_i.device) / 4.0)
```
mh-nguyen712:
```suggestion
                self._convolve2d(ref_i, torch.ones((2, 2), device=ref_i.device, dtype=ref_i.dtype) / 4.0)
```

mh-nguyen712: and other lines as well please

deepinv/physics/generator/base.py
```diff
@@ -115,7 +124,8 @@ def reset_rng(self):
         r"""
         Reset the random number generator to its initial state.
         """
-        self.rng.set_state(self.initial_random_state)
+        # NOTE: Generator.set_state expects a tensor living on the CPU.
+        self.rng.set_state(self.initial_random_state.cpu())
```
mh-nguyen712, line 128: Thanks!

deepinv/physics/noise.py
```diff
@@ -22,6 +22,14 @@ def __init__(self, noise_model: Callable = None, rng: torch.Generator = None):
         self.noise_model = noise_model
         self.rng = rng
         if rng is not None:
+            # NOTE: There is no use in moving RNG states from one device to another
```
mh-nguyen712, line 25: Another option could be:
- save the `initial_random_noise` as a regular attribute `self.initial_random_noise = rng.get_state()`
- then over-write the `self.state_dict()` and `self.load_state_dict()` method.
What do you think?

deepinv/tests/test_models.py
```diff
@@ -493,14 +493,22 @@ def test_wavelet_decomposition(channels, dimension, batch_size, device):
     # 1 decomposition
     out = model.dwt(x)
     x_hat = model.iwt(out)
-    assert x_hat.shape == x.shape and torch.allclose(x, x_hat, rtol=1e-5, atol=1e-5)
+
+    # For some reason the precision is more than 100x lower on GPU.
+    tol = 1e-3 if torch.device(device).type == "cuda" else 1e-5
```
mh-nguyen712, line 498: It passed on my GPU so it might by hardware dependent?

mh-nguyen712 (commented):
Many thanks Jérémy. Just have some minor comments

deepinv/physics/noise.py
jscanvic: I like the idea but I'm worried of breaking something!

Do you think this would work?

```python
def state_dict(self, *args, **kwargs):
    dict = super().state_dict(*args, **kwargs)
    dict["initial_random_noise"] = self.initial_random_noise
    return dict

def load_state_dict(self, state_dict, *args, **kwargs):
    self.initial_random_noise = dict.popitem("initial_random_noise")
    super().load_state_dict(state_dict)

```

deepinv/tests/test_models.py
jscanvic: Oh okay, let me revert it then!

jscanvic: The last error I had was related to this specific test - I think it's not just a tolerance issue anyway

deepinv/physics/noise.py
mh-nguyen712: It could work but I just realized that we also have to put `self.rng.set_state(self.initial_rng_state.cpu())`, to prevent error if someone load the state_dict with `torch. Load(..., map_device='cuda')`

mh-nguyen712: So I think it does not help much...  So a simple call `.cpu()` should be ok since this functions is not called regularly...

deepinv/tests/test_models.py
mh-nguyen712: Just checked it again on my side. It passed on a RTX5000 but not on my laptop GPU with tol=1e-5

deepinv/physics/noise.py
jscanvic: Hmm right! Let's keep it that way then.

deepinv/tests/test_models.py
jscanvic: Weird... What do you suggest we do?

mh-nguyen712: I think we could set tol=1e-3 for GPU as you did before you contributors dont have issues when testing. Sorry for the inconvenience.

jscanvic: I made the change! Should be good now.

Issue #563, opened 2025-06-24
Get GPU with freer memory - small bug to correct (in the doc)


fedor-goncharov:
While working on a station with 2 GPUs I noticed sometimes the utility function `get_freer_gpu` from `utils.nn.py`
does not select the right one, even if the message is correct. 

The issue well-known since `nvidia-smi` prints GPUs in order of the ids on the PCI-bus and CUDA Runtime 
prints those in the order of perfomance capabilities. 

https://discuss.pytorch.org/t/gpu-device-ordering/60785

To make `get_freer_gpu` work correctly it is needed simply to specify an environment variable:
```bash
export CUDA_DEVICE_ORDER=PCI_BUS_ID
```
Do it in the shell that runs your `.py` script or that is supporting your Jupyter notebook (e.g., use magic command `%env`). 

I think there is no need to change anything in the code itself, but to add a line a doc - "if you plan to select freer gpu 
by the library, then do set the environment variable ... ". 

P. S. Sorry, to put this as an issue, I wanted to make a PR on this but the docs compilation seems has some errors right now, 
so maybe one of main developers could add such comment in the docs. If you are lazy to do that, just push on me to make a PR :) 
I could do that in the end :)

jscanvic:
Thanks a lot for pointing out this oversight on our end!

Do you think you could write a minimal reproducible example? Something along the lines of:

1. Call `get_freer_gpu` and store the output device
2. Assert that it has more free memory than all other devices using `torch.cuda.mem_get_info` and `torch.cuda.device_count`

I understand that the test will still pass when the two orderings happen to be the same but it'd still be helpful!

Besides that, it'd be great if you could do a PR adding your fix to the documentation! The CI pipeline should work fine now - you just need to branch off your PR off of an up to date copy of the main branch (the fix was pushed this morning).

Apparently we can't change the ordering used by `nvidia-smi` but I'm really not fond of having a function return the wrong thing silently. I suggest we also add a warning in `get_freer_gpu` if the returned device has been obtained through `nvidia-smi` and the environment variable is not set, saying that it's unreliable along with a boolean parameter to skip `nvidia-smi` and use the slower but correct fallback instead. Do you think you can take care of it? It'd be really useful!

fedor-goncharov:
@jscanvic Yes, I will do a PR on this :) 

But I do not see it as unreliable method if you set the environmental as recommended in the link. 
[...]

PR #570, opened 2025-06-26
added a warning in the doc for get_freer_gpu 

fedor-goncharov:
This is a PR related to discussion in 

https://github.com/deepinv/deepinv/issues/563#issuecomment-3005644044


deepinv/utils/nn.py
```diff
@@ -4,6 +4,14 @@
 import numpy as np
 import torch
 
+_warn_msg_cuda_order = (
+    "Your attempt to choose GPU was not reliable because the environment variable "
+    "'CUDA_DEVICE_ORDER' either is not set or not equal to 'PCI_BUS_ID' (see also the doc). "
+    "Before loading Pytorch you can set the variable in your shell using the command: "
```
jscanvic: minor change for consistency: PyTorch*

deepinv/utils/nn.py
```diff
@@ -29,6 +44,9 @@ def get_freer_gpu(verbose=True):
         idx, mem = np.argmax(memory_available), np.max(memory_available)
         device = torch.device(f"cuda:{idx}")
 
+        env = os.environ.copy()
```
jscanvic: Is the copy needed?

deepinv/utils/nn.py
```diff
@@ -29,6 +44,9 @@ def get_freer_gpu(verbose=True):
...
+        if "CUDA_DEVICE_ORDER" not in env or env["CUDA_DEVICE_ORDER"] != "PCI_BUS_ID":
+            warn(_warn_msg_cuda_order)
```
jscanvic: Please put the string literal here instead of using a variable for better consistency with the rest of the code base

deepinv/utils/nn.py
```diff
@@ -4,6 +4,14 @@
...
+_warn_msg_cuda_order = (
+    "Your attempt to choose GPU was not reliable because the environment variable "
+    "'CUDA_DEVICE_ORDER' either is not set or not equal to 'PCI_BUS_ID' (see also the doc). "
+    "Before loading Pytorch you can set the variable in your shell using the command: "
+    "'export CUDA_DEVICE_ORDER=PCI_BUS_ID' "
+    "or use IPython's 'magic' command '%env CUDA_DEVICE_ORDER=PCI_BUS_ID'."
+)
+
 
 def get_freer_gpu(verbose=True):
```
jscanvic: Currently the functions tries to use `nvidia-smi` first and it if fails it uses `torch.cuda.mem_get_info` instead. Can you also add a new parameter `correct` which fallbacks to it if the environment variable is not set? The idea is some users might not want to set the environment variable for whatever reason but they might still want the correct output and to get rid of the warning.

jscanvic (changes requested):
It looks great so far (especially for a 1st PR!)

I suggest some minor changes & also adding a different fix for users' convenience.

deepinv/utils/nn.py
fedor-goncharov: I made a copy for security - if this code will be ever changed, it is to underline that one should not change current environment. This is potentially possible if the copy is not done. There is a separate command `os.getenv` for this - I think I will change to it.

deepinv/utils/nn.py
fedor-goncharov: I understand your point, your are right. But if the variable is not set, no need to have an extra parameter `correct`. 
I can just write `if env["CUDA_DEVICE_ORDER"]` - so it will generate an exception and switch to pytorch's routines. 
No warning will be printed. 

However, if the variable is set but to something else we can generate a warning but do nothing more. 
It is a responsibility of the user to choose the right device then.

What do you think?

deepinv/utils/nn.py
jscanvic: Understood - using `os.getenv` sounds good to me too 👍

deepinv/utils/nn.py
jscanvic: This is probably what we ultimately will aim for but I'd prefer not introducing a breaking change right away and let this other fix be opt-in for now

fedor-goncharov: But have you looked for the latest commit? I kept `os.environ.copy()` so that it is nicer to implement the logic 
on falling back to torch routines. It is following the logic already implemented in the code: if `try-catch` fails somewhere, fallback on torch routines.  

The other way, I am not sure I understand what do you mean on implementation of a `correct` flag. For me 
it means somewhat bypassing `the try-catch` that is already implemented. It is possible in principle, 
but then I would need bigger modifications of the code.

p.s. I did other corrections - put string literal / adjusted docs and text a bit. 

Sorry that a small makes a longer discussion :)

jscanvic: Sorry I might be missing something here! I'll take a closer look asap

jscanvic: My understanding of what you do is you use `env["CUDA_DEVICE_ORDER"]` which raises an exception if `CUDA_DEVICE_ORDER` is unset to fallback to using `torch.cuda.mem_get_info` in that case. Is that correct?

fedor-goncharov: Yes, that's right.

fedor-goncharov: Hi, is such logic is fine with the project?

jscanvic: Hi - sorry for the delay. I think it's a great fix! The thing is I believe it should be opt-in so users see the warning message first and decide which fix to rely on next. Do you see an issue with that?

fedor-goncharov: I see, you mean that in addition I should add a flag aka "torch_fallback" which automatically falls to torch routines without printing the warning message. If this is what you mean, yes, I agree with you. 

p.s. In this case I think it is better than move the torch routines in a separate function "_get_device_torch" and make calls to it. In that case I also do not need to make exception in ``env'' variable - writing logic with `try-except` is not really nice. 

I will proceed as above and submit a new commit if this is what you mean.

fedor-goncharov:
I refactored the code for get free gpu. I think larger changes are needed to make the code more clear and consistent:

1. before all logic was implemented only in 1 function only. Now I propose to put 2 different methods in two different functions `_get_freer_gpu_system` for nvidia-smi method and `_get_freer_gpu_torch` for using torch routines. Names of these functions reflect that these are internal methods.  
2. I propose using the PyTorch's method `_get_freer_gpu_torch` as the default one since it is more reliable in any generic situation and is less dependent on tricky parsing of strings from `nvidia-smi`. User can choose which method to choose by setting an appropriate boolean flag. 
3. I added boolean flag to supress all warnings
4. However, if user wants to use `nvidia-smi` method and an exception arises there, then by force the call fallbacks to torch routines and a warning is printed. I think in this situation the warning should never be supressed, as it reports of some problem. User can avoid this message if simply chooses torch's method + supresses all warnings with a flag.
5. The main call is now short and its logic is more clear

Sorry for so significant changes, but in the previous version I could not see how to make it clear with a small change. 
I think it is also not a very good style to implement logic that relies on exceptions, so I decided not to raise exceptions if 
some environment variables are not set.

jscanvic:
Thanks for making the changes and sorry for the delay (again!)

Final thing before it can be merged can you please update the changelog and fix the formatting using `black .` to make the checks pass?

fedor-goncharov:
> Thanks for making the changes and sorry for the delay (again!)
> 
> Final thing before it can be merged can you please update the changelog and fix the formatting using `black .` to make the checks pass?

Sorry, did it. I tested the code on my machine, seems to be fine, if you want maybe you can check on yours. Thanks for the answer btw!

fedor-goncharov:
I am sorry, I've just noticed, I did not resolve changelog conflicts, could you do it manually? :)

Andrewwango:
@jscanvic is it worth adding tests to fix this codecov?

jscanvic:
The coverage issue is mostly fixed by #555 - I just added support for the new arguments in the test case.

I also reordered the arguments for improved backwards compatibility.

It should be good for merging now!

PR #585, opened 2025-07-02
fix release one last time

tachella:
I've separated the release into two different actions: first, we manually run the release.yml which creates a PR, and once the PR is merged, this triggers the second action which uploads the new release to pipy

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,39 +45,18 @@ jobs:
             echo "❌ One or both workflows failed — aborting."
             exit 1
           fi
+          
+      - name: Install tools
+        run: |
+          pip install --upgrade tomlkit
 
       - name: Checkout code
         uses: actions/checkout@v4
-
-      - name: Check for recent release commit
-        id: check_release
-        run: |
-          LAST_COMMIT_MSG=$(git log -1 --pretty=%s)
-          echo "Last commit: $LAST_COMMIT_MSG"
-          if [[ "$LAST_COMMIT_MSG" =~ ^Release\ v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
-            echo "Latest commit is a release. Skipping release process."
-            echo "skip=true" >> $GITHUB_OUTPUT
-          else
-            echo "skip=false" >> $GITHUB_OUTPUT
-          fi
-
-      - name: Exit if no new changes
-        if: steps.check_release.outputs.skip == 'true'
-        run: |
-          echo "No new changes since last release. Skipping workflow."
-          exit 1
           
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
           python-version: '3.9'
-
-      - name: Install tools
-        run: |
-          pip install --upgrade build twine tomlkit
-
-      - name: Build the package
-        run: python -m build
 
       - name: Update changelog
         id: changelog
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -47,40 +45,19 @@ jobs:
             echo "❌ One or both workflows failed — aborting."
             exit 1
           fi
+
```
Andrewwango, line 48: why isn't this after "setup python"?

.github/workflows/release.yml
```diff
@@ -47,40 +45,19 @@ jobs:
...
           fi
+          
+      - name: Install tools
+        run: |
+          pip install --upgrade tomlkit
 
       - name: Checkout code
         uses: actions/checkout@v4
-
-      - name: Check for recent release commit
```
Andrewwango, line 54: so if I understand correctly, we don't need this anymore because we just trigger whenever we want

.github/workflows/release.yml
```diff
@@ -92,43 +69,21 @@ jobs:
...
-          
-      - name: Publish to PyPI
-        if: ${{ success() }}
-        env:
-          TWINE_USERNAME: __token__
-          TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
-        run: |
-          python -m twine --verbose upload dist/*
-
+      - name: create PR with changes
```
Andrewwango, line 72: I don't understand fully, don't we need to first create a branch called release-branch, checkout to it, before commiting the changelog and pyproject, and then creating the PR?

.github/workflows/release.yml
```diff
@@ -92,43 +69,21 @@ jobs:
...
           echo "version=$(python .github/scripts/get_version.py pyproject.toml)" >> "$GITHUB_OUTPUT"
 
-      - name: Stop if not on main
-        run: |
-          if [ "${{ github.ref_name }}" != "main" ]; then
-            echo "❌ Not on main branch — skipping release."
-            exit 1
-          fi
-
-      - name: Commit version bump
```
Andrewwango, line 102: Why did you remove the actual commit of the bumps?

Andrewwango (changes requested):

.github/workflows/release.yml
tachella: yes, we can check ourselves

.github/workflows/release.yml
tachella: my understanding is that this is automatically done by the github action I'm using

.github/workflows/release.yml
tachella: I think this is done by the action, but I'll double check

.github/workflows/release.yml
tachella: my bad, fixed

.github/workflows/release.yml
tachella: i've explicited which files to commit, to make sure that no other file, eg changelog.txt is pushed

.github/workflows/release.yml
tachella: should be fine https://github.com/peter-evans/create-pull-request/tree/v7/

Andrewwango (approved):

Issue #594, opened 2025-07-09
``fontsize`` parameter in ``deepinv.utils.plot`` has no effect


Tmodrzyk:
The ``fontsize`` parameter in the plotting function ``deepinv.utils.plot`` has no apparent effect on any text in the output figure. 
This behavior can be observed both on the saved images, and on the displayed images if you run it in a notebook.

I'm not very familiar with matplotlib's internals, but I guess the ``fontsize`` should be passed as a ``kwarg`` to most of the functions in ``plot``, instead of using ``config_matplotlib`` which seems to have no effect at the moment. 

Minimal example to reproduce the bug: 

```python
import torch
from deepinv.utils import plot
from pathlib import Path

x0 = torch.randn(1, 3, 256, 256)
x1 = torch.randn(1, 3, 256, 256)

save_dir = Path("./fontsize50/")

plot(
    img_list=[x0, x1],
    titles=["Image 1", "Image 2"],
    suptitle="Fontsize 50",
    figsize=(10, 4),
    fontsize=50,
    rescale_mode="clip",
    save_dir=save_dir,
)
save_dir = Path("./fontsize17/")

plot(
    img_list=[x0, x1],
    titles=["Image 1", "Image 2"],
    suptitle="Fontsize 17",
    figsize=(10, 4),
    fontsize=17,
    rescale_mode="clip",
    save_dir=save_dir,
)
```

jscanvic:
Thanks for raising this!

There are hard-coded font sizes in the current implementation:

https://github.com/deepinv/deepinv/blob/99379cfd164811895e745703cfac42e74c1ac413/deepinv/utils/plotting.py#L299

https://github.com/deepinv/deepinv/blob/99379cfd164811895e745703cfac42e74c1ac413/deepinv/utils/plotting.py#L311

I've managed to get the right font sizes by using this code instead:

```python
plt.suptitle(suptitle, size=12 * fontsize / 17, wrap=True)
...
axs[r, i].set_title(titles[i], size=9 * fontsize / 17, wrap=True)
```

but it'd probably be cleaner to set the default title/suptitle font size when the default base font size is set.

Tmodrzyk:
Hi Jeremy, I submitted a PR related to this issue :)

jscanvic:
I somehow didn't see that - sorry!

PR #596, opened 2025-07-09
Fix issue #594 related to the fontsize parameter in ``plot``

Tmodrzyk:
The ``fontsize`` parameter in plot has no effect (Issue #594). The fix was very simple, the ``size`` parameter in some matplotlib function calls was overwriting the ``config_matplotlib`` call.


samuro95:
Thanks for the correction! Now the config_matplotlib sets the font.size to 17, when it was set to 9 or 12 before. Have you checked that it still looks good ? thanks

samuro95 (dismissed):
Thanks for the correction! Now the config_matplotlib sets the font.size to 17, when it was set to 9 or 12 before. Have you checked that it still looks good ? thanks 

Moreover, I am wondering if there is a possibility to call config_matplotlib() in the setup.py file so that all the documentation is run with these options, rather than having it in every example file

Tmodrzyk:
Hi Sam, thanks for your feedback, I did not think about the documentation, I'll check it. 
As for the ``setup.py`` file you mention I can't seem to find it, where is it located in the library ?

samuro95:
> Hi Sam, thanks for your feedback, I did not think about the documentation, I'll check it. As for the `setup.py` file you mention I can't seem to find it, where is it located in the library ?

Sorry I mean the conf.py file

Tmodrzyk:
>  Now the config_matplotlib sets the font.size to 17, when it was set to 9 or 12 before. Have you checked that it still looks good ? 

So indeed it was not looking good with fontsize 17. Now it's supposed to be set at 9 again in the examples.

> Moreover, I am wondering if there is a possibility to call config_matplotlib() in the setup.py file so that all the documentation is run with these options, rather than having it in every example file

I did that at the top of the ``conf.py`` file. However, calling ``plot`` with default value ``fontsize=17`` overwrites this setting, so I changed slightly the way this parameter is handled.

The default fontsize is back to 17 except when generating the doc examples, then the fontsize is 9.

Tmodrzyk:
I built the doc locally and checked that everything looks ok. 
Now the **default fontsize is 9** for every ``title`` and ``suptitle`` in the doc's plots.
@Andrewwango @samuro95 could you maybe build the doc too just to double check that I did not break everything ? :smiley:

deepinv/utils/plotting.py
```diff
@@ -255,7 +260,10 @@ def plot(
     :param bool return_axs: return the axs object.
     """
     # Use the matplotlib config from deepinv
-    config_matplotlib(fontsize=fontsize)
+    if fontsize is not None:
```
samuro95: Acutally, if we call config_matplotlib every time we call plot, I wonder if we need the config_matplotlib in the conf.py ... do we ?

examples/basics/demo_remote_sensing.py
```diff
@@ -199,6 +199,7 @@
     compare_no_learning=True,
     no_learning_method="A_dagger",
     show_progress_bar=False,
+    device="cpu",
```
samuro95: maybe we could add the possibility to run this example on GPU too. Why did you add this line as this is the default device of the trainer ?

examples/basics/demo_tour_mri.py
```diff
@@ -96,7 +97,7 @@
 
 dinv.utils.plot(
     {
-        "x": (x := next(iter(DataLoader(knee_dataset)))),
+        "x": (x := next(iter(DataLoader(knee_dataset))).to(device)),
```
samuro95: why do you add this ? If the device is gpu you do not want to have the image going to gpu for plotting.

examples/sampling/demo_diffpir.py
```diff
@@ -288,8 +283,8 @@ def get_noise_schedule(sigma, lambda_=7.0, num_train_timesteps=1000, max_iter=ma
 # In the algorithm, we will only use sub-sampled versions of the noise and regularization schedules. Let's visualize
 # those.
 
-list_sigmas_algo = [sigmas[seq[i]] for i in range(max_iter)]
-list_rhos_algo = [rhos[seq[i]] for i in range(max_iter)]
+list_sigmas_algo = [sigmas[seq[i]].cpu().item() for i in range(max_iter)]
+list_rhos_algo = [rhos[seq[i]].cpu().item() for i in range(max_iter)]
 
 plt.figure(figsize=(6, 3))
 plt.rcParams.update({"font.size": 9})
```
samuro95: not necessary here (and in the rest of the file) if we call the config_matlotlib

samuro95 (commented):
I added new comments with the new changes.

deepinv/utils/plotting.py
Tmodrzyk: I think we still need it, otherwise we'd have to either call ``fontsize=9`` in every example in the doc, or set the default fontsize value to 9 (it is currently at 17). 
I find this solution more adapted since we keep the former default value, but also keep the fontsize value related to the doc examples, and don't have to set it to 9 in every example.

mh-nguyen712:
We just updated the preview docs for pr. Can your merge with deepinv/main to generate the preview doc @Tmodrzyk ? Thanks!

examples/basics/demo_tour_mri.py
Tmodrzyk: You're right, this was a quick fix I added because the example wasn't running on GPU.
Since ``physics`` is already on the device, ``x`` needs to be moved to the same device as well.
I updated the code and now send ``x`` to device just before passing it to ``physics`` [5b944cf](https://github.com/deepinv/deepinv/pull/596/commits/5b944cf2c3c366be2cb028116518e5de59eb4eb8)

examples/sampling/demo_diffpir.py
Tmodrzyk: Absolutely, removed in [d923a36](https://github.com/deepinv/deepinv/pull/596/commits/d923a36cf9037df53494e5ebd081f02f4e9b43af)

examples/basics/demo_remote_sensing.py
Tmodrzyk: This is not the default device of the Trainer when a GPU is available. 
Unfortunately, running the example on GPU broke the example, so that was again a quick fix to make it build.
I fixed the example now. There was in particular an error in the class ``PanNet`` in ``mutispectral.py``, the network was not sent on the device in the ``init``. [70e6164](https://github.com/deepinv/deepinv/pull/596/commits/70e616488de2b4c67b5bef9d5fca8708386acdc2)

Tmodrzyk:
@mh-nguyen712 It seems I'm up to date with main, I'm not sure about what I should do ? 
Also, is the CI broken ? I have a ``permission denied`` when building the docs

mh-nguyen712:
Sorry for this bug. We just fixed it @Tmodrzyk . Can you merge with main and push again, please? The generated docs then can be found in the Actions. Thanks!

Andrewwango:
I've just merged from main to remove existing changes from this PR

Tmodrzyk:
Looking at the docs generated by the CI, it seems the fontsize is back at 17 for some reasons. I'll double check locally.

Tmodrzyk:
Now the fontsize works as expected: 17 by default, ``fontsize`` if specified, and 9 in the docs.
I'm not sure that the solution was the most elegant but it works.

Tmodrzyk:
@Andrewwango @samuro95 I think I'm done with this PR, let me know if there's any issue. :)

Andrewwango (approved):
LGTM, thanks!

jscanvic (changes requested):
Thanks for the fix!

The font size is really small in the generated docs for the page on 3D diffraction PSFs. Can you check it out?

main:

Image

PR:

Image

Tmodrzyk:
I changed the CSS to fix Issue #673 causing figures to be plotted side by side instead of one after the other.
I also increased the default font size for the examples and fixed some errors in other plotting functions.
@jscanvic let me know if you spot any other issues in the examples :)

jscanvic (approved):
LGTM thanks for fixing that!

PR #603, opened 2025-07-11
Fix DRUNet doc and signature

romainvo:
The DRUNet class has an unused argument in its documentation ``nf: number of channels per convolutional layer.`` which actually corresponds to the ``nc`` parameter. I've removed ``nf`` from the documentation.

I've also completed the documentation, which was slightly misleading. I was expecting to control the number of scales instanciated in the DRUNet with the ``nc`` argument, while in practice the number of scales is hard-coded and fixed to 4.


tachella (approved):
LGTM, thanks for fixing this!

PR #608, opened 2025-07-14
More MRI, more downsampling, and more fixes

### LINKED ISSUE #606: DownsamplingGenerator batching
DownsamplingGenerator's batching is inconsistent. Currently, `step()` returns `factor: int` and `filter: Tensor` of shape `(1, 1, k, k)`, and the `batch_size` is unused. This means it can't generate batched params. Furthermore, the factor is incompatible with `generate_dataset`.

https://github.com/deepinv/deepinv/blob/f304b6adc964e0054fd182d1044c7af11d2d116a/deepinv/physics/generator/downsampling.py#L6

It should return `factor: Tensor` of shape `(B, 1)` and `filter: Tensor` of shape `(B, 1, k, k)`. `physics.Downsampling` should also be modified. It should either allow batched physics with different factors or filters, or raise a warning that physics params batch size must be one (or all params are constant in one batch).

Andrewwango:
Fix downsampling generator batching + DownsamplingMatlab + better FastMRI raw data loading + SKM-TEA + minor fixes

1. Closes #606 

How does this affect you?
- Online measurements with batch size 1: will work as before, unaffected
- Any kind of measurements with batch size > 1:
  - Where `len(factors) == 1 and len(filters) == 1`: now generates a batch where all entries are same
  - Else: will not work as before (because filters will be different sizes)
- Offline measurements with batch size 1: now works!

2. MATLAB imresize-style downsampling physics (via Git submodule from [this repo](https://github.com/sanghyun-son/bicubic_pytorch))

3. Better FastMRI raw data loading:
  - Load mask if present in test set
  - Optionally load target from different folder (due to fastMRI data structure)
  - Prewhiten raw kspace
  - Normalise raw kspace
  - MRI tour example has better support for raw data

4. [SKM-TEA](https://github.com/StanfordMIMI/skm-tea) dataset

5. Minor fixes:

- Fix CMRxRecon dtype
- Spelling mistakes in examples
- Trainer load model optionally unstrict
- Trainer load params allow non-Tensors
- Fix LPIPS device when on cuda: not equal to 0


deepinv/physics/blur.py
```diff
@@ -156,22 +180,61 @@ def splits(a, sf):
...
+            if factor.ndim > 1:
+                raise ValueError("Factor tensor must be 1D.")
+            elif len(torch.unique(factor)) > 1:
+                raise ValueError(
+                    f"Downsampling only supports one unique factor per batch, but got factors {torch.unique(factor).tolist()}."
+                )
+            elif factor.ndim == 1:
+                factor = factor[0]
+
+            return int(factor.item())
```
jscanvic, line 203: I think it can be made clearer w/ something like that:

```
assert factor.ndim <= 1, "Error message"
factor = torch.unique(factor)
assert factor.numel() == 1, "Error message"
return int(factor.item())
```

deepinv/physics/blur.py
```diff
@@ -95,26 +109,36 @@ def A_adjoint(self, y, filter=None, factor=None, **kwargs):
...
+
+        .. warning::
+
+            If `factor` is passed, `filter` must also be passed as a `str` or `Tensor`, in order to update the filter to the new factor.
+
         """
+        if factor is not None:
+            self.factor = self.check_factor(factor)
+
+        self.imsize_dynamic = (
```
jscanvic, line 125: This pattern is really thread-unsafe (and scary to me!) but we should probably deal with it in a separate issue given it's already used pretty much everywhere in the code base

deepinv/physics/blur.py
```diff
@@ -709,3 +774,60 @@ def bicubic_filter(factor=2):
...
+        :param int, float factor: downsampling factor. If not `None`, use this factor and store it as current factor.
+        """
+        self.update_parameters(factor=factor, **kwargs)
+        # Clone because of in-place ops
+        return self.imresize(
+            x.clone(),
+            scale=1 / self.factor,
+            antialiasing=True,
+            kernel="cubic",
+            padding_type="reflect",
```
jscanvic: Make those instance parameters?

deepinv/physics/functional/downsampling.py
jscanvic, line 1: Nice!

examples/basics/demo_train_inpainting.py
```diff
@@ -69,7 +69,7 @@
 )
 
 
-# Use parallel dataloader if using a GPU to fasten training,
```
jscanvic, line 72: Good catch!

deepinv/loss/metric/perceptual.py
```diff
@@ -77,9 +77,9 @@ class NIQE(Metric):
     def __init__(self, device="cpu", check_input_range=False, **kwargs):
         super().__init__(**kwargs)
         pyiqa = import_pyiqa()
-        self.niqe = pyiqa.create_metric("niqe", check_input_range=check_input_range).to(
-            device
-        )
+        self.niqe = pyiqa.create_metric(
+            "niqe", check_input_range=check_input_range, device=device
+        ).to(device)
```
jscanvic, line 82: Do we need both device=device and to(device)?

deepinv/tests/test_physics.py
```diff
@@ -1115,26 +1143,22 @@ def test_downsampling_adjointness(device):
...
-                    Axy = torch.sum(Ax * y)
-                    Atyx = torch.sum(Aty * x)
+                Ax = physics.A(x)
+                y = torch.rand_like(Ax)
+                Aty = physics.A_adjoint(y)
+                Axy = torch.sum(Ax * y)
+                Atyx = torch.sum(Aty * x)
 
-                    assert torch.abs(Axy - Atyx) < 1e-3
+                assert torch.abs(Axy - Atyx) < 1e-3
```
jscanvic, line 1143: Is there a reason we don't use adjointness_test here?

deepinv/tests/test_physics.py
```diff
@@ -810,6 +831,10 @@ def test_concatenation(name, device):
     if "pansharpen" in name:  # TODO: fix pansharpening
         return
     physics, imsize, _, dtype = find_operator(name, device)
+
+    if name == "super_resolution_matlab":
+        pytest.skip("DownsamplingMatlab has bad adjoint.")
```
jscanvic: Should we favor the autograd implementation instead?

deepinv/training/trainer.py
```diff
@@ -382,7 +385,7 @@ def load_model(self, ckpt_pretrained: Union[str, Path] = None) -> dict:
             checkpoint = torch.load(
                 ckpt_pretrained, map_location=self.device, weights_only=False
             )
-            self.model.load_state_dict(checkpoint["state_dict"])
+            self.model.load_state_dict(checkpoint["state_dict"], strict=strict)
```
jscanvic, line 395: Nice!

deepinv/datasets/fastmri.py
```diff
@@ -559,49 +616,110 @@ def generate_mask(
...
+            kspace.unsqueeze(0), calib_size=self.get_acs(metadata=metadata)
         ).squeeze(0)
 
+    def prewhiten_kspace(self, kspace: torch.Tensor) -> torch.Tensor:
+        """Prewhiten kspace using Cholesky decomposition.
+
+        :param torch.Tensor kspace: input multicoil kspace of shape (2, N, H, W)
+        :return: whitened kspace.
+        """
+        if len(kspace.shape) < 4:
```
jscanvic: kspace.ndim

deepinv/datasets/fastmri.py
```diff
@@ -559,49 +616,110 @@ def generate_mask(
...
+            raise ValueError("kspace must be multicoil for prewhitening.")
+
+        ksp = self.to_torch_complex(kspace.unsqueeze(0)).squeeze(0)  # N, H, W complex
+
+        n = ksp[:, self.prewhiten[0], self.prewhiten[1]].flatten(-2)
+        n -= n.mean(axis=-1, keepdim=True)
```
jscanvic: I believe this mutates the input `kspace` - can we use a n = n - n.mean(...) instead if it's not too expensive?

jscanvic (changes requested):
Looks really good so far - thanks Andrew!

deepinv/loss/metric/perceptual.py
Andrewwango: Probably not but I don't trust `pyiqa` so better to be safer?

deepinv/tests/test_physics.py
Andrewwango: Yes, I'm not sure the implemented adjoint is correct. Let `A = imresize(1 / factor, bicubic)`. Do you think it is appropriate that `A_adjoint = imresize(factor, bicubic)`? Idk what a better solution would be. This current solution is a bit dissatisfying, where we have to skip the adjointness tests

deepinv/tests/test_physics.py
Andrewwango: I'll benchmark it!

deepinv/loss/metric/perceptual.py
jscanvic: Right - agreed!

deepinv/physics/blur.py
Andrewwango: Ooh ok, that's a great point. The functionality should have not changed at all if you pass in imsize in the init as you do before. But if you want to use dynamic imsize, thanks for raising this! Please can you add to my new issue in #649 ?

deepinv/tests/test_physics.py
Andrewwango: But good point, I'm not sure the implemented adjoint is correct. Let `A = imresize(1 / factor, bicubic)`. Do you think it is appropriate that `A_adjoint = imresize(factor, bicubic)`? Idk what a better solution would be. This current solution is a bit dissatisfying, where we have to skip the adjointness tests

deepinv/tests/test_physics.py
Andrewwango: Not sure - we can raise the issue independently?

deepinv/tests/test_physics.py
Andrewwango: Results: the autograd doesn't seem great

```python
x = load_example("butterfly.png")
physics = dinv.physics.DownsamplingMatlab()
y = physics(x)

def adjoint1(y):
    return physics.A_adjoint(y)

def adjoint2(y):
    adjoint = dinv.physics.adjoint_function(physics.A, (1, 3, 256, 256))
    return adjoint(y)

m = dinv.metric.PSNR()
for adjoint in (adjoint1, adjoint2):
    print(m(physics(adjoint(y)), y))

tensor([34.5253])
tensor([8.0724])
```

jscanvic: Let's keep the imresize implementation then it shouldn't be too far from the true adjoint!

I'm a bit puzzled though... from what I checked:

* Adjointness test: imresize does not pass & autograd adjoint passes
* Spectral norm: 1 for imresize & 0.5 for autograd adjoint (I expected 1)
* The bad PSNR for A^\top A y vs y for autograd adjoint seems to be because A is not normalized

This is very weird!

jscanvic (approved):
LGTM

deepinv/tests/test_physics.py
Andrewwango: See #652

deepinv/tests/test_physics.py
mh-nguyen712: I think the adjoint function by autograd should be ok:
```python
for adjoint in (adjoint1, adjoint2):
    ax = physics(x)
    y = torch.randn_like(ax)
    aty = adjoint(y)
    left = (ax * y).sum()
    right = (x * aty).sum()
    print(f"Adjoint check: {left.item():.4f} vs {right. Item():.4f}")

Adjoint check: -19.6100 vs -78.4401
Adjoint check: -118.5649 vs -118.5649
```

However it's much slower than `imresize` (~4x slower in my PC)

mh-nguyen712 (changes requested):
Thanks for the hard work Andrew. I just have a comment on the adjoint of downsampling. I think the autograd adjoint gives good results but it's a bit slower. I think it's better to use autograd and then find a way to use imresize the the future.

deepinv/physics/blur.py
```diff
@@ -709,3 +774,74 @@ def bicubic_filter(factor=2):
...
+        antialiasing: bool = True,
+        **kwargs,
+    ):
+        super().__init__(filter=None, factor=factor, **kwargs)
+
+        try:
+            from bicubic_pytorch import imresize
+        except ImportError:
+            raise ImportError(
+                "Install 'bicubic_pytorch' using 'pip install bicubic-pytorch'"
```
mh-nguyen712: It seems that the core functions take ~500 lines of code. Could we add them directly in the library with a citation to the original repo. Just to remove a dependency...

mh-nguyen712: This is essentially what we did for defining denoiser models.

deepinv/tests/test_physics.py
Andrewwango: Thanks @mh-nguyen712 ! I will address this

deepinv/physics/blur.py
Andrewwango: Yes there are 3 options:
1. Add directly to library (I don't like this as then I want to test it and change all the formatting, docstrings etc.)
2. Add as submodule (I tried it initially, it's ok but it kind of sits alone as the only submodule, and maybe it suggests it's more important than it actually is)
3. Add as optional dependency - yes it adds another dependency but it sits alongside other optional specific dependencies like libcpab or sigpy

mh-nguyen712: I know that it's a lot of works to add it directly to the library... but then as we can test it, it will be more stable and compatible with deepinv then relying on a repo which is no longer maintained. 
We can just leave the formatting and docstring as it is now. If someone want to improve it, or add the adjoint function, it would be easier. 
 
Just my personal opinion.

Andrewwango: @tachella  any opinions? I'm happy to do either

tachella: I agree with Hai, ideally I would add it and test it

deepinv/tests/test_physics.py
Andrewwango: Done

deepinv/physics/blur.py
Andrewwango: Done

deepinv/datasets/fastmri.py
```diff
@@ -516,38 +539,72 @@ def save_simple_dataset(
...
+    :param tuple[slice, slice], bool prewhiten: if `True`, prewhiten kspace noise across coils,
+        defaults to using a 30x30 slice in the top left corner. Optionally set tuple of slices for custom location. Defaults to False.
+    :param bool normalise: if `True`, normalise kspace by 99th percentile of RSS reconstruction of kspace ACS block.
+        if `int` or `float`, normalise kspace by `normalise / kspace.max()`.
     """
 
     def __init__(
         self,
         mask_generator: Optional[BaseMaskGenerator] = None,
+        seed_mask_generator: bool = True,
```
mh-nguyen712, line 580: I think this is not necessary. If someone gives a seed in forward, it means they want set the seed of rng, otherwise it does nothing default.

deepinv/datasets/fastmri.py
```diff
@@ -516,38 +539,72 @@ def save_simple_dataset(
...
-        If `int`, pass this as auto-calibration size to `ESPIRiT <https://onlinelibrary.wiley.com/doi/10.1002/mrm.24751>`_. If `True`, use ACS size from `mask_generator`.
+    :param bool seed_mask_generator: if `True`, generated mask for given kspace is **always** the same.
+        This should be `True` for test set. For supervised training, set to `False` for higher diversity in kspace undersampling.
+    :param bool, int estimate_coil_maps: if `True`, estimate coil maps using :func:`deepinv.physics.MultiCoilMRI.estimate_coil_maps`.
+    :param int acs: optional number of low frequency lines for autocalibration. If `None`, look for acs lines in `mask_generator` attributes (if exists)
+        or in metadata (only available for FastMRI test/challenge data). If unavailable, and ACS required, then raises error.
+    :param tuple[slice, slice], bool prewhiten: if `True`, prewhiten kspace noise across coils,
+        defaults to using a 30x30 slice in the top left corner. Optionally set tuple of slices for custom location. Defaults to False.
+    :param bool normalise: if `True`, normalise kspace by 99th percentile of RSS reconstruction of kspace ACS block.
```
mh-nguyen712, line 573: Minor comments: I though that we're using American English everywhere. May be I'm wrong?

deepinv/datasets/fastmri.py
```diff
@@ -516,38 +539,72 @@ def save_simple_dataset(
...
+            return self.acs
+
+        if self.mask_generator is not None:
+            return self.mask_generator.n_center
+
+        if "acs" in metadata:
+            return metadata["acs"]
+
+        raise ValueError(
+            "ACS size not specified. Either define fixed acs, or ensure metadata has acs attribute, orpass in  mask_generator with fixed ACS size."
```
mh-nguyen712:
```suggestion
            "ACS size not specified. Either define fixed acs, or ensure metadata has acs attribute, or pass in  mask_generator with fixed ACS size."
```

deepinv/datasets/fastmri.py
Andrewwango: You're right. I can open an issue separately and leave this for now

deepinv/datasets/fastmri.py
Andrewwango: What do you mean? Passing `seed_mask_generator` as True or False gives different behaviours, both of which are desirable in different scenarios.

deepinv/datasets/fastmri.py
```diff
@@ -516,38 +539,72 @@ def save_simple_dataset(
...
         self.mask_generator = mask_generator
+        self.seed_mask_generator = seed_mask_generator
         self.estimate_coil_maps = estimate_coil_maps
+        self.acs = acs
+        self.prewhiten = prewhiten
+        if self.prewhiten is True:
+            self.prewhiten = (slice(0, 30), slice(0, 30))
+        self.normalise = normalise
+
+    def get_acs(self, metadata: dict = {}):
```
mh-nguyen712: Mutable default arg.

```suggestion
    def get_acs(self, metadata: dict = None):
```

deepinv/datasets/fastmri.py
```diff
@@ -559,49 +616,110 @@ def generate_mask(
...
         """
         return self.mask_generator.step(
-            seed=seed,
+            seed=seed if self.seed_mask_generator else None,
             img_size=kspace.shape[-2:],
             batch_size=0,
         )["mask"]
 
-    def generate_maps(self, kspace: torch.Tensor) -> torch.Tensor:
+    def generate_maps(self, kspace: torch.Tensor, metadata: dict = {}) -> torch.Tensor:
```
mh-nguyen712:
```suggestion
    def generate_maps(self, kspace: torch.Tensor, metadata: dict = None) -> torch.Tensor:
```

deepinv/datasets/fastmri.py
```diff
@@ -559,49 +616,110 @@ def generate_mask(
...
+            ksp = ksp.flatten(-2)  # N, H*W
+            ksp = torch.linalg.solve(torch.linalg.cholesky(cov), ksp)
+            return self.from_torch_complex(
+                ksp.unflatten(-1, kspace.shape[-2:]).unsqueeze(0)
+            ).squeeze(0)
+        except torch.linalg.LinAlgError:
+            return kspace  # Non-PSD due to ksp all zeros
+
+    def normalise_kspace(
+        self, kspace: torch.Tensor, metadata: dict = {}
```
mh-nguyen712:
```suggestion
        self, kspace: torch.Tensor, metadata: dict = None
```

deepinv/datasets/fastmri.py
```diff
@@ -559,49 +616,110 @@ def generate_mask(
...
+            self.kspace_to_im((kspace * mask).unsqueeze(0))
+        ).quantile(0.99)
+
     def __call__(
         self,
         target: torch.Tensor,
         kspace: torch.Tensor,
         mask: torch.Tensor = None,
         seed: Union[str, int] = None,
+        metadata: dict = {},
```
mh-nguyen712:
```suggestion
        metadata: dict = None,
```

deepinv/datasets/fastmri.py
Andrewwango: #656

deepinv/datasets/fastmri.py
```diff
@@ -559,49 +616,110 @@ def generate_mask(
...
+        cov = n @ n.H  # (N, N) complex
+
+        try:
+            ksp = ksp.flatten(-2)  # N, H*W
+            ksp = torch.linalg.solve(torch.linalg.cholesky(cov), ksp)
+            return self.from_torch_complex(
+                ksp.unflatten(-1, kspace.shape[-2:]).unsqueeze(0)
+            ).squeeze(0)
+        except torch.linalg.LinAlgError:
+            return kspace  # Non-PSD due to ksp all zeros
```
mh-nguyen712: It's a bit cleaner if we give a warning here?

deepinv/datasets/fastmri.py
```diff
@@ -149,33 +149,34 @@ def __len__(self):
 
 
 class FastMRISliceDataset(torch.utils.data.Dataset, MRIMixin):
-    """Dataset for `fastMRI <https://fastmri.med.nyu.edu/>`_ that provides access to raw MR image slices.
+    """Dataset for `fastMRI <https://fastmri.med.nyu.edu/>`_ that provides access to raw MR kspace data.
```
mh-nguyen712, line 159: Note related but there is a redundant import of `from typing import Any`, can you remove it please

deepinv/physics/functional/imresize.py
```diff
@@ -0,0 +1,360 @@
...
+        sample = unfold[..., base]
+        weight = weight.view(1, kernel_size, 1, sample.size(3))
+
+    # Apply the kernel
+    x = sample * weight
+    x = x.sum(dim=1, keepdim=True)
+    return x
+
+
+def imresize(
```
mh-nguyen712: call it imresize_matlab?

deepinv/physics/functional/imresize.py
```diff
@@ -0,0 +1,360 @@
...
+        if c is not None:
+            x = x.view(c, rh, rw)  # 3-dim
+        else:
+            x = x.view(rh, rw)  # 2-dim
+
+    return x
+
+
+def cast_input(x: torch.Tensor) -> tuple[torch.Tensor, _D]:
+    if x.dtype != torch.float32 or x.dtype != torch.float64:
```
mh-nguyen712: This logic will always be True I think.

```suggestion
    if x.dtype not in (torch.float64, torch.float32):
```

deepinv/physics/functional/imresize.py
```diff
@@ -0,0 +1,360 @@
...
+    x = sample * weight
+    x = x.sum(dim=1, keepdim=True)
+    return x
+
+
+def imresize(
+    x: torch.Tensor,
+    scale: Optional[float] = None,
+    sizes: Optional[tuple[int, int]] = None,
+    kernel: Union[str, torch.Tensor] = "cubic",
```
mh-nguyen712: in `resize_1d` only `str` is supported
```suggestion
    kernel: str = "cubic",
```

deepinv/physics/functional/imresize.py
```diff
@@ -0,0 +1,360 @@
...
+    sigma: float = 2,
+    padding_type: str = "reflect",
+    antialiasing: bool = True,
+) -> torch.Tensor:
+    """MATLAB imresize reimplementation.
+
+    A standalone PyTorch implementation for fast and efficient bicubic resampling.
+    The resulting values are the same to MATLAB function imresize('bicubic').
```
mh-nguyen712: It seems that only `reflect` padding is implemented?
```suggestion
    The resulting values are the same to MATLAB function `imresize('bicubic')` with `reflect` padding.
```

deepinv/physics/generator/downsampling.py
```diff
@@ -88,17 +89,36 @@ def step(
...
+        filter_indices = torch.randint(
+            low=0,
+            high=len(self.list_filters),
+            size=(batch_size,),
+            generator=self.rng,
+            **self.factory_kwargs,
+        )
+        factors = torch.Tensor(
+            [self.list_factors[int(i)] for i in factor_indices.tolist()]
+        )
```
mh-nguyen712: `torch.Tensor` will always return a FloatTensor, I'm not sure if it is intended
```suggestion
        factors = torch.tensor(self.list_factors, device=self.device)[factor_indices]
```

mh-nguyen712: Or using directly `dinv.physics.functional.rand.random_choice`

deepinv/physics/blur.py
```diff
@@ -709,3 +775,76 @@ def bicubic_filter(factor=2):
...
+
+        if self.adjoint_via_backprop:
+            adj = adjoint_function(
+                self.A,
+                (*y.shape[:2], y.shape[-2] * self.factor, y.shape[-1] * self.factor),
+                device=y.device,
+                dtype=y.dtype,
+            )
+            return adj(y)
+        else:
```
mh-nguyen712: I think we should only use autograd for now, since imresize with 1/factor is not the adjoint. If you're ok with that, update the docstring as well please

deepinv/tests/test_physics.py
```diff
@@ -1167,6 +1177,25 @@ def test_prox_l2_downsampling(device):
...
+    x = torch.rand(1, channels, *imsize, device=device)
+    physics = downsampling(device=device, factor=factor)
+    assert physics(x).shape == (1, channels, imsize[0] // factor, imsize[1] // factor)
+    assert physics.A_adjoint(x).shape == (
+        1,
+        channels,
+        imsize[0] * factor,
+        imsize[1] * factor,
+    )
+
```
mh-nguyen712, line 1194: Adjointness test?

deepinv/tests/test_physics_functional.py
```diff
@@ -134,3 +134,26 @@ def test_conv3d_adjointness(device):
...
+        sigma=sigma,
+        padding_type="reflect",
+        antialiasing=antialiasing,
+    )
+    assert y.shape == (
+        1,
+        img_size[0],
+        int(img_size[1] * scale),
+        int(img_size[2] * scale),
+    )
```
mh-nguyen712, line 159: Do you have Matlab license so we can test it? :D

deepinv/tests/test_trainer.py
```diff
@@ -737,7 +737,7 @@ def test_total_loss(dummy_dataset, imsize, device, dummy_model, tmpdir):
     train_data, eval_data = dummy_dataset, dummy_dataset
     dataloader = DataLoader(train_data, batch_size=2)
     eval_dataloader = DataLoader(eval_data, batch_size=2)
-    physics = dinv.physics.Inpainting(tensor_size=imsize, device=device, mask=0.5)
+    physics = dinv.physics.Inpainting(img_size=imsize, device=device, mask=0.5)
```
mh-nguyen712, line 744: Thanks!

deepinv/datasets/fastmri.py
mh-nguyen712: I meant these different behaviors can be achieved by providing `seed=int` or `seed=None` in `generate_mask`

Andrewwango: I guess `generate_mask` isn't a public method, the transform calls it when the transform is called, so I think the options should all be set in the transform init

deepinv/tests/test_physics_functional.py
Andrewwango: I don't think we can autotest it.

In my own tests I tested using div2k dataset and it's still not 100% exactly equal to the div2k downsampled images, the PSNR is like ~60dB

deepinv/physics/generator/downsampling.py
Andrewwango: You can't do this fix with filters below because they are strs, so I wanted to keep the factors the same

deepinv/tests/test_physics_functional.py
mh-nguyen712: I meant a hard test with a fix tensor, e.g. `matlab_output = tensor([1, 2, 3])`. But 60dB is good enough I think.

deepinv/tests/test_physics.py
```diff
@@ -345,6 +348,12 @@ def find_operator(name, device, get_physics_param=False):
             filter=None,
         )
         params = []
+    elif name == "super_resolution_matlab":
+        img_size = (1, 32, 32)
+        factor = 2
+        norm = 1.0 / factor**2
+        p = dinv.physics.DownsamplingMatlab(factor=factor, adjoint_via_backprop=True)
```
mh-nguyen712: I think you removed `adjoint_via_backprop`

mh-nguyen712 (commented):
Just a last minor comments otherwise LGTM! Thanks Andrew!

jscanvic (approved):
LGTM besides Hai's comment!

deepinv/tests/test_physics_functional.py
Andrewwango: Let me add a test!

Andrewwango: Done

deepinv/tests/test_physics.py
mh-nguyen712:
```suggestion
        p = dinv.physics.DownsamplingMatlab(factor=factor)
```

mh-nguyen712 (approved):
Can you just update the last comment in `test_physics`? Then we're good! Thanks for the PR Andrew!

Andrewwango:
I already removed it in https://github.com/deepinv/deepinv/pull/608/commits/d5b50fce111d6dc936bfe13a0ed15bfca99e7383? @mh-nguyen712

mh-nguyen712:
Sorry maybe I didn't reload the page. I think we can merge!

PR #612, opened 2025-07-16
DOC pass on DPS example

mathurinm:
This is just a cosmetic pass.

However I'm also considering modifying the example :
- to use a Face image, since the DiffU?et model was trained on faces
- to use inpainting with a square mask (in my limited experience, inpainting with random mask is not very hard and does not really require using a generative model) 
- to generate two samples according to the posterior, by changing the seed (this increases the runtime of the algorithm, maybe just specifying that different samples can be generated by changing the seed would be enough)

Here's what some early results with 2 differents seeds look like : 
image

@samuro95 @matthieutrs @mh-nguyen712 let me know if you think this would be itneresting, or if the current coesmtic change is enough.


mathurinm:
Also in the maths the current notation is $\hat x_t(x_t)$, e.g.
image

I wonder if it would not make more sense to use $\hat x_0(x_t)$ ? Since it's something that lives in the space of clean images, corresponding to time 0.
This would be in accordance with the notation of the DPS article : 
image

samuro95:
Sounds good to me !

examples/sampling/demo_dps.py
```diff
@@ -89,24 +89,8 @@
...
-# Utility function to let us easily retrieve \bar\alpha_t
-def compute_alpha(beta, t):
-    beta = torch.cat([torch.zeros(1).to(beta.device), beta], dim=0)
-    a = (1 - beta).cumprod(dim=0).index_select(0, t + 1).view(-1, 1, 1, 1)
-    return a
-
-
-betas = get_betas()
-
+betas = torch.linspace(1e-4, 1e-2, num_train_timesteps).to(device)
```
mathurinm:
```suggestion
betas = torch.linspace(1e-4, 2e-2, num_train_timesteps).to(device)
```

examples/sampling/demo_dps.py
```diff
@@ -284,9 +267,9 @@ def compute_alpha(beta, t):
 batch_size = 1
 eta = 1.0
 
-seq = range(0, num_train_timesteps, skip)
-seq_next = [-1] + list(seq[:-1])
-time_pairs = list(zip(reversed(seq), reversed(seq_next)))
+# seq = range(0, num_train_timesteps, skip)
+# seq_next = [-1] + list(seq[:-1])
+# time_pairs = list(zip(reversed(seq), reversed(seq_next)))
```
mathurinm:
```suggestion
```

Andrewwango:
Please can you just run `black .` before proceeidng

examples/sampling/demo_dps.py
```diff
@@ -89,24 +89,8 @@
...
-def compute_alpha(beta, t):
-    beta = torch.cat([torch.zeros(1).to(beta.device), beta], dim=0)
-    a = (1 - beta).cumprod(dim=0).index_select(0, t + 1).view(-1, 1, 1, 1)
-    return a
-
-
-betas = get_betas()
-
+betas = torch.linspace(1e-4, 2e-2, num_train_timesteps).to(device)
+alphas = (1 - betas).cumprod(dim=0)
```
mathurinm, line 93: @Andrewwango this seems way simpler to me, was I missing something?

examples/sampling/demo_dps.py
```diff
@@ -119,18 +103,18 @@ def compute_alpha(beta, t):
...
+#         \widehat{\mathbf{x}}_{0} (\mathbf{x}_t) &= \denoiser{\mathbf{x}_t}{\sqrt{1-\overline{\alpha}_t}/\sqrt{\overline{\alpha}_t}}
 #         \\
-#         \mathbf{g}_t &= \nabla_{\mathbf{x}_t} \log p( \widehat{\mathbf{x}}_{t}(\mathbf{x}_t) | \mathbf{y} ) \\
+#         \mathbf{g}_t &= \nabla_{\mathbf{x}_t} \log p( \widehat{\mathbf{x}}_{0}(\mathbf{x}_t) | \mathbf{y} ) \\
 #         \mathbf{\varepsilon}_t &= \mathcal{N}(0, \mathbf{I}) \\
 #         \mathbf{x}_{t-1} &= a_t \,\, \mathbf{x}_t
-#         + b_t \, \, \widehat{\mathbf{x}}_t
-#         + \tilde{\sigma}_t \, \, \mathbf{\varepsilon}_t + \mathbf{g}_t,
+#         + b_t \, \, \widehat{\mathbf{x}}_0
+#         + \tilde{\sigma}_t \, \, \mathbf{\varepsilon}_t - \mathbf{g}_t,
```
mathurinm: just to double check, previous gradient sign was wrong @Andrewwango ?

mh-nguyen712: I think if we define g_t as `\mathbf{g}_t &= \nabla_{\mathbf{x}_t} \log p( \widehat{\mathbf{x}}_{0}(\mathbf{x}_t) | \mathbf{y} )`, then the sign should be a `+`?

mh-nguyen712 (commented):
Thanks @mathurinm. Just have a minor comment about the sign of gradient

examples/sampling/demo_dps.py
mathurinm: you're right!

mathurinm:
```suggestion
#         + \tilde{\sigma}_t \, \, \mathbf{\varepsilon}_t + \mathbf{g}_t,
```

mathurinm:
@Andrewwango this is ready for merge! thanks for the feedbacks

mh-nguyen712:
Thanks @mathurinm! I look at the generated docs so finally you didn't change the image in the example to a face from FFHQ 256? I understand that this would increase significantly the runtime of the example (from ~1min30s to ~8mins).... What do you think @samuro95 @matthieutrs ?

mathurinm:
Yes in the end I kept the butterfly in 64x64, for me the running time of the examples is already a pain point so I don't want to increase it ; 8 mins for an example is a lot

It can be discussed in a subsequent PR (bc there are additional  problems : as you see below for the 256x256 butterfly + inpainting with a square mask the result is not good) ; but I prefer to merge those changes first and discuss the other issue separately

image

mh-nguyen712:
Agree to merge these changes first.

Issue #593, opened 2025-07-09
`Trainer` compatibility with `Denoiser`


tachella:
It is not currently straightforward to train/test a `deepinv.models.Denoiser` with the `deepinv.Trainer`, since it inputs a `y, sigma` instead of `y, physics` associated with `deepinv.models.Reconstructor`. Moreover, losses are not compatible with `y, sigma` inputs.

This can be solved by wrapping a denoiser in a `deepinv.models.ArtifactRemoval`, which converts a `deepinv.models.Denoiser` into a `deepinv.models.Reconstructor`

It would be nice to automatically wrap the denoiser when passed to Trainer, instead of asking the user to do it on their own.

Andrewwango:
Nice, and when accessing Trainer.model you should naturally retrieve the base denoiser

Tmodrzyk:
I submitted a PR for this issue (#614)

PR #575, opened 2025-06-29
Add bibtex style cite for references 

samuro95:
Change all references to a bibtex style cite. 

In order to cite a paper :
- Add the bibtex reference in the `docs/sources/refs.bib`
- use 

>  :footcite:t:`bibtex_name`

 for citing in the style :  `Author1 et. al [1] `
or just 

> :footcite:`bibtex_name`

 for citing only with the breakets : ` [1] `

I also added a paragraoh in the `contributing.rst` documentation file on how to write docstring.

### Checks to be done before submitting your PR

- [x] `python3 -m pytest deepinv/tests` runs successfully.
- [x] `black .` runs successfully.
- [x] `make html` runs successfully (in the `docs/` directory).
- [x] Updated docstrings related to the changes (as applicable).
- [x] Added an entry to the [CHANGELOG.rst](../CHANGELOG.rst).

PR #409, opened 2025-02-21
Adversarial + GAN fixes

Andrewwango:
- Fix UAIR adversarial loss
- Add tests for discriminator models
- Add another simple discriminator architecture from [Cole et al.](https://openaccess.thecvf.com/content/ICCV2021W/LCI/papers/Cole_Fast_Unsupervised_MRI_Reconstruction_Without_Fully-Sampled_Ground_Truth_Data_Using_ICCVW_2021_paper.pdf) + their loss function `MultiOperatorUnsupAdversarialLoss`
- Add discriminator from [Real-ESRGAN](https://github.com/xinntao/Real-ESRGAN) (UNetDiscriminatorSN)
- Update adversarial trainer to latest trainer format
- Remove adversarial trainer
- Discrimiantors are now trained in the losses
- Add MOI loss test
- Simplify adversarial training example
- Add adversarial finetuning example

### Checks to be done before submitting your PR
- [x] `python3 -m pytest tests/` runs successfully.
- [x] `black .` runs successfully.
- [x] `make html` runs successfully (in the `docs/` directory).
- [x] Updated docstrings related to the changes (as applicable).
- [x] Added an entry to the [CHANGELOG.rst](../CHANGELOG.rst).

TODO
- [x] fix julian comments
- [x] fix docs
- [x] add DiscriminatorMetric to User Guide/API
- [x] Discriminators: explain discrim loading mechanics
- [x] Discriminators: pretrained discriminator (e.g. that of stylegan3)
- [x] Remove adversarial trainer/optimizer/scheduler and port all the functionality (adversarial optimizer saving etc)
- [x] D step ratio (might be tricky)
- [x] loss store D train history
- [x] adversarial loss docstring params should include all for multioperator ones
- [x] multioperator mixin to mixins + generalise for MOI
- [x] **Fix example + new example for just fixed pretrained discrim perceptual FT**
- [x] Add tests for helper funcs (e.g. in adversarialloss) - note core functionality tested already

PR #614, opened 2025-07-17
add ``Trainer`` compatibility with ``Denoiser``

Tmodrzyk:
This PR follows the suggestion from @tachella in issue #593. 
When the model passed to the ``Trainer`` is a ``Denoiser``, it is wrapped in an ``ArtifactRemoval`` class. It is unwrapped at the end of the training, and during intermediate checkpoints to preserve the ``state_dict``.

:warning:  Note that this fixes the issue for the specific case of ``Denoiser``, but if we later want to support more general models (e.g., those conditioned on additional inputs such as segmentations or other images), the Trainer may require further generalization.
Let me know if I should add a gallery example or notebook on how to train a denoiser, I think that is a fairly common use case of the library. :smile: 

### DIFF
--- a/deepinv/training/trainer.py
+++ b/deepinv/training/trainer.py
@@ -15,6 +15,7 @@ from deepinv.physics.generator import PhysicsGenerator
 from deepinv.utils.plotting import prepare_images
 from torchvision.utils import save_image
 import inspect
+from deepinv.models import ArtifactRemoval, Denoiser
 
 
 @dataclass
@@ -233,6 +234,11 @@ class Trainer:
     verbose_individual_losses: bool = True
     show_progress_bar: bool = True
 
+    def setup_denoiser(self):
+        self.model = ArtifactRemoval(
+            backbone_net=self.model, mode="direct", device=self.device
+        )
+
     def setup_train(self, train=True, **kwargs):
         r"""
         Set up the training process.
@@ -366,6 +372,12 @@ class Trainer:
         self.save_folder_im = None
 
         _ = self.load_model()
+
+        if isinstance(self.model, Denoiser):
+            self.is_denoiser = True
+            self.setup_denoiser()
+        else:
+            self.is_denoiser = False
 
     def load_model(self, ckpt_pretrained: Union[str, Path] = None) -> dict:
         """Load model from checkpoint.
@@ -911,7 +923,11 @@ class Trainer:
         os.makedirs(str(self.save_path), exist_ok=True)
         state = state | {
             "epoch": epoch,
-            "state_dict": self.model.state_dict(),
+            "state_dict": (
+                self.model.backbone_net.state_dict()
+                if self.is_denoiser
+                else self.model.state_dict()
+            ),
             "loss": self.loss_history,
             "optimizer": self.optimizer.state_dict() if self.optimizer else None,
             "scheduler": self.scheduler.state_dict() if self.scheduler else None,
@@ -1156,6 +1172,9 @@ class Trainer:
         if self.wandb_vis:
             wandb.save("model.h5")
             wandb.finish()
+
+        if self.is_denoiser:
+            self.model = self.model.backbone_net
 
         return self.model
 
--- a/deepinv/tests/test_trainer.py
+++ b/deepinv/tests/test_trainer.py
@@ -864,3 +864,34 @@ def test_out_dir_collision_detection(
                 )
 
                 trainer.train()
+
+
+# Test that when training a denoiser, the model is unwrapped correctly at
+# the end of training.
+def test_denoiser_unwrap(dummy_dataset, imsize, device, dummy_model, tmpdir):
+    train_data, eval_data = dummy_dataset, dummy_dataset
+    dataloader = DataLoader(train_data, batch_size=2)
+    physics = dinv.physics.Denoising(
+        noise_model=dinv.physics.GaussianNoise(), device=device
+    )
+
+    model = dinv.models.DRUNet().to(device)
+
+    trainer = dinv.Trainer(
+        model,
+        device=device,
+        save_path=tmpdir,
+        verbose=True,
+        show_progress_bar=False,
+        physics=physics,
+        epochs=2,
+        losses=dinv.loss.SupLoss(),
+        optimizer=torch.optim.AdamW(model.parameters(), lr=1e-3),
+        train_dataloader=dataloader,
+        online_measurements=True,
+        check_grad=True,
+    )
+    model = trainer.train()
+
+    # Check that the model is unwrapped correctly
+    assert not hasattr(trainer.model, "backbone_net")


Tmodrzyk:
It seems that a lot of models / class inherit from ``Denoiser`` without ever using the noise level ``sigma``, for instance, `DnCNN`, `AutoEncoder`, `Restormer`, `SwinIR`, `UNet`, etc.
In this PR, I check whether a model is an instance of `Denoiser` before wrapping it in `ArtifactRemoval`, to ensure that `sigma` is passed instead of `physics`. 
But this test does not work as intended if some denoisers don't use ``sigma``.

I'm wondering if we really want these models to inherit from `Denoiser` if they don’t use the noise level? Wouldn't it be cleaner for them to simply subclass `nn.Module` unless they explicitly require `sigma`?

deepinv/training/trainer.py
```diff
@@ -233,6 +234,12 @@ class Trainer:
     verbose_individual_losses: bool = True
     show_progress_bar: bool = True
 
+    def setup_denoiser(self):
```
Andrewwango, line 243: just a very short one line docstring here please

deepinv/training/trainer.py
```diff
@@ -15,6 +15,7 @@
 from deepinv.utils.plotting import prepare_images
 from torchvision.utils import save_image
 import inspect
+from deepinv.models import ArtifactRemoval, Denoiser
 
 
 @dataclass
```
Andrewwango, line 28: Please can you add a small note to the docstring about `model` that if you pass in a denoiser, it automatically gets converted to an `ArtifactRemoval` for training?

Andrewwango (changes requested):
Hey Thibaut thanks for the PR!

To fix the tests and docs please can you apply the following hotfixes (these are planned to be fixed in #409 ):

1. Add `model(y, physics)` here:
https://github.com/deepinv/deepinv/blob/464ec13d13c0bbc6dfd0b7f133ea9b6a161d1636/deepinv/loss/adversarial/uair.py#L77
2. Wrap this unet in a `ArtifactRemoval(UNet(...), mode="direct"` (the PR should work without this fix, but it's good to be explicit about best practice in the docs):
https://github.com/deepinv/deepinv/blob/464ec13d13c0bbc6dfd0b7f133ea9b6a161d1636/examples/adversarial-learning/demo_gan_imaging.py#L112
3. Likewise, wrap the Unet here:
https://github.com/deepinv/deepinv/blob/464ec13d13c0bbc6dfd0b7f133ea9b6a161d1636/deepinv/tests/test_adversarial.py#L19

Andrewwango:
Additional comments:

> Let me know if I should add a gallery example or notebook on how to train a denoiser, I think that is a fairly common use case of the library. 😄

That would be super cool! I think the "Basics" section could really benefit from this example

> I'm wondering if we really want these models to inherit from Denoiser if they don’t use the noise level? Wouldn't it be cleaner for them to simply subclass nn.Module unless they explicitly require sigma?`

I think it's ok as is, as long as we are clear in the docs that these models you listed are by default Denoisers, which you can convert into a Reconstructor if you want.

deepinv/training/trainer.py
Tmodrzyk: Done in [0844921](https://github.com/deepinv/deepinv/pull/614/commits/0844921e8d8f788a8ee347bbc5fa9743c617b363)

Tmodrzyk:
Hi @Andrewwango, I think this is good to go ? Let me know if you think the example is enough :)

examples/basics/demo_train_denoiser.py
```diff
@@ -0,0 +1,144 @@
...
+#       The number of epochs and the batch size should be based on the available hardware and the size of the dataset.
+#
+
+verbose = True  # print training information
+wandb_vis = False  # plot curves and images in Weight&Bias
+
+loss = dinv.loss.SupLoss(metric=dinv.metric.MSE())
+metrics = [dinv.metric.PSNR(), dinv.metric.SSIM()]
+
+optimizer = torch.optim.Adam(
```
Andrewwango, line 110: How does this work? After the model gets wrapped, surely optimizer is tracking the wrong parameters?

examples/basics/demo_train_denoiser.py
```diff
@@ -0,0 +1,144 @@
+r"""
+Training a denoiser.
+====================================================================================================
+
+This example shows how to train a standard denoiser such as the [DRUNet by Zhang et al., (2020)](https://arxiv.org/abs/2008.13751)! using the :class:`deepinv.Trainer` class.
```
Andrewwango, line 5: Ref bib instead following #575

examples/basics/demo_train_denoiser.py
```diff
@@ -0,0 +1,144 @@
...
+CKPT_DIR = BASE_DIR / "ckpts"
+
+# Set the global random seed from pytorch to ensure reproducibility of the example.
+torch.manual_seed(0)
+device = dinv.utils.get_freer_gpu() if torch.cuda.is_available() else "cpu"
+
+# %%
+# Load train and eval datasets
+# --------------------------------------------------------------------------------------------
+# In this example, we use the CBSD68 dataset for training and the set3c dataset for validation.
```
Andrewwango, line 33: Ref dinv classes for the datasets

examples/basics/demo_train_denoiser.py
```diff
@@ -0,0 +1,144 @@
...
+    ckp_interval=2,
+    save_path=CKPT_DIR,
+)
+
+trainer.train()
+
+model = trainer.load_best_model()
+
+# %%
+# Test the network
```
Andrewwango, line 138: Can we plot some images please

examples/basics/demo_train_denoiser.py
```diff
@@ -0,0 +1,144 @@
...
+# --------------------------------------------
+# We can now test the trained network using the :func:`deepinv.test` function.
+#
+# Here the testing function will compute the PSNR and SSIM metrics on the evaluation dataset.
+
+trainer.test(eval_dataloader)
```
Andrewwango, line 144:
```
Test results:
PSNR no learning: 22.182 +- 14.746
PSNR: 15.195 +- 1.237
SSIM no learning: 0.526 +- 0.334
SSIM: 0.606 +- 0.029
```
The results aren't great - maybe should we train starting from a pretrained model? Or should we just test on the train set?

examples/basics/demo_train_denoiser.py
Tmodrzyk: From my understanding since ``ArtifactRemoval`` doesn't add any trainable parameters, the optimizer still tracks the same parameters ? 
When I check in the demo file ``model.named_parameters()`` and ``trainer.model.named_parameters()`` are the same.

Tmodrzyk: Ok so it seems that the model is not learning right now so you were probably right about the optimizer not tracking the right parameters :smiling_face_with_tear:

PR #615, opened 2025-07-17
FIX add `__all__` variable in `__about__.py` to fix undefined __version__ attribute

mathurinm:
Currently 
```
import deepinv
deepinv.__version__
```
fails with an AttributeError

This is because in the `__init__.py` we do `from .__about__ import *` and this command does not import variables in `__about__.py` whose name starts with a dunder, unless  `__all__` is defined

hence the fix


Andrewwango (approved):
Thanks @mathurinm for spotting and fixing this!

For some reason, these lines had been deleted in #569ba32

PR #635, opened 2025-07-25
Ruff zip strict

jscanvic:
Now that we dropped support for Python 3.9 we can start using strict zip calls:
https://docs.python.org/3.10/library/functions.html#zip

I suggest we make it mandatory to specify if a zip is strict or not by toggling ruff rule B905:
https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/

The goal is to catch bugs early on and to make the code clearer. (It's not always obvious if a zip call is meant to be strict.)

I checked the current zip calls individually and it seems to me that they are all meant to be strict so I made them strict.

### DIFF
--- a/deepinv/optim/optimizers.py
+++ b/deepinv/optim/optimizers.py
@@ -197,7 +197,8 @@ class BaseOptim(Reconstructor):
         # By default, each parameter in ``params_algo` is a list.
         # If given as a single number, we convert it to a list of 1 element.
         # If given as a list of more than 1 element, it should have lenght ``max_iter``.
-        for key, value in zip(params_algo.keys(), params_algo.values()):
+        # NOTE: The zip iterator appears to be an unidiomatic version of `metrics.items()`.
+        for key, value in zip(params_algo.keys(), params_algo.values(), strict=True):
             if not isinstance(value, Iterable):
                 params_algo[key] = [value]
             else:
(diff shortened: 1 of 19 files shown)


deepinv/optim/optimizers.py
```diff
@@ -197,7 +197,8 @@ def __init__(
         # By default, each parameter in ``params_algo` is a list.
         # If given as a single number, we convert it to a list of 1 element.
         # If given as a list of more than 1 element, it should have lenght ``max_iter``.
-        for key, value in zip(params_algo.keys(), params_algo.values()):
+        # NOTE: The zip iterator appears to be an unidiomatic version of `metrics.items()`.
+        for key, value in zip(params_algo.keys(), params_algo.values(), strict=True):
```
Andrewwango: Why can't we change this to `.items()`? Indeed seems very unidiomatic.

deepinv/utils/plotting.py
```diff
@@ -505,7 +505,8 @@ def wandb_imgs(imgs, captions, n_plot):
 
 
 def wandb_plot_curves(metrics, batch_idx=0, step=0):
-    for metric_name, metric_val in zip(metrics.keys(), metrics.values()):
+    # NOTE: The zip iterator appears to be an unidiomatic version of `metrics.items()`.
```
Andrewwango: likewise

Andrewwango (approved):
Nice one, thanks! Just a couple of comments.

FYI we also make use of `itertools.zip_longest` in a couple of places.

jscanvic:
Thanks for the review! I made the changes.

PR #637, opened 2025-07-26
Fix #634

### LINKED ISSUE #634: device of noise models
There are some inconsistencies when setting the device of the parameters of a noise model. The issue is that the device is decided by the random seed `torch.Generator` input of the noise model, instead of the device of the input parameter.

```
import deepinv as dinv
import torch

physics = dinv.physics.Denoising(noise_model=dinv.physics.GaussianNoise(sigma=0.01), device='cuda')
print(physics.noise_model.sigma.device)
# cpu while I would expect cuda

physics = dinv.physics.Denoising(noise_model=dinv.physics.GaussianNoise(sigma=torch.tensor(0.01, device='cuda')), device='cuda')
print(physics.noise_model.sigma.device)

# cpu while I would expect cuda

generator = torch.Generator(device='cuda')
physics = dinv.physics.Denoising(noise_model=dinv.physics.GaussianNoise(sigma=0.01, rng=generator), device='cuda')
print(physics.noise_model.sigma.device)
# cuda - OK
```

I believe that most users don't use the random seed, and thus can have unexpected device errors that are hard to spot. I propose that the device is decided by the input parameter instead of the generator.

jscanvic:
Fix #634 

### DIFF
--- a/deepinv/physics/noise.py
+++ b/deepinv/physics/noise.py
@@ -1,7 +1,9 @@
 import torch
 import torch.nn as nn
 from typing import Callable, Union, Optional
+from collections.abc import Iterable
 import warnings
+import deepinv as dinv
 
 
 class NoiseModel(nn.Module):
@@ -847,21 +849,26 @@ class SaltPepperNoise(NoiseModel):
 
 
 def _infer_device(
-    device_held_candidates, *, default: torch.device = torch.device("cpu")
+    device_held_candidates: Iterable, *, default: torch.device = torch.device("cpu")
 ) -> torch.device:
     """Infer the device from a list of candidates.
 
     Check that all candidates bound to a device are bound to the same device and return that device. If no candidate is bound to a device, then return a default device.
 
-    :param device_held_candidates: list of tensors or generators to infer the device from.
-    :param default: default device to return if no candidates are bound to a device (default: cpu).
+    Supported device-held types are ``torch.Tensor``, ``torch.Generator``, and ``deepinv.utils.TensorList``.
+
+    :param Iterable device_held_candidates: list of tensors or generators to infer the device from.
+    :param torch.device default: default device to return if no candidates are bound to a device (default: cpu).
     :raises RuntimeError: if more than one device is found among the inputs.
     :return: the device of the candidates or the default device if no candidates are bound to a device.
     """
     input_devices = set()
 
     for device_held_candidate in device_held_candidates:
-        if isinstance(device_held_candidate, (torch.Tensor, torch.Generator)):
+        if isinstance(
+            device_held_candidate,
+            (torch.Tensor, torch.Generator, dinv.utils.TensorList),
+        ):
             input_devices.add(device_held_candidate.device)
 
     if len(input_devices) > 1:


mh-nguyen712:
Thanks @jscanvic! We have the same bug for other noise model as well. Could you please fix them?
The bug comes from the way we infer the device of the noise model: `device = getattr(rng, "device", "cpu")` 
So correct this would fix the bug.

jscanvic:
Thanks @mh-nguyen712! Should be good now

mh-nguyen712:
At the current state the following error (not really an error) reported by Julian still not fixed:
```python
import deepinv as dinv
import torch

physics = dinv.physics.Denoising(noise_model=dinv.physics.GaussianNoise(sigma=0.01), device='cuda')
print(physics.noise_model.sigma.device)
# cpu while I would expect cuda
```
Since it will be ok in forward so I think it's not a problem. But we could call `self.to(device)` in the constructor of physics. I think it's a bit cleaner since we have a `device` parameter in the physics.

jscanvic:
Indeed it keeps returning cpu!

Sorry - I think there might be a misunderstanding there so let me try to clarify in case we're not on the same page!

There's actually two separate problems going on here 1) inferring the right device from input tensors which is what I change in this PR and 2) that device is **not** an argument Denoising or LinearPhysics/Physics take in the first place. In my understanding the whole premise of this code snippet was that it was and it was not doing its job properly but it's actually not and it wasn't clear from the get go because of a bug that used to silence errors related to using unexpected arguments. (It was fixed and is now merged)

Now there's two options as far as I can tell 1) we add a device argument to Denoising or LinearPhysics/Physics in order to match users' expectations that there is going to be one, or 2) we don't add it, we deem the code snippet incorrect and we don't change what it does.

I don't mind any of the two options but I have a preference for option 2 given Physics.to(device) already does the same.

What do you prefer we do?

mh-nguyen712:
Agree with you just make it as simple as possible, as long as it works! It was just a minor detail, not really a bug so I think we can leave as it for now and we can discuss in the future for this kind of stuff for all physics/models.

mh-nguyen712 (approved):

deepinv/physics/noise.py
```diff
@@ -11,7 +11,7 @@ class NoiseModel(nn.Module):
     Noise models can be combined via :func:`deepinv.physics.NoiseModel.__mul__`.
 
     :param Callable noise_model: noise model function :math:`N(y)`.
-    :param torch.Generator rng: (optional) a pseudorandom random number generator for the parameter generation.
+    :param torch.Generator | None rng: (optional) a pseudorandom random number generator for the parameter generation.
```
Andrewwango:
```suggestion
    :param torch.Generator, None rng: (optional) a pseudorandom random number generator for the parameter generation.
```

deepinv/physics/noise.py
```diff
@@ -228,16 +228,18 @@ class GaussianNoise(NoiseModel):
         >>> y = batch_gaussian_noise(x)
         >>> assert (t[0]*gaussian_noise).sigma.item() == batch_gaussian_noise.sigma[0].item(), "Wrong standard deviation value for the first GaussianNoise."
 
-    :param float sigma: Standard deviation of the noise.
-    :param torch.Generator rng: (optional) a pseudorandom random number generator for the parameter generation.
+    :param float | torch.Tensor sigma: Standard deviation of the noise.
```
Andrewwango:
```suggestion
    :param float, torch.Tensor sigma: Standard deviation of the noise.
```

deepinv/physics/noise.py
```diff
@@ -228,16 +228,18 @@ class GaussianNoise(NoiseModel):
...
+    :param torch.Generator | None rng: (optional) a pseudorandom random number generator for the parameter generation.
```
Andrewwango:
```suggestion
    :param torch.Generator, None rng: (optional) a pseudorandom random number generator for the parameter generation.
```

deepinv/physics/noise.py
```diff
@@ -228,16 +228,18 @@ class GaussianNoise(NoiseModel):
...
     """
 
-    def __init__(self, sigma=0.1, rng: torch.Generator = None):
+    def __init__(
+        self, sigma: float | torch.Tensor = 0.1, rng: torch.Generator | None = None
```
Andrewwango: Sorry, can we use `typing.Union` please as the pipe was introduced in 3.10. I know we are moving away from 3.9 but this is just to keep consistency in the codebase for now

deepinv/physics/noise.py
```diff
@@ -796,3 +840,27 @@ def forward(self, x, p=None, s=None, seed: int = None, **kwargs):
...
+    :param default: default device to return if no candidates are bound to a device (default: cpu).
+    :return: the device of the candidates or the default device if no candidates are bound to a device.
+    """
+    input_devices = set()
+
+    for device_held_candidate in device_held_candidates:
+        if isinstance(device_held_candidate, (torch.Tensor, torch.Generator)):
+            input_devices.add(device_held_candidate.device)
+
+    assert (
```
Andrewwango: RuntimeError?

deepinv/physics/noise.py
Andrewwango: etc for below noises

deepinv/physics/noise.py
Andrewwango: etc for below nosies

deepinv/physics/noise.py
```diff
@@ -796,3 +840,27 @@ def forward(self, x, p=None, s=None, seed: int = None, **kwargs):
...
+    Check that all candidates bound to a device are bound to the same device and return that device. If no candidate is bound to a device, then return a default device.
+
+    :param device_held_candidates: list of tensors or generators to infer the device from.
+    :param default: default device to return if no candidates are bound to a device (default: cpu).
+    :return: the device of the candidates or the default device if no candidates are bound to a device.
+    """
+    input_devices = set()
+
+    for device_held_candidate in device_held_candidates:
+        if isinstance(device_held_candidate, (torch.Tensor, torch.Generator)):
```
Andrewwango: tensorlist?

Andrewwango (changes requested):
Minor comments

Andrewwango:
Question: will the incorrect usage `Denoising(device=device)` now give a warning because device is unused? #423 #640

deepinv/physics/noise.py
jscanvic: I made the change!

jscanvic:
It should be good now!

> Question: will the incorrect usage `Denoising(device=device)` now give a warning because device is unused? #423 #640

It does:


```
In [1]: import deepinv as dinv
   ...: import torch
   ...: 
   ...: physics = dinv.physics.Denoising(noise_model=dinv.physics.GaussianNoise(sigma=0.01), device='cuda')
   ...: print(physics.noise_model.sigma.device)
   ...: # cpu while I would expect cuda

deepinv/physics/forward.py:65: UserWarning: Arguments {'device': 'cuda'} are passed to Denoising but are ignored.
  warnings.warn(
cpu

```

PR #640, opened 2025-07-26
Remove ignored arguments

jscanvic:
LinearPhysics ignores its unsupported arguments and as a result some of our code uses arguments that actually don't exist.

In this PR, I get rid of them.

This is related to #423 and #634.

**TODO**

* [x] Check for warnings in the docs generation

### DIFF
--- a/deepinv/physics/blur.py
+++ b/deepinv/physics/blur.py
@@ -447,6 +447,9 @@ class SpaceVaryingBlur(LinearPhysics):
         **kwargs,
     ):
         super().__init__(**kwargs)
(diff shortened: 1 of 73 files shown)


deepinv/physics/blur.py
```diff
@@ -447,6 +447,9 @@ def __init__(
         **kwargs,
     ):
         super().__init__(**kwargs)
+        # NOTE: The only expected value of self.method throughout the class is
+        # "product_convolution2d", which is set here. All of the if statements
+        # like this one are no-op.
```
Andrewwango: Can we just remove them then?

PR #644, opened 2025-07-27
Exclude type checking conditionals from coverage

### LINKED ISSUE #624: Remove `TYPE_CHECKING` from `codecov`
Towards #487 

The code:

```python
from __future__ import annotations
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from deepinv.xxx import yyy
    ...
```

... is important for static typing (#536) without circular import problems, but it is an extra three lines per file which don't need to be tested. Therefore they should be removed from codecov.

jscanvic:
Fix #624

I got the fix from https://coverage.readthedocs.io/en/7.10.1/excluding.html

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,6 +121,9 @@ omit = ['examples/*', 'deepinv/tests']  # define paths to omit
 [tool.coverage.report]
 show_missing = true
 skip_covered = true
+exclude_also = [
+    "if TYPE_CHECKING:",
+]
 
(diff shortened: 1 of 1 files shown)


Andrewwango (approved):
LGTM thanks!!

jscanvic:
Oh apparently upgrading to Python 3.10 already fixed the issue I didn't catch that sorry!

It's now excluded as default:
https://coverage.readthedocs.io/en/7.10.1/excluding.html#default-exclusions

PR #622, opened 2025-07-21
Overhaul examples + quickstarts + datasets

Andrewwango:
Changes as follows:

## Overhaul datasets
- Adds dataset base class with lots of checks
- Datasets inherit from base class, simplifying their implementations
- Adds `ImageFolder` and `TensorDataset` classes
- More unified dataset tests
- Much better dataset docs + guide
- PatchDataset returns tensors and not tuples
- Clean thorough example for defining datasets

## Overhaul examples + quickstart

- Quickstart examples + one very simple 5-minute quickstart tutorial (closes #571 )
- Organises examples: basics section is now indeed basics; add "Models" and "Physics" sections
- Defining physics -> better cleaner example
- Training your own model -> better cleaner example (to sit alongside #593 )
- Pretrained reconstructors Guide + Example
- Finetuning example (to clean up after #524 merged)
- README and index now cleaner
- Custom ordering

## Minor fixes
- Closes #621 
- Closes #669 
- Closes #672
- Closes #656
- Full-reference metrics can handle nans
- DPIR has optional denoiser param
- Spelling mistakes `standart` -> `standard`
- Tomography accepts tensor angles better
- `physics = GaussianNoise` -> `physics = Denoising(GaussianNoise)`
- some small utils loading funcs
- TensorList shape, all, any and isnan
- Docs code blocks on lightmode appear slightly less gray
- Add autogenerated "References" to every example (discussed in Slack)

Notes
- Quickstart requires the redirect (rather than inserting `auto_examples/basics/demo_quickstart` into `index` `toctree` otherwise the left sidebar doesn't show up in the quickstart example??

### Checks to be done before submitting your PR

- [x] `python3 -m pytest deepinv/tests` runs successfully.
- [x] `black .` runs successfully.
- [x] `make html` runs successfully (in the `docs/` directory).
- [x] Updated docstrings related to the changes (as applicable).
- [x] Added an entry to the [CHANGELOG.rst](../CHANGELOG.rst).

Todo

- [x] Add quickstart tutorial notebook to colab and replace original colab link in readme/index
- [x] Switch all temp models over to RAM after #524 merged
- [x] finetuning example has more modalities from pretrained example. finetuning example links out to custom physics + data at end + what's next?
- [x] "New to DeepInverse?" auto add to all examples?

Issue #621, opened 2025-07-20
Should this be ZeroNoise()?


gRox167:
https://github.com/deepinv/deepinv/blob/99379cfd164811895e745703cfac42e74c1ac413/deepinv/physics/forward.py#L401

LinearPhysics should act as Physics, default value of noise model should also be ZeroNoise().

Andrewwango:
Thanks @gRox167 , yes `ZeroNoise` would be a clearer way of writing this rather than the lambda!

Andrewwango:
Fixed in #622

gRox167:
Thanks, the problem of using a plain lambda is that when we compose Physics and LinearPhysics, pytorch doesn't allow lambda to be module attribute.

PR #658, opened 2025-08-01
various noise fixes

tachella:
- default `ZeroNoise` to `Physics` class (beyond `LinearPhysics`)

For some reason, `ZeroNoise` was only added as default to LinearPhysics and not to Physics, causing errors when trying to compose a Physics with a LinearPhysics (as the default callable noise of Physics is not compatible with the torch module of the LinearPhysics)

- more robust `PoissonGaussian` noise by lowe bound `gain` and clipping by default

PoissonGaussian noise gives wrong results if gain=0, and most users would believe that this should be the case of 0 Poisson noise (only Gaussian). I added a clipping such that this is approximately true. I've also set positive clipping by default, because this was failing silently when using CUDA for some reason, leading to some very annoying debugging (which happened to me a couple of times)

- fix wrong `footcite`s

Just searching in the docs footcite, we can easily find all the wrong ones.

- fix docstring order of noise classes

By default, the doctest examples should appear after the parameter definition (see eg Inpainting class), and for some reason, they were always before the parameter definition in the noise modules.

- set `lsqr` as default conjugate gradient solver (much more robust than `CG`)

The lsqr solver is a much more robust choice than CG. This change is non-backwards compatible and might break some code, but I still believe this is better for future users.

- fix `release.yml`

The current github action for releasing a new version of the library is not working

- add automatic `A_adjoint` definition in `LinearPhysics`, and `V` and `U_adjoint` in `DecomposablePhysics`

This simplifies the implementation of new user-defined physics


Andrewwango:
What issues does this close?

tachella:
> What issues does this close?

all new issues - I can be more specific in the PR instead of opening 5 new issues

Andrewwango:
> default ZeroNoise to Physics class (beyond LinearPhysics)

This was already raised in #621  and I fixed it in #622 .

deepinv/tests/test_physics.py
```diff
@@ -1593,6 +1593,18 @@ def test_composed_linear_physics(device):
         rtol=1e-4,
     )
 
+    # test non-linear physics - checking for possible bugs in noise model
+    non_lin_physics = dinv.physics.Physics(A=lambda x: x**2)
```
Andrewwango: define this with a noise model too or no?

deepinv/tests/test_physics.py
```diff
@@ -1593,6 +1593,18 @@ def test_composed_linear_physics(device):
...
+    p = physics * non_lin_physics
+
+    y_2 = physics(non_lin_physics.A(x))
+    assert torch.allclose(y_2, p(x))
```
Andrewwango, line 1821: assert isinstance physics  no longer linearphysics, but physics?

Andrewwango (changes requested):
minor comments

tachella:
> > default ZeroNoise to Physics class (beyond LinearPhysics)
> 
> This was already raised in #621 and I fixed it in #622 .

sorry I missed that issue! should I rollback my fixes of that issue? Or we keep them if they are the same as yours?

deepinv/tests/test_physics.py
tachella: added

deepinv/tests/test_physics.py
tachella: added one and checked validity

deepinv/physics/forward.py
```diff
@@ -331,13 +336,13 @@ class LinearPhysics(Physics):
...
-
-            A_adjoint can be generated automatically using the :func:`deepinv.physics.adjoint_function`
-            method which relies on automatic differentiation, at the cost of a few extra computations per adjoint call.
-
+    :param None | Callable A_adjoint: transpose of the forward operator, which should verify the adjointness test.
+        By default, it is set to `None`, which means that the adjoint is computed automatically using :func:`deepinv.physics.adjoint_function`.
+        This automatic adjoint is computed using automatic differentiation, which is slower than a closed form adjoint, and can
+        have a larger memory footprint. If you want to use the automatic adjoint, you should set the `img_size` parameter
+        If you have a closed form for the adjoint, you can pass it as a callable function or rewrite the class method.
+    :param tuple img_size: size of the input signal, e.g. `(C, ...)` where `C` is the number of channels and `...` are the spatial dimensions,
```
Andrewwango: add a small comment to the start of this param docstring saying optional, only required when A_adjoint is not defined

Andrewwango: "size of the input signal `x`" 
("input signal") might get confused for y for the adjoint input

deepinv/physics/forward.py
```diff
@@ -812,10 +829,18 @@ class DecomposablePhysics(LinearPhysics):
...
+    :param None | Callable V_adjoint: transpose of V. If `None` (default), it is set to the identity function.
+    :param None | Callable U_adjoint: transpose of U. If `None` (default), it is computed automatically using :func:`deepinv.physics.adjoint_function`
+        from the `U` function and the `img_size` parameter.
+        This automatic adjoint is computed using automatic differentiation, which is slower than a closed form adjoint, and can
+        have a larger memory footprint. If you want to use the automatic adjoint, you should set the `img_size` parameter.
+    :param None | Callable V: If `None` (default), it is computed automatically using :func:`deepinv.physics.adjoint_function`
+        from the `V_adjoint` function and the `img_size` parameter.
+        This automatic adjoint is computed using automatic differentiation, which is slower than a closed form adjoint, and can
+        have a larger memory footprint. If you want to use the automatic adjoint, you should set the `img_size` parameter.
+    :param tuple img_size: size of the input signal, e.g. `(C, ...)` where `C` is the number of channels and `...` are the spatial dimensions,
```
Andrewwango: same here

deepinv/physics/forward.py
```diff
@@ -947,9 +974,21 @@ def V(self, x):
 
         :param torch.Tensor x: input tensor
         """
-        return self._V(x)
+        if self._V is None:
```
Andrewwango, line 991: but in your code now self._V is never None?

deepinv/physics/forward.py
```diff
@@ -959,7 +998,17 @@ def U_adjoint(self, x):
 
         :param torch.Tensor x: input tensor
         """
-        return self._U_adjoint(x)
+        if self._U_adjoint is None:
```
Andrewwango, line 1015: same here

deepinv/physics/forward.py
tachella: done

deepinv/physics/forward.py
tachella: done

deepinv/physics/forward.py
tachella: there was a typo (fixed now)

the idea is that by default U = identity and V_adjoint =  identity, and so U_adjoint=V=identity. If the user changes U or V_adjoint only, then U_adjoint and/or V will be computed via autodiff

deepinv/physics/forward.py
tachella: fixed too

Andrewwango (approved):

PR #663, opened 2025-08-02
Make ruff more convenient to deal with

jscanvic:
Currently ruff does not automatically use the settings it uses in the CI pipeline when executed locally. This is because the configuration is hard-coded as flags in the ruff command that gets executed in the GitHub action. In this PR, we fix that by moving those to the pyproject.toml file.

Benefits:
* Being able to run `ruff check --fix` directly like we do for `black .` without having to specify the settings manually

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -20,4 +20,4 @@ jobs:
       - name: Install and run ruff
         run: |
           pip install ruff
-          ruff check --target-version "py310" --select "UP006" --select "F401" --select "B905"
+          ruff check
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -20,4 +20,4 @@ jobs:
       - name: Install and run ruff
         run: |
           pip install ruff
-          ruff check --target-version "py310" --select "UP006" --select "F401" --select "B905"
+          ruff check
```
Andrewwango, line 23: Please can you add this to the "Code quality" bulletpoint in "Contributing" page? Thanks!

Andrewwango (changes requested):
LGTM except one comment, thanks! This is much cleaner + more expandable for future rules to add

.github/workflows/lint.yml
jscanvic: Done!

PR #665, opened 2025-08-04
Small fix in the user guide on Markov Chain Monte Carlo

HenKlei:
If I am not mistaken, the arguments to `sampling_builder` changed at some point and that was not reflected in the respective user guide so far.


jscanvic:
Indeed it's an oversight on our part - thanks for fixing it!

Issue #553, opened 2025-06-18
Trainer improvements


jscanvic:
Here's a few improvements I think would be useful to make:

* [x] Saving the eval loss #777
* [ ] Saving the actual best checkpoint (it's not actually the best when resuming training, it's only the best for the current training segment)
* [x] Make it so we can use the eval loss as criterion for the best checkpoint #777
* [ ] Make it possible to save what's saved to wandb on disk as well
* [x] Support MLflow in addition to wandb #668
* [ ] Make the trainer easier to work with for a train/val/test workflow instead of the current train/val-test workflow (progress in #777 )
* [ ] Improve the post-training (loading the best weights automatically instead of using the last weights)
* [ ] Log the learning rate used at every epoch
* [x] "No learning" metrics should be computed by passing the "no learning" reconstructor instead of the trained one #788
* [ ] Use 1-based indexing for reporting (Epochs 1 to 500, not 0 to 499)
* [ ] Save the path to the run directory in mlops tools
* [ ] Save the path to the best checkpoint in mlops tools
* [ ] Log test images to mlops tools
* [ ] Save the last checkpoint in addition to/instead of historical checkpoints
* [ ] Add an option to disable periodic checkpointing
* [ ] Add support for mixed-precision training #1205

**Dubious code**

1. [x] #904

https://github.com/deepinv/deepinv/blob/ce2b1a33f3afc72f9ee0b59375b57a74d5928e9f/deepinv/training/trainer.py#L903-L913

* It iterates on the batch dimension but does basically the same thing every iteration.
* It appears to save batched images despite iterating on the batch dimension (intended?)
* It increments the counter in the inner loop but probably shouldn't.

PR #668, opened 2025-08-05
Add support for mlflow

jscanvic:
Towards #553 

**Why?**

MLFlow is a popular alternative to wandb which is open-source unlike wandb and which is better supported on certain super-computers like the super-computer Jean-Zay.

**TODO**

* [x] Make logging images work
* [x] Thorough testing


Andrewwango:
Wow this is cool...

Andrewwango:
How does this compare to e.g. tensorboard which seems supported in JZ too?

jscanvic:
I have no idea to be fair - I've never met anyone using it but MLflow is what my colleagues use. It looks more popular in terms of GitHub stars too! Do you think we should try to work out a tensorboard integration as well?

Andrewwango:
I have used quite a lot of MLFlow in my previous job but not really for research...

My impression is that's it's geared more towards ML tracking/monitoring/serving (CD) for production (which is what we used it for) with nice integrations into cloud services such as Azure, rather than for research. You could also spin up a local server.

wandb is quite nice in that it does away with lots of engineering boilerplate, allowing you to focus on research.

I would say let's go for whatever is most popular in your research labs where you are working on big supercomputers like JZ!

Andrewwango:
Thanks for doing this Jeremy! Please can you explain here and in the trainer docstring how to setup mlflow to start using this feature? Is it any more difficult than setting up wandb? (Ie you just need wandb.login)

deepinv/training/trainer.py
```diff
@@ -400,6 +423,8 @@ def load_model(
             if "wandb_id" in checkpoint and self.wandb_vis:
                 self.wandb_setup["id"] = checkpoint["wandb_id"]
                 self.wandb_setup["resume"] = "allow"
+            if "mlflow_run_id" in checkpoint and self.mlflow_vis:  # pragma: no cover
```
Andrewwango: mlflow_id to be consistent with wandb_id?

deepinv/training/trainer.py
```diff
@@ -887,6 +914,10 @@ def plot(self, epoch, physics, x, y, x_net, train=True):
                 log_dict_post_epoch["step"] = epoch
                 wandb.log(log_dict_post_epoch, step=epoch)
 
+            if self.mlflow_vis:
```
Andrewwango, line 946: do we need to do the logging here too like in wandb?

Andrewwango (changes requested):
Minor comments!

deepinv/training/trainer.py
jscanvic: Do you mean adding `mlflow.log_metrics({ "step": epoch }, step=epoch)`? I'll add it


Mlflow doesn't have a generic log function like wandb, it uses log_metrics and log_image as far as I can tell

deepinv/training/trainer.py
```diff
@@ -1183,6 +1235,9 @@ def train(
         if self.wandb_vis:
             wandb.save("model.h5")
             wandb.finish()
+        if self.mlflow_vis:
+            mlflow.log_artifact("model.h5")
```
jscanvic:
```
  File "train.py", line 392, in <module>
    model = trainer.train()
            ^^^^^^^^^^^^^^^
  File "/lib/python3.12/site-packages/deepinv/training/trainer.py", line 1267, in train
    mlflow.log_artifact("model.h5")
  File "/lib/python3.12/site-packages/mlflow/tracking/fluent.py", line 1179, in log_artifact
    MlflowClient().log_artifact(run_id, local_path, artifact_path)
  File "/lib/python3.12/site-packages/mlflow/tracking/client.py", line 2379, in log_artifact
    self._tracking_client.log_artifact(run_id, local_path, artifact_path)
  File "/lib/python3.12/site-packages/mlflow/tracking/_tracking_service/client.py", line 931, in log_artifact
    artifact_repo.log_artifact(local_path, artifact_path)
  File "/lib/python3.12/site-packages/mlflow/store/artifact/local_artifact_repo.py", line 45, in log_artifact
    shutil.copy2(local_file, os.path.join(artifact_dir, os.path.basename(local_file)))
  File "/lib/python3.12/shutil.py", line 463, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/lib/python3.12/shutil.py", line 260, in copyfile
    with open(src, 'rb') as fsrc:
         ^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'model.h5'

```

jscanvic:
I get rid of `wandb.save("model.h5")` because it hasn't been doing anything in a while (it doesn't actually save the model)

Issue #679, opened 2025-08-11
Import time very long


Andrewwango:
### Environment

All envs

### Describe the bug

`import deepinv` takes a long time:

<img width="2392" height="1466" alt="Image" src="https://github.com/user-attachments/assets/f8fed457-faf8-47bc-9188-0d3baed28750" />

### Steps to reproduce

Based on the figure above, we should:
- [ ] Delay importing `timm` to when it's used
- [ ] Delay importing HF `datasets`
- [ ] Figure out if `timm` needs `wandb` (it imports it [here](https://github.com/huggingface/pytorch-image-models/blob/main/timm/utils/summary.py#L9) but it's not a required dep)
- [ ] Try resolve our own long import chains involving `DIP` and `loss.MC`

Creds to @matthieutrs @jscanvic  for the analysis!

### Expected behavior

Our import time should be very marginally more than that of torch, not much more

### Full traceback (if applicable)

You can generate the plot using 
```
python3 -m pip install tuna
```
and then run
```
python3 -X importtime -c "import deepinv" 2> log_deepinv.log && python3 -m tuna log_deepinv.log
```

tachella:
Some info from 19/08/2025

Torch 1623.64 ms

**Without dependencies:**
3784.68 ms
**With dependencies:**
6291.94 ms

PR #681, opened 2025-08-11
Decrease import time

jscanvic:
Towards fixing #679 

In this PR, I move imports to where they're used instead of being top-level which makes for longer startups.

**Results**

Without optional deps:

```
Torch mean time: 2790.55 ms
Main mean time: 4029.39 ms
PR mean time:   3014.49 ms
Rel difference (PR-main)/main: -25.19 %
Failure threshold >+5 %
✅ No significant performance regression detected.
```

With optional deps:

```
Torch mean time: 2342.21 ms
Main mean time: 5835.27 ms
PR mean time:   2711.42 ms
Rel difference (PR-main)/main: -53.53 %
Failure threshold >+5 %
✅ No significant performance regression detected.
```

How other libraries deal with this problem?

* https://github.com/numpy/numpy/pull/14083

TODO
- ~~[ ] rerun tuna profiling to check that our overhead on top of torch is minimal~~
- [x] add CI check that our overhead with all optional depedncies installed is minimal compared to torch import 

### DIFF
--- a/deepinv/training/trainer.py
+++ b/deepinv/training/trainer.py
@@ -929,8 +929,6 @@ class Trainer:
         :param None, float eval_metrics: Evaluation metrics across epochs.
         :param dict state: custom objects to save with model
         """
-        import wandb
-
         if state is None:
             state = {}
 
@@ -1064,8 +1062,6 @@ class Trainer:
 
         :returns: The trained model.
         """
-        import wandb
-
         self.setup_train()
         stop_flag = False
         for epoch in range(self.epoch_start, self.epochs):


deepinv/datasets/lidc_idri.py
```diff
@@ -97,6 +91,8 @@ def __init__(
         transform: Optional[Callable] = None,
         hounsfield_units: bool = False,
     ) -> None:
+        import pandas as pd
```
tachella, line 97: this is not dangerous, ie removing the try except?

deepinv/models/scunet.py
```diff
@@ -40,6 +27,8 @@ def __init__(self, input_dim, output_dim, head_dim, window_size, type):
 
         self.linear = nn.Linear(self.input_dim, self.output_dim)
 
+        from timm.layers import trunc_normal_
```
tachella, line 31: same here

deepinv/datasets/lidc_idri.py
jscanvic: Thanks for the review! It makes less sense to try-except them if they're in classes. Try-expecting is used to delay raising the error to when imports are used, but if imports are used in classes the error is raised at the right time by default anyway so there's no need to delay it.

The only difference it makes is we won't have the custom error message but a regular "couldn't find x module" without pip instructions.

I'm happy with putting them back if you prefer to keep the custom error messages!

(If we leave them in we get something like:)

```python
try:
    from datasets import load_dataset as load_dataset_hf, load_from_disk
except ImportError:  # pragma: no cover
    raise ImportError(
        "datasets is not available. Please install the datasets package with `pip install datasets`."
    )  # pragma: no cover
```

deepinv/datasets/cbsd68.py
```diff
@@ -5,12 +5,6 @@
 from deepinv.datasets.base import ImageDataset
 
 error_import = None
```
Andrewwango: Is this still needed?

deepinv/datasets/cbsd68.py
```diff
@@ -62,8 +56,7 @@ def __init__(
         download: bool = False,
         transform: Callable = None,
     ) -> None:
-        if error_import is not None and isinstance(error_import, ImportError):
-            raise error_import
+        from datasets import load_dataset as load_dataset_hf, load_from_disk
```
Andrewwango, line 61: should we have a try except like we do for other optional deps here?

deepinv/datasets/lidc_idri.py
Andrewwango: oh I seee now. Yes I would prefer keeping the custom message (if it's not much effort). However this solution will give `TypeError: 'ImportError' object is not callable`. Not sure what the standard solution is here?

Andrewwango: Tbh I'm not even that fussed. The default python import error is pretty good. You would have to be illiterate to not realise that the importerror means you should go and find how to install the package

deepinv/training/trainer.py
```diff
@@ -868,6 +864,8 @@ def plot(self, epoch, physics, x, y, x_net, train=True):
         :param torch.Tensor x_net: Network reconstruction.
         :param bool train: If ``True``, the model is trained, otherwise it is evaluated.
         """
+        import wandb
```
Andrewwango: do you think that the wandb import here should be somehow optional? so that one can use the trainer without wandb and not get errors

deepinv/utils/plotting.py
```diff
@@ -527,6 +533,7 @@ def plot_parameters(model, init_params=None, save_dir=None, show=True):
     :param str, pathlib.Path save_dir: the directory where to save the plot. Defaults to ``None``.
     :param bool show: whether to show the plot. Defaults to ``True``.
     """
+    import matplotlib.pyplot as plt
```
Andrewwango, line 537: Is matplotlib not a required dependency anyway? Or is this just to defer importing to save on import time?

pyproject.toml
```diff
@@ -131,8 +131,13 @@ skip_covered = true
...
-select = ["UP006", "F401", "B905"]
+select = ["UP006", "F401", "B905", "TID253"]
 
 [tool.ruff.lint.per-file-ignores]
 # NOTE: Rule F401 is ruff's unused imports rule.
 "__init__.py" = ["F401"]
+"examples/**/*" = ["TID253"]
+
+[tool.ruff.lint.flake8-tidy-imports]
+# Modules that slow down startup and should be imported lazily
```
Andrewwango, line 142: nice

deepinv/utils/plotting.py
jscanvic: Indeed it's just deferring it

deepinv/training/trainer.py
jscanvic: I messed this up let me fix that

deepinv/datasets/cbsd68.py
jscanvic: Good catch!

deepinv/training/trainer.py
jscanvic: I made the imports in the if wandb_vis so that it fails only if the user wants to use wandb but doesn't have it installed - should be fixed

deepinv/datasets/lidc_idri.py
jscanvic: It shouldn't be too much work to be honest!

I'm not sure I follow what you say about the TypeError though - it appears to work on my end:

```
>>> try:
...     from datasets import load_dataset as load_dataset_hf, load_from_disk
... except ImportError:  # pragma: no cover
...     raise ImportError(
...         "datasets is not available. Please install the datasets package with `pip install datasets`."
...     )  # pragma: no cover
...
  File "<stdin>", line 2, in <module>
ImportError: cannot import name 'load_dataset' from 'datasets' (unknown location)

During handling of the above exception, another exception occurred:

  File "<stdin>", line 4, in <module>
ImportError: datasets is not available. Please install the datasets package with `pip install datasets`.
```

deepinv/datasets/cbsd68.py
jscanvic: Marking this as resolved & responded in the other thread

Andrewwango (changes requested):
looks good and happy to approve when tests are fixed.

One general comment: can we add a warning/note/info to each class which lazy imports an optional depedency, to say that the class requires the dep and to install using `pip install xxx`?

jscanvic:
Thanks for the review @Andrewwango 

Should be good now!

PR #691, opened 2025-08-19
Client model

Andrewwango:


deepinv/models/client.py
```diff
@@ -0,0 +1,211 @@
...
+            return jsonify({
+                "output": {
+                    "file": Client.serialize(x_hat)
+                }
+            })
+
+        if __name__ == "__main__":
+            app.run()
+
+    **Server using RunPod** ::
```
matthieutrs: is this restricted to serverless, or can this also be used in the general case? If not general, I would specify it (serverless runpod)

deepinv/models/client.py
```diff
@@ -0,0 +1,211 @@
...
+            raise ValueError("Response output missing 'file'")
+
+        return Client.deserialize(result["output"]["file"])
+
+    def to(self, *args, **kwargs):
+        if args[0] and args[0] != "cpu":
+            warn("`.to()` has no effect on remote models. Ignoring.")
+        return self
+
+    def train(self, mode=True):
```
matthieutrs, line 284: why is it not possible to use client in training mode? (I would not use it in training mode, but is there a reason to prevent it?)

deepinv/models/client.py
```diff
@@ -0,0 +1,211 @@
...
+            }
+
+        if __name__ == '__main__':
+            runpod.serverless.start({'handler': handler })
+
+    :param str endpoint: endpoint URL.
+    :param str api_key: API key.
+    """
+
+    def __init__(self, endpoint: str, api_key: str = ""):
```
matthieutrs: Is it safe to pass an api_key as a variable, instead of exporting the variable?

e.g. 
```bash
export MY_KEY="my_key"
```
(I am afraid this code incites to unsafe practices)

docs/source/user_guide/reconstruction/denoisers.rst
```diff
@@ -177,3 +177,17 @@ using :class:`deepinv.models.TimeAveragingNet`.
 To adapt any existing network to take dynamic data as independent time-slices, :class:`deepinv.models.TimeAgnosticNet`
 creates a time-agnostic wrapper that flattens the time dimension into the batch dimension.
 
+Using models in the cloud
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The client model :class:`deepinv.models.Client` allows users to perform inference on models hosted in the cloud directly from DeepInverse.
+
+The client allows contributors to disseminate their reconstruction models, without requiring the user to have high GPU resources
```
matthieutrs: Maybe I would remove: "without requiring the user to have high GPU ressources": it is maybe a bit vague, is the user hte contributor? If an api key is needed, this can be seen as "having resources"?

```suggestion
The client allows contributors to disseminate their reconstruction models.
```

matthieutrs (changes requested):
Thanks a lot for this PR, this is great! I just have a few questions/suggestions.

@tachella, wdyt?

deepinv/models/client.py
Andrewwango: the standard way to do this on client-side is using python dotenv i.e. 

.env:
```
API_KEY=...
```

client script:
```
import os
from dotenv import load_dotenv
load_dotenv()

model = Client(api_key=os.getenv("API_KEY"))
```

Of course, if the client is called from a web server then you can just add the var as a secret.

deepinv/models/client.py
Andrewwango: Well it is illogical to allow training because you can't differentiate through the API call so it's to prevent confusion in case some people try to use it with Trainer.train (and not just Trainer.test)

deepinv/models/client.py
```diff
@@ -0,0 +1,211 @@
...
+            k: Client._check_value(v) for k, v in kwargs.items() if isinstance(k, str)
+        }
+        if len(safe_kwargs) != len(kwargs):
+            raise TypeError("All kwarg keys must be strings")
+
+        payload = {"input": {"file": Client.serialize(y), **safe_kwargs}}
+
+        headers = {"Content-Type": "application/json"}
+        if self.api_key != "":
+            headers |= {"Authorization": f"Bearer {self.api_key}"}
```
jscanvic, line 249: Do you intend authentication to be done at the application level or elsewhere like in a user-facing nginx server? I think it's worth clarifying

deepinv/models/client.py
jscanvic, line 1: Can we emphasize security somehow somewhere? Web APIs are among the biggest targets when it comes to cybersecurity attacks and they easily end up vulnerable

jscanvic (commented):
Thanks Andrew! Some comments

deepinv/models/client.py
Andrewwango: done

docs/source/user_guide/reconstruction/denoisers.rst
Andrewwango: Ok I reworded it, thanks!

deepinv/models/client.py
Andrewwango: Thank you for point this out! I added a note

deepinv/models/client.py
Andrewwango: Thank you for point this out! I added a note

deepinv/models/client.py
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+    * The API must return the following output response:
+
+    .. code-block:: python
+
+        {
+            "output": {
+                "file": "<b64 serialized file>",
+                "other_outputs": "such as inference time",
```
jscanvic: Can you change the format so that user-specified fields can't clash with our supported fields? Like, instead of output.time having output.metadata.time or metadata.time with the metadata field being explicitely supported by us?

The current format is good as long as we don't introduce new fields, but if we do we might end up introducing bugs if the fields we introduce clash with the fields introduced by our users.

deepinv/models/client.py
```diff
@@ -0,0 +1,225 @@
...
+    .. code-block:: python
+
+        {
+            "input": {
+                "file": <b64 serialized file>,
+                "param1": "such as a config str",
+                "param2": <or a b64 serialized param>,
+                ...
+            }
+        }
```
jscanvic, line 66: Same here maybe?

deepinv/tests/test_models.py
```diff
@@ -1126,3 +1129,72 @@ def test_denoiser_perf(device):
...
+        buffer.seek(0)
+        return base64.b64encode(buffer.read()).decode("utf-8")
+
+    model = dinv.models.Client(endpoint="http://example.com", api_key="test_key")
+
+    y = torch.ones(1, 3, 16, 16)
+
+    resp = MagicMock()
+    resp.status_code = 200
+    resp.json.return_value = {"output": {"file": make_b64_tensor(y), "time": 0.1}}
```
jscanvic: Just a suggestion

Use Client.serialize here, get rid of make_b64_tensor, and test Client.serialize and deserialize separately with something like:

```python
assert isinstance(Client.serialize(tensor), str)
assert torch.equal(Client.deserialize(Client.serialize(tensor)), tensor)
```

That way we rely less on the implementation in the tests

deepinv/models/client.py
```diff
@@ -0,0 +1,225 @@
...
+        self.api_key = api_key
+        self.endpoint = endpoint
+        self.training = False
+
+    @staticmethod
+    def serialize(tensor: torch.Tensor) -> str:
+        buffer = io.BytesIO()
+        torch.save(tensor.cpu(), buffer)
+        buffer.seek(0)
+        return base64.b64encode(buffer.read()).decode("utf-8")
```
jscanvic: I'm worried base64 is going to be a significant bottleneck here in terms of bandwidth. According to Wikipedia it has a 33% overhead in terms of bytes which can be fine for small blobs of data like compressed images (PNG, JPEG) but probably not for larger blobs like raw images (numpy arrays, tensors).

Ideally we should stream the blob separately from the JSON and I can think of several ways to do so:

* The "REST" way: return a short-lived URL pointing to the blob
* The "HTTP-wizard" way: HTTP2/3 might support streaming a JSON document plus a blob

I'm happy with keeping the serialization as is for now as long as we document the whole thing as experimental and subject to breaking changes and if we set to make the change in the future!

jscanvic (commented):
Thanks for making the changes! A few more comments

deepinv/models/client.py
```diff
@@ -0,0 +1,225 @@
...
+        :param torch.Tensor y: input measurements tensor
+        :param kwargs: any optional params depending on the API input requirements e.g. `physics` string, `config`, `sigma`, `mask` etc.
+        :return: torch.Tensor output reconstruction tensor
+        """
+        if self.training:
+            raise RuntimeError("Model client can only be used in evaluation mode.")
+
+        params = {k: Client._check_value(v) for k, v in kwargs.items()}
+
+        payload = {"input": {"file": Client.serialize(y), **params}}
```
jscanvic: With regard to my other comment, I'd prefer if we used a base64-encoded data URI instead of raw base64:

https://en.wikipedia.org/wiki/Data_URI_scheme

deepinv/models/client.py
```diff
@@ -0,0 +1,225 @@
...
+    def serialize(tensor: torch.Tensor) -> str:
+        buffer = io.BytesIO()
+        torch.save(tensor.cpu(), buffer)
+        buffer.seek(0)
+        return base64.b64encode(buffer.read()).decode("utf-8")
+
+    @staticmethod
+    def deserialize(data: str) -> torch.Tensor:
+        buffer = io.BytesIO(base64.b64decode(data))
+        return torch.load(buffer, map_location="cpu")
```
jscanvic: Please specify weights_only=True, it is not the default in older versions of PyTorch and it leads to the Flask example being vulnerable to remote code execution (RCE):

```python
# Send a request to a DeepInverse API to make it print a message and then
# terminate as an illustration of the remote code execution vulnerability.

import requests
import base64
import pickle

class MaliciousInput(object):
    def __reduce__(self):
        return (exec, ("import os, signal; print('RCE!'); os.kill(os.getpid(), signal.SIGINT)",))

malicious_input = base64.b64encode(pickle.dumps(MaliciousInput())).decode("utf-8")

payload = {"input": {"file": malicious_input}}
resp = requests.post("http://127.0.0.1:5000/", json=payload, timeout=60)
resp.raise_for_status()
```

deepinv/models/client.py
Andrewwango: done, good idea, thanks

deepinv/models/client.py
Andrewwango: done

deepinv/tests/test_models.py
Andrewwango: good idea, done, thanks

deepinv/models/client.py
Andrewwango: done, thanks - i just add a mimetype in serialize and then strip it in deserialize. I don't add any checks to check the mimetype - AFAIK it's more for info/debugging rather than a security check.

deepinv/models/client.py
Andrewwango: done thanks!

deepinv/models/client.py
Andrewwango: Thanks for the insight and the research!

You are right, and the current solution isn't very production ready at all.

If this feature gets traction in future then these are points that we definitely need to consider. 

For now, I added a warning.

deepinv/models/client.py
jscanvic: Thanks!

deepinv/models/client.py
```diff
@@ -0,0 +1,245 @@
...
+    @staticmethod
+    def serialize(tensor: torch.Tensor) -> str:
+        buffer = io.BytesIO()
+        torch.save(tensor.cpu(), buffer)
+        buffer.seek(0)
+        b64 = base64.b64encode(buffer.read()).decode("utf-8")
+        return f"data:application/octet-stream;base64,{b64}"
+
+    @staticmethod
+    def deserialize(data: str) -> torch.Tensor:
```
jscanvic, line 184: I managed to find a data URI reader in the standard library - I think it's better to use it instead of a custom implementation!

https://docs.python.org/3/library/urllib.request.html#datahandler-objects

I suggest to use the implementation below which has other small changes

* I added checks for the media type to make the intent clearer and because it doesn't cost much imo
* I got rid of the raw base64 fallback - I think it's better to have one way of doing things PEP8-style

```suggestion
    def deserialize(data: str) -> torch.Tensor:
        """
        Helper function to deserialize client inputs and outputs

        Instances of torch.Tensor are serialized by first `pickling
        <https://docs.python.org/3/library/pickle.html>`_ them using
        :func:`torch.save` and then returning a URI pointing to the pickle
        file. For now, only data URIs are supported, but in the future
        short-lived URLs may also be supported. The media type for the pickled
        documents is expected to be ``application/octet-stream``.

        :param str data: input serialized using :meth:`serialize`
        :return: torch.Tensor deserialized Tensor
        """
        import urllib
        from urllib.request import build_opener, UnknownHandler, DataHandler
        from email.message import EmailMessage

        # A URI reader that only reads data URIs
        opener = urllib.request.OpenerDirector()
        handlers = [
            # Handle data URIs
            DataHandler(),
            # Fallback for better error reporting
            UnknownHandler(),
        ]
        for handler in handlers:
            opener.add_handler(handler)

        f: urllib.response.addinfourl
        with opener.open(data) as f:
            headers: EmailMessage = f.headers
            ctype: str = headers.get_content_type()

            if ctype != "application/octet-stream":
                raise RuntimeError(f"Unexpected media type: {ctype}, expected 'application/octet-stream'")

            # The function torch.load can return objects of various types, even
            # when weights_only=True is set. Here, we expect a torch.Tensor
            # specifically.
            obj = torch.load(f, map_location="cpu", weights_only=True)

            if not isinstance(obj, torch.Tensor):
                raise RuntimeError(f"Expected a torch.Tensor, got {type(obj).__name__}")

            return obj
```

deepinv/models/client.py
```diff
@@ -0,0 +1,245 @@
...
+        if data.startswith("data:"):
+            _, data = data.split(",", 1)
+
+        buffer = io.BytesIO(base64.b64decode(data))
+        return torch.load(buffer, map_location="cpu", weights_only=True)
+
+    @staticmethod
+    def _check_value(v: Any):
```
jscanvic: (Sorry for the nitpick!) Can we change "check" to "sanitize"?

deepinv/models/client.py
```diff
@@ -0,0 +1,245 @@
...
+        """Client model forward pass.
+
+        :param torch.Tensor y: input measurements tensor
+        :param kwargs: any optional params depending on the API input requirements e.g. `physics` string, `config`, `sigma`, `mask` etc.
+        :return: torch.Tensor output reconstruction tensor
+        """
+        if self.training:
+            raise RuntimeError("Model client can only be used in evaluation mode.")
+
+        params = {k: Client._check_value(v) for k, v in kwargs.items()}
```
jscanvic: The more I look at it the more I'm unhappy with mixing a JSON document and pickled tensors. Can we drop JSON and pickle (args, kwargs) and the outputs directly?

deepinv/models/client.py
```diff
@@ -0,0 +1,245 @@
...
+        if "output" not in result or not isinstance(result["output"], dict):
+            raise ValueError("Response missing 'output' field")
+
+        if "file" not in result["output"]:
+            raise ValueError("Response output missing 'file'")
+
+        out = Client.deserialize(result["output"]["file"])
+
+        if self.return_metadata:
+            return out, result["output"].get("metadata")
```
jscanvic: We should probably deserialize the output metadata the same way we serialize the input metadata

jscanvic (changes requested):

deepinv/models/client.py
Andrewwango: Hmm, not sure I agree with that but happy to be persuaded otherwise!

In my opinion, it's more natural and readable to pass and return structured json where one field may be a data field consisting of a b64 string, so users can debug the return value without having to deserialise the metadata fields.

Furthermore, it feels more transferable this way - for example your API might have another route that instead takes serialised PNGs instead of tensors (which can be called for example by a webapp) - in which case you pass a serialised PNG in your data field along with other raw metadata fields - as it feels wrong to pickle a PNG.

jscanvic: Nevermind then!

deepinv/models/client.py
Andrewwango: The output metadata is arbitrary, and I think we shouldn't expect the API to return deserializable tensor strs in the metadata, because it would be difficult to check for this (unless we do a try/except?). However, I add the sanitize values to the metadata.

deepinv/models/client.py
Andrewwango: Thank you for this!

deepinv/models/client.py
jscanvic: Makes sense - I like your solution!

jscanvic (commented):
Thanks for making the changes Andrew!

I'm happy with what we've converged to!

There's probably some more work to do to have a stable interface but it can be left off for a separate PR, until we get some initial feedback on the feature. What we have so far is a strong foundation anyway!

Happy to approve once CI passes

jscanvic (approved):

matthieutrs (approved):

docs/source/user_guide/reconstruction/denoisers.rst
```diff
@@ -177,3 +177,17 @@ using :class:`deepinv.models.TimeAveragingNet`.
 To adapt any existing network to take dynamic data as independent time-slices, :class:`deepinv.models.TimeAgnosticNet`
 creates a time-agnostic wrapper that flattens the time dimension into the batch dimension.
 
+Using models in the cloud
```
tachella: I wouldn't add this here, which is a section a bit specific to denoisers.

I would maybe add a new section to the User Guide (in 'other') called Cloud API or similar

Andrewwango: moved to end of introduction

docs/source/user_guide/reconstruction/introduction.rst
```diff
@@ -85,3 +85,17 @@ For the models that require training, you can do this using the :ref:`trainer <t
...
+Using models in the cloud
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The client model :class:`deepinv.models.Client` allows users to perform inference on models hosted in the cloud directly from DeepInverse.
+
+The client allows contributors to disseminate their reconstruction models, without requiring the user to have high GPU resources
+or to accurately define their physics. As a contributor, all you have to do is:
+
+  * Define your model to take tensors as input and output tensors (like :class:`deepinv.models.Reconstructor`)
+  * Create a simple API (see below for example)
```
tachella:
```suggestion
  * Create a simple API
```

tachella (approved):
final comment and then I'm happy to merge

docs/source/user_guide/reconstruction/introduction.rst
tachella: there is no example below

tachella (approved):

PR #695, opened 2025-08-21
Remove factor check in downsampling init

Andrewwango:
Factor doesn't need to be checked in downsampling init as it's checked in check_factor which is called by update_parameters.

Currently this causes a bug when one initialises downsampling with a tensor factor, which may arise when you have e.g. `params = {"factor":tensor([2])}; physics = Downsampling(**params)` where your params are batched (returned by a dataloader).


matthieutrs (approved):
thanks a lot!!

Issue #662, opened 2025-08-02
Failures when building the docs with GPUs


jscanvic:
### Environment

Up to date

### Describe the bug

In #556 @Tmodrzyk reports that building the docs fails when GPUs are available, notably `demo_diffpir.py`.

### Steps to reproduce

Build the docs

### Expected behavior

No failure should happen

mh-nguyen712:
I'm assigning this to myself, as I've started fixing some examples that aren't running on the GPU, but I haven't finished yet , and since many PRs are currently modifying the examples, I'm waiting a bit before continuing.

PR #696, opened 2025-08-22
Fix example on GPU

mh-nguyen712:

Fix some examples that fail on GPU. 

### DIFF
--- a/deepinv/datasets/cmrxrecon.py
+++ b/deepinv/datasets/cmrxrecon.py
@@ -261,7 +261,7 @@ class CMRxReconSliceDataset(FastMRISliceDataset, MRIMixin):
             kspace = self.noise_model(kspace) * mask
 
         if self.apply_mask:
-            kspace = kspace * mask + 0.0
+            kspace = kspace * mask.to(device=kspace.device) + 0.0
             return target, kspace.float(), {"mask": mask.float()}
         else:
             return target, kspace.float()
--- a/deepinv/datasets/fastmri.py
+++ b/deepinv/datasets/fastmri.py
@@ -735,7 +735,9 @@ class MRISliceTransform(MRIMixin):
             params["mask"] = mask
         if self.mask_generator is not None:
             params["mask"] = self.generate_mask(kspace, seed)
-            kspace = kspace * params["mask"]  # TODO remove signed zeros
+            kspace = kspace * params["mask"].to(
+                device=kspace.device
+            )  # TODO remove signed zeros
         if self.estimate_coil_maps:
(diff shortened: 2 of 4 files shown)


deepinv/datasets/cmrxrecon.py
```diff
@@ -261,7 +261,7 @@ def __getitem__(self, i: int) -> tuple[Tensor, Tensor, dict[str, Tensor]]:
             kspace = self.noise_model(kspace) * mask
 
         if self.apply_mask:
-            kspace = kspace * mask + 0.0
+            kspace = kspace * mask.to(device=kspace.device) + 0.0
```
Andrewwango, line 264: What is this for? Generally all datasets should only ever load onto cpu, so everything should be cpu here, and then trainer loads onto gpu

deepinv/datasets/fastmri.py
```diff
@@ -735,7 +735,9 @@ def __call__(
             params["mask"] = mask
         if self.mask_generator is not None:
             params["mask"] = self.generate_mask(kspace, seed)
-            kspace = kspace * params["mask"]  # TODO remove signed zeros
+            kspace = kspace * params["mask"].to(
+                device=kspace.device
```
Andrewwango, line 739: likewise, see comment in cmrxrecon

examples/basics/demo_custom_physics.py
```diff
@@ -30,6 +30,8 @@
 import deepinv as dinv
 import torch
 
+device = dinv.utils.get_freer_gpu() if torch.cuda.is_available() else "cpu"
```
Andrewwango, line 33: Do we need this? This example works fine all on CPU. But if you think this example would be better if it demonstrated physics on devices, then we can do that

Andrewwango (changes requested):
Thanks Hai! 
Does this fix the demo_diffpir issue reported in #662?

deepinv/datasets/cmrxrecon.py
mh-nguyen712: There is a problem when I run it on my machine. Because the mask_generator is on GPU

examples/basics/demo_custom_physics.py
mh-nguyen712: We can then show that `physics.to(device)` will work, as in torch

mh-nguyen712:
I think it has been fixed. I dont see any error. We can close #662

deepinv/datasets/cmrxrecon.py
mh-nguyen712: It was just a quick fix, I'm not sure if it's the best way to do, as you said everything should on CPU.

PR #699, opened 2025-08-25
Fix the condition for break in compute_norm

qbarthelemy:
In current code, when the relative variation criterion is reached, `break` is achieved only if `verbose` is activated.
This PR separates these two conditions, first for the relative variation criterion for the `break`, and if so, `verbose` to print.

Also, I transformed the `for` loop into a `for-else` loop, to easily detect when convergence is not reached.


Andrewwango (approved):
Looks good to me, thanks @qbarthelemy !

Issue #702, opened 2025-08-26
Docs CI flakes: external image downloads intermittently fail during gallery builds


Tmodrzyk:
### Environment

Github VM, latest deepinv

### Describe the bug

Sometimes the doc building step in the CI fails because images cannot be downloaded from the internet. 
This happens randomly and is not tied to the content of a specific commit.
Here's a typical example where the whole pipeline works but the CI fails on the doc building only because of this bug: https://github.com/deepinv/deepinv/actions/runs/17219304845/job/48850615907

### Steps to reproduce

The bug is very hard to reproduce, here's an example: https://github.com/deepinv/deepinv/actions/runs/17219304845/job/48850615907

### Expected behavior

The CI should not fail randomly

### Full traceback (if applicable)

```shell
Traceback
=========

      File "/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/sphinx_gallery/gen_gallery.py", line 1558, in summarize_failing_examples
        raise ExtensionError(fail_message)
    sphinx.errors.ExtensionError: Here is a summary of the problems encountered when running the examples:
    
    Unexpected failing examples (4):
    
        ../../examples/sampling/demo_sampling.py failed leaving traceback:
    
        Traceback (most recent call last):
          File "/home/runner/work/deepinv/deepinv/examples/sampling/demo_sampling.py", line 37, in <module>
            x = load_url_image(url=url, img_size=32).to(device)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          File "/home/runner/work/deepinv/deepinv/deepinv/utils/demo.py", line 227, in load_url_image
            img = Image.open(load_url(url))
                             ^^^^^^^^^^^^^
          File "/home/runner/work/deepinv/deepinv/deepinv/utils/demo.py", line 204, in load_url
            response.raise_for_status()
          File "/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/requests/models.py", line 1026, in raise_for_status
            raise HTTPError(http_error_msg, response=self)
        requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://upload.wikimedia.org/wikipedia/commons/b/b4/Lionel-Messi-Argentina-2022-FIFA-World-Cup_%28cropped%29.jpg
    
        ../../examples/optimization/demo_dip.py failed leaving traceback:
    
        Traceback (most recent call last):
[...]

PR #703, opened 2025-08-27
Fix the Messi image from examples

Tmodrzyk:
As stated in issue #702 the CI appears to fail randomly with an error 403 whenever examples try to download an image from wikipedia, namely the image of Messi. 
I replaced it with the leaves image which is available on deepinv's huggingface, and provides interesting details even when center-cropped to 32 by 32. 

### DIFF
--- a/examples/optimization/demo_dip.py
+++ b/examples/optimization/demo_dip.py
@@ -28,11 +28,11 @@ from deepinv.utils import load_example
 # Load image from the internet
 # ----------------------------
 #
-# This example uses an image of the set3c dataset.
+# This example uses an image of Messi.
 
 device = dinv.utils.get_freer_gpu() if torch.cuda.is_available() else "cpu"
 
-x = load_example("leaves.png", img_size=32).to(device)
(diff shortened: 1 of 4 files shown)


tachella:
Many thanks for fixing this!

I believe it is safer to keep the messi image (to make sure that the center crop is interesting, and algorithms perform the same), but download it from huggingface.

I've just uploaded `messi.jpg` to [huggingface ](https://huggingface.co/datasets/deepinv/images/tree/main)

could you please modify the set3 image for that one?

Tmodrzyk:
It seems that the center crop doesn't capture the "10" on Messi's jersey. Is that ok ? 
image

Andrewwango (approved):
LGTM

Tmodrzyk:
I did a merge thinking I was on another PR, I reverted it in the following commit. 
Please ignore the last 2 commits :)

tachella:
I've uploaded a new Messi image to hugging face whose center crop should be the same as before

PR #715, opened 2025-09-05
Fix behaviour of Tomography operator when normalize=True

romainvo:
Remove the ad-hoc procedure use to normalize the Tomography operator, i.e. the procedure did not yield an operator with a spectral norm equal to one (checked with `compute_norm`).

Instead of scaling each call to `A` and `A_adjoint` by the `img_width`, the operator norm is computed during initialization and stored afterwards to scale every computation.

**Warning**: this will break backward compatibility with any network/method optimized with normalize=True, e.g. I encountered the problem with RAM (@tachella @matthieutrs ). Not sure how you want to handle this ? 
- An easy solution for the user is to manually set the `operator_norm` attribute to whatever value is needed, thus overwriting the value computed during initialization.

PS: I also fixed a small bug related to the import chain of `astra.experimental` in `deepinv/physics/functional/astra.py` (not covered by the mock test as it deliberately skips these calls), everything should work fine now.

### DIFF
--- a/deepinv/physics/tomography.py
+++ b/deepinv/physics/tomography.py
@@ -93,24 +93,24 @@ class Tomography(LinearPhysics):
         >>> seed = torch.manual_seed(0)  # Random seed for reproducibility
         >>> x = torch.randn(1, 1, 4, 4)  # Define random 4x4 image
         >>> angles = torch.linspace(0, 45, steps=3)
-        >>> physics = Tomography(angles=angles, img_width=4, circle=True)
+        >>> physics = Tomography(angles=angles, img_width=4, circle=True, normalize=True)
         >>> physics(x)
-        tensor([[[[ 0.0000, -0.1791, -0.1719],
-                  [-0.5713, -0.4521, -0.5177],
-                  [ 0.0340,  0.1448,  0.2334],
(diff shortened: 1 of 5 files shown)


deepinv/tests/test_physics.py
```diff
@@ -1316,8 +1316,13 @@ def test_tomography(
     )
 
     x = torch.randn(imsize, device=device).unsqueeze(0)
```
jscanvic: Can you seed the `randn` to make CI deterministic?

jscanvic:
Looks nice! Happy to approve once the CI passes

deepinv/physics/tomography.py
```diff
@@ -158,7 +157,7 @@ def __init__(
         self.img_width = img_width
         self.device = device
         self.dtype = dtype
-        self.normalize = normalize
+        self.normalize = False
```
tachella: I prefer to normalize by default, even if not backwards compatible

romainvo: The default behavior now defaults to True although if ``normalize`` is not specified explicitly, it triggers a warning to inform the user. I've also updated the documentation for the user to note that normalizing the operator does change the range of values in the reconstruction.

PR #723, opened 2025-09-08
Spatial unwrapping

bemc22:
This PR introduces the following features:

- `SpatialUnwrapping`: Implements a new forward operator for spatial unwrapping, supporting both "round" and "floor" wrapping modes.
- `ItohFidelity`: Adds a data-fidelity term tailored for spatial unwrapping problems, compatible with the new operator.
- Demo: Provides `demo_spatial_unwrapping.py` to showcase the usage and performance of the new operator and fidelity term.


deepinv/physics/spatial_unwrapping.py
```diff
@@ -0,0 +1,147 @@
...
+    The forward operator is defined as:
+
+    .. math::
+
+        y = W_t(x) = x - t \cdot \mathrm{q}(x / t)
+
+    where :math:`W_t` is the wrapping operator, :math:`t` is the threshold, and :math:`\mathrm{q}` is either the rounding or flooring function depending on the mode.
+
+
+    :param float threshold: The threshold value for the modulo operation (default: 1.0).
```
tachella:
```suggestion
    :param float threshold: The threshold :math:`t` value for the modulo operation (default: 1.0).
```

deepinv/physics/spatial_unwrapping.py
```diff
@@ -0,0 +1,147 @@
...
+    :param str mode: Wrapping mode, either 'round' or 'floor' (default: 'round').
```
tachella:
```suggestion
    :param str mode: Wrapping function :math:`q(\cdot)`, either 'round' or 'floor' (default: 'round').
```

deepinv/physics/spatial_unwrapping.py
```diff
@@ -0,0 +1,147 @@
...
+    def __init__(self, threshold=1.0, mode="round", **kwargs):
+
+        super().__init__(**kwargs)
+
+        self.threshold = threshold
+        self.mode = mode
+
+        self.B = modulo_floor if mode == "floor" else modulo_round
+
+    def forward(self, x, **kwargs):
```
tachella, line 52: do you need to define the forward? this looks the same as the parent class, in which case we can avoid the definition

deepinv/physics/spatial_unwrapping.py
```diff
@@ -0,0 +1,147 @@
...
+        return self.sensor(self.A(self.noise(x, **kwargs), **kwargs))
+
+    def A(self, x):
```
tachella: add docstring

deepinv/physics/spatial_unwrapping.py
```diff
@@ -0,0 +1,147 @@
...
+        return self.B(x, self.threshold)
+
+    def finite_differences(self, x):
```
tachella: add docstring

deepinv/physics/spatial_unwrapping.py
```diff
@@ -0,0 +1,147 @@
...
+        # apply spatial finite differences
+        Dh_x = F.pad(torch.diff(x, 1, dim=-1), (0, 1))
+        Dv_x = F.pad(torch.diff(x, 1, dim=-2), (0, 0, 0, 1))
+        return Dh_x, Dv_x
+
+    def D(self, x):
```
tachella: add docstring

deepinv/physics/spatial_unwrapping.py
```diff
@@ -0,0 +1,147 @@
...
+        Dh_x, Dv_x = self.finite_differences(x)
+        out = torch.stack((Dh_x, Dv_x), dim=-1)
+        return out
+
+    def WD(self, x):
```
tachella: add docstring

deepinv/physics/spatial_unwrapping.py
```diff
@@ -0,0 +1,147 @@
...
+        Dx = self.D(x)
+        WDx = modulo_round(Dx, self.threshold)
+        return WDx
+
+    def A_dagger(self, y):
```
tachella: add docstring

deepinv/physics/spatial_unwrapping.py
```diff
@@ -0,0 +1,147 @@
...
+        phi = self.prox_l2(None, y, self.threshold)
+        return phi
+
+    def A_adjoint(self, y):
```
tachella: add docstring

deepinv/physics/spatial_unwrapping.py
```diff
@@ -0,0 +1,147 @@
...
+
+        out = self.WD(y)
+        out = self.A_transpose(out)
+
+        return out
+
+    def A_transpose(self, y):
```
tachella: add docstring

deepinv/physics/spatial_unwrapping.py
```diff
@@ -0,0 +1,147 @@
...
+
+        tmp = torch.zeros_like(y)[..., 0]
+        return self.A_vjp(tmp, y)
+
+    def A_vjp(self, x, v):
```
tachella: add docstring

deepinv/physics/spatial_unwrapping.py
```diff
@@ -0,0 +1,59 @@
...
+
+    This class implements a forward operator for spatial unwrapping, where the input is wrapped modulo a threshold value.
+    The operator can use either floor or round mode for the wrapping operation. It is useful for problems where the observed data is wrapped,
+    such as in phase imaging, modulo imaging, or interferometry.
+
+    The forward operator is defined as:
+
+    .. math::
+
+        y = W_t(x) = x - t \cdot \mathrm{q}(x / t)
```
tachella: I would use lowercase for w, since it is not a matrix but a function

deepinv/physics/spatial_unwrapping.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __init__(self, threshold=1.0, mode="round", **kwargs):
+
+        super().__init__(**kwargs)
+
+        self.threshold = threshold
+        self.mode = mode
+
+        self.B = modulo_floor if mode == "floor" else modulo_round
+
```
tachella, line 51: I would recommend adding some standard reconstruction as A_dagger (which doesn't use an advanced regularizer)

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,137 @@
+r"""
+Spatial Unwrapping Demo
+====================================
+
+This demo shows the use of the :class:`deepinv.physics.SpatialUnwrapping` forward model and the :class:`deepinv.optim.ItohFidelity` for phase unwrapping problems.
+It shows how to generate a wrapped phase image, apply blur and noise, and reconstruct the original phase using both DCT inversion and ADMM optimization.
+
+Sections:
```
tachella: I don't think this outline is necessary, I would remove

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,137 @@
+r"""
+Spatial Unwrapping Demo
```
tachella: I think we can avoid the word 'demo' since it is clearly a demo
```suggestion
Spatial unwrapping and modulo imaging
```

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,137 @@
...
+====================================
+
+This demo shows the use of the :class:`deepinv.physics.SpatialUnwrapping` forward model and the :class:`deepinv.optim.ItohFidelity` for phase unwrapping problems.
```
tachella:
```suggestion
This demo shows the use of the :class:`deepinv.physics.SpatialUnwrapping` forward model and the :class:`deepinv.optim.ItohFidelity` for unwrapping problems, which occur in modulo imaging, interferometry SAR and other imaging applications.
```

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,137 @@
...
+img_size = (size, size)
+mode = "floor"  # available modes: "round", "floor"
+
+
+x_rgb = load_example(
+    "CBSD_0010.png", grayscale=False, device=device, dtype=dtype, img_size=img_size
+)
+x_rgb = channel_norm(x_rgb) * dr
+
+factor = 3
```
tachella: why do you need this factor? do you need oversampled images? In which case I think it would be good to explain in the text

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,137 @@
...
+def channel_norm(x):
+    x = x - x.min(dim=-1, keepdim=True)[0].min(dim=-2, keepdim=True)[0]
+    x = x / x.max(dim=-1, keepdim=True)[0].max(dim=-2, keepdim=True)[0]
+    return x
+
+
+# %%
+# Load image and preprocess
+# -------------------------------------------------------
+# Load example image from deepinv and preprocess
```
tachella:
```suggestion
# Load example image and preprocess
```

tachella: I would explain better how you preprocess to simulate a given dynamic range

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,137 @@
...
+resize = transforms.Resize(size=(img_size[0] * factor, img_size[1] * factor))
+x_rgb = resize(x_rgb)
+
+if mode == "round":
+    x_rgb = x_rgb - dr / 2
+
+# %%
+# Apply blur
+# -------------------------------------------------------
+# We apply a slight Gaussian blur to the image to simulate the low bandwidth assumption of the phase map.
```
tachella: why do we need this? the assumption should be explained at the beginning

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,137 @@
...
+threshold = 1.0
+noise_model = dinv.physics.GaussianNoise(sigma=0.1)
+physics = SpatialUnwrapping(threshold=threshold, mode=mode, noise_model=noise_model)
+phase_map = x_rgb
+wrapped_phase = physics(phase_map)
+
+# %%
+# Invert with DCT and ADMM (ItohFidelity)
+# -------------------------------------------------------
+# We provide two inversion methods: a simple DCT-based inversion and an ADMM-based inversion using the Itoh fidelity term and TV prior.
```
tachella, line 164: explain why we need Itoh fidelity with some text and equations to non-experts of phase unwrapping - maybe you can also point to a paper

pyproject.toml
```diff
@@ -34,6 +34,7 @@ dependencies = [
     "requests",
     "h5py",
     "natsort",
+    "torch_dct",
```
tachella: I wouldn't add this as a mandatory dependency, but rather as an optional one (see the list below)

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +281,182 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
...
+    This is based on the Itoh condition for phase unwrapping :footcite:t:`itoh1982analysis`.
+    It is designed to be used in conjunction with the :class:`deepinv.physics.SpatialUnwrapping` class for spatial unwrapping tasks.
+
+    The data-fidelity term is defined as:
+
+    .. math::
+
+        f(x) = \frac{1}{2\sigma^2} \| D x - W_t(D y) \|^2
+
+    where :math:`D` denotes the spatial finite differences operator and :math:`W_t` denotes the wrapping operator.
```
tachella:
```suggestion
    where :math:`D` denotes the spatial finite differences operator, :math:`W_t` denotes the wrapping operator, and :math:`\sigma` denotes the noise level.
```

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +281,182 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
...
+
+    |sep|
+
+    :Example:
+
+        >>> import torch
+        >>> from deepinv.physics.spatial_unwrapping import SpatialUnwrapping
+        >>> from deepinv.optim.data_fidelity import ItohFidelity
+        >>> x = torch.ones(1, 1, 3, 3)
+        >>> y = torch.ones(1, 1, 3, 3)
```
tachella:
```suggestion
        >>> y = x
```

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +281,182 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
...
+        if physics.__class__.__name__ != "SpatialUnwrapping":
+            print(
+                "Warning: ItohFidelity is designed to be used with SpatialUnwrapping physics."
+            )
+
+        Dx = self.D(x)
+        WDy = self.WD(y)
+        return super().fn(Dx, WDy, physics, *args, **kwargs)
+
+    def grad(self, x, y, *args, **kwargs):
```
tachella, line 350: add docstrings to all methods

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +281,182 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
...
+
+    This class implements a data-fidelity term based on the normalized :math:`\ell_2` norm, but applied to the spatial finite differences of the variable and the wrapped differences of the data.
+    This is based on the Itoh condition for phase unwrapping :footcite:t:`itoh1982analysis`.
+    It is designed to be used in conjunction with the :class:`deepinv.physics.SpatialUnwrapping` class for spatial unwrapping tasks.
+
+    The data-fidelity term is defined as:
+
+    .. math::
+
+        f(x) = \frac{1}{2\sigma^2} \| D x - W_t(D y) \|^2
```
tachella:
```suggestion
        f(x,y) = \frac{1}{2\sigma^2} \| D x - W_t(D y) \|^2
```

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +281,182 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
...
+        """
+
+        Dh_x, Dv_x = torch.unbind(x, dim=-1)
+        rho = -(
+            torch.diff(F.pad(Dh_x, (1, 0)), 1, dim=-1)
+            + torch.diff(F.pad(Dv_x, (0, 0, 1, 0)), 1, dim=-2)
+        )
+        return rho
+
+    def D_dagger(self, y):
```
tachella: I would move this to A_dagger of the forward operator, otherwise it sounds like it is the pseudoinverse of the DCT operator (which is not if I'm not wrong)

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +281,182 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
         return physics.prox_l2(x, y, self.norm * gamma)
 
 
+class ItohFidelity(L2):
+    r"""
+    Itoh data-fidelity term for spatial unwrapping problems.
+
+    This class implements a data-fidelity term based on the normalized :math:`\ell_2` norm, but applied to the spatial finite differences of the variable and the wrapped differences of the data.
+    This is based on the Itoh condition for phase unwrapping :footcite:t:`itoh1982analysis`.
```
tachella:
```suggestion
    This is based on the Itoh condition for phase unwrapping :footcite:p:`itoh1982analysis`.
```

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +281,182 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
...
+
+        :param torch.Tensor x: Input tensor to which the spatial finite differences and wrapping are applied.
+        :return: (:class:`torch.Tensor`) The wrapped finite differences of the input tensor.
+        """
+
+        Dx = self.D(x)
+        WDx = self.modulo_round(Dx)
+        return WDx
+
+    def D_transpose(self, x):
```
tachella:
```suggestion
    def D_adjoint(self, x):
```

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +281,182 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
...
+        r"""
+        Applies the adjoint (transpose) of the spatial finite difference operator to the input tensor.
+
+        This function computes the adjoint operation corresponding to spatial finite differences,
+        typically used in image processing and variational optimization problems. The input `x`
+        is expected to have its last dimension of size 2, representing the horizontal and vertical
+        finite differences (Dh_x, Dv_x).
```
tachella:
```suggestion
        finite differences :math:`(D_h x, D_v x)`.
```

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +281,182 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
...
+        Dh_x = F.pad(torch.diff(x, 1, dim=-1), (0, 1))
+        Dv_x = F.pad(torch.diff(x, 1, dim=-2), (0, 0, 0, 1))
+        out = torch.stack((Dh_x, Dv_x), dim=-1)
+        return out
+
+    def WD(self, x):
+        r"""
+        Applies spatial finite differences to the input and wraps the result.
+
+        This method computes the spatial finite differences of the input tensor :param torch.Tensor x: using the :meth:`D` operator,
```
tachella:
```suggestion
        This method computes the spatial finite differences of the input tensor :math:`x` using the :math:`D` operator,
```

deepinv/physics/spatial_unwrapping.py
bemc22: It is slightly different: in the parent class, the A operator is applied first, followed by the noise.

In spatial unwrapping, however, the noise is applied first, and then the A operator (the modulo operator) is used.

deepinv/optim/data_fidelity.py
bemc22: Yes, this is essentially the pseudoinverse of the spatial finite difference (D) operator, which admits a fast closed-form solution using the DCT operator.

Since the forward operator does not involve the D operator (as it is used only for the Itoh condition), I suggest keeping it here.

I am open to your suggestions!

deepinv/physics/spatial_unwrapping.py
bemc22: done!

deepinv/physics/spatial_unwrapping.py
bemc22: done!

deepinv/physics/spatial_unwrapping.py
bemc22: done!

deepinv/physics/spatial_unwrapping.py
bemc22: done!

deepinv/physics/spatial_unwrapping.py
bemc22: done!

deepinv/physics/spatial_unwrapping.py
bemc22: done!

deepinv/physics/spatial_unwrapping.py
bemc22: done!

deepinv/physics/spatial_unwrapping.py
bemc22: done!

deepinv/physics/spatial_unwrapping.py
bemc22: done!

deepinv/physics/spatial_unwrapping.py
bemc22: mmm the standard reconstruction it's the same proximal of itoh fidelity with the `x_tilde` set in zeros

examples/physics/demo_spatial_unwrapping.py
bemc22: done!

examples/physics/demo_spatial_unwrapping.py
bemc22: yes, in some cases its necessary oversampling, for the demo it's just to shows nice recons.

pyproject.toml
bemc22: done! removed

deepinv/optim/data_fidelity.py
bemc22: done!

examples/physics/demo_spatial_unwrapping.py
bemc22: done!

examples/physics/demo_spatial_unwrapping.py
bemc22: done!

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +282,234 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
...
+    The data-fidelity term is defined as:
+
+    .. math::
+
+        f(x,y) = \frac{1}{2\sigma^2} \| D x - w_{t}(Dy) \|^2
+
+    where :math:`D` denotes the spatial finite differences operator, :math:`w_t` denotes the wrapping operator, and :math:`\sigma` denotes the noise level.
+
+    :param float sigma: Standard deviation of the noise to be used as a normalisation factor.
+    :param float threshold: Threshold value used in the wrapping operator (default: 1.0).
```
tachella:
```suggestion
    :param float threshold: Threshold value :math:`t` used in the wrapping operator (default: 1.0).
```

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +282,234 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
...
+        r"""
+        Calculates the gradient of the data fidelity term :math:`\datafidname` at :math:`x`.
+
+        The gradient is computed using the chain rule:
+
+        .. math::
+
+            \nabla_x \distance{Dx}{w_{t}(Dy)} = \left. \frac{\partial D}{\partial x} \right|_x^\top \nabla_u \distance{u}{w_{t}(Dy)},
+
+        where :math:`\left. \frac{\partial D}{\partial x} \right|_x` is the Jacobian of :math:`D` at :math:`x`, and :math:`\nabla_u \distance{u}{w_{t}(Dy)}` is computed using ``grad_d`` with :math:`u = Dx`. The multiplication is computed using the ``D_adjoint`` method of the class.
```
tachella:
```suggestion
        where :math:`\left. \frac{\partial D}{\partial x} \right|_x` is the Jacobian of :math:`D` at :math:`x`, and :math:`\nabla_u \distance{u}{w_{t}(Dy)}` is computed using ``grad_d`` with :math:`u = Dx`. The multiplication is computed using the :func:`D_adjoint <deepinv.optim.ItohFidelity.D_adjoint>` method of the class.
```

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +282,234 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
...
+        :param torch.Tensor u: Variable :math:`u` at which the gradient is computed.
+        :param torch.Tensor y: Data :math:`y` of the same dimension as :math:`u`.
+        :return: (:class:`torch.Tensor`) gradient of :math:`d` in :math:`u`, i.e. :math:`\nabla_u\distance{u}{w_{t}(Dy)}`.
+        """
+        WDy = self.WD(y)
+        return self.d.grad(u, WDy, *args, **kwargs)
+
+    def prox_d(self, u, y, *args, **kwargs):
+        r"""
+        Computes the proximity operator :math:`\operatorname{prox}_{\gamma\distance{\cdot}{w_{t}(Dy)}}(u)`, computed in :math:`u`.
```
tachella:
```suggestion
        Computes the proximity operator :math:`\operatorname{prox}_{\gamma\distance{\cdot}{w_{t}(Dy)}}(u)`, computed at :math:`u`.
```

deepinv/physics/spatial_unwrapping.py
tachella: we can do that then

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,201 @@
...
+np.random.seed(0)
+torch.manual_seed(0)
+from deepinv.physics.spatial_unwrapping import SpatialUnwrapping
+import deepinv as dinv
+from deepinv.utils.demo import load_example
+import torchvision.transforms as transforms
+import matplotlib.pyplot as plt
+
+
+def channel_norm(x):
```
tachella, line 37: I would move this function definition to the subsection below

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,201 @@
...
+    img_size=img_size,
+)
+x_rgb = channel_norm(x_rgb) * dr
+
+
+# %%
+# Itoh condition
+# -------------------------------------------------------
+# The Itoh condition requires that the difference between adjacent pixels is less than half of the threshold (here, 1.0)
+# to enable perfect unwrapping. Specifically, :math:`|Dx| < t / 2`
```
tachella:
```suggestion
# to enable perfect unwrapping. Specifically, :math:`\|Dx\|_{\infty} < t / 2`
```
isn't it the \ell infinity norm?

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,201 @@
+r"""
+Spatial unwrapping and modulo imaging
+==============================================
+
+This demo shows the use of the :class:`deepinv.physics.SpatialUnwrapping` forward model and the :class:`deepinv.optim.ItohFidelity` for unwrapping problems, which occur in modulo imaging, interferometry SAR and other imaging applications.
+It shows how to generate a wrapped phase image, apply blur and noise, and reconstruct the original phase using both DCT inversion and ADMM optimization.
```
tachella, line 6: I would recall the math definition of the forward problem here

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,201 @@
...
+# By varying the amount of Gaussian blur, we observe that the Itoh condition is satisfied when the blur is sufficiently large.
+
+modulo_round = lambda x: x - torch.round(x)
+modulo_fn = lambda x: x - torch.floor(x) if mode == "floor" else modulo_round(x)
+row = 120
+
+row_sel = x_rgb[0, 0, row, :]
+
+
+def plot_itoh(sigma_blur):
```
tachella, line 88: image
for some reason the legends are cropped from the image - I wonder if we can fix this reducing the fig_size?

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,201 @@
...
+        row_x = row_x - dr / 2
+
+    row_dx = row_x[1:] - row_x[:-1]
+    row_y = modulo_fn(row_x)
+    row_wdy = modulo_round(row_y[1:] - row_y[:-1])
+
+    plt.figure(figsize=(10, 3))
+    plt.plot(row_x.cpu(), label="Pixel values", linewidth=3, color="g")
+    plt.plot(row_dx.cpu(), label="Dx", linewidth=3, color="k")
+    plt.plot(row_wdy.cpu(), label="w_t(Dy)", linewidth=3, color="b", linestyle="--")
```
tachella, line 119: why do you plot this? it is not related to the itoh condition explained above?

examples/physics/demo_spatial_unwrapping.py
tachella: both cases verify the itoh condition?

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+plot_itoh(sigma_blur=0.1)
+plot_itoh(sigma_blur=2.0)
+
+
+# %%
+# Apply Resize and Gaussian blur
+# -------------------------------------------------------
+# To satisfy the Itoh condition, we resize the image and apply a slight Gaussian blur to ensure that adjacent pixel differences are small enough.
```
tachella: I would relate the blur used here to the analysis in 1D above

tachella (changes requested):

examples/physics/demo_spatial_unwrapping.py
bemc22: yes! this plot visualizes the pixel values, their differences, and the wrapped differences for a single row.
It directly illustrates the Itoh condition by showing when the differences exceed the threshold and how the wrapped differences behave.

examples/physics/demo_spatial_unwrapping.py
bemc22: done!

examples/physics/demo_spatial_unwrapping.py
bemc22: done!

examples/physics/demo_spatial_unwrapping.py
bemc22: it should be correctly plotted now

for sigma=0.1 do not satisfy itoh condition (Dx are above the threshold) while for sigma=2.0 it does (differences within the threshold)

examples/physics/demo_spatial_unwrapping.py
bemc22: done!

bemc22:
Sorry for the delay; I will promptly address any feedback.

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,219 @@
...
+from deepinv.utils.demo import load_example
+import torchvision.transforms as transforms
+import matplotlib.pyplot as plt
+
+
+# %%
+# Load image and preprocess
+# -------------------------------------------------------
+# Load example image and preprocess to emulate a high dynamic range image.
+# images are normalized to [0, 1] and then scaled to the desired dynamic range.
```
tachella:
```suggestion
# Images are normalized to [0, 1] and then scaled to the desired dynamic range.
```

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,219 @@
...
+# to enable perfect unwrapping. Specifically, :math:`\|Dx\|_{\infty} < t / 2`
+# where :math:`D` denotes the spatial finite difference operator and :math:`x` is the original image :footcite:p:`itoh1982analysis`.
+# When this condition is satisfied, the high dynamic range (HDR) image can be recovered from
+# the wrapped differences of the modulo image by minimizing the Itoh fidelity term:
+#
+# .. math::
+#         \begin{equation}  \underset{x}{\arg\min} \quad \| Dx - w_t(Dy) \|^2_2. \end{equation}
+#
+# Below, we illustrate this for a single row of the image by visualizing the pixel values, their differences, and the wrapped differences.
+# By varying the amount of Gaussian blur, we observe that the Itoh condition is satisfied when the blur is sufficiently large.
```
tachella:
```suggestion
# We can understand the condition by artificially blurring the images with a Gaussian kernel to reduce their high frequencies (and thus the :math:`\|Dx\|_{\infty}) until the condition is verified.
```

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,219 @@
...
+# For instace, with a blur of 0.1 it can be seen that the differences :math:`Dx` exceed the threshold (red dotted lines),
+# and consequenly, missmatch with the wrapped differences :math:`w_t(Dy)`.
```
tachella:
```suggestion
# and consequently, we observe a mismatch with the wrapped differences :math:`w_t(Dy)`,
```

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,219 @@
...
+# while with a blur of 2.0, the differences :math:`Dx` remain within the threshold are match the wrapped differences :math:`w_t(Dy)`,
```
tachella:
```suggestion
# while with a blur of 2.0, the differences :math:`Dx` remain within the threshold, matching the wrapped differences :math:`w_t(Dy)`,
```

examples/physics/demo_spatial_unwrapping.py
tachella: can you add labels to the different graphs?

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,219 @@
...
+    plt.plot(row_x.cpu(), label="Pixel values", linewidth=3, color="g")
+    plt.plot(row_dx.cpu(), label="Dx", linewidth=3, color="k")
+    plt.plot(row_wdy.cpu(), label="w_t(Dy)", linewidth=3, color="b", linestyle="--")
+    plt.axhline(threshold / 2, color="r", linestyle="--", label="t/2")
+    plt.axhline(-threshold / 2, color="r", linestyle="--")
+    plt.legend(loc="upper center", bbox_to_anchor=(0.5, -0.18), ncol=4)
+    plt.title(f"Itoh Condition, sigma={sigma_blur}, mode={mode}")
+    plt.xlabel("Pixel Index")
+    plt.ylabel("Difference")
+    plt.show()
```
tachella, line 126: plt.legend()

(after adding labels to the plots)

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,219 @@
...
+params_algo = {"stepsize": stepsize, "lambda": lam, "g_param": 1.0}
+model = dinv.optim.optim_builder(
+    iteration="ADMM",
+    prior=prior,
+    data_fidelity=fidelity,
+    max_iter=10,
+    verbose=False,
+    params_algo=params_algo,
+)
+x_model = model(wrapped_phase, physics, compute_metrics=False)
```
tachella:
```suggestion
x_model = model(wrapped_phase, physics)
```

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +282,234 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
         return physics.prox_l2(x, y, self.norm * gamma)
 
 
+class ItohFidelity(L2):
+    r"""
+    Itoh data-fidelity term for spatial unwrapping problems.
+
+    This class implements a data-fidelity term based on the normalized :math:`\ell_2` norm, but applied to the spatial finite differences of the variable and the wrapped differences of the data.
```
tachella:
```suggestion
    This class implements a data-fidelity term based on the :math:`\ell_2` norm, but applied to the spatial finite differences of the variable and the wrapped differences of the data.
```

not sure why 'normalized'?

deepinv/physics/functional/dct.py
```diff
@@ -0,0 +1,126 @@
...
+
+def dct(x, norm=None):
+    r"""
+    Discrete Cosine Transform, Type II (a.k.a. the DCT)
+
+    For the meaning of the parameter `norm`, see:
+    https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.fftpack.dct.html
+
+    :param torch.Tensor x: the input signal
+    :param norm: the normalization, None or 'ortho'
```
tachella:
```suggestion
    :param None, str norm: the normalization, `None` or `'ortho'`
```

deepinv/physics/functional/dct.py
```diff
@@ -0,0 +1,126 @@
...
+
+def idct(x, norm=None):
+    r"""
+    The inverse to DCT-II, which is a scaled Discrete Cosine Transform, Type III
+
+    For the meaning of the parameter `norm`, see:
+    https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.fftpack.dct.html
+
+    :param torch.Tensor x: the input signal
+    :param norm: the normalization, None or 'ortho'
```
tachella:
```suggestion
    :param None, str norm: the normalization, `None` or `'ortho'`
```

deepinv/physics/functional/dct.py
```diff
@@ -0,0 +1,126 @@
...
+
+def dct_2d(x, norm=None):
+    r"""
+    2-dimentional Discrete Cosine Transform, Type II (a.k.a. the DCT)
+
+    For the meaning of the parameter `norm`, see:
+    https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.fftpack.dct.html
+
+    :param torch.Tensor x: the input signal
+    :param norm: the normalization, None or 'ortho'
```
tachella:
```suggestion
    :param None, str norm: the normalization, `None` or `'ortho'`
```

deepinv/physics/functional/dct.py
```diff
@@ -0,0 +1,126 @@
...
+    r"""
+    The inverse to 2D DCT-II, which is a scaled Discrete Cosine Transform, Type III
+
+    Our definition of idct is that idct_2d(dct_2d(x)) == x
+
+    For the meaning of the parameter `norm`, see:
+    https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.fftpack.dct.html
+
+    :param torch.Tensor x: the input signal
+    :param norm: the normalization, None or 'ortho'
```
tachella:
```suggestion
    :param None, str norm: the normalization, `None` or `'ortho'`
```

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,219 @@
...
+"""
+
+# %%
+# Imports and setup
+# -------------------------------------------------------
+import numpy as np
+import torch
+from deepinv.utils.plotting import plot
+
+np.random.seed(0)
```
tachella: why this? the example is not using numpy right?

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +282,234 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
...
+        r"""
+        Computes the data fidelity term :math:`\datafid{x}{y} = \distance{Dx}{w_{t}(Dy)}`.
+
+        :param torch.Tensor x: Variable :math:`x` at which the data fidelity is computed.
+        :param torch.Tensor y: Data :math:`y`.
+        :param deepinv.physics.Physics physics: physics model.
+        :return: (:class:`torch.Tensor`) data fidelity :math:`\datafid{x}{y}`.
+        """
+
+        if physics.__class__.__name__ != "SpatialUnwrapping":
```
Andrewwango:
```suggestion
        if not isinstance(physics, SpatialUnwrapping):
```
you will need the import too

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +282,234 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
...
+            print(
+                "Warning: ItohFidelity is designed to be used with SpatialUnwrapping physics."
+            )
```
Andrewwango:
```suggestion
            warnings.warn(
                "ItohFidelity is designed to be used with SpatialUnwrapping physics."
            )
```
Or simply raise a ValueError if it really shouldn't be used with any other physics

deepinv/physics/functional/dct.py
```diff
@@ -0,0 +1,126 @@
...
+Parts of this code are adapted from the 'torch-dct' repository by zh217:
+https://github.com/zh217/torch-dct
+
+Original code is licensed under the MIT License.
+Modifications have been made for integration with the DeepInverse library.
+"""
+
+import torch
+
+pi = torch.acos(torch.zeros(1)).item() * 2
```
Andrewwango: why not just torch.pi

deepinv/physics/functional/dct.py
```diff
@@ -0,0 +1,126 @@
+"""
+This module implements the Discrete Cosine Transform (DCT) and its inverse (IDCT) for PyTorch tensors.
+
+Parts of this code are adapted from the 'torch-dct' repository by zh217:
```
Andrewwango, line 4: If only parts, I think you should just directly ref them from the functions themselves, so that they show up in the user guide. The attribution here is quite hidden

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,219 @@
...
+torch.manual_seed(0)
```
Andrewwango: We shouldn't need this if we pass `torch.Generator` appropriately to all random functions

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,219 @@
...
+from deepinv.physics.spatial_unwrapping import SpatialUnwrapping
+import deepinv as dinv
```
Andrewwango, line 23: Clean up the imports - you can probably just call `dinv.utils.plot` instead of populating the imports

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+def channel_norm(x):
+    x = x - x.min(dim=-1, keepdim=True)[0].min(dim=-2, keepdim=True)[0]
+    x = x / x.max(dim=-1, keepdim=True)[0].max(dim=-2, keepdim=True)[0]
+    return x
+
+
+size = 256
+dr = 2  # dynamic range
```
Andrewwango: Can we just call the variable `dynamic_range`?

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,219 @@
...
+threshold = 1.0  # threshold for spatial unwrapping
+factor = 2  # oversampling factor to ensure Itoh condition
+device = "cpu"
```
Andrewwango: Support GPU too?

examples/physics/demo_spatial_unwrapping.py
```diff
@@ -0,0 +1,219 @@
...
+modulo_round = lambda x: x - torch.round(x)
+modulo_fn = lambda x: x - torch.floor(x) if mode == "floor" else modulo_round(x)
+row = 120
+
+row_sel = x_rgb[0, 0, row, :]
+
+
+def plot_itoh(sigma_blur):
+
+    row_x = row_sel.clone()
```
Andrewwango, line 91: There's a lot of algebra going on in this example - please could you explain what's going on to the user?

deepinv/tests/test_physics_functional.py
```diff
@@ -164,3 +164,17 @@ def test_imresize_div2k():
     y = dinv.utils.load_example("div2k_valid_lr_bicubic_0877x4.png") * 255.0
     y2 = dinv.physics.functional.imresize_matlab(x, scale=1 / 4).round()
     assert dinv.metric.PSNR()(y2 / 255.0, y / 255.0) > 59
+
+
+def test_dct_idct(device):
+    torch.manual_seed(0)
```
Andrewwango: There is nothing stochastic about this test right? Then we could probably remove this

examples/physics/demo_spatial_unwrapping.py
bemc22: true

deepinv/physics/functional/dct.py
bemc22: Sorry, I wasn’t aware that Torch already has a native implementation of pi. Thanks!

deepinv/tests/test_physics_functional.py
bemc22: true

deepinv/optim/data_fidelity.py
bemc22: Sorry, my bad, it was a typo.

examples/physics/demo_spatial_unwrapping.py
bemc22: done!

deepinv/optim/data_fidelity.py
bemc22: done!

deepinv/optim/data_fidelity.py
bemc22: done, included as local import to avoid circular import

examples/physics/demo_spatial_unwrapping.py
bemc22: okey

examples/physics/demo_spatial_unwrapping.py
bemc22: done!

examples/physics/demo_spatial_unwrapping.py
bemc22: done!

examples/physics/demo_spatial_unwrapping.py
bemc22: done!

examples/physics/demo_spatial_unwrapping.py
bemc22: done!

deepinv/physics/functional/dct.py
bemc22: done!

bemc22:
Hi, I’ve uploaded a new update, I’m open to your feedback, @tachella @Andrewwango

deepinv/optim/data_fidelity.py
```diff
@@ -278,6 +282,237 @@ def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
...
+        r"""
+        Proximal operator of :math:`\gamma \datafid{x}{y}`
+
+        Compute the proximal operator of the fidelity term :math:`\operatorname{prox}_{\gamma \datafidname}`, i.e.
+
+        .. math::
+
+           \operatorname{prox}_{\gamma \datafidname} = \underset{u}{\text{argmin}} \frac{\gamma}{2\sigma^2}\|Du-w_{t}(Dy)\|_2^2+\frac{1}{2}\|u-x\|_2^2
+
+        these can be computed using DCT with the close-form solution of :footcite:t:`ramirez2024phase` as follows
```
tachella, line 471:
```suggestion
        using the DCT-based closed-form solution of :footcite:t:`ramirez2024phase` as follows
```

Issue #716, opened 2025-09-05
Matplotlib backend/dependency issues when running quickstart example


ScepticalRabbit:
### Environment

Ubuntu 22.04, Python 3.11, deepin 0.3.3

### Describe the bug

Running the "5 minute quickstart" example as is using a blank python 3.11 venv with just deepinv installed will not show any plots in interactive mode. Installing the required dependencies and then changing the matplotlib backend causes additional errors when running the example.

### Steps to reproduce

1. Create a blank python 3.11 venv and activate it
2. pip install deepinv
3. Run the following python script:
```python
import deepinv as dinv
import torch

device = dinv.utils.get_freer_gpu() if torch.cuda.is_available() else "cpu"
print(device)
print()

x = dinv.utils.load_example("butterfly.png", device=device)
print(x.shape)

physics = dinv.physics.Inpainting(x.shape[1:], mask=0.3, device=device)
y = physics(x)

physics.noise_model = dinv.physics.GaussianNoise(sigma=0.1)
y = physics(x)

dinv.utils.plot({"GT": x, "Noisy inpainting measurement": y})
```
4. No plots are produced and the following is given in the terminal: UserWarning: FigureCanvasAgg is non-interactive, and thus cannot be shown plt.show().
5. Installing a different backend with `pip install PyQt5` and then invoking this at the start of the script with 
```
import matplotlib
matplotlib.use('Qt5Agg')
```
Produced an error:
```
QSocketNotifier: Can only be used with threads started with QThread
qt.qpa.wayland: Wayland does not support QWindow::requestActivate()
Traceback (most recent call last):
  File "/home/lloydf/reviews/rev-deepinv/.venv/lib/python3.11/site-packages/matplotlib/backends/backend_qt.py", line 522, in _draw_idle
    self.draw()
``` 

### Expected behavior

The quickstart example should produce plots out of the box if deepinv is installed in a blank venv

### Full traceback (if applicable)

QSocketNotifier: Can only be used with threads started with QThread
qt.qpa.wayland: Wayland does not support QWindow::requestActivate()
Traceback (most recent call last):
  File "/home/lloydf/reviews/rev-deepinv/.venv/lib/python3.11/site-packages/matplotlib/backends/backend_qt.py", line 522, in _draw_idle
    self.draw()
  File "/home/lloydf/reviews/rev-deepinv/.venv/lib/python3.11/site-packages/matplotlib/backends/backend_agg.py", line 382, in draw
[...]

PR #726, opened 2025-09-08
Add disable_tex function

Andrewwango:

### DIFF
--- a/deepinv/utils/plotting.py
+++ b/deepinv/utils/plotting.py
@@ -24,6 +24,7 @@ from mpl_toolkits.axes_grid1 import make_axes_locatable
 from PIL import Image
 
 _DEFAULT_PLOT_FONTSIZE = 17
+_ENABLE_TEX = True
 
 
 def set_default_plot_fontsize(fontsize: int):
@@ -37,6 +38,23 @@ def get_default_plot_fontsize() -> int:
     return _DEFAULT_PLOT_FONTSIZE
 
 
+def disable_tex():
+    """Globally disable LaTeX"""
+    global _ENABLE_TEX
+    _ENABLE_TEX = False
+
+
+def enable_tex():
+    """Globally enable LaTeX"""
+    global _ENABLE_TEX
+    _ENABLE_TEX = True
+
+
+def get_enable_tex() -> bool:
+    """Get whether LaTeX is globally enabled"""
+    return _ENABLE_TEX
+
+
 def config_matplotlib(fontsize=17):
     """Config matplotlib for nice plots in the examples."""
     if fontsize is None:
@@ -46,10 +64,14 @@ def config_matplotlib(fontsize=17):
     plt.rcParams["axes.titlesize"] = fontsize
     plt.rcParams["figure.titlesize"] = fontsize
     plt.rcParams["lines.linewidth"] = 2
-    plt.rcParams["text.usetex"] = True if shutil.which("latex") else False
-    plt.rcParams["text.latex.preamble"] = (
-        r"\usepackage{amsmath}" if plt.rcParams["text.usetex"] else ""
-    )
+
+    if shutil.which("latex") and get_enable_tex():
+        plt.rcParams["text.usetex"] = True
+        plt.rcParams["text.latex.preamble"] = (
+            r"\usepackage{amsmath}" if plt.rcParams["text.usetex"] else ""
+        )
+    else:
+        plt.rcParams["text.usetex"] = False
 
 
 def resize_pad_square_tensor(tensor, size):
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ New Features
 - Dataset base class + ImageFolder and TensorDataset classes (:gh:`622` by `Andrew Wang`_)
 - Added GitHub action checking import time (:gh:`680` by `Julian Tachella`_)
 - Client model for server-side inference for using models in the cloud (:gh:`691` by `Andrew Wang`_)
+- Add dinv.utils.disable_tex to disable LaTeX (:gh:`726` by `Andrew Wang`_)
 
 Changed
 ^^^^^^^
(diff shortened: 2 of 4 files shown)


Andrewwango:
@Tmodrzyk @jscanvic  2 questions:

Q1: do you think we should add use_tex as a param to locally fix the problem when plotting? e.g. `dinv.utils.plot(..., use_tex=False)`. Advantage: doesn't require adding an extra command. Disadvantage: using tex is a global problem, either it globally causes a problem or globally it's fine!

Q2: should we set global enable_tex to True or False by default? Advantage: currently it's True, so True will be backward compatible. The vast majority of users and the CI servers have no problem. However, False will be safer for all users and will have avoided the problem in #716 .

deepinv/utils/plotting.py
```diff
@@ -41,10 +59,14 @@ def config_matplotlib(fontsize=17):
...
     plt.rcParams["lines.linewidth"] = 2
-    plt.rcParams["text.usetex"] = True if shutil.which("latex") else False
-    plt.rcParams["text.latex.preamble"] = (
-        r"\usepackage{amsmath}" if plt.rcParams["text.usetex"] else ""
-    )
+
+    if shutil.which("latex") and get_enable_tex():
+        plt.rcParams["text.usetex"] = True
+        plt.rcParams["text.latex.preamble"] = (
+            r"\usepackage{amsmath}" if plt.rcParams["text.usetex"] else ""
```
jscanvic: This if should be gone

deepinv/utils/plotting.py
```diff
@@ -41,10 +59,14 @@ def config_matplotlib(fontsize=17):
     plt.rcParams["axes.titlesize"] = fontsize
     plt.rcParams["figure.titlesize"] = fontsize
     plt.rcParams["lines.linewidth"] = 2
-    plt.rcParams["text.usetex"] = True if shutil.which("latex") else False
-    plt.rcParams["text.latex.preamble"] = (
-        r"\usepackage{amsmath}" if plt.rcParams["text.usetex"] else ""
-    )
+
+    if shutil.which("latex") and get_enable_tex():
```
jscanvic, line 93: I don't really like this solution - I don't think we can expect users deal with the TeX state themselves. I think we should be able to fix the issue by simply try making one plot with LaTeX once, catching errors and disabling it in the back if it fails. Said otherwise, simply improving the detection of a working LaTeX env. Thoughts?

Andrewwango: done

deepinv/utils/plotting.py
```diff
@@ -31,20 +33,66 @@ def get_default_plot_fontsize() -> int:
...
     plt.rcParams["lines.linewidth"] = 2
-    plt.rcParams["text.usetex"] = True if shutil.which("latex") else False
-    plt.rcParams["text.latex.preamble"] = (
-        r"\usepackage{amsmath}" if plt.rcParams["text.usetex"] else ""
-    )
+
+    # If plot gives TeX errors, force disable TeX globally
+    if not get_checked_tex():
+        try:
+            _ = plot({r"$\mathbf{x}$": torch.rand(1, 1, 2, 2)}, show=True, close=True)
```
jscanvic: Use regular matplotlib code instead of our plotting code - it should be more robust and avoid the infinite recursions we're having in CI

docs/source/user_guide/other/utils.rst
```diff
@@ -19,6 +19,9 @@ which can be used to quickly plot a list of tensor images.
     >>> x2 = torch.rand(4, 3, 16, 16)
     >>> plot([x1, x2], titles=['x1', 'x2'])
 
+.. hint::
+
+  Do you get a matplotlib LaTeX error when plotting? Disable LaTeX using `dinv.utils.disable_tex()`
```
jscanvic, line 24: Should we leave this in?

deepinv/utils/plotting.py
Andrewwango: What would you suggest @jscanvic ?
i'm worried to use something that might not reproduce the error on someone's setup, because I cannot reproduce the error.

examples/basics/demo_quickstart.py
```diff
@@ -63,6 +63,9 @@
 # DeepInverse implements
 # :ref:`many different types of physics <physics>` across various imaging modalities.
 # Physics also possess noise models such as Gaussian or Poisson noise.
+#
+# .. hint::
+#     Do you get a matplotlib LaTeX error when plotting? Disable LaTeX using `dinv.utils.disable_tex()`
```
jscanvic, line 68: This too

docs/source/user_guide/other/utils.rst
Andrewwango: I guess just in case

deepinv/utils/plotting.py
Andrewwango: I guess I could just do imshow; show()

jscanvic: Hmm right tbh I'm pretty confident the error doesn't depend much on what we plot as long as it renders LaTeX. I think a simple figure with plt.text rendered on it and saved to a in-memory file (can't remember the Python name for it) should work.

If you think we should do something closer to the code that's failing you can use the code from deepinv with a mock / global variable / something to avoid recursive calls

I'm happy with both

Andrewwango: I added a thing where it temp sets check to True then uses a finally to revert, to prevent recursion.

jscanvic: Thanks

Tmodrzyk:
Hi @Andrewwango sorry for the late review, @jscanvic was way faster than me. Q1 I definitely think this should be global, and that's the solution you went with so that's great. Q2 I also think the default value should be True because most users do have a working tex environment and probably just want tex to be supported in their titles / labels by default. 
All in all, looks good to me :)

PR #728, opened 2025-09-08
``Trainer`` refactor

### LINKED ISSUE #545: Make Trainer visualizations have plug-able interface
Problem:
I am looking for a way to log images to `wandb`.  I pass `wandb_vis=True` to the `Trainer(...)` but I have to set `plot_images=True`  to get images, however this means I get blocking matplotlib plots.  See: https://github.com/deepinv/deepinv/blob/0c6f5da176ffaef547afbb72c20bdc2429791199/deepinv/training/trainer.py#L852

Proposal:
Can we decouple these two visualization tools `wandb` and `matplotlib` and potentially other visualization tools. 
Perhaps a `Class VisLogger: ...` which has an interface of `should_plot(epoch)` and `plot(args, kwargs)`. Then I can pass this to the trainer and get the plots I want. 
```
dinv.Trainer(
    ...
-    plot_images=True,
-    wandb_vis=True
+   vis=VisLogger( "here we hide implementation details. "),
...    
)
```
Now the trainer no-longer has to handle combinations of booleans to determine what and where to plot. Users can easily create a custom `VisPlotter` and log to `wandb` or `rerun` or wherever is convenient for them. 
The `VisPlotter` interface would need some careful thinking and the names also can be better (hehehe) but that is the idea.

If others are in the same boat my current approach is (monkey patch):
```
from types import MethodType
...
def plot(self: dinv.training.Trainer, epoch, physics, x, y, x_net, train=True): ...
...
trainer.plot = MethodType(plot, trainer)
model = trainer.train()
```
Which is actually fine for my current experiments.
  
If the authors of this super useful library would entertain a PR in this spirit I'd be happy to give it a go. If you not looking for these kinds of contributions feel free to just close this issue.

Thanks for the amazing library

### LINKED ISSUE #553: Trainer improvements

Tmodrzyk:
This PR aims at rewriting the ``Trainer`` class to make it easier for users to customize, and also easier to maintain / extend for contributors. 
Right now the code is bloated by logging logic, so we propose a standalone ``RunLogger`` class which allows us to implement a disk logger, a weights and biases logger as well as an MLFlow logger.

We will also try to address some of the points in Issue #553. 
An audit / checklist of issues we found in the current ``Trainer``: [https://docs.google.com/document/d/12fZKxPQKAYreWcBQT6pHV2ojfIqtu0b3sphNdBNJKNQ/edit?tab=t.0]( https://docs.google.com/document/d/12fZKxPQKAYreWcBQT6pHV2ojfIqtu0b3sphNdBNJKNQ/edit?tab=t.0)


Andrewwango:
- [x] TODO: absorb #668 ?

deepinv/training/run_logger.py
```diff
@@ -0,0 +1,108 @@
...
+    Defines the interface for logging metrics, losses, images, and other
+    training artifacts during model training and evaluation.
+    """
+    
+    def __init__(self, run_name: Optional[str] = None, config: Optional[dict[str, Any]] = None):
+        """
+        Initialize the logger.
+        
+        :param run_name: Optional name for the training run
+        :param config: Configuration dictionary for the logger
```
Andrewwango: Can we be explicit about what config contains? Is it only for wandb/mlflow?

I'm generally against dicts as arguments where it could just be **kwargs or even better naming all the args. The config dict feels a bit like TF notation

deepinv/training/run_logger.py
```diff
@@ -0,0 +1,108 @@
...
+import pickle
+from pathlib import Path
+from datetime import datetime
+import numpy as np
+import torch
+import matplotlib.pyplot as plt
+from torchvision.utils import save_image
+import warnings
+from typing import Any, Optional, Union
+
```
Andrewwango, line 25: Can you link in the PR description related issues?

deepinv/training/run_logger.py
```diff
@@ -0,0 +1,108 @@
...
+        :param run_name: Optional name for the training run
+        :param config: Configuration dictionary for the logger
+        """
+        self.run_name = run_name or f"run_{datetime.now().strftime('%Y%m%d_%H%M%S')}"
+        self.config = config or {}
+        self.current_epoch = 0
+        self.current_step = 0
+        
+    @abstractmethod
+    def start_run(self, hyperparams: Optional[dict[str, Any]] = None):
```
Andrewwango: Again, can we be explicit at all about what keys are in the hyperparam dict?

deepinv/training/run_logger.py
```diff
@@ -0,0 +1,108 @@
...
+        pass
+    
+    @abstractmethod
+    def finish_run(self):
+        """
+        Finalize and close the training run.
+        """
+        pass
+    
+    def set_step(self, step: int):
```
Andrewwango: Could you make it a property?

Andrewwango (changes requested):
Looks good so far!

deepinv/training/run_logger.py
Tmodrzyk: I tend to agree, we choose this solution for now because it's what is used to configure wandb in the Trainer right now.
We'll try to name the the arguments. :)

Andrewwango:
- [ ] TODO logging should fix #809

examples/self-supervised-learning/demo_artifact2artifact.py
```diff
@@ -229,14 +229,12 @@
     losses=loss,
     optimizer=optimizer,
     train_dataloader=train_dataloader,
-    compute_eval_losses=True,  # use self-supervised loss for evaluation
```
Andrewwango, line 232: why removing these two lines?

examples/self-supervised-learning/demo_ei_transforms.py
```diff
@@ -129,17 +129,15 @@
     physics=physics,
     online_measurements=True,
     train_dataloader=train_dataloader,
-    eval_dataloader=test_dataloader,
-    compute_eval_losses=True,  # use self-supervised loss for evaluation
```
Andrewwango, line 133: likewise

examples/self-supervised-learning/demo_equivariant_imaging.py
```diff
@@ -184,17 +185,12 @@
     losses=losses,
     optimizer=optimizer,
     train_dataloader=train_dataloader,
-    eval_dataloader=test_dataloader,
```
Andrewwango, line 187: so this means this example no longer does early stopping on the test dataloader? maybe for now it's ok that test_dataloader = val_dataloader

Andrewwango (changes requested):
I reviewed all files except trainer, run_logger, adversarial and test_trainer. Minor comments so far:

deepinv/training/run_logger.py
```diff
@@ -0,0 +1,499 @@
...
+
+
+@dataclass
+class RunLogger(ABC):
+    """
+    Abstract base class for logging training runs.
+    """
+
+    @abstractmethod
+    def init_logger(self, hyperparams: dict[str, Any] | None = None) -> None:
```
Andrewwango, line 34: you expose the hyperparams argument yet all examples seem to be only using Logger.init_logger(). is there a valid use case for passing a dict here that we can exemplify in some examples? or at least document the high level interface for loggers in the user guide somewhere alongside the trainer docs?

Andrewwango (commented):
small comment

examples/self-supervised-learning/demo_artifact2artifact.py
MaximeSongIdris: compute_eval_losses is replaced by `compute_losses_on`, but the change isn't done yet on this file

examples/self-supervised-learning/demo_ei_transforms.py
MaximeSongIdris: same, compute_eval_losses is depreciated in favor of compute_losses_on (can be used to compute losses on train, val and test)

deepinv/training/run_logger.py
MaximeSongIdris: hyperparams are experience settings that we want to save for our run (it is used in WandbLogger and in LocalLogger)

MaximeSongIdris: it can be lr, batch_size, model name, ...

deepinv/training/run_logger.py
```diff
@@ -0,0 +1,499 @@
...
+        """
+        Start a new training run.
+
+        :param dict hyperparams: Dictionary of hyperparameters to log.
+        """
+        pass
+
+    @abstractmethod
+    def setLevel(self, level: str) -> None:
```
Andrewwango: 1. This could be nicer as a property?
2. `set_level` instead of `setLevel`

deepinv/training/run_logger.py
```diff
@@ -0,0 +1,499 @@
...
+
+
+class WandbLogger(RunLogger):
+    """
+    TODO
+    """
+
+    def __init__(
+        self,
+        log_dir: str = "logs",
```
Andrewwango, line 122: Seems like there are a lot of kwargs here, and these are all replicated in get_wandb_setup + init_logger, resulting in many lines of code. 

Are these all specific to wandb?

deepinv/training/run_logger.py
```diff
@@ -0,0 +1,499 @@
...
+            wandb_setup = {
+                "dir": wandb_save_dir,
+                "mode": wandb_logging_mode,
+                "project": wandb_proj_name,
+                "name": wandb_run_name,
+                "config": wandb_hp_config,
+            }
+        return wandb_setup
+
+    def init_logger(self, hyperparams: dict[str, Any] | None = None) -> None:
```
Andrewwango, line 177: What's the purpose of init_logger when the class has an init? I'm generally against "init" methods as ideally all init is done in the init.

deepinv/training/run_logger.py
```diff
@@ -0,0 +1,499 @@
...
+        # {loss_name_1: loss_value_1, ...} -> {phase/loss_name_1: loss_value_1, ...}
+        logs = {
+            f"{phase}/{loss_name}": loss_value
+            for loss_name, loss_value in losses.items()
+        }
+
+        # default x-axis is the current training step
+        self.wandb_run.log(logs, step=step)
+
+    def log_metrics(
```
Andrewwango: It seems like log_metrics will generally always duplicate a lot of code to log_losses. Could they become one method or do you think they are sufficiently conceptually separate?

deepinv/training/run_logger.py
```diff
@@ -0,0 +1,499 @@
...
+
+        self.run_log_dir = Path(log_dir) / Path(self.project_name) / self.run_name
+        self.loss_dir = self.run_log_dir / "losses"
+        self.metrics_dir = self.run_log_dir / "metrics"
+        self.images_dir = self.run_log_dir / "images"
+
+        self.stdout_logger = getLogger("stdout_logger")
+        self.stdout_logger.setLevel("INFO")
+
+    def init_logger(self, hyperparams: dict[str, Any] | None = None) -> None:
```
Andrewwango, line 320: Again it's unclear what the distinction is between the logger init and the init logger

deepinv/training/run_logger.py
```diff
@@ -0,0 +1,499 @@
...
+        if step is not None:
+            dir_path = dir_path / f"step_{step}"
+        os.makedirs(dir_path, exist_ok=True)
+
+        for k, (name, img) in enumerate(images.items()):
+            for i in range(img.size(0)):
+                img_name = f"{dir_path}/{name}_{k}_{self.img_counter + i}.png"
+                save_image(img[i], img_name)
+
+    def load_from_checkpoint(self, checkpoint_dict: dict[str, Any]):
```
Andrewwango, line 406: Given the name of this method, I would expect it to load the checkpoint: model weights. optimizer weights... Is this not true?

deepinv/training/trainer.py
```diff
@@ -1,624 +1,473 @@
+# Python standard modules
```
Andrewwango: remove comment

deepinv/training/trainer.py
```diff
@@ -1,624 +1,473 @@
...
-    :Basics:
+    **Dataset interface**
 
     The **dataloaders** should return data in the correct format for DeepInverse: see :ref:`datasets user guide <datasets>` for
     how to use predefined datasets, create datasets, or generate datasets. These will be checked automatically with :func:`deepinv.datasets.check_dataset`.
 
-    If the dataloaders do not return
-    measurements `y`, then you should use the `online_measurements=True` option which generates measurements in an online manner (optionally with parameters), running
-    under the hood `y=physics(x)` or `y=physics(x, **params)`. Otherwise if dataloaders do return measurements `y`, set `online_measurements=False` (default) otherwise
-    `y` will be ignored and new measurements will be generated online.
```
Andrewwango, line 56: keep this warning because it is not unpermissable - I do this often

deepinv/training/trainer.py
```diff
@@ -1,624 +1,473 @@
...
+    Losses can be logged at every step or at every epoch.
 
-    |sep|
-
-    :Evaluation:
-
-    .. note::
+    Training state is saved automatically every ``ckpt_interval`` epochs, and can be resumed with ``ckpt_pretrained``.
 
-        - **Supervised evaluation**: If ground-truth data is available for validation, use any
```
Andrewwango, line 109: Please can you carefully reintegrate all the newest updates to the trainer docstring, such as this one?

deepinv/training/trainer.py
```diff
@@ -771,553 +672,534 @@ def model_inference(self, y, physics, x=None, train=True, **kwargs):
...
-                if self.plot_convergence_metrics:
-                    x_net, self.conv_metrics = self.model(
-                        y, physics, x_gt=x, compute_metrics=True, **kwargs
-                    )
-                else:
-                    x_net = self.model(y, physics, **kwargs)
-        else:
-            x_net = self.model(y, physics, **kwargs)
+        with torch.set_grad_enabled(train):
+            if self.iterative_model_returns_different_outputs:
```
Andrewwango, line 697: Hmm I don't like the use of a flag that the user must set rather than a user preference. How about just:

```
out = self.model(...)
x_net = out[0] if isinstance(out, tuple) else out
```

deepinv/training/trainer.py
```diff
@@ -1,624 +1,473 @@
...
-    ckpt_pretrained: str | None = None
-    save_path: str | Path | None = "."
-    compare_no_learning: bool = False
+    compute_losses_on: list[str] = field(default_factory=lambda: ["train", "val"])
+    metrics: Metric | list[Metric] = field(default_factory=PSNR)
+    compute_metrics_on: list[str] = field(default_factory=lambda: ["val", "test"])
     no_learning_method: str = "A_adjoint"
-    grad_clip: float = None
-    check_grad: bool = False
-    wandb_vis: bool = False
```
Andrewwango, line 289: We are now deprecating lots of args, which means that a lot of users will be confused. We either need to very clearly and smoothly deprecate these (e.g. with deprecated alias + warning + internally construct loggers based on the args), or provide very clear instructions on how to transition

deepinv/training/run_logger.py
```diff
@@ -0,0 +1,499 @@
...
+    def setLevel(self, level: str) -> None:
+        """
+        Set the logging level.
+
+        :param str level: Logging level (e.g., 'DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL').
+        """
+        pass
+
+    @abstractmethod
+    def log_losses(
```
Andrewwango: where is the `eval_interval`, `plot_interval` etc. ?

deepinv/training/trainer.py
```diff
@@ -1,624 +1,473 @@
...
-    ckp_interval: int = 1
-    eval_interval: int = 1
-    plot_interval: int = 1
-    freq_plot: int = None
-    plot_images: bool = False
-    plot_measurements: bool = True
-    plot_convergence_metrics: bool = False
+    compute_comparison_on: list[str] = field(default_factory=lambda: ["val", "test"])
+
+    ## Checkpointing & Persistence (at the end of epoch)
```
Andrewwango, line 256: Why can't this checkpointing also be handled by the same run loggers?

deepinv/training/trainer.py
```diff
@@ -1,624 +1,473 @@
...
+    ckpt_dir: str | Path = Path("./checkpoints")
+    ckpt_pretrained: str | Path | None = None
+    ckpt_interval: int = 1
+
+    ## Logging & Monitoring
+    loggers: RunLogger | list[RunLogger] | None = field(
+        default_factory=lambda: [LocalLogger(Path("./logs"))]
```
Andrewwango, line 263: does this make the dir `logs` or not?

deepinv/training/trainer.py
```diff
@@ -1,624 +1,473 @@
...
+    )
+    log_images: bool = False
```
Andrewwango, line 265: what is the use of these `log_...` params when it should be handled by setting the appropriate args in the run logger?

deepinv/training/trainer.py
```diff
@@ -1,624 +1,473 @@
...
+                for l in self.losses
+            }
+            for phase in self.compute_losses_on
+        }  # 1 AverageMeter per loss term and 1 for the total loss
 
-        # make physics and data_loaders of list type
-        if type(self.physics) is not list:
-            self.physics = [self.physics]
+        if not hasattr(self, "train_loss_history_per_epoch"):
+            self.train_loss_history_per_epoch = []
```
Andrewwango, line 359: Why do we need these specific lists? Can we just store the history in the meters?

deepinv/training/trainer.py
```diff
@@ -771,553 +672,534 @@ def model_inference(self, y, physics, x=None, train=True, **kwargs):
...
-                f"No learning reconstruction method {self.no_learning_method} not recognized or physics does not implement it"
-            )
-
-        return x_nl
-
-    class _NoLearningModel(Reconstructor):
-        def __init__(self, *, trainer: Trainer):
-            super().__init__()
-            self.trainer = trainer
+    def _process_gradients(self) -> None:
```
```suggestion
    def _clip_gradients(self) -> None:
```
"process" is vague

deepinv/training/trainer.py
```diff
@@ -1,624 +1,473 @@
...
-                "Update progress bar frequency of 1 may slow down training on GPU. Consider setting freq_update_progress_bar > 1."
-            )
+                for m in self.metrics
+            }
+            for phase in self.compute_comparison_on
+        }
 
-        _ = self.load_model()
+        # Init logger specific for the Trainer
+        self.train_logger = getLogger("train_logger")
```
Andrewwango, line 388: It's confusing for me here the difference between this logger and the run loggers

deepinv/training/trainer.py
```diff
@@ -1,624 +1,473 @@
...
+        default_factory=lambda: [LocalLogger(Path("./logs"))]
+    )
+    log_images: bool = False
     rescale_mode: str = "clip"
-    display_losses_eval: bool = None
-    compute_eval_losses: bool = False
-    log_train_batch: bool = False
-    verbose: bool = True
-    verbose_individual_losses: bool = True
+    log_grad: bool = False
```
Andrewwango, line 267: I'm under the impression that the new run loggers should be able to handle everything related to logging. Can we remove this argument too?

deepinv/training/trainer.py
```diff
@@ -1,624 +1,473 @@
...
                 )
-            else:
-                self.early_stop = None
-        elif self.early_stop is not None:
-            assert (
-                isinstance(self.early_stop, int) and self.early_stop > 0
-            ), "early_stop should be a positive integer or None."
-
-        if self.early_stop:
-            if self.early_stop_on_losses:
```
Andrewwango, line 526: You will need to carefully look at the newest trainer and incorporate the flexibility of early stopping, becasue it is nuanced but quite crucial!

deepinv/tests/test_trainer.py
```diff
@@ -791,137 +837,32 @@ def test_loss_logging(
...
     trainer = dinv.Trainer(
         model=dummy_model,
         losses=losses,
         metrics=metrics,
-        epochs=2,
+        epochs=epochs,
         physics=physics,
         device=device,
         train_dataloader=dataloader,
-        eval_dataloader=eval_dataloader,
```
Andrewwango, line 805: Again it seems like there's huge negative diffs in these tests where you will need to merge from main

Andrewwango (changes requested):
Hey guys, thanks for the big changes. I have some comments below!

deepinv/training/trainer.py
MaximeSongIdris: yes, it is vague because we optionally clip gradient and optionally "print" gradient norm

deepinv/training/run_logger.py
MaximeSongIdris: 1. Since the method isn't about an attribute, but a change in the object behavior, it is preferrable to keep it as method
2. yes, will change it

deepinv/training/run_logger.py
MaximeSongIdris: yes they are specific to wandb

get_wandb_setup is a method that is used in init_logger, but it was also intended as a normal function that could be used outside of the class  (c.f. classmethod), that's why it requires a lot of args to work as standalone.

deepinv/training/run_logger.py
MaximeSongIdris: Ideally we want to :
```
logger = RunLogger(...)
trainer = Trainer(logger)
trainer.train() -> logger.init_logger
```

In init, we specify the base config of the logger, then in init_logger, we use trainer-specific arg to add more config, before setting up the logger effectively. It makes resuming wandb possible, since we use the trainer to load a checkpoint, use this checkpoint to get the resume id of wandb, give it to wandb, then setting up.

Why then not instanciating directly wandbLogger inside of the Trainer ? For modularity, I prefer that the logger is an external object of the Trainer, and the only interation is through the public methods of RunLogger.

deepinv/training/run_logger.py
MaximeSongIdris: it is for clarity, since metric and loss are conceptually different
also in case when people want to write their own custom RunLogger and want to differentiate the treatment between loss and metric

it is true that it is useless here, i will probably refactor it by writing :
def log_metrics():
   self.log_losses()

deepinv/training/trainer.py
MaximeSongIdris: This logger display on the terminal what's happening in the class.
WandBLogger display specific things to WANDB of what's happening during training.
FileLogger write specific things to a file in your disk of what's happening during training.

deepinv/training/trainer.py
MaximeSongIdris: RunLogger is in charge of interacting with outside object (WANDB server, File in your disk).
Here, log_grad tells us to compute and display gradient norm in the terminal (so under the charge of Trainer own logger.

deepinv/training/trainer.py
MaximeSongIdris: Ok, will look at it

deepinv/tests/test_trainer.py
MaximeSongIdris: Yes, will look at it later, I think we need to first validate the new design of the Trainer.

deepinv/training/run_logger.py
jscanvic: (Nice dependency inversion! 👀)

Andrewwango:
@Tmodrzyk @MaximeSongIdris feel free to re-request my review when ready! Thanks for the hard work!

deepinv/training/trainer.py
MaximeSongIdris: Done

deepinv/training/trainer.py
MaximeSongIdris: image
Is it fine to do it like this ?

deepinv/training/trainer.py
MaximeSongIdris: Done

deepinv/training/trainer.py
MaximeSongIdris: Yes, I think a guide will be cleaner.

deepinv/training/trainer.py
MaximeSongIdris: Actually, I tried at first. I had the LocalLogger do the main checkpointing, but  I also needed to have the WandbLogger write in the checkpoint to save its run number. And it becomes quite messy because: 
- 1) each RunLogger requires to properly deal with file path
- 2) there were unnecessary read/write on disk (each RunLogger reads and modifies the checkpoint file separately)

deepinv/training/trainer.py
MaximeSongIdris: I don't know what you mean exactly.
This line instanciate the LocalLogger object and create dir at the path ".logs"

deepinv/training/trainer.py
MaximeSongIdris: It is the Trainer class that deal with whether we want to log gradients or images, and RunLogger objects provide only the mean to do the logging.

deepinv/training/trainer.py
MaximeSongIdris: Actually, the meters saved the losses and metrics during a whole epoch to compute epoch statistics, then it is reset at the begining of each epoch. So we use train_loss_history_per_epoch and val_metrics_history_per_epoch to have the list of all statistics accross epochs.

deepinv/training/trainer.py
MaximeSongIdris: For now, the early stopping only use the ``max_batch_steps`` and ``stop_criterion``, if we want to have the early stopping on losses, the current way is to overwrite ``stop_criterion``.

deepinv/training/trainer.py
MaximeSongIdris: initially it was an option that we could use to display some intermediate results from iterative model, but now it is obsolete

as to whether we use a flag or not, I am not sure, I prefer personally to have an explicit flag, so user are informed that their interative model works too

MaximeSongIdris: can you confirm that you prefer without flag ?

deepinv/training/run_logger.py
MaximeSongIdris: All the control logics are in Trainer class, while in RunLogger, we have only the execution logics.
I removed the plot interval and the eval_interval and instead I do the plotting and the validation at each epoch

Should I put it again ?

deepinv/training/run_logger.py
MaximeSongIdris: here it only takes from a loaded checkpoint necessary information for the RunLogger

MaximeSongIdris:
Sry I was on vacation @Andrewwango , but I am back now. I did some small changes and mostly revamp the docstring in Trainer. Can you tell me if you are fine with the trainer.py and run_logger.py ?

Tmodrzyk:
Hi @Andrewwango, sorry I was very busy and did not take the time to look at this. :disappointed: 
I'll have a look this week.

deepinv/training/run_logger.py
Tmodrzyk: I applied the changes.

deepinv/training/run_logger.py
Tmodrzyk: You're absolutely right, it was almost entirely duplicated code. I factored them into a single ``log_scalar`` and propagated the change to the other files. (in [0c694c5](https://github.com/deepinv/deepinv/pull/728/commits/0c694c5339149e0e4e92187cb6f005b3da882ee2))

deepinv/training/run_logger.py
Tmodrzyk: ``init_logger`` is called when a run begins and is used to create the log directories for this specific run, whereas ``init`` is called when the object is created, typically when you instantiate the trainer.

deepinv/training/trainer.py
Tmodrzyk: Totally agree that the naming / purpose was too vague. I split it into two different private functions in [58a09bf](https://github.com/deepinv/deepinv/pull/728/commits/58a09bfc3520e32ebd40b652fae484fea2db9926)

deepinv/training/trainer.py
Tmodrzyk: I agree this is confusing but we still wanted to display some debug info about what's happening in the class. I'm not sure we choose the best solution.

Tmodrzyk:
Hi @Andrewwango, I implemented some of your comments. I also agree with the general comment that the loggers should handle _all_ the logging. I'll try to come up with a solution. Do you think we should try to merge main into this branch before continuing ? I saw you mentioned some huge diffs in the tests.

Tmodrzyk:
Update: this PR has unfortunately gone stale. In retrospect, the PR tried to fix too many things at once. 
But the need for a refactor remains, and proper logging would still probably de-clutter a lot of the current implementation.

Andrewwango:
@Tmodrzyk definitely a challenging PR, but focussing purely on the logging is not a bad idea, and hopefully we can finish that without affecting too many other parts of the trainer, and I think it will have a big impact anyway. I liked your idea of the loggers, what are the blockers now?

The other option is to start from scratch again and look at a fundamental redesign of trainer for 2026 onwards, maybe we have some more interesting perspectives compared to last year. Happy to meet or work on it during hackathon!

PR #735, opened 2025-09-08
Reduced resolution loss

Andrewwango:
The reduced resolution loss for blur/downsampling problems which uses the measurement as the GT and a redowngraded image as the input, following e.g. Irani et al., Scanvic et al., and in pansharpening the classic reduced resolution "Wald" protocol.

### DIFF
--- a/deepinv/training/trainer.py
+++ b/deepinv/training/trainer.py
@@ -813,13 +813,10 @@ class Trainer:
             x_net = x_net.detach()
 
             # Log metrics
-            logs = (
-                logs
-                if self.disable_train_metrics and train
-                else self.compute_metrics(
+            if not (self.disable_train_metrics and train):
+                logs = self.compute_metrics(
                     x, x_net, y, physics_cur, logs, train=train, epoch=epoch
                 )
-            )
 
             # Update the progress bar
             progress_bar.set_postfix(logs)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ New Features
 
 Changed
 ^^^^^^^
+- (Breaking) Make HDF5Dataset similar to Trainer in the unsupervised setting by using NaNs for ground truths instead of a copy of the measurements (:gh:`761` by `Jérémy Scanvic`_)
 
 Fixed
 ^^^^^
@@ -26,6 +27,7 @@ New Features
 - Added GitHub action checking import time (:gh:`680` by `Julian Tachella`_)
 - Client model for server-side inference for using models in the cloud (:gh:`691` by `Andrew Wang`_)
 - Reduced resolution self-supervised loss (:gh:`735` by `Andrew Wang`_)
+- Add dinv.utils.disable_tex to disable LaTeX (:gh:`726` by `Andrew Wang`_)
 
 Changed
 ^^^^^^^
--- a/deepinv/datasets/datagenerator.py
+++ b/deepinv/datasets/datagenerator.py
@@ -112,12 +112,13 @@ class HDF5Dataset(ImageDataset):
         else:
             y = self.cast(torch.from_numpy(self.y[index]))
 
-        x = y
         if not self.unsupervised:
             x = self.cast(torch.from_numpy(self.x[index]))
 
-        if self.transform is not None:
-            x = self.transform(x)
+            if self.transform is not None:
+                x = self.transform(x)
+        else:
+            x = torch.tensor(torch.nan, dtype=y.dtype, device=y.device)
 
         if self.load_physics_generator_params:
             params = {
--- a/deepinv/datasets/utils.py
+++ b/deepinv/datasets/utils.py
@@ -9,7 +9,6 @@ from pathlib import Path
 import requests
 from tqdm.auto import tqdm
 
-from scipy.io import loadmat as scipy_loadmat
 from numpy import ndarray
 
 from torch import randn, Tensor, stack, zeros_like
@@ -109,11 +108,17 @@ def extract_tarball(file_path: Union[str, Path], extract_dir: Union[str, Path]) 
 def loadmat(fname: str, mat73: bool = False) -> dict[str, ndarray]:
     """Load MATLAB array from file.
 
+    .. note::
+
+        This function depends on the ``scipy`` package. You can install it with ``pip install scipy``.
+
     :param str fname: filename to load
     :param bool mat73: if file is MATLAB 7.3 or above, load with ``mat73``. Requires
         ``mat73``, install with ``pip install mat73``.
     :return: dict with str keys and array values.
     """
+    from scipy.io import loadmat as scipy_loadmat
+
     if mat73:
         try:
             from mat73 import loadmat as loadmat73
--- a/deepinv/optim/data_fidelity.py
+++ b/deepinv/optim/data_fidelity.py
@@ -476,7 +476,7 @@ class LogPoissonLikelihood(DataFidelity):
 
         \datafid{z}{y} =  N_0 (1^{\top} \exp(-\mu z)+ \mu \exp(-\mu y)^{\top}x)
 
-    Corresponds to LogPoissonNoise with the same arguments N0 and mu.
+    Corresponds to :class:`deepinv.physics.LogPoissonNoise` with the same arguments :math:`N_0` and :math:`\mu`.
     There is no closed-form of the proximal operator known.
 
     :param float N0: average number of photons
--- a/deepinv/optim/dpir.py
+++ b/deepinv/optim/dpir.py
@@ -6,7 +6,7 @@ from deepinv.optim.data_fidelity import L2
 from deepinv.optim.prior import PnP
 from deepinv.optim.optimizers import create_iterator
 from deepinv.optim import BaseOptim
-import numpy as np
+import torch
 
 
 def get_DPIR_params(noise_level_img):
@@ -19,12 +19,17 @@ def get_DPIR_params(noise_level_img):
     max_iter = 8
     s1 = 49.0 / 255.0
     s2 = noise_level_img
-    sigma_denoiser = np.logspace(np.log10(s1), np.log10(s2), max_iter).astype(
-        np.float32
+    sigma_denoiser = torch.logspace(
+        torch.log10(torch.tensor(s1, dtype=torch.float32)),
+        torch.log10(torch.tensor(s2, dtype=torch.float32)),
+        steps=max_iter,
+        dtype=torch.float32,
     )
+
     stepsize = (sigma_denoiser / max(0.01, noise_level_img)) ** 2
     lamb = 1 / 0.23
-    return list(sigma_denoiser), list(lamb * stepsize), max_iter
+
+    return sigma_denoiser, lamb * stepsize, max_iter
 
 
 class DPIR(BaseOptim):
--- a/deepinv/tests/conftest.py
+++ b/deepinv/tests/conftest.py
@@ -68,7 +68,7 @@ def non_blocking_plots():
         # Use a non-interactive backend to avoid blocking the tests
         matplotlib.use("Agg", force=True)
         plt.close("all")
-        # NOTE: Why do we reload matplotlib.pyplot?
+        # Reload matplotlib.pyplot to force usage
         importlib.reload(plt)
         yield
     finally:
--- a/deepinv/tests/test_datasets.py
+++ b/deepinv/tests/test_datasets.py
@@ -668,6 +668,13 @@ def download_nbu():
 
 def test_load_nbu_dataset(download_nbu):
     """Check that dataset correct length and type."""
+    pytest.importorskip(
+        "scipy",
+        reason="This test requires scipy. It should be "
+        "installed with `pip install "
+        "scipy`",
+    )
+
     dataset = NBUDataset(download_nbu, satellite="gaofen-1", download=False)
     check_dataset_format(dataset, length=5, dtype=Tensor, shape=(4, 256, 256))
     assert torch.all(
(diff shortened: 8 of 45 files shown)


deepinv/loss/__init__.py
```diff
@@ -1,7 +1,7 @@
 from .mc import MCLoss
 from .ei import EILoss
 from .moi import MOILoss, MOEILoss
-from .sup import SupLoss
+from .sup import SupLoss, ReducedResolutionLoss
```
jscanvic, line 4: I'm happy with leaving it as is but it might be misleading to put it in `deepinv.loss.sup` when it's not really a supervised loss

deepinv/loss/sup.py
```diff
@@ -41,3 +43,80 @@ def forward(self, x_net, x, **kwargs):
...
+        to the measurements :math:`y`.
+
+    .. info::
+
+        During training, consider using the `disable_train_metrics` option in :class:`deepinv.Trainer` to prevent a shape
+        mismatch during metric computation since the reduced resolution output will smaller than ground truth.
+
+    :param Metric, torch.nn.Module metric: metric used for computing data consistency,
+        which is set as the mean squared error by default.
+    """
```
jscanvic, line 80: Add a reference to Irani's paper?

deepinv/loss/sup.py
```diff
@@ -41,3 +43,80 @@ def forward(self, x_net, x, **kwargs):
...
+    class ReducedResolutionModel(Reconstructor):
+        def __init__(self, model):
+            super().__init__()
+            self.model = model
+
+        def forward(self, y, physics, **kwargs):
+            with torch.set_grad_enabled(self.training):
+                if self.training:
+                    try:
+                        y_rr = physics.A(y)
```
jscanvic: In Michaeli's paper they don't have noise but in ours for deblurring we have noise and we use the noise model to generate the doubly corrupted measurements. Might make more sense to use `physics(y)` instead, or adding an option to compute either version. 

Also: `z` instead of `y_rr`?

deepinv/loss/sup.py
```diff
@@ -41,3 +43,80 @@ def forward(self, x_net, x, **kwargs):
...
+    .. info::
+
+        During training, consider using the `disable_train_metrics` option in :class:`deepinv.Trainer` to prevent a shape
+        mismatch during metric computation since the reduced resolution output will smaller than ground truth.
+
+    :param Metric, torch.nn.Module metric: metric used for computing data consistency,
+        which is set as the mean squared error by default.
+    """
+
+    def forward(self, x_net, y, *args, **kwargs):
```
jscanvic: Random thought: there's this pattern in the losses where we either expect the caller to feed in the pre-computed tensors (x_net) or we compute them in the loss directly (y_rr). Arguably, the 1st pattern enables more global optimizations (sharing variables between multiple loss for x_net, parallelization in data loading for y_rr) while the 2nd pattern is more convenient (we don't have to deal with computing stuff). I think it'd be nice to always have the choice to do one or the other. Not sure how though!

jscanvic: Typing?

deepinv/loss/sup.py
```diff
@@ -41,3 +43,80 @@ def forward(self, x_net, x, **kwargs):
...
+            return model
+        else:
+            return self.ReducedResolutionModel(model)
+
+    class ReducedResolutionModel(Reconstructor):
+        def __init__(self, model):
+            super().__init__()
+            self.model = model
+
+        def forward(self, y, physics, **kwargs):
```
jscanvic: Typing?

deepinv/training/trainer.py
```diff
@@ -691,6 +694,9 @@ def compute_metrics(
         :param int epoch: current epoch.
         :returns: The logs with the metrics.
         """
+        if train and self.disable_train_metrics:
```
jscanvic: I'm happy with leaving it as is but I'd prefer guarding the calls instead of hijacking the behavior of the function

deepinv/training/trainer.py
```diff
@@ -117,6 +117,8 @@ class Trainer:
     :param Metric, list[Metric] metrics: Metric or list of metrics used for evaluating the model.
         They should have ``reduction=None`` as we perform the averaging using :class:`deepinv.utils.AverageMeter` to deal with uneven batch sizes.
         :ref:`See the libraries' evaluation metrics <metric>`. Default is :class:`PSNR <deepinv.loss.metric.PSNR>`.
+    :param bool disable_train_metrics: if `True`, do not compute metrics on train set. Useful if your model output during training
+        is in the wrong shape for metric computation.
```
jscanvic, line 122: Can you elaborate on what's going on here? I'm not entirely sure

deepinv/tests/test_loss.py
```diff
@@ -666,3 +677,41 @@ def test_stacked_loss(device, imsize):
...
+        physics = dinv.physics.BlurFFT(
+            x.shape[1:], filter=dinv.physics.blur.gaussian_blur(0.4), device=device
+        )
+    else:
+        raise ValueError()
+
+    y = physics(x)
+
+    model.eval()
+    x_hat_eval = model(y, physics)  # just asjoint
```
jscanvic: adjoint*?

deepinv/tests/test_loss.py
```diff
@@ -666,3 +677,41 @@ def test_stacked_loss(device, imsize):
...
+    elif physics_name == "blur":
+        physics = dinv.physics.Blur(
+            filter=dinv.physics.blur.gaussian_blur(0.4), device=device
+        )
+    elif physics_name == "blurfft":
+        physics = dinv.physics.BlurFFT(
+            x.shape[1:], filter=dinv.physics.blur.gaussian_blur(0.4), device=device
+        )
+    else:
+        raise ValueError()
```
jscanvic: Add a message?

deepinv/tests/test_loss.py
```diff
@@ -666,3 +677,41 @@ def test_stacked_loss(device, imsize):
...
+
+    model.eval()
+    x_hat_eval = model(y, physics)  # just asjoint
+    assert x_hat_eval.shape == x.shape
+    assert metric(x_hat_eval, x) < 50
+
+    model.train()
+    x_hat_train = model(y, physics)
+    assert x_hat_train.shape == y.shape
+    assert metric(x_hat_train, y) < 50
```
jscanvic, line 723: I'm not sure I understand the `metric(...) < 50`. What is it we're testing? That it doesn't work too well?

deepinv/tests/test_loss.py
```diff
@@ -666,3 +677,41 @@ def test_stacked_loss(device, imsize):
...
+        physics = dinv.physics.Downsampling(filter=None, factor=2, device=device)
+    elif physics_name == "downsamplingmatlab":
+        physics = dinv.physics.DownsamplingMatlab(factor=2, device=device)
+    elif physics_name == "blur":
+        physics = dinv.physics.Blur(
+            filter=dinv.physics.blur.gaussian_blur(0.4), device=device
+        )
+    elif physics_name == "blurfft":
+        physics = dinv.physics.BlurFFT(
+            x.shape[1:], filter=dinv.physics.blur.gaussian_blur(0.4), device=device
```
jscanvic, line 706: Just wondering, could we leave out the `device=device` in every condition and have `physics.to(device)` after the conditional?

deepinv/tests/test_loss.py
```diff
@@ -319,15 +322,22 @@ def test_losses(
...
 
     save_dir = tmp_path / "dataset"
     # choose backbone denoiser
     backbone = dinv.models.AutoEncoder(
         dim_input=imsize[0] * imsize[1] * imsize[2], dim_mid=64, dim_hid=16
-    ).to(device)
+    )
+
+    if loss_name == "reducedresolution":
+        backbone = dinv.models.UNet(1, 1, scales=2)
```
jscanvic, line 341: Can you elaborate on why it's needed?

deepinv/tests/test_loss.py
```diff
@@ -319,15 +322,22 @@ def test_losses(
...
 
     # choose a reconstruction architecture
-    model = dinv.models.ArtifactRemoval(backbone, device=device)
+    model = dinv.models.ArtifactRemoval(backbone.to(device), device=device)
```
jscanvic: Again, just wondering, could we leave out the `to(device)` and the `device=device` and add a `model.to(device)` below?

deepinv/loss/sup.py
```diff
@@ -41,3 +43,80 @@ def forward(self, x_net, x, **kwargs):
...
+            with torch.set_grad_enabled(self.training):
```
jscanvic: Does it make sense to disable grad here specifically?

deepinv/loss/sup.py
```diff
@@ -41,3 +43,80 @@ def forward(self, x_net, x, **kwargs):
         :return: (:class:`torch.Tensor`) loss.
         """
         return self.metric(x_net, x)
+
+
+class ReducedResolutionLoss(SupLoss):
```
jscanvic, line 52: Creative name - I like it!

deepinv/loss/sup.py
```diff
@@ -41,3 +43,80 @@ def forward(self, x_net, x, **kwargs):
...
+    r"""
+    Reduced resolution loss for blur and downsampling problems.
+
+    The reduced resolution loss is defined as
+
+    .. math::
+
+        \frac{1}{n}\|y-\inverse{\forw{y}}\|^2
+
+    where :math:`\forw{y}` is the further-degraded measurement, and the measurement :math:`y` is used a supervisory signal.
```
jscanvic: Happy to leave it as is but if we go down the "reduced resolution" route, we might want to emphasize resolution reduction in the docstrings more than the idea of twice measuring? I feel like "measurement" might be too broad here

jscanvic (commented):
Thanks for the great feature! A few comments

deepinv/loss/__init__.py
Andrewwango: true but mechanistically it's sup I guess

deepinv/loss/sup.py
Andrewwango: Hmm I agree and there were two design choices. IMO the biggest flaw in the second approach is to waste the original `x_net` which is always computed, and perform another forward pass. Am I right in thinking this?

deepinv/training/trainer.py
Andrewwango: What do you mean by guarding?

deepinv/tests/test_loss.py
Andrewwango: That the x_hat_train is a `f(A(y))` where `f = A^T` i.e. a degraded version of `y`, testing that the loss adapted model actually performs degradataion

deepinv/training/trainer.py
jscanvic: Instead of

```
def elsewhere(...):
    val = compute_metrics(...)

def compute_metrics(...):
    if ...:
       return dummy_value
    pass
```

have


```
def elsewhere(...):
    if ...:
       compute_metrics(...)

def compute_metrics(...):
    pass
```

Lemme know if that doesn't make sense lol

deepinv/loss/sup.py
jscanvic: Exactly

deepinv/tests/test_loss.py
Andrewwango: I guess but it's not a big deal I don't think

deepinv/loss/sup.py
Andrewwango: done

deepinv/loss/sup.py
Andrewwango: done

deepinv/loss/sup.py
Andrewwango: Not sure if this is needed

Andrewwango: done

deepinv/loss/sup.py
Andrewwango: done

deepinv/training/trainer.py
Andrewwango: done, thanks!

deepinv/tests/test_loss.py
Andrewwango: done

deepinv/tests/test_loss.py
Andrewwango: done

deepinv/tests/test_loss.py
Andrewwango: done

deepinv/tests/test_loss.py
Andrewwango: done

deepinv/loss/sup.py
Andrewwango: done, gthanks

deepinv/loss/sup.py
Andrewwango: done

deepinv/training/trainer.py
Andrewwango: doen

deepinv/training/trainer.py
```diff
@@ -808,8 +813,12 @@ def step(
...
             # Log metrics
-            logs = self.compute_metrics(
-                x, x_net, y, physics_cur, logs, train=train, epoch=epoch
+            logs = (
+                logs
+                if self.disable_train_metrics and train
+                else self.compute_metrics(
+                    x, x_net, y, physics_cur, logs, train=train, epoch=epoch
+                )
             )
```
jscanvic: Sorry for the nitpick

```suggestion
            if not (self.disable_train_metrics and train):
                logs = self.compute_metrics(
                    x, x_net, y, physics_cur, logs, train=train, epoch=epoch
                )
```

deepinv/tests/test_loss.py
jscanvic: Oh so that's what DummyModel does, gotcha!

deepinv/tests/test_loss.py
jscanvic: Gotcha!

PR #739, opened 2025-09-09
Linear solver backward 

mh-nguyen712:

Write custom backward function for `least_squares` solver:
- Can differentiate w.r.t all input and physics parameters
- Less memory consumption
- Sometimes faster than backprop
- But not support TensorList  

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ New Features
 - Dataset base class + ImageFolder and TensorDataset classes (:gh:`622` by `Andrew Wang`_)
 - Added GitHub action checking import time (:gh:`680` by `Julian Tachella`_)
 - Client model for server-side inference for using models in the cloud (:gh:`691` by `Andrew Wang`_)
+- O(1) memory backprop for linear solver (:gh:`TODO` by `Minh Hai Nguyen`_)
 
 Changed
 ^^^^^^^
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ New Features
 
 Changed
 ^^^^^^^
+- (Breaking) Make HDF5Dataset similar to Trainer in the unsupervised setting by using NaNs for ground truths instead of a copy of the measurements (:gh:`761` by `Jérémy Scanvic`_)
 
 Fixed
 ^^^^^
@@ -25,7 +26,9 @@ New Features
 - Dataset base class + ImageFolder and TensorDataset classes (:gh:`622` by `Andrew Wang`_)
 - Added GitHub action checking import time (:gh:`680` by `Julian Tachella`_)
 - Client model for server-side inference for using models in the cloud (:gh:`691` by `Andrew Wang`_)
+- Add dinv.utils.disable_tex to disable LaTeX (:gh:`726` by `Andrew Wang`_)
 - O(1) memory backprop for linear solver and example (:gh:`739` by `Minh Hai Nguyen`_)
+
 
 Changed
 ^^^^^^^
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,18 @@ Current
 
 New Features
 ^^^^^^^^^^^^
+
+Changed
+^^^^^^^
+
+Fixed
+^^^^^
+
+
+v0.3.4
+------
+New Features
+^^^^^^^^^^^^
 - Quickstart tutorials + clean examples (:gh:`622` by `Andrew Wang`_)
 - Dataset base class + ImageFolder and TensorDataset classes (:gh:`622` by `Andrew Wang`_)
 - Added GitHub action checking import time (:gh:`680` by `Julian Tachella`_)
@@ -28,6 +40,8 @@ Fixed
 - NaN motion blur generator (:gh:`685` by `Matthieu Terris`_)
 - Fix the condition for break in compute_norm (:gh:`699` by `Quentin Barthélemy`_)
 - Python 3.9 backward compatibility and zip_strict (:gh:`707` by `Andrew Wang`_)
+
+
 
 v0.3.3
 ------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,7 +25,7 @@ New Features
 - Dataset base class + ImageFolder and TensorDataset classes (:gh:`622` by `Andrew Wang`_)
 - Added GitHub action checking import time (:gh:`680` by `Julian Tachella`_)
 - Client model for server-side inference for using models in the cloud (:gh:`691` by `Andrew Wang`_)
-- O(1) memory backprop for linear solver (:gh:`TODO` by `Minh Hai Nguyen`_)
+- O(1) memory backprop for linear solver and example (:gh:`739` by `Minh Hai Nguyen`_)
 
 Changed
 ^^^^^^^
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Current
 
 New Features
 ^^^^^^^^^^^^
+- Add statistics for SAR imaging + fix variance of GammaNoise in doc (:gh:`740` by `Louise Friot Giroux`)
 
 Changed
 ^^^^^^^
@@ -16,6 +17,7 @@ Changed
 
 Fixed
 ^^^^^
+- Reduced CI python version tests (:gh:`746` by `Mathieu Terris`_)
 
 
 v0.3.4
@@ -26,9 +28,10 @@ New Features
 - Dataset base class + ImageFolder and TensorDataset classes (:gh:`622` by `Andrew Wang`_)
 - Added GitHub action checking import time (:gh:`680` by `Julian Tachella`_)
 - Client model for server-side inference for using models in the cloud (:gh:`691` by `Andrew Wang`_)
+- Reduced resolution self-supervised loss (:gh:`735` by `Andrew Wang`_)
 - Add dinv.utils.disable_tex to disable LaTeX (:gh:`726` by `Andrew Wang`_)
+- Add BSDS500 dataset (:gh:`749` by `Johannes Hertrich`_ and `Sebastian Neumayer`_)
 - O(1) memory backprop for linear solver and example (:gh:`739` by `Minh Hai Nguyen`_)
-
 
 Changed
 ^^^^^^^
@@ -43,6 +46,7 @@ Fixed
 - NaN motion blur generator (:gh:`685` by `Matthieu Terris`_)
 - Fix the condition for break in compute_norm (:gh:`699` by `Quentin Barthélemy`_)
 - Python 3.9 backward compatibility and zip_strict (:gh:`707` by `Andrew Wang`_)
+- Fix numerical instability of Bicgstab solver(:gh:`739` by `Minh Hai Nguyen`_)
 
 
 
@@ -448,3 +452,4 @@ Authors
 .. _Sebastian Neumayer: https://www.tu-chemnitz.de/mathematik/invimg/index.en.php
 .. _Romain Vo: https://github.com/romainvo
 .. _Quentin Barthélemy: https://github.com/qbarthelemy
+.. _Louise Friot Giroux: https://github.com/Louisefg
--- a/deepinv/optim/utils.py
+++ b/deepinv/optim/utils.py
@@ -308,17 +308,31 @@ def bicgstab(
     max_iter = int(max_iter)
 
     tol = dot(b, b, dim=dim).real * (tol**2)
+    eps = torch.finfo(b.dtype).eps  # Breakdown tolerance, to avoid division by zero
     flag = False
     for i in range(max_iter):
         y = right_precon(left_precon(p))
         v = A(y)
-        alpha = rho / dot(r_hat, v, dim=dim)
+        # Safeguard: avoid division by small/zero
+        alpha_denom = dot(r_hat, v, dim=dim)
+        alpha = torch.where(
+            torch.abs(alpha_denom) > eps, rho / alpha_denom, torch.zeros_like(rho)
+        )
+
         h = x + alpha * y
         s = r - alpha * v
         z = right_precon(left_precon(s))
         t = A(z)
-        omega = dot(left_precon(t), left_precon(s), dim=dim) / dot(
-            left_precon(t), left_precon(t), dim=dim
+
+        # Safeguard: avoid division by small/zero
+        left_s = left_precon(s)
+        left_t = left_precon(t)
+        omega_num = dot(left_t, left_s, dim=dim)
+        omega_denom = dot(left_t, left_t, dim=dim)
+        omega = torch.where(
+            torch.abs(omega_denom) > eps,
+            omega_num / omega_denom,
+            torch.zeros_like(omega_num),
         )
 
         x = h + omega * z
@@ -330,7 +344,11 @@ def bicgstab(
             break
 
         rho_new = dot(r, r_hat, dim=dim)
-        beta = (rho_new / rho) * (alpha / omega)
+        # Safeguard for beta: if rho or omega small, set beta=0
+        beta_mask = (torch.abs(rho) > eps) & (torch.abs(omega) > eps)
+        beta = torch.where(
+            beta_mask, (rho_new / rho) * (alpha / omega), torch.zeros_like(rho_new)
+        )
         p = r + beta * (p - omega * v)
         rho = rho_new
 
--- a/deepinv/optim/utils.py
+++ b/deepinv/optim/utils.py
@@ -4,10 +4,10 @@ from torch import Tensor
 from torch.autograd.function import once_differentiable
 from tqdm import tqdm
 import torch.nn as nn
-from typing import Callable
+from typing import Callable, Union
 from deepinv.utils.tensorlist import TensorList
 from deepinv.utils.compat import zip_strict
-
+import warnings
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
@@ -89,7 +89,6 @@ def least_squares(
     :param kwargs: Keyword arguments to be passed to the solver.
     :return: (class:`torch.Tensor`) :math:`x` of shape (B, ...).
     """
-
     if isinstance(parallel_dim, int):
         parallel_dim = [parallel_dim]
 
@@ -115,7 +114,7 @@ def least_squares(
     else:
         Aty = AT(y)
         complete = Aty.shape == y.shape
-        overcomplete = Aty.flatten().shape[0] < y.flatten().shape[0]
+        overcomplete = Aty.numel() < y.numel()
 
         if complete and (solver == "BiCGStab" or solver == "minres"):
             H = lambda x: A(x)
@@ -173,7 +172,7 @@ def least_squares(
                 f"Solver {solver} not recognized. Choose between 'CG', 'lsqr' and 'BiCGStab'."
             )
 
-        if gamma is None and not overcomplete and not complete:
+        if (gamma is None or gamma == 0) and not overcomplete and not complete:
             x = AT(x)
     return x
 
@@ -219,7 +218,6 @@ def conjugate_gradient(
     :return: torch.Tensor :math:`x` of shape (B, ...) verifying :math:`Ax=b`.
 
     """
-
     if isinstance(parallel_dim, int):
         parallel_dim = [parallel_dim]
     if parallel_dim is None:
@@ -809,6 +807,7 @@ class LeastSquaresSolver(torch.autograd.Function):
     ):
 
         kwargs = extra_kwargs if extra_kwargs is not None else {}
+
         with torch.no_grad():
             solution = least_squares(
                 A=physics.A,
@@ -827,6 +826,7 @@ class LeastSquaresSolver(torch.autograd.Function):
         # Save other non-tensor contexts
         ctx.physics = physics
         ctx.kwargs = kwargs
+
         return solution
 
     @staticmethod
@@ -897,7 +897,7 @@ def least_squares_implicit_backward(
     y: Tensor,
     z: Tensor = None,
     init: Tensor = None,
-    gamma: float = None,
+    gamma: Union[float, Tensor] = None,
     **kwargs,
 ) -> Tensor:
     r"""
@@ -918,6 +918,10 @@ def least_squares_implicit_backward(
 
         Implicit gradients can be incorrect if the least squares solver does not converge sufficiently. Make sure to set the `max_iter` and `tol` parameters of the least squares solver appropriately to ensure convergence. You can monitor the convergence by setting `verbose=True` in the least squares solver via `kwargs`. If the solver does not converge, the implicit gradients can be very inaccurate and lead to divergence of the training.
 
+    .. warning::
+
+        This function does not support :class:`deepinv.utils.TensorList` inputs yet. If you use :class:`deepinv.utils.TensorList` as inputs, the function will fall back to standard least squares with full backpropagation.
+
     .. tip::
 
         If you do not need gradients with respect to the physics parameters, you can set `requires_grad=False` for all parameters of the physics operator to avoid the additional backward pass. This can save some computation time.
@@ -930,12 +934,28 @@ def least_squares_implicit_backward(
     :param deepinv.physics.LinearPhysics: physics operator :class:`deepinv.physics.LinearPhysics`.
     :param torch.Tensor y: input tensor of shape (B, ...)
     :param torch.Tensor z: input tensor of shape (B, ...). Default is `None`, which corresponds to a zero tensor.
-    :param Optional[torch.Tensor] init: Optional initial guess. Default is `None`, which corresponds to a zero initialization.
+    :param Optional[torch.Tensor] init: Optional initial guess, only used for the forward pass. Default is `None`, which corresponds to a zero initialization.
     :param Optional[float, torch.Tensor] gamma: regularization parameter :math:`\gamma > 0`. Default is `None`.
     :param kwargs: additional arguments to be passed to the least squares solver.
 
     :return: (:class:`torch.Tensor`) :math:`x` of shape (B, ...), the solution of the least squares problem.
     """
+    # NOTE: TensorList not supported by autograd function, we fall back to standard least_squares in this case for now
+    if isinstance(y, TensorList):
+        warnings.warn(
+            "Warning: least_squares_implicit_backward does not support TensorList inputs. Falling back to standard least_squares with full backpropagation."
+        )
+        return least_squares(
+            A=physics.A,
+            AT=physics.A_adjoint,
+            y=y,
+            z=z,
+            init=init,
+            gamma=gamma,
+            AAT=physics.A_A_adjoint,
+            ATA=physics.A_adjoint_A,
+            **kwargs,
+        )
 
     if z is None:
         # To get correct shape
@@ -946,6 +966,8 @@ def least_squares_implicit_backward(
     physics_requires_grad_params = any(
         p.requires_grad for p in getattr(physics, "buffers", lambda: [])()
     )
+    # NOTE: backward is triggered if any of the inputs require grad
+    # When input tensors (y, z, gamma) do not require grad and we want to do backward w.r.t physics parameters, we need to trigger it manually by a dummy tensor.
     trigger_backward = (
         y.requires_grad
         or z.requires_grad
@@ -959,9 +981,9 @@ def least_squares_implicit_backward(
     else:
         trigger = torch.ones(1, device=y.device, dtype=y.dtype)
     extra_kwargs = kwargs if kwargs else None
-    return LeastSquaresSolver.apply(
-        physics, y, z, init, gamma, trigger, extra_kwargs
-    )
+    if gamma is None:
+        gamma = torch.tensor(0, device=y.device, dtype=y.dtype)
+    return LeastSquaresSolver.apply(physics, y, z, init, gamma, trigger, extra_kwargs)
 
 
 class GaussianMixtureModel(nn.Module):
--- a/deepinv/optim/utils.py
+++ b/deepinv/optim/utils.py
@@ -1,10 +1,17 @@
 from deepinv.utils import zeros_like
 import torch
+from torch import Tensor
+from torch.autograd.function import once_differentiable
 from tqdm import tqdm
 import torch.nn as nn
 from typing import Callable
 from deepinv.utils.tensorlist import TensorList
 from deepinv.utils.compat import zip_strict
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from deepinv.physics import LinearPhysics
 
 
 def check_conv(X_prev, X, it, crit_conv="residual", thres_conv=1e-3, verbose=False):
@@ -71,6 +78,7 @@ def least_squares(
     :param Callable AT: Adjoint operator :math:`A^{\top}` as a callable function.
     :param torch.Tensor y: input tensor of shape (B, ...)
     :param torch.Tensor z: input tensor of shape (B, ...) or scalar.
+    :param torch.Tensor init: (Optional) initial guess for the solver. If None, it is set to a tensor of zeros.
     :param None, float gamma: (Optional) inverse regularization parameter.
     :param str solver: solver to be used.
     :param Callable AAT: (Optional) Efficient implementation of :math:`A(A^{\top}(x))`. If not provided, it is computed as :math:`A(A^{\top}(x))`.
@@ -87,7 +95,7 @@ def least_squares(
 
     if solver == "lsqr":  # rectangular solver
 
-        if gamma is not None:
+        if gamma is not None and gamma > 0:  # prevent division by zero
             eta = 1 / gamma
         else:
             eta = 0
@@ -118,7 +126,7 @@ def least_squares(
             if ATA is None:
                 ATA = lambda x: AT(A(x))
 
-            if gamma is not None:
+            if gamma is not None and gamma > 0:
                 b = AT(y) + 1 / gamma * z
                 H = lambda x: ATA(x) + 1 / gamma * x
                 overcomplete = False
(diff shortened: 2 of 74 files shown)


CHANGELOG.rst
```diff
@@ -25,6 +25,7 @@ New Features
 - Dataset base class + ImageFolder and TensorDataset classes (:gh:`622` by `Andrew Wang`_)
 - Added GitHub action checking import time (:gh:`680` by `Julian Tachella`_)
 - Client model for server-side inference for using models in the cloud (:gh:`691` by `Andrew Wang`_)
+- O(1) memory backprop for linear solver (:gh:`TODO` by `Minh Hai Nguyen`_)
```
jscanvic: Put the PR number

deepinv/optim/utils.py
```diff
@@ -71,6 +78,7 @@ def least_squares(
     :param Callable AT: Adjoint operator :math:`A^{\top}` as a callable function.
     :param torch.Tensor y: input tensor of shape (B, ...)
     :param torch.Tensor z: input tensor of shape (B, ...) or scalar.
+    :param torch.Tensor init: (Optional) initial guess for the solver. If None, it is set to a tensor of zeros.
```
jscanvic, line 80: Good catch!

deepinv/optim/utils.py
```diff
@@ -81,13 +89,12 @@ def least_squares(
...
     :return: (class:`torch.Tensor`) :math:`x` of shape (B, ...).
     """
-
     if isinstance(parallel_dim, int):
         parallel_dim = [parallel_dim]
 
     if solver == "lsqr":  # rectangular solver
 
-        if gamma is not None:
+        if gamma is not None and gamma > 0:  # prevent division by zero
```
jscanvic: Add this at the top

```
if gamma is None:
    gamma = 0.0
```

and then get rid of the `gamma is not None` and `gamma is None` everywhere else

deepinv/optim/utils.py
```diff
@@ -118,7 +125,7 @@ def least_squares(
             if ATA is None:
                 ATA = lambda x: AT(A(x))
 
-            if gamma is not None:
+            if gamma is not None and gamma > 0:
```
jscanvic: Here too

deepinv/optim/utils.py
```diff
@@ -165,7 +172,7 @@ def least_squares(
                 f"Solver {solver} not recognized. Choose between 'CG', 'lsqr' and 'BiCGStab'."
             )
 
-        if gamma is None and not overcomplete and not complete:
+        if (gamma is None or gamma == 0) and not overcomplete and not complete:
```
jscanvic: Here too

deepinv/optim/utils.py
```diff
@@ -107,7 +114,7 @@ def least_squares(
     else:
         Aty = AT(y)
         complete = Aty.shape == y.shape
-        overcomplete = Aty.flatten().shape[0] < y.flatten().shape[0]
+        overcomplete = Aty.numel() < y.numel()
```
jscanvic, line 113: Nice!

deepinv/physics/forward.py
```diff
@@ -10,8 +10,9 @@
 import torch.nn as nn
 from deepinv.physics.noise import NoiseModel, GaussianNoise, ZeroNoise
 from deepinv.utils.tensorlist import randn_like, TensorList
-from deepinv.optim.utils import least_squares, lsqr
+from deepinv.optim.utils import least_squares, lsqr, least_squares_implicit_backward
```
jscanvic, line 13: Do we expect end-users to use `least_squares_implicit_backward`? If not, what do you think of prefixing it as `_least_squares_implicit_backward`?

deepinv/physics/forward.py
```diff
@@ -417,6 +419,7 @@ def __init__(
         max_iter=50,
         tol=1e-4,
         solver="lsqr",
+        implicit_backward_solver: bool = True,
```
jscanvic, line 422: Is it safe in terms of backwards compatibility? I mean to have it set by default

deepinv/tests/test_optim.py
```diff
@@ -985,3 +987,106 @@ def A_adjoint(self, y, **kwargs):
     gt_cond = c.max() / c.min()
     rel_error = (cond - gt_cond).abs() / gt_cond
     assert rel_error < 0.1
+
+
+from deepinv.optim.utils import least_squares_implicit_backward
```
jscanvic: Move imports to the top

deepinv/utils/tensorlist.py
```diff
@@ -288,6 +288,12 @@ def isnan(self):
         """
         return TensorList([torch.isnan(xi) for xi in self.x])
 
+    def numel(self):
+        """
+        Returns the total number of elements in the TensorList.
+        """
+        return sum([xi.numel() for xi in self.x])
```
jscanvic, line 295: Nice

deepinv/physics/forward.py
```diff
@@ -430,6 +433,12 @@ def __init__(
         )
         self.A_adj = A_adjoint
         self.img_size = img_size
+        self.implicit_backward_solver = implicit_backward_solver
+
+        if self.implicit_backward_solver and self.max_iter < 20:
+            warnings.warn(
+                "Using implicit_backward_solver with a low number of iterations may produce inaccurate gradients during the backward pass. Consider increasing max_iter for better training accuracy."
```
jscanvic: Does the new implementation take more steps to converge than the previous one?

deepinv/optim/utils.py
```diff
@@ -769,6 +775,221 @@ def gradient_descent(grad_f, x, step_size=1.0, max_iter=1e2, tol=1e-5):
...
+        z: Tensor,
+        init: Tensor,
+        gamma: float,
+        trigger: Tensor = None,
+        extra_kwargs: dict = None,
+    ):
+
+        kwargs = extra_kwargs if extra_kwargs is not None else {}
+
+        with torch.no_grad():
```
jscanvic, line 825: Is it needeed? I feel like the torch engine should run autograd Function code in no_grad by default

deepinv/optim/utils.py
```diff
@@ -769,6 +775,221 @@ def gradient_descent(grad_f, x, step_size=1.0, max_iter=1e2, tol=1e-5):
...
+                y=y,
+                z=z,
+                init=init,
+                gamma=gamma,
+                AAT=physics.A_A_adjoint,
+                ATA=physics.A_adjoint_A,
+                **kwargs,
+            )
+        # Save tensors only
+        gamma = torch.as_tensor(gamma, dtype=y.dtype, device=y.device)
```
jscanvic, line 839: If possible: don't convert `gamma` to a tensor and save it in `ctx.gamma = gamma` instead of in the `ctx.save_for_backward` call

deepinv/optim/utils.py
```diff
@@ -769,6 +775,221 @@ def gradient_descent(grad_f, x, step_size=1.0, max_iter=1e2, tol=1e-5):
...
+        return solution
+
+    @staticmethod
+    @once_differentiable
+    def backward(ctx, grad_output: Tensor) -> tuple[Tensor, ...]:
+        h, y, z, gamma = ctx.saved_tensors
+        physics = ctx.physics
+
+        # Solve (A^T A + I/gamma) mv = grad_output
+        with torch.no_grad():
```
jscanvic, line 854: Is no_grad needed here as well?

deepinv/optim/utils.py
```diff
@@ -769,6 +775,221 @@ def gradient_descent(grad_f, x, step_size=1.0, max_iter=1e2, tol=1e-5):
...
+                AAT=physics.A_A_adjoint,
+                ATA=physics.A_adjoint_A,
+                **ctx.kwargs,
+            )
+
+        # Build grads aligned with inputs: (physics, y, z, gamma, solver, max_iter, tol, ..., **kwargs)
+        needs = ctx.needs_input_grad
+        grads = [None] * len(needs)
+
+        # dL/dy = A mv
```
jscanvic, line 870: I'm not sure we should need to do much more than running `least_squares` here. It looks like we're computing backwards of related nodes manually but I'm pretty sure we could rely on existing backwards implementations somehow, with a loop or something. Thoughts?

jscanvic (changes requested):
Thanks Hai for this really nice PR, it looks really promising! Some comments

deepinv/physics/forward.py
tachella: it is not super safe, but I think it is a cool improvement that could pass unnoticed if we leave as False
I think the warning is enough

deepinv/physics/forward.py
```diff
@@ -355,6 +356,7 @@ class LinearPhysics(Physics):
     :param float tol: If the operator does not have a closed form pseudoinverse, a least squares algorithm
         is used for computing it, and this parameter fixes the relative tolerance of the least squares algorithm.
     :param str solver: least squares solver to use. Choose between `'CG'`, `'lsqr'`, `'BiCGStab'` and `'minres'`. See :func:`deepinv.optim.utils.least_squares` for more details.
+    :param bool implicit_backward_solver: If `True`, uses implicit differentiation for computing gradients through the least squares solver using :func:`deepinv.optim.utils.least_squares_implicit_backward`. This can reduce memory consumption significantly, especially when using many iterations. If `False`, uses the standard autograd mechanism, which can be memory intensive. Default is `True`.
```
tachella: This should go in `LinearPhysics`, not in `Physics`, which doesn't use linear least square solvers.

I would also modify to

If `True`, uses implicit differentiation for computing gradients through the :func:`deepinv.LinearPhysics.A_dagger` (pseudoinverse using a least-squares solver), applying :func:`deepinv.optim.utils.least_squares_implicit_backward` instead of :func:`deepinv.optim.utils.least_squares`. This can significantly reduce memory consumption, especially when using many iterations. If `False`, uses the standard autograd mechanism, which can be memory-intensive. Default is `True`.

docs/source/user_guide/reconstruction/unfolded.rst
```diff
@@ -56,6 +56,34 @@ evaluated with any forward model (e.g., denoising, deconvolution, inpainting, et
     >>> y = physics(x)
     >>> x_hat = model(y, physics)
 
+Memory-efficient back-propagation
+-------------------------------------------
+The unfolded architectures can be trained using standard backpropagation or using a custom implementation of the backward pass.
```
tachella:
```suggestion
Some unfolded architectures rely on a least-squares solver to compute the proximal step w.r.t. the data-fidelity term (e.g., :class:`ADMM <deepinv.optim.optim_iterators.ADMMIteration>` or :class:`HQS <deepinv.optim.optim_iterators.HQSIteration>`). During backpropagation, a naive implementation requires storing the gradients of every intermediate step of the least squares solver, resulting in significant memory and computational costs.
```

docs/source/user_guide/reconstruction/unfolded.rst
```diff
@@ -56,6 +56,34 @@ evaluated with any forward model (e.g., denoising, deconvolution, inpainting, et
     >>> y = physics(x)
     >>> x_hat = model(y, physics)
 
+Memory-efficient back-propagation
+-------------------------------------------
+The unfolded architectures can be trained using standard backpropagation or using a custom implementation of the backward pass. 
+We provide a memory-efficient back-propagation strategy that reduces the memory footprint during training.
```
tachella:
```suggestion
We provide a memory-efficient back-propagation strategy that reduces the memory footprint during training, by computing the gradients of the least squares solver in closed-form, without storing any intermediate steps. This closed-form computation requires evaluating the least-squares solver one additional time during the gradient computation.
```

docs/source/user_guide/reconstruction/unfolded.rst
```diff
@@ -56,6 +56,34 @@ evaluated with any forward model (e.g., denoising, deconvolution, inpainting, et
...
     >>> x_hat = model(y, physics)
 
+Memory-efficient back-propagation
+-------------------------------------------
+The unfolded architectures can be trained using standard backpropagation or using a custom implementation of the backward pass. 
+We provide a memory-efficient back-propagation strategy that reduces the memory footprint during training. 
+This is particularly useful when training deep unfolded architectures with many iterations. 
+To enable this feature, set the argument ``implicit_backward_solver=True`` (default is `True`) when creating the physics. It is supported for all linear physics
+(:class:`deepinv.physics.LinearPhysics`).  
+Note that when setting ``implicit_backward_solver=True``, we need to use a large enough number of iterations in the physics solver to ensure convergence, otherwise the gradient might be inaccurate.
```
tachella:
```suggestion
Note that when setting ``implicit_backward_solver=True``, we need to use a large enough number of iterations in the physics solver to ensure convergence (as the closed-form gradients assume that we converged to the least squares minimizer), otherwise the gradient might be inaccurate.
```

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,251 @@
+r"""
+Training unfolded neural networks without out-of-memory issues
```
tachella:
```suggestion
Reducing the memory and computational complexity of unfolded network training
```

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,251 @@
...
+====================================================================================================
+
```
tachella, line 4: I think here we need to provide a more detailed explanation.

what about something similar to the docs above:

```

Some unfolded architectures rely on a `least-squares solver <deepinv.optim.utils.least_squares>` to compute the proximal step w.r.t. the data-fidelity term (e.g., :class:`ADMM <deepinv.optim.optim_iterators.ADMMIteration>` or :class:`HQS <deepinv.optim.optim_iterators.HQSIteration>`):

.. math::

     \operatorname{prox}_{\gamma f}(x_k)  = \underset{x}{\arg\min} \; \frac{\gamma}{2}\|Ax-y\|^2 + \frac{1}{2}\|x-z\|^2

During backpropagation, a naive implementation requires storing the gradients of every intermediate step of the least squares solver, resulting in significant memory and computational costs which are proportional to number of iterations done by the solver.

The library provides a memory-efficient back-propagation strategy that reduces the memory footprint during training, by computing the gradients of the proximal step in closed-form, without storing any intermediate steps. This closed-form computation requires evaluating the least-squares solver one additional time during the gradient computation.

.. note::

    Linear forward operators that have a :class:`closed-form singular value decomposition <deepinv.physics.DecomposablePhysics>` benefit from a :func:`closed-form formula <deepinv.physics.DecomposablePhysics.prox_l2>` for computing the proximal step, and thus we shouldn't expect speed-ups in these specific cases.


```

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,251 @@
...
+This example shows how to train an unfolded neural network with a memory complexity that is independent of the number of iterations in least squares solver (O(1) memory complexity).
```
tachella:
```suggestion
This example shows how to train an unfolded neural network with a memory complexity that is independent of the number of iterations in least squares solver (O(1) memory complexity) used for computing the data-fidelity proximal step.
```

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,251 @@
...
+# Define the base train and test datasets of clean images.
+train_base_dataset = load_dataset(train_dataset_name, transform=train_transform)
+test_base_dataset = load_dataset(test_dataset_name, transform=test_transform)
+
+# Use parallel dataloader if using a GPU to speed up training, otherwise, as all computes are on CPU, use synchronous
+# dataloading.
+num_workers = 4 if torch.cuda.is_available() else 0
+
+# Degradation parameters
+factor = 2
```
tachella: this two parameters are not used

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,251 @@
...
+    implicit_backward_solver=False,
+)
+my_dataset_name = "demo_unfolded_memory"
+n_images_max = (
+    1000 if torch.cuda.is_available() else 10
+)  # maximal number of images used for training
+measurement_dir = DATA_DIR / train_dataset_name / "deblurring"
+generated_datasets_path = dinv.datasets.generate_dataset(
+    train_dataset=train_base_dataset,
+    test_dataset=test_base_dataset,
```
tachella: the test set is not used right?

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,251 @@
...
+img_size = 64 if torch.cuda.is_available() else 32
+n_channels = 3  # 3 for color images, 1 for gray-scale images
+
+# %%
+# Generate a dataset of low resolution images and load it.
+# ----------------------------------------------------------------------------------------
+# We use the Blur class from the physics module to generate a dataset of blurry images.
+# For simplicity, we use a small dataset for training.
+train_dataset_name = "CBSD500"
+test_dataset_name = "set3c"
```
tachella: should we fully remove the test set?

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,251 @@
...
+# ----------------------------------------------------------------------------------------
+
+# Unrolled optimization algorithm parameters
+max_iter = 5  # number of unfolded layers
+
+# Select the data fidelity term
+data_fidelity = L2()
+stepsize = [1] * max_iter  # stepsize of the algorithm
+sigma_denoiser = [0.01] * max_iter  # noise level parameter of the denoiser
+beta = 1  # relaxation parameter of the Douglas-Rachford splitting
```
tachella: beta is not used in HQS?

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,251 @@
...
+    avg_loss += loss.item()
+    loss.backward()
+    optimizer.step()
+sync()
+end = time.perf_counter()
+auto_peak_memory_mb = peak_memory() / (10**6)
+auto_time_per_iter = (end - start) / len(train_dataloader)
+auto_avg_loss = avg_loss / len(train_dataloader)
+# %%
+# Setting this parameter to True to use implicit differentiation
```
tachella:
```suggestion
# We now train the model using the closed-form gradients of the proximal step.
We can do this by setting `implicit_backward_solver` to `True`.
```

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,251 @@
...
+    max_iter=50,
+    tol=1e-8,
+    implicit_backward_solver=False,
+)
+my_dataset_name = "demo_unfolded_memory"
+n_images_max = (
+    1000 if torch.cuda.is_available() else 10
+)  # maximal number of images used for training
+measurement_dir = DATA_DIR / train_dataset_name / "deblurring"
+generated_datasets_path = dinv.datasets.generate_dataset(
```
tachella: you could remove this step, and simply generate y during the training loop?

docs/source/user_guide/reconstruction/unfolded.rst
```diff
@@ -56,6 +56,34 @@ evaluated with any forward model (e.g., denoising, deconvolution, inpainting, et
...
+
```
tachella, line 68:
```suggestion
See also :ref:`sphx_glr_auto_examples_unfolded_demo_unfolded_constant_memory.py` for more details.
```

tachella (changes requested):
I think this is adding really nice functionalities! I added some suggestions for the docs, I think we can give much more details.

I also realised that we were testing the wrong operators here

https://github.com/deepinv/deepinv/blob/eeae34cc57a29c587dbce54fab9c62717c71e63b/deepinv/tests/test_optim.py#L883C40-L883C50

since most of them have SVD decompositions, and the least squares solver is not called!

deepinv/optim/utils.py
mh-nguyen712: thanks! good point!

deepinv/optim/utils.py
mh-nguyen712: You're right! Thanks!

deepinv/optim/utils.py
mh-nguyen712: It is converted to a tensor for differentiating indeed, ie computing the gradient w.r.t gamma

deepinv/optim/utils.py
mh-nguyen712: Good catch! No with @once_differentiable. But in the future we should check if it's can be twice differentiable or not

deepinv/optim/utils.py
mh-nguyen712: I donn't understand what you mean. This implements the VJPs for each tensor input, including y, z, gamma and potentially physics parameters. So indeed we need to implement case-by-case manually. Am I missing something?

deepinv/physics/forward.py
mh-nguyen712: Currently we only used in linear physics and prox. But it might be a useful function. To be discussed.

deepinv/physics/forward.py
mh-nguyen712: Thanks for the clear doc. What do you mean by "should go in LinearPhysics"? I'm adding it to LinearPhysics not Physics. Or you meant something else?

deepinv/physics/forward.py
mh-nguyen712: Yes I do agree with Julian

deepinv/physics/forward.py
mh-nguyen712: No It uses the same solver under the hood. But It has more impact in the backward, especially if the solver does not converge

examples/unfolded/demo_unfolded_constant_memory.py
mh-nguyen712: Done

deepinv/physics/forward.py
jscanvic: Sorry I meant the backward function, I think it's only used in LinearPhysics but not in proximal operators right?

jscanvic: I'm happy to leave it that way in any case

deepinv/optim/utils.py
jscanvic: Makes sense, thanks!

deepinv/optim/utils.py
jscanvic: I'm probably missing something sorry! I'll have a closer look at it later

deepinv/optim/utils.py
```diff
@@ -769,6 +771,221 @@ def gradient_descent(grad_f, x, step_size=1.0, max_iter=1e2, tol=1e-5):
...
+
+    .. tip::
+
+        If you do not need gradients with respect to the physics parameters, you can set `requires_grad=False` for all parameters of the physics operator to avoid the additional backward pass. This can save some computation time.
+
+    .. tip::
+
+        Training unfolded network with implicit differentiation can reduce memory consumption significantly, especially when using many iterations.
+        TODO: link to example notebook when available.
+
```
tachella, line 973: I would add the math formulas you use to compute the closed-form derivatives (the ones that you showed in the slides during the hackathon, with minor adaptations)

mh-nguyen712: I've added it into the example. Do you think it's better to reference to the example or add it directly here?

deepinv/physics/forward.py
mh-nguyen712: It is used in LinearPhysics.A_dagger and LinearPhysics.prox_l2, which will be internally called by data_fid.prox I think. Yeah anw I think we can leave this function

jscanvic: My bad I misread the diff. Sounds good to me then - thanks!

mh-nguyen712:
Thanks for your review. I just updated the examples and the test (I changed a bit a the test and lower tolerance). I also fix the first issue in #769.

jscanvic (approved):

deepinv/optim/utils.py
```diff
@@ -769,6 +789,221 @@ def gradient_descent(grad_f, x, step_size=1.0, max_iter=1e2, tol=1e-5):
...
+    physics: "LinearPhysics",
+    y: Tensor,
+    z: Tensor = None,
+    init: Tensor = None,
+    gamma: Union[float, Tensor] = None,
+    **kwargs,
+) -> Tensor:
+    r"""
+    Least squares solver with O(1) memory backward propagation using implicit differentiation.
+    The functions is similar to :func:`deepinv.optim.utils.least_squares` for the forward pass, but uses implicit differentiation for the backward pass, which reduces memory consumption to O(1) in the number of iterations. See :func:`deepinv.optim.utils.least_squares` for more details about the problem solved in the forward pass.
```
tachella:
```suggestion
    The function is similar to :func:`deepinv.optim.utils.least_squares` for the forward pass, but uses implicit differentiation for the backward pass, which reduces memory consumption to O(1) in the number of iterations. 
```

deepinv/optim/utils.py
```diff
@@ -769,6 +789,221 @@ def gradient_descent(grad_f, x, step_size=1.0, max_iter=1e2, tol=1e-5):
...
+        This function does not support :class:`deepinv.utils.TensorList` inputs yet. If you use :class:`deepinv.utils.TensorList` as inputs, the function will fall back to standard least squares with full backpropagation.
+
+    .. tip::
+
+        If you do not need gradients with respect to the physics parameters, you can set `requires_grad=False` for all parameters of the physics operator to avoid the additional backward pass. This can save some computation time.
+
+    .. tip::
+
+        Training unfolded network with implicit differentiation can reduce memory consumption significantly, especially when using many iterations.
+        TODO: link to example notebook when available.
```
tachella: I would link to the demo before the notes and the tips, in the main text of the function

deepinv/optim/utils.py
tachella: I think it is never too bad to repeat explanations, because people often don't click on the link

deepinv/physics/forward.py
```diff
@@ -355,6 +356,7 @@ class LinearPhysics(Physics):
     :param float tol: If the operator does not have a closed form pseudoinverse, a least squares algorithm
         is used for computing it, and this parameter fixes the relative tolerance of the least squares algorithm.
     :param str solver: least squares solver to use. Choose between `'CG'`, `'lsqr'`, `'BiCGStab'` and `'minres'`. See :func:`deepinv.optim.utils.least_squares` for more details.
+    :param bool implicit_backward_solver: If `True`, uses implicit differentiation for computing gradients through the :meth:`deepinv.physics.LinearPhysics.A_dagger` and :meth:`deepinv.physics.LinearPhysics.prox_l2`, using :func:`deepinv.optim.utils.least_squares_implicit_backward` instead of :func:`deepinv.optim.utils.least_squares`. This can significantly reduce memory consumption, especially when using many iterations. If False, uses the standard autograd mechanism, which can be memory-intensive. Default is True.
```
tachella:
```suggestion
    :param bool implicit_backward_solver: If `True`, uses implicit differentiation for computing gradients through the :meth:`deepinv.physics.LinearPhysics.A_dagger` and :meth:`deepinv.physics.LinearPhysics.prox_l2`, using :func:`deepinv.optim.utils.least_squares_implicit_backward` instead of :func:`deepinv.optim.utils.least_squares`. This can significantly reduce memory consumption, especially when using many iterations. If `False`, uses the standard autograd mechanism, which can be memory-intensive. Default is `True`.
```

deepinv/tests/test_optim.py
```diff
@@ -985,3 +987,116 @@ def A_adjoint(self, y, **kwargs):
...
+    )
+
+    # ------------------------------------------------------------
+    # Check gradients physics parameters
+
+    # Enable gradients w.r.t physics parameters
+    buffers = copy.deepcopy(dict(physics.named_buffers()))
+    parameters = {k: v for k, v in buffers.items() if k in params}
+    for k, v in parameters.items():
+        if isinstance(v, torch.Tensor) and v.dtype in [
```
tachella, line 1040: maybe add a comment explaining that we do this to only backprop wrt to continuous variables

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,277 @@
...
+import time
+import numpy as np
+import matplotlib.pyplot as plt
+
+# %%
+# Setup paths for data loading and results.
+# ----------------------------------------------------------------------------------------
+#
+
+BASE_DIR = Path(".")
```
tachella: these 2 lines are not used I believe

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,277 @@
...
+from deepinv.optim.prior import PnP
+from deepinv.unfolded import unfolded_builder
+from torchvision import transforms
+from deepinv.utils.demo import load_dataset
+import time
+import numpy as np
+import matplotlib.pyplot as plt
+
+# %%
+# Setup paths for data loading and results.
```
tachella: I guess we can remove this subsection if we remove the 2 lines below

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,277 @@
...
+    dinv.utils.get_freer_gpu() if torch.cuda.is_available() else torch.device("cpu")
+)
+device = torch.device("cpu")
+# %%
+# Load base image datasets and degradation operators.
+# ----------------------------------------------------------------------------------------
+# In this example, we use the CBSD500 dataset for training and the Set3C dataset for testing.
+
+img_size = 64 if torch.cuda.is_available() else 32
+n_channels = 3  # 3 for color images, 1 for gray-scale images
```
tachella: this is not used

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,277 @@
...
+
+# %%
+# Generate a dataset of low resolution images and load it.
+# ----------------------------------------------------------------------------------------
+# We use the Blur class from the physics module to generate a dataset of blurry images.
```
tachella: maybe we can explain that we choose blur with 'valid' padding since it its proximal operator does not have a closed form solution and requires using a least squares solver

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,277 @@
+r"""
+Reducing the memory and computational complexity of unfolded network training
+====================================================================================================
+
+Some unfolded architectures rely on a `least-squares solver <deepinv.optim.utils.least_squares>` to compute the proximal step w.r.t. the data-fidelity term (e.g., :class:`ADMM <deepinv.optim.optim_iterators.ADMMIteration>` or :class:`HQS <deepinv.optim.optim_iterators.HQSIteration>`):
```
tachella:
```suggestion
Some unfolded architectures rely on a :func:`least-squares solver <deepinv.optim.utils.least_squares>` to compute the proximal step w.r.t. the data-fidelity term (e.g., :class:`ADMM <deepinv.optim.optim_iterators.ADMMIteration>` or :class:`HQS <deepinv.optim.optim_iterators.HQSIteration>`):  
```

tachella: please double check that this is showing well

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,277 @@
...
+
+.. math::  
+
+     \operatorname{prox}_{\gamma f}(z)  = \underset{x}{\arg\min} \; \frac{\gamma}{2}\|A_\theta x-y\|^2 + \frac{1}{2}\|x-z\|^2  
+
+During backpropagation, a naive implementation requires storing the gradients of every intermediate step of the least squares solver (which is an iterative method), resulting in significant memory and computational costs which are proportional to number of iterations done by the solver.  
+
+The library provides a memory-efficient back-propagation strategy that reduces the memory footprint during training, by computing the gradients of the proximal step in closed-form, without storing any intermediate steps. This closed-form computation requires evaluating the least-squares solver one additional time during the gradient computation.  
+
+Let :math:`h(z, y, \theta, \gamma) = \operatorname{prox}_{\gamma f}(z)` be the proximal operator. During the backward pass, we need to compute the vector-Jacobian products (VJPs), in the input variables :math:`(z, y, \theta, \gamma)` is required for backpropagation:
```
tachella:
```suggestion
Let :math:`h(z, y, \theta, \gamma) = \operatorname{prox}_{\gamma f}(z)` be the proximal operator. During the backward pass, we need to compute the vector-Jacobian products (VJPs), wrt the input variables :math:`(z, y, \theta, \gamma)` required for backpropagation:
```

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,277 @@
...
+
+.. math::
+
+    \left( \frac{\partial h}{\partial z} \right)^T v, \quad \left( \frac{\partial h}{\partial y} \right)^T v, \quad \left( \frac{\partial h}{\partial \theta} \right)^T v, \quad \left( \frac{\partial h}{\partial \gamma} \right)^T v
```
tachella:
```suggestion
    \left( \frac{\partial h}{\partial z} \right)^{\top} v, \quad \left( \frac{\partial h}{\partial y} \right)^{\top} v, \quad \left( \frac{\partial h}{\partial \theta} \right)^{\top} v, \quad \left( \frac{\partial h}{\partial \gamma} \right)^{\top} v
```

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,277 @@
...
+
+and :math:`v` is the upstream gradient. The VJPs can be computed in closed-form by solving a second least-squares problem, as shown in the following. 
+When the forward least-squares solver converges to the exact minimizer, we have the following closed-form expressions for :math:`h(z, y, \theta, \gamma)`:
+
+.. math::
+
+    h(z, y, \theta, \gamma) = \left( A_\theta^T A_\theta + \frac{1}{\gamma} I \right)^{-1} \left( A_\theta^T y + \frac{1}{\gamma} z \right)
```
tachella:
```suggestion
    h(z, y, \theta, \gamma) = \left( A_\theta^{\top} A_\theta + \frac{1}{\gamma} I \right)^{-1} \left( A_\theta^{\top} y + \frac{1}{\gamma} z \right)
```

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,277 @@
...
+
+Let :math:`M` denote the inverse :math:`\left( A_\theta^T A_\theta + \frac{1}{\gamma} I \right)^{-1}`. The VJPs can be computed as follows:
+
+.. math::
+
+    \left( \frac{\partial h}{\partial z} \right)^T v               &= \frac{1}{\gamma} M v \\
```
tachella:
```suggestion
    \left( \frac{\partial h}{\partial z} \right)^{\top} v               &= \frac{1}{\gamma} M v \\
```

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,277 @@
...
+    \left( \frac{\partial h}{\partial y} \right)^T v               &= A_\theta M v \\
```
tachella:
```suggestion
    \left( \frac{\partial h}{\partial y} \right)^{\top} v               &= A_\theta M v \\
```

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,277 @@
...
+    \left( \frac{\partial h}{\partial \gamma} \right)^T v          &= \langle M v, h - z \rangle / \gamma^2 \\
```
tachella:
```suggestion
    \left( \frac{\partial h}{\partial \gamma} \right)^{\top} v          &=   (h - z)^\top M  v / \gamma^2 \\
```

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,277 @@
...
+    \left( \frac{\partial h}{\partial \theta} \right)^T v          &= \frac{\partial p}{\partial \theta}
```
tachella:
```suggestion
    \left( \frac{\partial h}{\partial \theta} \right)^{\top} v          &= \frac{\partial p}{\partial \theta} 
```

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,277 @@
...
+Let :math:`M` denote the inverse :math:`\left( A_\theta^T A_\theta + \frac{1}{\gamma} I \right)^{-1}`. The VJPs can be computed as follows:
+
+.. math::
+
+    \left( \frac{\partial h}{\partial z} \right)^T v               &= \frac{1}{\gamma} M v \\
+    \left( \frac{\partial h}{\partial y} \right)^T v               &= A_\theta M v \\
+    \left( \frac{\partial h}{\partial \gamma} \right)^T v          &= \langle M v, h - z \rangle / \gamma^2 \\
+    \left( \frac{\partial h}{\partial \theta} \right)^T v          &= \frac{\partial p}{\partial \theta} 
+    
+where :math:`p = \langle M v, A_\theta^T (y - A_\theta h) \rangle` and :math:`\frac{\partial p}{\partial \theta}` can be computed using the standard backpropagation mechanism (autograd).
```
tachella:
```suggestion
where :math:`p =  (y - A_\theta h)^{\top} A_\theta M v ` and :math:`\frac{\partial p}{\partial \theta}` can be computed using the standard backpropagation mechanism (autograd).
```

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,277 @@
...
+    print(f"Full backpropagation: peak memory usage: {auto_peak_memory_mb:.1f} MB")
+    print(
+        f"Implicit differentiation: peak memory usage: {implicit_peak_memory_mb:.1f} MB"
+    )
+    print(
+        f"Memory reduction factor: {auto_peak_memory_mb/implicit_peak_memory_mb:.1f}x"
+    )
+
+
+# Compare the training loss
```
tachella, line 257: image

- I cannot see the blue curve, is it just behind the orange? in which case is it possible to increase the fontsize to see both?
- unfortunately it seems that the time difference is minimal (see print below) - it it because we are using cpu? is there a way to accentuate the difference? otherwise we should write somewhere that by default the example runs on cpu and we don't expect to observe a strong difference? Another option is to choose a more ill-posed blur kernel, ie increase the sigma of the gaussian blur, such that the least squares solver needs more iterations to converge?

tachella (changes requested):
many thanks for all the updates!

I have some additional comments, but I think we should converge soon :)

deepinv/optim/utils.py
mh-nguyen712: done

examples/unfolded/demo_unfolded_constant_memory.py
mh-nguyen712: - Yes the blur curve is just behind the orange one. I adjust the linestyle to see it better
- I increase the number of iteration to see a time difference (~1.1x-1.2x speed up, not much). The most interesting part is 2x-3x memory reduction factor, but I can only trace these info in GPU.

mh-nguyen712: I just added a note at the beginning on the example and in the function as well

deepinv/tests/test_optim.py
mh-nguyen712: done

deepinv/optim/utils.py
mh-nguyen712: Done

deepinv/optim/utils.py
```diff
@@ -769,6 +789,243 @@ def gradient_descent(grad_f, x, step_size=1.0, max_iter=1e2, tol=1e-5):
...
+    **kwargs,
+) -> Tensor:
+    r"""
+    Least squares solver with O(1) memory backward propagation using implicit differentiation.
+    The function is similar to :func:`deepinv.optim.utils.least_squares` for the forward pass, but uses implicit differentiation for the backward pass, which reduces memory consumption to O(1) in the number of iterations. 
+
+    This function supports backpropagation with respect to the inputs :math:`y`, :math:`z` and :math:`\gamma` and also with respect to the parameters of the physics operator :math:`A_\theta` if they require gradients. See :ref:`sphx_glr_auto_examples_unfolded_demo_unfolded_constant_memory.py` and the notes below for more details. 
+    
+    Let :math:`h(z, y, \theta, \gamma)` denote the output of the least squares solver, i.e. the solution of the following problem:
+    .. math::
```
tachella, line 929: add a new line before this, otherwise it is not showing right

image

deepinv/optim/utils.py
```diff
@@ -769,6 +789,243 @@ def gradient_descent(grad_f, x, step_size=1.0, max_iter=1e2, tol=1e-5):
...
+    Let :math:`M` denote the inverse :math:`\left( A_\theta^T A_\theta + \frac{1}{\gamma} I \right)^{-1}`. In the forward, we need to compute the vector-Jacobian products (VJPs), which can be computed as follows:
+
+    .. math::
+
+        \left( \frac{\partial h}{\partial z} \right)^{\top} v               &= \frac{1}{\gamma} M v \\
+        \left( \frac{\partial h}{\partial y} \right)^{\top} v               &= A_\theta M v \\
+        \left( \frac{\partial h}{\partial \gamma} \right)^{\top} v          &=   (h - z)^\top M  v / \gamma^2 \\
+        \left( \frac{\partial h}{\partial \theta} \right)^{\top} v          &= \frac{\partial p}{\partial \theta} 
+        
+    where :math:`p =  (y - A_\theta h)^{\top} A_\theta M v ` and :math:`\frac{\partial p}{\partial \theta}` can be computed using the standard backpropagation mechanism (autograd).
```
tachella:
```suggestion
    where :math:`p =  (y - A_\theta h)^{\top} A_\theta M v` and :math:`\frac{\partial p}{\partial \theta}` can be computed using the standard backpropagation mechanism (autograd).
```
This equation is not showing correctly, please double check that this change fixes the issuea

examples/unfolded/demo_unfolded_constant_memory.py
```diff
@@ -0,0 +1,273 @@
...
+Let :math:`M` denote the inverse :math:`\left( A_\theta^T A_\theta + \frac{1}{\gamma} I \right)^{-1}`. The VJPs can be computed as follows:
+
+.. math::
+
+    \left( \frac{\partial h}{\partial z} \right)^{\top} v               &= \frac{1}{\gamma} M v \\
+    \left( \frac{\partial h}{\partial y} \right)^{\top} v               &= A_\theta M v \\
+    \left( \frac{\partial h}{\partial \gamma} \right)^{\top} v          &=   (h - z)^\top M  v / \gamma^2 \\
+    \left( \frac{\partial h}{\partial \theta} \right)^{\top} v          &= \frac{\partial p}{\partial \theta} 
+    
+where :math:`p =  (y - A_\theta h)^{\top} A_\theta M v ` and :math:`\frac{\partial p}{\partial \theta}` can be computed using the standard backpropagation mechanism (autograd).
```
tachella, line 37:
```suggestion
    where :math:`p =  (y - A_\theta h)^{\top} A_\theta M v` and :math:`\frac{\partial p}{\partial \theta}` can be computed using the standard backpropagation mechanism (autograd).
```
This equation is not showing correctly, please double check that this change fixes the issue

docs/source/user_guide/reconstruction/unfolded.rst
```diff
@@ -56,6 +56,34 @@ evaluated with any forward model (e.g., denoising, deconvolution, inpainting, et
...
+We provide a memory-efficient back-propagation strategy that reduces the memory footprint during training, by computing the gradients of the least squares solver in closed-form, without storing any intermediate steps. This closed-form computation requires evaluating the least-squares solver one additional time during the gradient computation.
+This is particularly useful when training deep unfolded architectures with many iterations. 
+To enable this feature, set the argument ``implicit_backward_solver=True`` (default is `True`) when creating the physics. It is supported for all linear physics
+(:class:`deepinv.physics.LinearPhysics`).  
+Note that when setting ``implicit_backward_solver=True``, we need to use a large enough number of iterations in the physics solver to ensure convergence (as the closed-form gradients assume that we converged to the least squares minimizer), otherwise the gradient might be inaccurate.
+See also :ref:`sphx_glr_auto_examples_unfolded_demo_unfolded_constant_memory.py` for more details.
+.. doctest::
```
tachella, line 69: I think you need to add a new line before doctest, as it is not showing well

image

examples/unfolded/demo_unfolded_constant_memory.py
tachella: perfect, thank you!

deepinv/optim/utils.py
mh-nguyen712: Sorry for this.

mh-nguyen712:
Should be good now @tachella.

PR #747, opened 2025-09-10
Update integration with ASTRA Toolbox

askorikov:
This PR updates integration with ASTRA Toolbox introduced in (#474).

One thing I'm not completely certain about is a breaking change in the interface of `TomographyWithAstra` physics introduced by 47e1ba84feb7973dab9d248c8fcaa267623fb5e3. The motivation there is to make it as close as possible to the existing PyTorch-based `Tomography` physics, which should make it simpler to use the new ASTRA-based physics as a drop-in replacement. However, I'm not sure if it's worth a breaking change.

### DIFF
--- a/deepinv/physics/functional/astra.py
+++ b/deepinv/physics/functional/astra.py
@@ -359,7 +359,7 @@ class AutogradTransform(torch.autograd.Function):
 
 def create_projection_geometry(
     geometry_type: str,
-    detector_spacing: Union[int, tuple[int, int]],
+    detector_spacing: Union[float, tuple[float, float]],
     n_detector_pixels: Union[int, tuple[int, int]],
     angles: torch.Tensor,
     is_2d: bool = False,
--- a/deepinv/physics/functional/astra.py
+++ b/deepinv/physics/functional/astra.py
@@ -126,7 +126,7 @@ class XrayTransform:
                 return self.projection_geometry["DistanceOriginDetector"]
 
     @property
-    def magnitude(self) -> float:
+    def magnification_factor(self) -> float:
         """The magnification factor induced by the fan/cone geometry."""
         if "cone" in self.projection_geometry["type"]:
             return (self.detector_radius + self.source_radius) / self.source_radius
@@ -196,7 +196,7 @@ class XrayTransform:
                 parent._backprojection(x, out)
                 if self.is_2d:
                     # necessary scaling in fanbeam to obtain decent approximated adjoint
-                    out /= parent.magnitude
+                    out /= parent.magnification_factor
 
                 return out
 
--- a/deepinv/physics/functional/astra.py
+++ b/deepinv/physics/functional/astra.py
@@ -227,7 +227,7 @@ class XrayTransform:
         return _Adjoint()
 
     def _forward_projection(self, x: torch.Tensor, out: torch.Tensor) -> None:
-        import astra
+        import astra.experimental
 
         assert (
             x.shape == self.domain_shape
@@ -248,7 +248,7 @@ class XrayTransform:
         )
 
     def _backprojection(self, y: torch.Tensor, out: torch.Tensor) -> None:
-        import astra
+        import astra.experimental
 
         assert (
             y.shape == self.range_shape
--- a/deepinv/physics/functional/astra.py
+++ b/deepinv/physics/functional/astra.py
@@ -82,9 +82,9 @@ class XrayTransform:
     def detector_cell_v_length(self) -> float:
         """The vertical length of a detector cell."""
         if "vec" in self.projection_geometry["type"]:
-            return np.sqrt(
-                (self.projection_geometry["Vectors"][0, [9, 10, 11]] ** 2).sum()
-            )
+            return np.linalg.norm(
+                self.projection_geometry["Vectors"][0, [9, 10, 11]],
+            ).item()
         else:
             return self.projection_geometry["DetectorSpacingY"]
 
@@ -92,9 +92,9 @@ class XrayTransform:
     def detector_cell_u_length(self) -> float:
         """The horizontal length of a detector cell."""
         if "vec" in self.projection_geometry["type"]:
-            return np.sqrt(
-                (self.projection_geometry["Vectors"][0, [6, 7, 8]] ** 2).sum()
-            )
+            return np.linalg.norm(
+                self.projection_geometry["Vectors"][0, [6, 7, 8]]
+            ).item()
         else:
             return self.projection_geometry["DetectorSpacingX"]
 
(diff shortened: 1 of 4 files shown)


deepinv/physics/functional/astra.py
```diff
@@ -277,6 +257,7 @@ def _create_astra_link(data: torch.Tensor) -> object:
     """
     import astra
 
+    assert data.device.type == "cuda", "Data must be on a CUDA device"
```
jscanvic: I'm not entirely sure why but this appears to break the tests. Got any idea why?

romainvo: It's because the current tests do not take into account that during initialization and when ``normalize=True``, the ``compute_norm`` function is called with cpu Tensors as input (the wrapper only handles cuda), which did not raise any error before @askorikov's fix. Wrapping the ``physics`` instanciation into the mock context manager should fix the issue.

romainvo:
Thanks for improving the integration ! I like the idea of harmonizing the interface with the initial ``Tomography`` operator. From the feedback I received, a lot of ``Tomography`` users have yet to try the ``TomographyWithAstra`` wrapper, and the break would only concern how angles are passed to the constructor, so I guess it should have minimal impact ?

deepinv/physics/tomography.py
```diff
@@ -371,9 +370,9 @@ class TomographyWithAstra(LinearPhysics):
             >>> x = torch.randn(1, 1, 5, 5, device='cuda') # Define random 5x5 image
             >>> physics = TomographyWithAstra(
             ...        img_size=(5,5),
-            ...        num_angles=10,
+            ...        angles=10,
             ...        angular_range=(0, 2*torch.pi),
```
romainvo: A few changes to update the examples in the doctest strings accordingly, e.g. angles from radii to degrees here

```suggestion
            ...        angular_range=(0, 360.),
```

deepinv/tests/test_external_libraries.py
```diff
@@ -45,14 +45,14 @@ def test_tomography_with_astra_logic(self, is_2d, geometry_type, normalize):
         ## Test 2d transforms
```
romainvo, line 65: With respect to my comment on the test breaking https://github.com/deepinv/deepinv/pull/747#discussion_r2339729967, one solution could be to wrap line 45 to 79 into this mock context manager
```suggestion
        with (
            mock.patch.object(
                dinv.physics.functional.XrayTransform,
                "_forward_projection",
                new=self.dummy_projection if device != "cuda" else dinv.physics.functional.XrayTransform._forward_projection,
            ),
            mock.patch.object(
                dinv.physics.functional.XrayTransform,
                "_backprojection",
                new=self.dummy_projection if device != "cuda" else dinv.physics.functional.XrayTransform._backprojection,
            ),
            mock.patch.object(
                dinv.physics.TomographyWithAstra,
                "compute_norm",
                new=self.dummy_compute_norm if device != "cuda" else dinv.physics.TomographyWithAstra.compute_norm
            ),
        ):

            ## Test 2d transforms
            ...
```

deepinv/physics/tomography.py
askorikov: Thanks, I've found a few more of these

deepinv/tests/test_external_libraries.py
askorikov: In the end, I decided to use [monkeypatch](https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch.setattr) to do this conveniently.

romainvo: I didn't know about this feature ! this is way better than what I proposed

deepinv/physics/functional/astra.py
```diff
@@ -18,10 +18,10 @@ class XrayTransform:
 
         This transform does not handle batched and multi-channel inputs. It is
         handled by a custom :class:`torch.autograd.Function` that wraps the :class:`XrayTransform`.
-        To handle standard PyTorch pipelines, :class:`XrayTransform` is instanciated inside a :class:`deepinv.physics.TomographyWithAstra` operator.
+        To handle standard PyTorch pipelines, :class:`XrayTransform` is instantiated inside a :class:`deepinv.physics.TomographyWithAstra` operator.
 
-    :param dict[str, Any] projection_geometry: Dictionnary containing the parameters of the projection geometry. It is passed to the ``astra.create_projector()`` function to instanciate the projector.
-    :param dict[str, Any] object_geometry:  Dictionnary containing the parameters of the object geometry. It is passed to the ``astra.create_projector()`` function to instanciate the projector.
+    :param dict[str, Any] projection_geometry: Dictionary containing the parameters of the projection geometry. It is passed to the ``astra.create_projector()`` function to instantiate the projector.
```
Andrewwango: Could you provide a description of structure of the format in which projection_geometry should be passed? Or perhaps could you have a default value if the user does not pass it? I understand that this may be trivial for experienced Astra users, but to a novice, the object is an uninformative collection of tensors e.g. `self.projection_geometry["Vectors"][0, [9, 10, 11]]`

Andrewwango: Thank you for the contribution btw @askorikov ! Exciting to have a more mature Astra wrapper and should be ready to go after this comment has been resolved.

askorikov: Hi! I added a small clarification, but ultimately I think this class will be very Astra-specific and the end users are expected to use the `TomographyWithAstra` physics. There, the format and semantics of these arguments is described in much more detail. Let me know what you think.

Andrewwango:
Maybe you could add the deepinv deprecated alias to decorate these deprecated params - so if there is a user who uses the old params it catches it

Andrewwango (approved):
LGTM thanks for your contribution @askorikov !

askorikov:
@jscanvic @Andrewwango Hi! Pinging to ask for approving the workflows.

askorikov:
I'm not sure if this codecov diff is relevant. Let me know!

askorikov:
Let me know if I should do something about the cedcov, else can be merged I guess 🙂

Andrewwango:
Hi @askorikov , thanks for notifying me - it seems like codecov is picking up untested changes in the astra file - please can you explain why the codecov diff is not relevant? Thanks!

askorikov:
@Andrewwango Well the main diff is for deepinv/physics/functional/astra.py, where we went from ~60% to ~50% coverage. The things we didn't test before are what errors are thrown and some non-default args. I've added a few lines to these cases. What I meant as not relevant is going from ~50% back to ~60% coverage. On the other hand, I could expand the test suite if we aim for ~100% coverage here.

Andrewwango:
Thanks @askorikov - don't worry, we can add pragma no cover to the mundane cases that don't need testing.

askorikov:
@Andrewwango Done! Let me know if you disagree about what can be excluded from the coverage tough :)

Andrewwango:
Thanks @askorikov ! Let's go!

askorikov:
Thanks for reviewing!

PR #749, opened 2025-09-10
BSDS500 Dataset

johertrich:
Add the BSDS500 dataset.

### DIFF
--- a/deepinv/datasets/bsds500.py
+++ b/deepinv/datasets/bsds500.py
@@ -3,7 +3,6 @@ import zipfile
 import os
 import os.path
 from PIL import Image
-import numpy as np
 from deepinv.datasets.utils import calculate_md5
 from deepinv.datasets.base import ImageDataset
 
--- a/deepinv/datasets/bsds500.py
+++ b/deepinv/datasets/bsds500.py
@@ -46,7 +46,7 @@ class BSDS500(ImageDataset):
             to load the "train" and "val" splits. None for using the splits defined by the `train` parameter. Default None.
         :param Callable transform: (optional) A function/transform that takes in a PIL image
             and returns a transformed version. E.g, ``torchvision.transforms.RandomCrop``.
-        :param bool rotate: If set to ``True`` images are rotated to have all shape 481 x 321. This can be important to use a torch dataloader.
+        :param bool rotate: If set to ``True`` images are rotated to have all the same orientation. This can be important to use a torch dataloader.
             Default at False.
         """
         checksum = "7bfe17302a219367694200a61ce8256c"
@@ -110,8 +110,8 @@ class BSDS500(ImageDataset):
     def __len__(self):
         return len(self.file_list)
 
-    def __getitem__(self, IDX):
-        img = Image.open(self.file_list[IDX]).convert("RGB")
+    def __getitem__(self, idx):
+        img = Image.open(self.file_list[idx]).convert("RGB")
         if self.rotate:
             width, height = img.size
             if width > height:
--- a/deepinv/datasets/bsds500.py
+++ b/deepinv/datasets/bsds500.py
@@ -4,6 +4,7 @@ from PIL import Image
 from pathlib import Path
 from deepinv.datasets.utils import calculate_md5, download_archive, extract_zipfile
 from deepinv.datasets.base import ImageDataset
+from natsort import natsorted
 
 
 class BSDS500(ImageDataset):
--- a/deepinv/datasets/bsds500.py
+++ b/deepinv/datasets/bsds500.py
@@ -113,14 +113,10 @@ class BSDS500(ImageDataset):
 
     def __getitem__(self, IDX):
         img = Image.open(self.file_list[IDX]).convert("RGB")
-        img = np.array(img) / 255.0
+        if self.rotate:
+            width, height = img.size
+            if width > height:
+                img = img.transpose(Image.ROTATE_90)
         if self.transforms is not None:
             img = self.transforms(img)
-        if self.rotate:
-            if isinstance(img, (tuple, list)):
-                img = [
-                    i.transpose(-2, -1) if i.shape[-1] > i.shape[-2] else i for i in img
-                ]
-            else:
-                img = img.transpose(-2, -1) if img.shape[-1] > img.shape[-2] else img
         return img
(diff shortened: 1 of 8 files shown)


docs/source/api/deepinv.datasets.rst
```diff
@@ -56,6 +56,7 @@ Image Datasets
     deepinv.datasets.DIV2K
     deepinv.datasets.Urban100HR
     deepinv.datasets.Set14HR
+    deepinv.datasets.BSDS500
```
Andrewwango, line 59: Please can you add it to user guide too?

deepinv/tests/test_datasets.py
```diff
@@ -442,6 +443,48 @@ def test_load_cbsd68_dataset(download_cbsd68):
...
+
+def test_load_bsds500_dataset(download_bsds500):
+    """Check that dataset contains 400 + 100 PIL images."""
+
+    pytest.importorskip(
+        "datasets",
+        reason="This test requires datasets. It should be "
+        "installed with `pip install datasets`",
+    )
+    for totensor in [ToTensor(), None]:
```
Andrewwango: Thanks for writing these tests! Please can you add totensor as a fixture? And similarly for train

deepinv/tests/test_datasets.py
```diff
@@ -442,6 +443,48 @@ def test_load_cbsd68_dataset(download_cbsd68):
...
+    # After the test function complete, any code after the yield statement will run
+    if download:
+        shutil.rmtree(tmp_data_dir)
+
+
+def test_load_bsds500_dataset(download_bsds500):
+    """Check that dataset contains 400 + 100 PIL images."""
+
+    pytest.importorskip(
+        "datasets",
```
Andrewwango: Seems like you didn't use `datasets` in the dataset?

deepinv/datasets/bsds500.py
```diff
@@ -0,0 +1,121 @@
...
+            urllib.request.urlretrieve(
+                "https://github.com/BIDS/BSDS500/archive/refs/heads/master.zip",
+                zip_path,
+            )
+            download_sum = calculate_md5(fpath=zip_path)
+            if not download_sum == checksum:
+                return ValueError(
+                    "Verification of the dataset failed (unexpected md5 checksum of the downloaded zip-file)"
+                )
+            with zipfile.ZipFile(zip_path, "r") as zip_ref:
```
Andrewwango: is there any way you could replace the downloading and extracting using our `deepinv.datasets.utils.download_archive` and `deepinv.datasets.utils.extract_zipfile`?

deepinv/datasets/bsds500.py
```diff
@@ -0,0 +1,121 @@
...
+        )
+        image_path_test = os.path.join(
+            self.base_path, "BSDS500-master/BSDS500/data/images/test"
+        )
+        image_path_val = os.path.join(
+            self.base_path, "BSDS500-master/BSDS500/data/images/val"
+        )
+        self.file_list = []
+        if "train" in splits:
+            file_list = os.listdir(image_path_train)
```
Andrewwango: Could you inherit from `ImageFolder` and pass a glob to `x_path` instead of having to iterate the paths yourself? as is done in e.g. `deepinv.datasets.DIV2K`

If you really don't want to inherit from ImageFolder, you could use a pathlib glob here anyway

deepinv/datasets/bsds500.py
```diff
@@ -0,0 +1,121 @@
...
+            self.file_list = self.file_list + [
+                os.path.join(image_path_val, f) for f in file_list if f.endswith("jpg")
+            ]
+
+        self.file_list.sort()
+
+    def __len__(self):
+        return len(self.file_list)
+
+    def __getitem__(self, IDX):
```
Andrewwango:
```suggestion
    def __getitem__(self, idx):
```

deepinv/datasets/bsds500.py
```diff
@@ -0,0 +1,121 @@
...
+        img = Image.open(self.file_list[IDX]).convert("RGB")
```
Andrewwango: Again, if you inherited from `ImageFolder` it will handle the PIL opening for you.

deepinv/datasets/bsds500.py
johertrich: I don't think that inherting from `ImageFolder` makes sense here for two reasons:

- the data is gathered from different base folders (because the train set consists out of two different splits) which cannot naturally put into a glob pattern. This is currently not supported by the `ImageFolder` .

- there is this rotate option to enforce that all images have the same orientation (without it I cannot load several images at once with a torch dataloader without cropping them to the same shape). Integrating that into the `ImageFolder` class would be hacky and produce unnecessary code.

Don't get me wrong, I am not generally against using `ImageFolder`. I just think that due to the two above reasons, squeezing the dataset into this format would make the code longer and harder to understand (and to maintain). And it would only replace is the PIL open (which is one line of code). I adopted your suggestion with using glob.

johertrich:
Hi @Andrewwango thank you for your comments!

I added the changes, despite your suggestion to inherit from `ImageFolder` (see answer to your comment for details).

I also added the rotate option to CBSD (and fixed in the docs the shape of the images which is not always 481x321 pixels, but for some images 321x481pixels).

deepinv/datasets/bsds500.py
```diff
@@ -0,0 +1,111 @@
...
+            self.base_path, "BSDS500-master/BSDS500/data/images/test"
+        )
+        image_path_val = Path(self.base_path, "BSDS500-master/BSDS500/data/images/val")
+
+        try:
+            from natsort import natsorted
+        except ImportError:
+            natsorted = ImportError(
+                "natsort is not available. In order to use ImageFolder, please install the natsort package with `pip install natsort`."
+            )
```
mh-nguyen712: I'm not sure if we still need this. If not, can you just remove these lines please.

johertrich: I looked it up, it seems like you are right: `natsort` is now listed in the `pyproject.toml` as non-optional dependency. So, we shouldn't need the check. I removed it (also from the `ImageFolder` class, where I copied it from).

Andrewwango (approved):
Looks good to me, thanks @johertrich !

Issue #755, opened 2025-09-11
Prevent source pages from being indexed (SEO)


jscanvic:
### Environment

Latest

### Describe the bug

Bing indexes source pages in the documentation and while I haven't see Google indexing them, I fear it might still affect our SEO if Google decides they are low-quality pages. 

I suggest to prevent indexation of those pages.

Related: #505 

### Steps to reproduce

/

### Expected behavior

/

Issue #744, opened 2025-09-09
Add a robots.txt file to (hopefully) improve SEO


jscanvic:
### Environment

Latest

### Describe the bug

There is currently no robots.txt file which might hinder SEO.

@tachella and I suggest to add one.

### Steps to reproduce

/

### Expected behavior

/

Issue #710, opened 2025-09-03
Remove scipy dependency


Andrewwango:
### Environment

All

### Describe the bug

We can delegate scipy to an optional dependency, as it seems like the only place it is used is in `deepinv.utils.loadmat` for loading MATLAB arrays, which is quite a specific use case.

### Steps to reproduce

-

### Expected behavior

-

Issue #698, opened 2025-08-22
Get rid of wandb in the examples


jscanvic:
### Environment

Up to date

### Describe the bug

Two examples pop up when looking for "wandb" in the docs:

https://deepinv.github.io/deepinv/search.html?q=wandb

https://deepinv.github.io/deepinv/auto_examples/unfolded/demo_DEQ.html

https://deepinv.github.io/deepinv/auto_examples/unfolded/demo_vanilla_unfolded.html

It might not be very meaningful to mention it (and other mlops tools) there.

@tachella and I suggest to get rid of all the text mentioning wandb in the examples (except in examples that would be specifically dedicated to integration with mlops tools) and possibly also of all the code mentioning it as well (e.g. `wandb_vis = False` scattered throughout the examples).

### Steps to reproduce

/

### Expected behavior

/

PR #756, opened 2025-09-11
Fix a couple of small issues

### LINKED ISSUE #754: `AverageMeter` still using numpy
`AverageMeter` which is used in the Trainer should use torch rather than numpy. This problem surfaces when you inspect `trainer.test(...)["PSNR"]` which is a numpy float64!

jscanvic:
Fixes #754 #744 #755 #698 #710

* AverageMeter aggregates types

Cast aggregates to builtin types (list, float)

* Enable ruff UP004

It detects classes that subclass `object` (which is a no-op)

* Add a permissive robots.txt

It enables every crawler to crawl any page

* Prevent _modules pages from being indexed

Add a `<meta>` tag which prevents those pages from being indexed

* Get rid of wandb in unrelated examples

* Make scipy optional

The single scipy import is moved in the function using it

**TODO**

* [x] Make sure that the interface change in AverageMeter doesn't break the trainer
* [x] Skip tests involving scipy when it's not installed (to make CI pass!)

### DIFF
--- a/deepinv/utils/logger.py
+++ b/deepinv/utils/logger.py
@@ -89,7 +89,6 @@ class ProgressMeter:
         return "[" + fmt + "/" + fmt.format(num_epochs) + "]"
 
 
-@_deprecated_func
 def get_timestamp() -> str:
     """Get current timestamp string.
 


deepinv/utils/logger.py
```diff
@@ -29,34 +38,35 @@ def reset(self):
...
-        self.std = np.sqrt(var) if var > 0 else 0
+        self.std = float(np.sqrt(var) if var > 0 else 0)
 
     def __str__(self):
         fmtstr = "{name}={avg" + self.fmt + "}"
         return fmtstr.format(**self.__dict__)
 
 
-class ProgressMeter(object):
+class ProgressMeter:
```
Andrewwango, line 71: Should probably just remove unused classes like progress meter

deepinv/tests/test_utils.py
```diff
@@ -965,3 +965,34 @@ def test_default_tex(latex_exists, monkeypatch):
...
+def test_AverageMeter_aggregates_types(to_float):
+    rng = torch.Generator().manual_seed(0)
+    vals = torch.randn(10, generator=rng)
+
+    meter = deepinv.utils.AverageMeter("DummyValue", fmt=":f")
+    for val in vals:
+        meter.update(to_float(val.item()))
+
+    # Scalar aggregates should be instances of the builtin float type
+    scalar_attr_names = [
```
Andrewwango: Maybe check mean of vals is mean etc?

deepinv/utils/logger.py
jscanvic: I'll add a deprecation warning

deepinv/tests/test_utils.py
jscanvic: I'm surprised it wasn't already tested - I'll cover that

PR #761, opened 2025-09-12
Make HDF5Dataset follow the Trainer's convention for unsupervised training

### LINKED ISSUE #512: HDFDataset unsupervised loading inconsistent with trainer
In HDFDataset, if there is no ground truth ("unsupervised"), then the dataset effectively gives `y,y`:

https://github.com/deepinv/deepinv/blob/bde6e513eea13c8adfab43dc3bd7a877e810b752/deepinv/datasets/datagenerator.py#L101

This is inconsistent with trainer, which exepects ground-truth-free datasets to instead load `nan,y`:

https://github.com/deepinv/deepinv/blob/bde6e513eea13c8adfab43dc3bd7a877e810b752/deepinv/training/trainer.py#L47

jscanvic:
Fixes #512

This is technically a breaking change but probably unlikely to cause problems

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ New Features
 
 Changed
 ^^^^^^^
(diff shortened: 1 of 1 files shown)


jscanvic:
Thanks for the review! I updated the changelog

PR #764, opened 2025-09-12
Revamp HDF5Dataset

### LINKED ISSUE #765: HDF5Dataset doesn't return test ground truths when train ground truths are missing
### Environment

Latest

### Describe the bug

Due to a logic bug in `HDF5Dataset`, it won't always return the test ground truths when the train ground truths are missing, even when the test ground truths are present. More specifically, it happens when setting `split="test"` without setting `train=False` which is the case in certain demos and in third-party user code:

https://github.com/deepinv/deepinv/blob/1a4803087be766b564441c50f0b0887b88d808aa/examples/physics/demo_mri_tour.py#L134

https://github.com/Andrewwango/ssibench/blob/44af995f4ce3e09bfbc346c1646f8dc71dde9618/train.py#L72

**Cause**

The condition `train or split == "train"` is used to detect if the user is asking for the train split. In that case, it checks if there are training ground truths and if not it sets the attribute `unsupervised` to true. But in practice, it is possible that `train = True` (the default value) and `split = "test"` (which usually defaults to None). In that case, the right condition should be `(train and split is None) or split == "train"` to let `split` take precedence over `train`.

https://github.com/deepinv/deepinv/blob/1a4803087be766b564441c50f0b0887b88d808aa/deepinv/datasets/datagenerator.py#L82-L88

**Severity**

The attribute `x` still ends up being set using `f"x{prefix}"` which is set correctly, despite `"x_train"` being hardcoded in the previous line. This means that there is no data contamination between the train and test splits and the only observable consequence is the test split might end up being marked as unsupervised when it should really always be supervised.

**Reproducing**

Adapted from https://github.com/deepinv/deepinv/blob/1a4803087be766b564441c50f0b0887b88d808aa/examples/physics/demo_mri_tour.py

```python
import deepinv as dinv
import torch, torchvision

device = dinv.utils.get_freer_gpu() if torch.cuda.is_available() else "cpu"
rng = torch.Generator(device=device).manual_seed(0)

transform = torchvision.transforms.Resize(128)
knee_dataset = dinv.datasets.SimpleFastMRISliceDataset(
    dinv.utils.get_data_home(),
    anatomy="knee",
    transform=transform,
    train=True,
    download=True,
)
brain_dataset = dinv.datasets.SimpleFastMRISliceDataset(
    dinv.utils.get_data_home(),
    anatomy="brain",
    transform=transform,
    train=True,
    download=True,
)

img_size = knee_dataset[0].shape[-2:]  # (128, 128)
dinv.utils.plot({"knee": knee_dataset[0], "brain": brain_dataset[0]})

physics_generator = dinv.physics.generator.GaussianMaskGenerator(
    img_size=img_size, acceleration=4, rng=rng, device=device
)
mask = physics_generator.step()["mask"]

physics = dinv.physics.MRI(mask=mask, img_size=img_size, device=device)

dinv.utils.plot(
    {
        "x": (x := next(iter(DataLoader(knee_dataset)))),
        "mask": mask,
        "y": physics(x.to(device)).clamp(-1, 1),
    }
)
print("Shapes:", x.shape, physics.mask.shape)

dataset_path = dinv.datasets.generate_dataset(
    train_dataset=knee_dataset,
    test_dataset=brain_dataset,
    val_dataset=None,
    physics=physics,
    physics_generator=physics_generator,
    save_physics_generator_params=True,
    overwrite_existing=True,
    device=device,
    save_dir=dinv.utils.get_data_home(),
    batch_size=1,
    supervised=False,
)

test_dataset = dinv.datasets.HDF5Dataset(
    dataset_path, split="test", load_physics_generator_params=True
)

# This first assert passes
assert "x_test" in test_dataset.hd5, "There are test ground truths in the dataset"

x, y, params = test_dataset[0]

# The second assert fails
assert not torch.isnan(x), "The test ground truth should not be NaN"
```

### Steps to reproduce

/

### Expected behavior

/

_No response_

### Additional context

_No response_

jscanvic:
* Move h5py imports to where they are used

* Add static typing

* Fixes #765

The split name is now correctly determined by the parameters `train: bool` and `split: Optional[str]`. If `split` is set it takes the precedence, otherwise `train` is used. This fixes the bug.

* Process all the splits the same way

Non-training splits used to have `unsupervised` unset by default, even if there is no member `x_{split}` in the h5 file. On the other hand, training splits have `unsupervised` set when a member `x_train` is present in the h5 file. Now, all the splits are processed the same way in this regard and `unsupervised` is always set if and only if `x_{split}` is a member in the h5 file.

* Deprecate redundant and unused attributes

1. `load_physics_generator_params` is redundant with `params`
2. `stacked` is redundant with `self.y`
3. `data_info` and `data_cache` are unused entirely

* Refactor the implementation

Make the logic more streamlined instead of being unstructured. It makes the (existing) collisions and incompatibilities between features (stacked measurements and forward operator parameters) more obvious. It looks more complex now but the complexity was always there, it was simply hidden in unstructured code that is difficult to apprehend in its entirety.

**Breaking changes**

The code should do exactly the same as before. Fixing the bug shouldn't break anything. Changing the way `unsupervised` works shouldn't break anything either for normal uses.

**TODO**

* [x] Add tests
* [x] Possibly, fix the weird interaction between stacked measurements and forward operator parameters - see #767 
* [ ] Explain the case of stacked physics / measurements in the docs

### DIFF
--- a/deepinv/tests/test_datasets.py
+++ b/deepinv/tests/test_datasets.py
@@ -438,10 +438,14 @@ def test_hdf5dataset_generate_dataset(tmpdir, physgen, stacked, supervised):
         assert torch.isnan(x_train).all(), "Unsupervised train split should have NaN x."
 
     if stacked:
-        assert isinstance(y_train, TensorList), "Stacked physics should return TensorList."
+        assert isinstance(
+            y_train, TensorList
+        ), "Stacked physics should return TensorList."
         assert len(y_train) == 2, "Stacked measurements should have two elements."
     else:
-        assert isinstance(y_train, torch.Tensor), "Unstacked physics should return Tensor."
+        assert isinstance(
+            y_train, torch.Tensor
+        ), "Unstacked physics should return Tensor."
 
     if physgen is None:
         assert params_train == {}, "Params should be empty when no generator is used."
@@ -463,10 +467,14 @@ def test_hdf5dataset_generate_dataset(tmpdir, physgen, stacked, supervised):
     assert not torch.isnan(x_test).all(), "Test split should have x."
 
     if stacked:
-        assert isinstance(y_test, TensorList), "Stacked physics should return TensorList."
+        assert isinstance(
+            y_test, TensorList
+        ), "Stacked physics should return TensorList."
         assert len(y_test) == 2, "Stacked measurements should have two elements."
     else:
-        assert isinstance(y_test, torch.Tensor), "Unstacked physics should return Tensor."
+        assert isinstance(
+            y_test, torch.Tensor
+        ), "Unstacked physics should return Tensor."
 
     if physgen is None:
         assert params_test == {}, "Params should be empty when no generator is used."


jscanvic:
@Andrewwango The PR is pretty much done but I'm not happy with it - it looks difficult to review. Maybe I should split it into smaller PRs. What do you think?

Andrewwango:
@jscanvic Happy to keep as is and not split. Don't worry, it will be easy to review when the tests are there!

jscanvic:
Alright, thanks!

deepinv/datasets/base.py
```diff
@@ -129,7 +129,7 @@ class ImageDataset(Dataset):
     If using DeepInverse with your own custom dataset, you should inherit from this class and use :func:`check_dataset` to check your dataset is compatible.
```
Andrewwango, line 128: Changelog? Pretty major fix I would say

deepinv/datasets/datagenerator.py
```diff
@@ -23,6 +17,49 @@
     from deepinv.physics import Physics
     from deepinv.physics.generator import PhysicsGenerator
     from deepinv.transform import Transform
+    from typing import Any
+
+
+def _register_deprecated_attr(
```
Andrewwango, line 25: If this is general boilerplate code, should this live alongside other utils like our deprecated_alias decorator?

deepinv/datasets/datagenerator.py
```diff
@@ -23,6 +17,49 @@
...
+    :param Any attr_initial_value: The initial value of the attribute.
+    :param str deprecation_message: The deprecation warning message to be shown.
+    :param str, None doc: The docstring for the deprecated attribute.
+    """
+    setattr(self, attr_underscore_name, attr_initial_value)
+
+    def fget(self) -> bool:
+        val = getattr(self, attr_underscore_name)
+        # warn last in case retrieval fails
+        warn(deprecation_message, DeprecationWarning, sacklevel=2)
```
Andrewwango: stacklevel?

Andrewwango: tbh we should maybe just keep all warnings consistently on the same stacklevel throughout the codebase

deepinv/datasets/datagenerator.py
```diff
@@ -23,6 +17,49 @@
...
+        return val
+
+    def fset(self, value: bool) -> None:
+        setattr(self, attr_underscore_name, value)
+        # warn last in case setting fails
+        warn(deprecation_message, DeprecationWarning, sacklevel=2)
```
Andrewwango: likewise

deepinv/datasets/datagenerator.py
```diff
@@ -23,6 +17,49 @@
...
+
+    def fdel(self) -> None:
+        delattr(self, attr_underscore_name)
+        # warn last in case deletion fails
+        warn(deprecation_message, DeprecationWarning, sacklevel=2)
```
Andrewwango: likewise

deepinv/tests/test_datasets.py
```diff
@@ -202,8 +203,111 @@ def test_base_dataset():
...
+
+    # Test HDF5Dataset.__len__
+    assert (
+        len(dataset) == length
+    ), f"Dataset length should be {length} but got {len(dataset)}."
+
+    # Test HDF5Dataset.__getitem__
+    idx = 0
+    entry = dataset[idx]
+    expected_entry_length = 3 if load_physics_generator_params else 2
```
Andrewwango, line 317: thanks for the really clean tests

deepinv/datasets/datagenerator.py
```diff
@@ -58,69 +95,168 @@ def __init__(
         dtype: torch.dtype = torch.float,
         complex_dtype: torch.dtype = torch.cfloat,
     ):
+        import h5py
+
```
Andrewwango, line 189: I think we should better enumerate in the docstring what are the possibilities of data that is returned, when unsupervised, when stacked etc. - plus how you arrive at these possibilities depending on how you use `generate_dataset` - I know that unsupervised isn't a user-set param but it still confuses me a little

deepinv/datasets/datagenerator.py
```diff
@@ -58,69 +95,168 @@ def __init__(
...
 
-    def __getitem__(self, index):
-        r"""
-        Returns the measurement and signal pair ``(x, y)`` at the given index.
+    def __getitem__(self, index: int) -> tuple:
+        r"""Get an entry in the dataset.
 
-        If there is no training ground truth (i.e. ``x_train``) in the dataset file,
-        the dataset returns the measurement again as the signal.
+        Return the measuremet and signal pair ``(x, y)`` at the given index, in
```
Andrewwango:
```suggestion
        Return the measurement and signal pair ``(x, y)`` at the given index, in
```

deepinv/datasets/datagenerator.py
```diff
@@ -129,28 +265,81 @@ def __getitem__(self, index):
...
+    def close(self) -> None:
         """
         Closes the HDF5 dataset. Use when you are finished with the dataset.
         """
         if hasattr(self, "hd5") and self.hd5:
             self.hd5.close()
             self.hd5 = None
 
+    @property
+    def unsupervised(self) -> bool:
```
Andrewwango, line 445: nice

deepinv/datasets/datagenerator.py
```diff
@@ -129,28 +265,81 @@ def __getitem__(self, index):
...
+        import h5py
+
+        # The attribute load_physics_generator_params is redundant with the attribute params.
+        # Indeed, it is true if and only if the attribute params exists.
+        _register_deprecated_attr(
+            self,
+            attr_name="load_physics_generator_params",
+            attr_underscore_name="_load_physics_generator_params",
+            attr_initial_value=hasattr(self, "params"),
+            deprecation_message="The attribute 'load_physics_generator_params' is deprecated and will be removed in future versions. Use the attribute 'params' instead.",
```
Andrewwango: so just to check I understand, this warning is more for developers to not use `self.load_physics_generator_params`, but users can still use the `load_physics_generator_params` arg?

deepinv/datasets/datagenerator.py
```diff
@@ -58,69 +95,168 @@ def __init__(
...
         self.transform = transform
-        self.load_physics_generator_params = load_physics_generator_params
         self.cast = lambda x: x.type(complex_dtype if x.is_complex() else dtype)
 
-        if isinstance(h5py, ImportError):
-            raise h5py
-
-        self.hd5 = h5py.File(path, "r")
-        suffix = ("_train" if train else "_test") if split is None else f"_{split}"
+        self._register_deprecated_attributes()
```
Andrewwango: Why is this a separate method when it's just used once? Surely it will read better if it's in line. No worries if you disagree,  it's a minor point

deepinv/datasets/datagenerator.py
```diff
@@ -129,28 +265,81 @@ def __getitem__(self, index):
...
-        if self.stacked > 0:
-            return len(self.y[0])
-        else:
-            return len(self.y)
+        import h5py
 
-    def close(self):
+        y = self.y
+        if not isinstance(y, h5py.Dataset):
+            y = y[0]
```
Andrewwango: should we enforce that all datasets have the same len?

deepinv/datasets/datagenerator.py
```diff
@@ -58,69 +95,168 @@ def __init__(
...
+        r"""Get an entry in the dataset.
 
-        If there is no training ground truth (i.e. ``x_train``) in the dataset file,
-        the dataset returns the measurement again as the signal.
+        Return the measuremet and signal pair ``(x, y)`` at the given index, in
+        the selected split. If forward operator parameters are available, it
+        returns ``(x, y, params)`` where ``params`` is a dict of parameters.
+
+        If the split contains no ground truth, it returns a scalar NaN tensor
+        as the ground truth, similarly to :class:`deepinv.training.Trainer`.
```
Andrewwango: instead of referencing trainer which is just something that uses datasets, I would ref the datasets user guide https://deepinv.github.io/deepinv/user_guide/training/datasets.html

deepinv/datasets/datagenerator.py
```diff
@@ -58,69 +95,168 @@ def __init__(
...
+        import h5py
+
         super().__init__()
-        self.data_info = []
-        self.data_cache = {}
-        self.unsupervised = False
+
+        f = h5py.File(path, "r")
+
+        self._register_members(
```
Andrewwango: Again, is there a reason why this is a separate method rather than in-line or is it stylistic?

deepinv/datasets/datagenerator.py
```diff
@@ -58,69 +95,168 @@ def __init__(
...
+                raise ValueError(
+                    f"Dataset marked as stacked with {stacked} parts but some parts are missing."
+                )
+        self.y = y
 
-        if self.load_physics_generator_params:
-            self.params = {}
-            for k in self.hd5:
-                if suffix in k and k not in (f"x{suffix}", f"y{suffix}"):
-                    self.params[k.replace(suffix, "")] = self.hd5[k]
```
Andrewwango, line 95: thanks for removing this hacky crap that I wrote

deepinv/datasets/datagenerator.py
Andrewwango, line 445: wait so this `unsupervised` attribute is no longer used at all in the code? is it just for users in case they want to check the dataset? if no use surely we should deprecate it too

deepinv/datasets/datagenerator.py
```diff
@@ -58,69 +95,168 @@ def __init__(
...
+        # Process forward operator parameters
+        if load_physics_generator_params:
+            params = {}
         else:
-            self.x = self.hd5[f"x{suffix}"]
+            params = None
+
+        # Register members in the HDF5 file as ground truths, parts of stacked
+        # measurements or regular measurements and forward operator parameters
+        split_name = split if split is not None else ("train" if train else "test")
```
Andrewwango: should we add a warning when the arg is overridden to prevent jump-scares in future?

deepinv/datasets/datagenerator.py
```diff
@@ -58,69 +95,168 @@ def __init__(
...
-            if "x_train" in self.hd5:
-                self.x = self.hd5[f"x{suffix}"]
-            else:
-                self.unsupervised = True
+        # Process forward operator parameters
+        if load_physics_generator_params:
+            params = {}
         else:
-            self.x = self.hd5[f"x{suffix}"]
+            params = None
```
Andrewwango: oneliner?

deepinv/datasets/datagenerator.py
```diff
@@ -58,69 +95,168 @@ def __init__(
...
+                                    )
+                            else:
+                                warn(
+                                    f"Found stacked measurement member {member_name} with stacking index {stacking_index} outside of the expected range [0, {stacked}). There is probably an error with the dataset.",
+                                    UserWarning,
+                                    stacklevel=2,
+                                )
+
+                    # Register the forward operator parameters
+                    if params is not None and not registered_as_measurements:
```
Andrewwango, line 305: Thanks for this precise logic! but even though I originally contributed to parts of this I'm confusing myself again. Just to check, is it possible to have params and also stacked measurements, or not? 

Either way, we should make this clear in the docstring (as per previous comment)

deepinv/datasets/datagenerator.py
```diff
@@ -58,69 +95,168 @@ def __init__(
...
+                        params[attr_name] = member
```
Andrewwango, line 306: also here we assume that all leftover members are params, right? I know this is the original implementation; it feels like we should have a warning when there are unused members but I guess it doesn't really work with this current implementation

deepinv/tests/test_datasets.py
```diff
@@ -202,8 +203,111 @@ def test_base_dataset():
...
+@pytest.mark.parametrize("with_transform", [False, True])
+@pytest.mark.parametrize("load_physics_generator_params", [True, False])
+@pytest.mark.parametrize("dtype", [torch.float32, torch.float64])
+@pytest.mark.parametrize("complex_dtype", [torch.complex64, torch.complex128])
+@pytest.mark.parametrize("length", [10])
+@pytest.mark.parametrize("with_params", [False, True])
+@pytest.mark.parametrize("unsupervised", [False, True])
+@pytest.mark.parametrize("close", [False, True])
+@pytest.mark.parametrize("stack_size", [1, 2, 3])
+def test_hdf5dataset(
```
Andrewwango, line 231: should we also have an end-to-end dataset test showing all the possibilities of data you can save and load (stacked, params, unsupervised etc.), and simply test that the input dataset -> save -> load = output dataset?

deepinv/tests/test_datasets.py
Andrewwango, line 317: we should also check that the param has the correct name, no?

Andrewwango: and we also should test the interaction between params and stacked

deepinv/tests/test_datasets.py
```diff
@@ -202,8 +203,111 @@ def test_base_dataset():
...
+
+    # Populate a HDF5 file
+    with h5py.File(path, "w") as f:
+        # The stacked attribute is expected to be present only if stack_size > 1,
+        # even though stacked = 1 could be meaningful.
+        if stack_size > 1:
+            f.attrs["stacked"] = stack_size
+
+        def populate_dummy_data(field_name):
+            data = np.random.randn(length, 1, 4, 4)
```
Andrewwango: Do you think we could make the data fixed and different for train (x, y, kernel) and test (x, y, kernel) so that we can test the right split is returned on the right occassion, and x and y and kernel aren't mixed up?

deepinv/datasets/datagenerator.py
```diff
@@ -58,69 +95,168 @@ def __init__(
...
+                    # Other members are understood as parameters except if they are stacked measurements
+                    registered_as_measurements = False
+
+                    # Register part of the stacked measurements
+                    # Example target values for member_name: "y0_train", "y1_train", ...
+                    if attr_name.startswith("y"):
+                        stacking_needle = attr_name[1:]
+                        # If needle is the canonical base 10 representation of a non-negative integer
+                        if (
+                            re.fullmatch(r"0|[1-9]\d*", stacking_needle, re.ASCII)
```
Andrewwango, line 278: I know you are a 10x engineer who can recite regex but for mortals like me, does it not make sense just to check `stacking_needle.is_digit()`?

Andrewwango (changes requested):
Thanks Jérémy for the major overhaul! I have some comments, some questions and some confusions - but it looks good so far! Sorry for some of the demands

deepinv/datasets/datagenerator.py
jscanvic: Short answer: yes, always has been.

Long answer: `params` used to contain stacked measurements when it shouldn't due to incorrect fallback logic. It doesn't anymore!

deepinv/datasets/datagenerator.py
jscanvic: Exactly! ikr! It's the point I make in #771

deepinv/datasets/datagenerator.py
jscanvic: Exactly!

Why I do that? It's the whole _making illegal states unrepresentable_ ¹. Here, it's _impossible_ to have (successfully) loaded the params `self.load_physics_generator_params = True` yet the params not being loaded `not hasattr(self, "params")`, so I make it _unrepresentable_ by getting rid of (deprecating) `self.load_physics_generator_params`. (Hope it's clearer 😅)

¹ As described in, e.g., Domain Modeling Made Functional, Wlaschin, 2018

deepinv/datasets/datagenerator.py
jscanvic: I know it's gonna sound extra pedantic but please bear with me. 🙏 I didn't use `str.isdigit` because of 1) non-ascii digits: `"𐩀𐩁𐩂𐩃".isdigit() == True`, and 2) integers in non-canonical form: `"001".isdigit() == True`.

Case 1) makes `int(stacking_needle)` crash, and case 2) enables silent overwrites in `y[stacking_index] = member`, _and_ it's also a (pedantic) breaking change: `y001_train` used to be considered a parameter, and would be considered a measurement with `isdigit`.

The code is technical because our format is difficult to process and I'd prefer to leave it that way until we update the format itself as suggested in #771. We'd be fooling ourselves into thinking the format is not that clunky by writing simpler but incorrect code.

Happy to use `isdigit` anyway if you don't find the argument convincing!

deepinv/datasets/datagenerator.py
jscanvic: What do you mean? I thought we were using stacklevel=1 by default and 2 in warning helpers

deepinv/tests/test_datasets.py
jscanvic: You mean basically completing the test I renamed `test_hdf5dataset_generate_dataset` to have the same options as `test_hdf5dataset`?

deepinv/datasets/datagenerator.py
jscanvic: I don't mind either version - I inlined it

deepinv/datasets/datagenerator.py
jscanvic: I inlined it

deepinv/datasets/datagenerator.py
jscanvic: Happy to do it if we discontinue `deepinv.utils.decorators` in favor of a new `deepinv.utils._deprecation`

deepinv/datasets/datagenerator.py
jscanvic: Random thought passing by it would be neat to support semi-supervised training in the future

deepinv/datasets/datagenerator.py
Andrewwango: `𐩀𐩁𐩂𐩃` what

deepinv/datasets/datagenerator.py
Andrewwango: this is still todo?

deepinv/datasets/datagenerator.py
Andrewwango: where has this been done?

deepinv/datasets/datagenerator.py
Andrewwango: Ok would be great to include this in docstring

deepinv/tests/test_datasets.py
Andrewwango: I guess it would become a superset of the test_hdf5dataset_generate_dataset yes. but we can keep them separate for now, it would only add minimal overhead to do what I suggested

deepinv/datasets/datagenerator.py
jscanvic: Should be good now!

Issue #536, opened 2025-06-13
Static typing everywhere


Andrewwango:
Would be nice to have everything statically typed. Note that all params and return values are all implicitly typed from the docstrings (and hence the docs), so for those that aren't typed, it would be just a matter of corroborating docstrings and type hints.

Also, we should add `mypy` to the CI to check this.

This will complement the `ruff` checker added in #501

jscanvic:
We're a long way to go!

```bash
$ mypy deepinv
...
Found 699 errors in 93 files (checked 187 source files)
```

Maybe we should focus on a subset of the errors first?

There's [has-type]:

```
deepinv/physics/generator/base.py:76: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/generator/base.py:77: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/generator/base.py:112: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/generator/base.py:219: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/singlepixel.py:395: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/singlepixel.py:396: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/singlepixel.py:429: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/compressed_sensing.py:127: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/compressed_sensing.py:128: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/compressed_sensing.py:138: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/compressed_sensing.py:143: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/compressed_sensing.py:152: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/phase_retrieval.py:168: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/phase_retrieval.py:179: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/phase_retrieval.py:182: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/generator/inpainting.py:176: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/generator/inpainting.py:190: error: Cannot determine type of "rng"  [has-type]
deepinv/physics/generator/inpainting.py:206: error: Cannot determine type of "rng"  [has-type]
[...]

PR #768, opened 2025-09-13
`dinv.io` module and clean up utils

Andrewwango:
- input/output module for images and different types of data
- clean up `utils.demo` file and move some funcs to `io`

### DIFF
--- a/deepinv/utils/demo.py
+++ b/deepinv/utils/demo.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 from typing import Union, Callable, TYPE_CHECKING
+from warnings import warn
 import os, shutil, zipfile, requests
 from io import BytesIO
 
@@ -10,6 +11,7 @@ from PIL import Image
 import numpy as np
 import torch
 from torchvision import transforms
+from deepinv.utils.io import load_np, load_torch, load_url as _load_url
 
 if TYPE_CHECKING:
     from deepinv.datasets.base import ImageFolder
@@ -196,13 +198,10 @@ def load_image(
 
 
 def load_url(url: str) -> BytesIO:
-    """Load URL to a buffer.
-
-    :param str url: URL of the file to load
-    """
-    response = requests.get(url)
-    response.raise_for_status()
-    return BytesIO(response.content)
+    warn(
+        "deepinv.utils.demo.load_url is deprecated. Use deepinv.utils.load_url instead."
+    )
+    return _load_url(url)
 
 
 def load_url_image(
@@ -223,27 +222,24 @@ def load_url_image(
     :param str device: Device on which to load the image (gpu or cpu).
     :return: :class:`torch.Tensor` containing the image.
     """
-
-    img = Image.open(load_url(url))
-    transform_list = []
-    if img_size is not None:
-        if resize_mode == "crop":
-            transform_list.append(transforms.CenterCrop(img_size))
-        elif resize_mode == "resize":
-            transform_list.append(transforms.Resize(img_size))
-        else:
-            raise ValueError(
-                f"resize_mode must be either 'crop' or 'resize', got {resize_mode}"
-            )
-    if grayscale:
-        transform_list.append(transforms.Grayscale())
-    transform_list.append(transforms.ToTensor())
-    transform = transforms.Compose(transform_list)
-    x = transform(img).unsqueeze(0).to(device=device, dtype=dtype)
-    return x
+    return load_image(
+        _load_url(url),
+        img_size=img_size,
+        grayscale=grayscale,
+        resize_mode=resize_mode,
+        device=device,
+        dtype=dtype,
+    )
 
 
-def load_example(name, **kwargs):
+def load_example(
+    name,
+    img_size=None,
+    grayscale=False,
+    resize_mode="crop",
+    device="cpu",
+    dtype=torch.float32,
+):
     r"""
     Load example image from the `DeepInverse HuggingFace <https://huggingface.co/datasets/deepinv/images>`_ using
     :func:`deepinv.utils.load_url_image` if image file or :func:`deepinv.utils.load_torch_url` if torch tensor in `.pt` file
@@ -313,17 +309,27 @@ def load_example(name, **kwargs):
 
 
     :param str name: filename of the image from the HuggingFace dataset.
-    :param dict kwargs: keyword args to pass to :func:`deepinv.utils.load_url_image`
+    :param int, tuple[int] img_size: Size of the image to return.
+    :param bool grayscale: Whether to convert the image to grayscale.
+    :param str resize_mode: If ``img_size`` is not None, options are ``"crop"`` or ``"resize"``.
+    :param str device: Device on which to load the image (gpu or cpu).
     :return: :class:`torch.Tensor` containing the image.
     """
     url = get_image_url(name)
 
     if name.split(".")[-1].lower() == "pt":
-        return load_torch_url(url, **kwargs)
+        return load_torch_url(url, device=device)
     elif name.split(".")[-1].lower() in ("npy", "npz"):
-        return load_np_url(url, **kwargs)
+        return load_np_url(url)
 
-    return load_url_image(url, **kwargs)
+    return load_url_image(
+        url,
+        img_size=img_size,
+        grayscale=grayscale,
+        resize_mode=resize_mode,
+        device=device,
+        dtype=dtype,
+    )
 
 
 def download_example(name: str, save_dir: Union[str, Path]):
@@ -341,22 +347,22 @@ def download_example(name: str, save_dir: Union[str, Path]):
         f.write(data)
 
 
-def load_torch_url(url, device="cpu", **kwargs):
+def load_torch_url(url: str, device="cpu", **kwargs) -> torch.Tensor:
     r"""
     Load an array from url and read it by torch.load.
 
     :param str url: URL of the image file.
     :param str, torch.device device: Device on which to load the tensor.
-    :return: whatever is pickled in the file.
+    :return: weights or tensors contained in file.
     """
-    return torch.load(load_url(url), weights_only=True, map_location=device)
+    return load_torch(_load_url(url), device=device)
 
 
-def load_np_url(url=None, **kwargs):
+def load_np_url(url: str, **kwargs) -> torch.Tensor:
     r"""
-    Load a numpy array from url.
+    Load a numpy array from url and convert to tensor.
 
     :param str url: URL of the image file.
-    :return: numpy ndarray containing the data.
+    :return: torch rensor containing the data.
     """
-    return np.load(load_url(url), allow_pickle=False)
+    return load_np(_load_url(url))
--- a/deepinv/utils/io.py
+++ b/deepinv/utils/io.py
@@ -1,0 +1,62 @@
+from typing import Union
+from pathlib import Path
+from io import BytesIO
+import requests
+import numpy as np
+import torch
+
+# TODO user guide and API for all loaders here + in demo
+# TODO move over load_image and load_url_image?
+# TODO promote load_url as can use with any of the below
+# TODO docs and docstrings
+
+
+def load_dicom(path: Union[str, Path]) -> torch.Tensor:
+    """Load image from DICOM file.
+
+    Requires `pydicom` to be installed.
+
+    :param str, Path path: path to DICOM file.
+    :return: torch float tensor of shape `(B, 1, ...)` where `...` are the DICOM image dimensions.
+    """
+    try:
+        import pydicom
+    except ImportError:
+        raise ImportError(
+            "load_dicom requires pydicom, which is not installed. Please install it with `pip install pydicom`."
+        )
+    return torch.from_numpy(pydicom.dcmread(str(path)).pixel_array).float().unsqueeze(0)
+
+
+def load_nifti(path: Union[str, Path]) -> torch.Tensor:
+    try:
+        import nibabel as nib
+    except ImportError:
+        raise ImportError(
+            "load_nifti requires nibabel, which is not installed. Please install it with `pip install nibabel`."
+        )
+    return torch.from_numpy(nib.load(path).get_fdata())
+
+
+# TODO loadmat, loadmat scipy
+# TODO load h5 ISMRM
+# TODO load GDAL COSAR
+# TODO load GEOTIFF
+
+
+def load_torch(path: Union[str, Path], device=None) -> torch.Tensor:
+    return torch.load(path, weights_only=True, map_location=device)
+
+
+def load_np(path: Union[str, Path]) -> torch.Tensor:
+    return torch.from_numpy(np.load(path, allow_pickle=False))
+
+
+def load_url(url: str) -> BytesIO:
+    """Load URL to a buffer.
+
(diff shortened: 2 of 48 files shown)


nucli-vicky:
Two comments:
(i) .nii.gz files are loaded way faster when the `indexed-gzip` package is available. Since not all users are aware of this, it might be useful to flag this somewhere or include it when the install dependencies per use case functionality is included

(ii) For 3D data very often a subset / patch of the volume is loaded. Loading the full volume into memory therefore is very inefficient and often leads to dataloading becoming a bottleneck during training. `open_memmap` (numpy) and nibabel's equivalent `nib.load(...).dataobj` are ways to avoid such bottlenecks.

Andrewwango:
> Two comments: (i) .nii.gz files are loaded way faster when the `indexed-gzip` package is available. Since not all users are aware of this, it might be useful to flag this somewhere or include it when the install dependencies per use case functionality is included
> 
> (ii) For 3D data very often a subset / patch of the volume is loaded. Loading the full volume into memory therefore is very inefficient and often leads to dataloading becoming a bottleneck during training. `open_memmap` (numpy) and nibabel's equivalent `nib.load(...).dataobj` are ways to avoid such bottlenecks.

Hi @nucli-vicky thanks so much for the suggestions, it's great to have an expert opinion!

I fully agree that both suggestions (speeding up by suggesting installing indexed-gzip and streaming subset/patch data) would be really useful. Would you be interested in contributing by pushing to this branch? Let me know if you would like any help or alternatively if you can post a code snippet I can use it!

nucli-vicky:
@Andrewwango sure, I'll do so in a few days

pyproject.toml
```diff
@@ -141,3 +143,6 @@ select = ["UP006", "F401", "B905", "B006", "RUF008", "RUF012", "TID253", "UP004"
 [tool.ruff.lint.flake8-tidy-imports]
 # Modules that slow down startup and should be imported lazily
 banned-module-level-imports = ["timm", "torchmetrics", "matplotlib", "datasets", "FrEIA", "bm3d", "einops", "wandb", "pandas", "kornia", "ptwt", "pywt", "astra", "torchkbnufft"]
+
+[tool.ruff.lint.flake8-builtins]
+strict-checking = true
```
jscanvic: Should we follow that? torch and numpy don't with `torch.random` and `numpy.random`, torchvision neither with `torchvision.io`

deepinv/utils/io.py
```diff
@@ -0,0 +1,268 @@
...
+    :return: torch float tensor of shape `(1, ...)` where `...` are the DICOM image dimensions.
+    """
+    try:
+        import pydicom
+    except ImportError:  # pragma: no cover
+        raise ImportError(
+            "load_dicom requires pydicom, which is not installed. Please install it with `pip install pydicom`."
+        )
+    fname = str(fname) if isinstance(fname, Path) else fname
+    return torch.from_numpy(pydicom.dcmread(fname).pixel_array).float().unsqueeze(0)
```
jscanvic: I think we should use `pydicom.pixel_data_handlers.apply_rescale` to map the stored values (SV) to output values like we do in LidcIdriSliceDataset

Related: https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.11.html#sect_C.11.1

deepinv/utils/io_utils.py
```diff
@@ -0,0 +1,268 @@
...
+    :param str, Path fname: path to NIFTI `.nii.gz` file.
+    :return: torch float tensor of shape `(1, ...)` where `...` are the NIFTI image dimensions.
+    """
+    try:
+        import nibabel as nib
+    except ImportError:  # pragma: no cover
+        raise ImportError(
+            "load_nifti requires nibabel, which is not installed. Please install it with `pip install nibabel`."
+        )
+    return torch.from_numpy(nib.load(fname).get_fdata()).float().unsqueeze(0)
```
jscanvic: What's the intent of the `.float()`? The function `get_fdata` always returns float arrays, if you want to convert specifically to a single precision float, can you use `get_fdata(dtype=np.float32)` instead please? See: https://github.com/nipy/nibabel/blob/132ded6fa7576207648bd7d88fd8fea0f0866511/nibabel/dataobj_images.py#L226

jscanvic (changes requested):
Thanks Andrew, I really like the idea of grouping those in a single module! Just a few comments for now, I'll get back to it soon

pyproject.toml
Andrewwango: That's a good point, but calling our files with stdlib names messes up our doctests (e.g. try renaming io_utils to io.py, include a failing doctest in the doctests and run the doctests). Idk how numpy/torch/torchvision do that with their doctests?

Andrewwango: If we could find a solution that would be great, as I really want to call the file io.py

jscanvic: That's unfortunate... I'll have a look at it

Andrewwango:
- [x] TODO try io.py again

deepinv/utils/io_utils.py
nucli-vicky: Additionally, nifti images can have channels or a time axis. Automatically performing `unsqueeze(0)` may lead to some unintended or quiet bugs in these cases. Either we need to figure out a better way of handling this (in principle, these cases can be handled on the dataset side), or we should add a warning in the docs for users working with such nifti files.

nucli-vicky: Quite a lot of things can be learned from the image object (result of `nib.load`) prior to calling `get_fdata` or `dataobj` by looking at the [header](https://nipy.org/nibabel/nifti_images.html).
For instance the header already gives us the dtype which is important i.e. if we want to [gracefully handle complex values](https://github.com/nipy/nibabel/issues/975).

So it can help both with handling shapes other than H, W, D and with handling dtypes

Andrewwango: Done. Will add the smarter loading to the header

pyproject.toml
Andrewwango: fixed thanks, changed to io.py, turns out it's just the basic python `doctest` command isn't great but the sphinx one catches it

deepinv/utils/io.py
Andrewwango: done thanks! also let lidc use load_dicom

jscanvic (commented):
LGTM just one question is there a reason why `deepinv/utils/io.py` instead of `deepinv/io.py`? Happy to approve after that and once CI passes

deepinv/utils/demo.py
```diff
@@ -223,27 +225,24 @@ def load_url_image(
...
 
-def load_example(name, **kwargs):
+def load_example(
+    name,
+    img_size=None,
+    grayscale=False,
+    resize_mode="crop",
+    device="cpu",
+    dtype=torch.float32,
+):
```
nucli-vicky, line 245: Should we make this typed too?

deepinv/utils/demo.py
```diff
@@ -313,17 +312,27 @@ def load_example(name, **kwargs):
 
 
     :param str name: filename of the image from the HuggingFace dataset.
-    :param dict kwargs: keyword args to pass to :func:`deepinv.utils.load_url_image`
+    :param int, tuple[int] img_size: Size of the image to return.
+    :param bool grayscale: Whether to convert the image to grayscale.
+    :param str resize_mode: If ``img_size`` is not None, options are ``"crop"`` or ``"resize"``.
+    :param str device: Device on which to load the image (gpu or cpu).
```
nucli-vicky: Add `dtype` to docstring

deepinv/utils/io.py
```diff
@@ -0,0 +1,370 @@
...
+    :param np.dtype, str dtype: data type to use when loading the numpy array. This is ignored if `as_memmap` is `True`.
+    :return: :class:`torch.Tensor` containing loaded numpy array. If `as_memmap` is `True`, returns a numpy `memmap` object instead.
+    """
+    if as_memmap:
+        return open_memmap(fname)
+    else:
+        return torch.from_numpy(np.load(fname, allow_pickle=False).astype(dtype))
+
+
+def load_torch(fname: str | Path, device=None, **kwargs) -> torch.Tensor:
```
nucli-vicky: typing device

deepinv/utils/io.py
```diff
@@ -0,0 +1,370 @@
...
+    """
+    if mat73:
+        try:
+            from mat73 import loadmat as loadmat73
+
+            return loadmat73(fname)
+        except ImportError:  # pragma: no cover
+            raise ImportError("mat73 is required, install with 'pip install mat73'.")
+        except TypeError:  # pragma: no cover
+            pass
```
nucli-vicky: Do we want this to pass silently? I get it now falls back to the scipy function, but precise error message is now not visible to the user, and maybe we should at least add a warning?

deepinv/utils/io.py
```diff
@@ -0,0 +1,370 @@
...
+        return im.dataobj
+    else:
+        return torch.from_numpy(im.get_fdata(dtype=dtype))
+
+
+def load_blosc2(
+    fname: str | Path,
+    as_memmap: bool = False,
+    dtype: np.dtype | str = np.float32,
+    **kwargs,
```
nucli-vicky, line 358: What is `**kwargs` purpose here?

examples/external-libraries/demo_io.py
```diff
@@ -0,0 +1,115 @@
+r"""
+Loading scientific images
+=========================
+
+This example presents the various input/output functions provided by DeepInverse
+for handling medical and scientific imaging formats. We demonstrate loading and
+plotting from DICOM, NIfTI, ISMRMRD, PyTorch, NumPy and raster data sources.
```
nucli-vicky, line 7: This is moreso a general question/note than a comment :)
 I get we will not include everything here, and it's likely difficult to get some data which is natively in `blosc2` format (maybe there is some cool astronomy dataset?). 
But how will we make a general guideline of what loaders are mentioned here, and which will not be mentioned? 

In the case of `blosc2`, a tutorial might be nice just to show the type of speedups it offers, but that could just as well be a blog post. Not sure how to handle.

deepinv/utils/io.py
```diff
@@ -0,0 +1,370 @@
...
+    response.raise_for_status()
+    return BytesIO(response.content)
+
+
+def load_dicom(
+    fname: str | Path,
+    as_tensor: bool = True,
+    apply_rescale: bool = False,
+    dtype: np.dtype | str = np.float32,
+    **kwargs,
```
nucli-vicky, line 68: Just a note: dicom can contain useful metadata in the tags. I think some users might appreciate being able to obtain these from the loader too, but think this is outside this PR's scope

deepinv/utils/io.py
```diff
@@ -0,0 +1,370 @@
...
+        # the new function pydicom.pixels.apply_rescale. It is currently
+        # kept for compatibility with Python 3.9 which is only compatible
+        # with versions of pydicom older than version 3.0.0.
+        if not hasattr(pydicom, "pixel_data_handlers") or not hasattr(
+            pydicom.pixel_data_handlers, "apply_rescale"
+        ):
+            raise ImportError(
+                "pydicom version is unsupported. Please install a version of pydicom ≥ 2.0.0 and < 4.0.0"
+            )
+        x = pydicom.pixel_data_handlers.apply_rescale(x, data)
```
nucli-vicky, line 109: PET imaging typically wants SUV values, which are now not calculated and cannot be recovered from the array alone (needs some DICOM tags). 
This makes this code difficult for applying it to PET; I don't think however we need to bring code specific for each modality, so fine with leaving as is. After all users can also convert to SUV themselves and store as e.g. a `.nii` file

nucli-vicky (changes requested):
Thanks @Andrewwango for this much needed work! I have a few comments, questions, and also just notes we need to keep in mind. All minor, the core LGTM!

deepinv/utils/io.py
Andrewwango: to be compatible to be able to be passed as a loader to 3D patch sampler :)

I envision these loader functions being drop in functions for datasets, demos, examples etc. But if you think this is premature, then I can remove them

examples/external-libraries/demo_io.py
Andrewwango: Yes I would love to include one!

Do you have a demo volume which we could show? and/or some sample text

Secondly, I realise that we don't test blosc2 in test_utils now, only test_datasets (your code). Do you think we should add blosc2 test to test_utils?

deepinv/utils/io.py
Andrewwango: Agreed! Lot's to do

deepinv/utils/io.py
Andrewwango: would you suggest adding more to the apply_rescale docstring to say it's only useful for limited cases like Lidc-idri?

deepinv/utils/demo.py
Andrewwango: I will leave this to #536 because the rest of the demo file is not statically typed too

deepinv/utils/demo.py
Andrewwango: done

deepinv/utils/io.py
nucli-vicky: Makes sense, I have a personal bias against such patterns but it really is grounded in no true reason. Fine to keep as is!

examples/external-libraries/demo_io.py
nucli-vicky: Problem is I'm not aware of any dataset that is natively in `blosc2`. We could ofcourse convert some image/volume to it for the sake of it, but this seems less useful. Maybe best kept for a true blosc2 tutorial, outside of test_utils? 

Regarding tests: I would add it, simply because otherwise we do not test `return torch.from_numpy(arr[:].astype(dtype))`

deepinv/utils/io.py
nucli-vicky: This would be useful, yes. Maybe we can explicitly mention SUV/PET as an example, and suggest the user to perform conversion themselves and store to blosc2/nifti if the `apply_rescale` is not sufficient for their dataset?

deepinv/utils/io.py
```diff
@@ -0,0 +1,372 @@
...
+        import blosc2
+    except ImportError:  # pragma: no cover
+        raise ImportError(
+            "load_blosc2 requires blosc2, which is not installed. Please install it with `pip install blosc2`."
+        )
+    arr = blosc2.open(fname)
+    if as_memmap:
+        return arr
+    else:
+        return torch.from_numpy(arr[:].astype(dtype))
```
nucli-vicky, line 377: Test the else statement instead of just relying on `RandomPatch` tests which only trigger `as_memmap`

deepinv/utils/io.py
Andrewwango: done

examples/external-libraries/demo_io.py
Andrewwango: added a test

deepinv/utils/io.py
Andrewwango: done

jscanvic (approved):
Thanks for the hard work!

PR #777, opened 2025-09-17
add self-sup eval + history fixes

tachella:
This PR removes model.eval() from the evaluation phase of the trainer, to allow for self-supervised evaluation and early stopping. 

- I also modified the self-supervised examples accordingly. 
- I've added history loss saving, which was missing for some weird reason

### DIFF
--- a/deepinv/training/trainer.py
+++ b/deepinv/training/trainer.py
@@ -114,7 +114,7 @@ class Trainer:
 
     :param None, torch.utils.data.DataLoader, list[torch.utils.data.DataLoader] eval_dataloader: Evaluation data loader(s),
         see :ref:`datasets user guide <datasets>` for how we expect data to be provided.
-    :param Metric, list[Metric] metrics: Metric or list of metrics used for evaluating the model.
+    :param Metric, list[Metric], Loss, list[Loss] metrics: Metric or list of metrics used for evaluating the model.
         They should have ``reduction=None`` as we perform the averaging using :class:`deepinv.utils.AverageMeter` to deal with uneven batch sizes.
         :ref:`See the libraries' evaluation metrics <metric>`. Default is :class:`PSNR <deepinv.loss.metric.PSNR>`.
     :param bool disable_train_metrics: if `False` (default), metrics are computed both during training and testing on their respective data.
@@ -220,11 +220,11 @@ class Trainer:
     physics_generator: Union[PhysicsGenerator, list[PhysicsGenerator]] = None
     loop_random_online_physics: bool = False
     optimizer_step_multi_dataset: bool = True
-    metrics: Union[Metric, list[Metric]] = field(default_factory=PSNR)
+    metrics: Union[Metric, list[Metric], Loss, list[Loss]] = field(default_factory=PSNR)
     disable_train_metrics: bool = False
     device: Union[str, torch.device] = "cuda" if torch.cuda.is_available() else "cpu"
     ckpt_pretrained: Union[str, None] = None
-    save_path: Union[str, Path] = "."
+    save_path: Union[str, Path, None] = "."
     compare_no_learning: bool = False
     no_learning_method: str = "A_adjoint"
     grad_clip: float = None
@@ -330,6 +330,12 @@ class Trainer:
             for l in self.losses
         ]
 
+        self.train_loss_history = {}
+        self.eval_loss_history = {}
+        for l in self.losses:
+            self.train_loss_history[l.__class__.__name__] = []
+            self.eval_loss_history[l.__class__.__name__] = []
+
         # metrics
         self.logs_metrics_train = [
             AverageMeter("Training metric " + l.__class__.__name__, ":.2e")
@@ -380,8 +386,6 @@ class Trainer:
         ):
             self.physics_generator = [self.physics_generator]
 
-        if train:
-            self.loss_history = []
         self.save_folder_im = None
 
         _ = self.load_model()
@@ -945,11 +949,12 @@ class Trainer:
         state = state | {
             "epoch": epoch,
             "state_dict": self.model.state_dict(),
-            "loss": self.loss_history,
+            "loss": self.train_loss_history,
+            "eval_loss": self.eval_loss_history,
+            "eval_metrics": self.eval_metrics_history,
             "optimizer": self.optimizer.state_dict() if self.optimizer else None,
             "scheduler": self.scheduler.state_dict() if self.scheduler else None,
         }
-        state["eval_metrics"] = self.eval_metrics_history
         if self.wandb_vis:
             import wandb
 
@@ -1132,7 +1137,10 @@ class Trainer:
                         ]
                     )
 
-                    self.model.eval()
+                    # self.model.eval()
+                    # The eval mode is removed to allow for
+                    # self-supervised losses used as eval metrics (e.g. R2R)
+
                     # close train progress bar
                     progress_bar.update(1)
                     progress_bar.close()
@@ -1155,13 +1163,21 @@ class Trainer:
                             last_batch=(j == eval_batches - 1),
                         )
 
+                    # store losses history
+                    for l in self.losses:
+                        self.train_loss_history[l.__class__.__name__].append(
+                            self.logs_losses_train[self.losses.index(l)].avg
+                        )
+                        self.eval_loss_history[l.__class__.__name__].append(
+                            self.logs_losses_eval[self.losses.index(l)].avg
+                        )
+
+                    # store metrics history
                     for k in range(len(self.metrics)):
                         metric = self.logs_metrics_eval[k].avg
                         self.eval_metrics_history[
                             self.metrics[k].__class__.__name__
-                        ].append(
-                            metric
-                        )  # store metrics history
+                        ].append(metric)
 
                     self.save_best_model(epoch, train_ite)
 
@@ -1173,8 +1189,6 @@ class Trainer:
                     progress_bar.update(1)
                     progress_bar.close()
                     break
-
-            self.loss_history.append(self.logs_total_loss_train.avg)
 
             if self.scheduler:
                 self.scheduler.step()
@@ -1201,6 +1215,7 @@ class Trainer:
         save_path: Union[str, Path] = None,
         compare_no_learning: bool = True,
         log_raw_metrics: bool = False,
+        metrics: Union[Metric, list[Metric], None] = None,
     ) -> dict:
         r"""
         Test the model, compute metrics and plot images.
@@ -1210,8 +1225,15 @@ class Trainer:
         :param str save_path: Directory in which to save the plotted images.
         :param bool compare_no_learning: If ``True``, the linear reconstruction is compared to the network reconstruction.
         :param bool log_raw_metrics: if `True`, also return non-aggregated metrics as a list.
+        :param deepinv.metrics.Metric, list[deepinv.metrics.Metric] metrics: Metric or list of metrics used for evaluation. If
+            ``None``, uses the metrics provided during Trainer initialization.
         :returns: dict of metrics results with means and stds.
         """
+        if metrics is not None:
+            if not isinstance(metrics, list):
+                metrics = [metrics]
+            self.metrics = metrics
+
         self.compare_no_learning = compare_no_learning
         self.setup_train(train=False)
 
--- a/examples/self-supervised-learning/demo_artifact2artifact.py
+++ b/examples/self-supervised-learning/demo_artifact2artifact.py
@@ -201,6 +201,10 @@ model = loss.adapt_model(model)
 # train from scratch, simply comment out the model loading code and
 # increase the number of epochs.
 #
+# .. tip::
+#
+#       We can use the same self-supervised loss for evaluation, as it does not require clean images,
+#       to monitor the training process (e.g. for early stopping).
 
 optimizer = torch.optim.Adam(model.parameters(), lr=1e-3, weight_decay=1e-8)
 
@@ -222,7 +226,7 @@ trainer = dinv.Trainer(
     losses=loss,
     optimizer=optimizer,
     train_dataloader=train_dataloader,
-    metrics=[dinv.metric.PSNR(), dinv.metric.SSIM()],
+    metrics=loss,
     online_measurements=True,
     device=device,
     save_path=None,
@@ -239,7 +243,7 @@ model = trainer.train()
 #
 
 trainer.plot_images = True
-trainer.test(test_dataloader)
+trainer.test(test_dataloader, metrics=[dinv.metric.PSNR(), dinv.metric.SSIM()])
 
 # %%
 # :References:
--- a/examples/self-supervised-learning/demo_ei_transforms.py
+++ b/examples/self-supervised-learning/demo_ei_transforms.py
@@ -105,6 +105,10 @@ physics = dinv.physics.Inpainting((3, 256, 256), mask=0.6, device=device)
 #
 #       We only train for a single epoch in the demo, but it is recommended to train multiple epochs in practice.
 #
+# .. tip::
+#
+#       We can use the same self-supervised loss for evaluation, as it does not require clean images,
+#       to monitor the training process (e.g. for early stopping).
 
 model = dinv.models.UNet(
     in_channels=3, out_channels=3, scales=2, circular_padding=True, batch_norm=False
@@ -125,6 +129,8 @@ model = dinv.Trainer(
     eval_dataloader=test_dataloader,
     epochs=1,
     losses=losses,
+    metrics=losses,
+    early_stop=True,  # we can use early stopping as we have a validation set
     optimizer=optimizer,
     verbose=True,
     show_progress_bar=False,
--- a/examples/self-supervised-learning/demo_equivariant_imaging.py
+++ b/examples/self-supervised-learning/demo_equivariant_imaging.py
@@ -158,7 +158,10 @@ optimizer.load_state_dict(ckpt["optimizer"])
 # Train the network
 # --------------------------------------------
 #
-
+# .. tip::
+#
+#       We can use the same self-supervised loss for evaluation, as it does not require clean images,
+#       to monitor the training process (e.g. for early stopping).
 
 verbose = True  # print training information
 
@@ -178,6 +181,9 @@ trainer = dinv.Trainer(
     losses=losses,
     optimizer=optimizer,
     train_dataloader=train_dataloader,
+    eval_dataloader=test_dataloader,
+    metrics=losses,
+    early_stop=True,  # early stop using the self-supervised loss on the test set
     plot_images=True,
     device=device,
     save_path=str(CKPT_DIR / operation),
@@ -194,7 +200,7 @@ model = trainer.train()
 #
 #
 
-trainer.test(test_dataloader)
+trainer.test(test_dataloader, metrics=dinv.metric.PSNR())
 
 # %%
 # :References:
--- a/examples/self-supervised-learning/demo_multioperator_imaging.py
+++ b/examples/self-supervised-learning/demo_multioperator_imaging.py
@@ -160,8 +160,10 @@ optimizer.load_state_dict(ckpt["optimizer"])
 # Train the network
 # --------------------------------------------
 #
+# .. tip::
 #
-
+#       We can use the same self-supervised loss for evaluation, as it does not require clean images,
+#       to monitor the training process (e.g. for early stopping).
 
 verbose = True  # print training information
 
@@ -185,6 +187,8 @@ trainer = dinv.Trainer(
     device=device,
     train_dataloader=train_dataloader,
     eval_dataloader=test_dataloader,
+    metrics=losses,
+    early_stop=True,  # early stop using the self-supervised loss on the test set
     save_path=str(CKPT_DIR / operation),
     verbose=verbose,
     plot_images=True,
@@ -200,7 +204,7 @@ model = trainer.train()
 # --------------------------------------------
 #
 
-trainer.test(test_dataloader)
+trainer.test(test_dataloader, metrics=dinv.metric.PSNR())
 
 # %%
 # :References:
--- a/examples/self-supervised-learning/demo_n2n_denoising.py
+++ b/examples/self-supervised-learning/demo_n2n_denoising.py
@@ -144,7 +144,10 @@ optimizer.load_state_dict(ckpt["optimizer"])
 # Train the network
 # --------------------------------------------
 #
+# .. tip::
 #
+#       We can use the same self-supervised loss for evaluation, as it does not require clean images,
+#       to monitor the training process (e.g. for early stopping).
 
 verbose = True  # print training information
 
@@ -166,6 +169,8 @@ trainer = dinv.Trainer(
     device=device,
     train_dataloader=train_dataloader,
     eval_dataloader=test_dataloader,
+    metrics=loss,
+    early_stop=True,  # early stop using the self-supervised loss on the test set
     plot_images=True,
     save_path=str(CKPT_DIR / operation),
     verbose=verbose,
@@ -180,7 +185,7 @@ model = trainer.train()
 #
 #
 
-trainer.test(test_dataloader)
+trainer.test(test_dataloader, metrics=dinv.metric.PSNR())
 
 # %%
 # :References:
--- a/examples/self-supervised-learning/demo_r2r_denoising.py
+++ b/examples/self-supervised-learning/demo_r2r_denoising.py
@@ -149,6 +149,11 @@ if noise_name == "poisson":
 # --------------------------------------------
 #
 #
+# .. tip::
+#
+#       We can use the same self-supervised loss for evaluation, as it does not require clean images,
+#       to monitor the training process (e.g. for early stopping).
+
 
 verbose = True  # print training information
 
@@ -170,11 +175,13 @@ trainer = dinv.Trainer(
     optimizer=optimizer,
     device=device,
     train_dataloader=train_dataloader,
-    eval_dataloader=None,
+    eval_dataloader=test_dataloader,
+    early_stop=True,  # early stop using the self-supervised loss on the test set
+    metrics=loss,
     plot_images=True,
     save_path=str(CKPT_DIR / operation),
     verbose=verbose,
-    show_progress_bar=False,  # disable progress bar for better vis in sphinx gallery.
+    show_progress_bar=True,  # disable progress bar for better vis in sphinx gallery.
 )
 
 # Train the network
@@ -186,8 +193,7 @@ model = trainer.train()
 # --------------------------------------------
 #
 #
-
-trainer.test(test_dataloader)
+trainer.test(test_dataloader, metrics=dinv.metric.PSNR())
 
 # %%
 # :References:
--- a/examples/self-supervised-learning/demo_splitting_loss.py
+++ b/examples/self-supervised-learning/demo_splitting_loss.py
@@ -147,6 +147,10 @@ optimizer.load_state_dict(ckpt["optimizer"])
 # Train and test network
 # ----------------------
 #
+# .. tip::
+#
+#       We can use the same self-supervised loss for evaluation, as it does not require clean images,
+#       to monitor the training process (e.g. for early stopping).
 
 trainer = dinv.Trainer(
     model=model,
@@ -156,6 +160,9 @@ trainer = dinv.Trainer(
     optimizer=optimizer,
     device=device,
     train_dataloader=train_dataloader,
+    eval_dataloader=test_dataloader,
+    metrics=loss,
+    early_stop=True,  # we can use early stopping as we have a validation loss
     plot_images=False,
     save_path=None,
     verbose=True,
@@ -173,7 +180,7 @@ model = trainer.train()
 #
 
 trainer.plot_images = True
-trainer.test(test_dataloader)
+trainer.test(test_dataloader, metrics=dinv.metric.PSNR())
 
 
 # %%
@@ -183,7 +190,7 @@ trainer.test(test_dataloader)
 #
 
 model.eval_n_samples = 1
-trainer.test(test_dataloader)
+trainer.test(test_dataloader, metrics=dinv.metric.PSNR())
 
 
 # %%
@@ -194,7 +201,7 @@ trainer.test(test_dataloader)
 #
 
 model.eval_split_input = False
-trainer.test(test_dataloader)
+trainer.test(test_dataloader, metrics=dinv.metric.PSNR())
 
 # %%
 # :References:
--- a/examples/self-supervised-learning/demo_sure_denoising.py
+++ b/examples/self-supervised-learning/demo_sure_denoising.py
@@ -152,7 +152,10 @@ optimizer.load_state_dict(ckpt["optimizer"])
 # Train the network
 # --------------------------------------------
 #
+# .. tip::
 #
+#       We can use the same self-supervised loss for evaluation, as it does not require clean images,
+#       to monitor the training process (e.g. for early stopping).
 
 verbose = True  # print training information
 
@@ -174,6 +177,8 @@ trainer = dinv.Trainer(
     device=device,
     train_dataloader=train_dataloader,
     eval_dataloader=test_dataloader,
+    metrics=loss,
+    early_stop=True,  # early stop using the self-supervised loss on the test set
     plot_images=True,
     save_path=str(CKPT_DIR / operation),
     verbose=verbose,
@@ -190,4 +195,4 @@ model = trainer.train()
 #
 #
 
-trainer.test(test_dataloader)
+trainer.test(test_dataloader, metrics=dinv.metric.PSNR())


deepinv/training/trainer.py
```diff
@@ -1132,7 +1137,10 @@ def train(
                         ]
                     )
 
-                    self.model.eval()
+                    # self.model.eval()
```
jscanvic: Just making sure I understand this well: it makes the eval loss and also the metrics be computed in training mode? Does it also affect `Trainer.test`?

jscanvic (commented):
Thanks Julian it's a really great improvement! I'll have a closer look later

deepinv/training/trainer.py
tachella: Yes, this allows to compute R2R or splitting losses during eval mode
`Trainer.test` has another `model.eval`, which is untouched

tachella: I've ended up changing the strategy - see details in the PR description

deepinv/training/trainer.py
```diff
@@ -696,6 +722,12 @@ def compute_metrics(
         :returns: The logs with the metrics.
         """
 
+        if len(self.metrics) > 0 and (
+            self.compute_metrics_on_eval_mode or x_net is None
+        ):
+            # re-evaluate the model in eval mode if needed
+            x_net = self.model_inference(y=y, physics=physics, x=x, train=False)
```
jscanvic, line 865: Thanks Julian, I really like having the option to compute eval losses in train mode but eval metrics in eval mode. It makes a lot of sense to me!

deepinv/loss/sup.py
```diff
@@ -118,7 +120,8 @@ def forward(self, y: Tensor, physics: Physics, **kwargs):
                         str(e),
                     )
                 try:
-                    return self.model(z, physics)
+                    x_hat = self.model(z, physics)
+                    return x_hat
```
jscanvic: Why this change?

deepinv/loss/sup.py
```diff
@@ -93,7 +93,9 @@ def forward(self, x_net: Tensor, y: Tensor, *args, **kwargs):
             return self.metric(x_net, y)
         except BaseException as e:
             raise RuntimeError(
-                f"Metric error. Check that the reconstruction (of shape {x_net.shape}) and y (of shape {y.shape}) can be used to calculate the metric. Full error:",
+                f"Metric error. Check that the reconstruction (of shape {x_net.shape}) and y (of shape {y.shape}) can be used to calculate the metric. "
+                f"If using deepinv.Trainer, this issue can be solved by setting compute_metrics_on_eval_mode=True. "
```
jscanvic: It used to fail using e.g. a splitting model right? Just to make sure I understand that correctly

deepinv/models/artifactremoval.py
```diff
@@ -61,6 +61,8 @@ def __init__(
         if type(self.backbone_net).__name__ == "UNetRes":
             for _, v in self.backbone_net.named_parameters():
                 v.requires_grad = False
+
+        if device is not None:
```
jscanvic, line 65: Nice

deepinv/models/gan.py
```diff
@@ -342,25 +343,26 @@ def optimize_z(self, z: Tensor, y: Tensor, physics: Physics):
...
-            optimizer.step()
-
-            err_curr = error.item()
-            err_perc = abs(err_curr - err_prev) / err_curr
-            err_prev = err_curr
-            pbar.set_postfix({"err_curr": err_curr, "err_perc": err_perc})
-
-            if err_perc < self.inf_tol:
-                break
+        with torch.enable_grad():
```
jscanvic: Can you put it as a decorator instead since it applies to the entire function? https://docs.pytorch.org/docs/2.8/generated/torch.enable_grad.html

deepinv/tests/test_loss.py
```diff
@@ -366,15 +366,22 @@ def test_losses(
         plot_images=(loss_name == LOSSES[0]),  # save time
         verbose=False,
         log_train_batch=(loss_name == "sup_log_train_batch"),
-        disable_train_metrics=(loss_name == "reducedresolution"),
+        # disable_train_metrics=(loss_name == "reducedresolution"),
```
jscanvic: Can you remove this line instead of commenting it?

deepinv/tests/test_loss.py
```diff
@@ -366,15 +366,22 @@ def test_losses(
...
 
     # test the untrained model
-    initial_test = trainer.test(test_dataloader=test_dataloader)
+    initial_test = trainer.test(
+        test_dataloader=test_dataloader, metrics=deepinv.metric.PSNR()
+    )
 
     # train the network
+    trainer.metrics = (
+        [] if (loss_name == "reducedresolution") else deepinv.metric.PSNR()
```
jscanvic: Don't you need to wrap PSNR() in an array here?

deepinv/tests/test_trainer.py
```diff
@@ -731,24 +731,37 @@ def forward(self, *args, **kwargs):
         )
 
 
-def test_total_loss(dummy_dataset, imsize, device, dummy_model, tmpdir):
+class ConstantLoss2(ConstantLoss):
+    def __init__(self, value, device):
```
jscanvic: The constructor isn't needed here I believe, can you put `pass` instead of it?

deepinv/tests/test_trainer.py
```diff
@@ -731,24 +731,37 @@ def forward(self, *args, **kwargs):
...
     eval_dataloader = DataLoader(eval_data, batch_size=2)
     physics = dinv.physics.Inpainting(img_size=imsize, device=device, mask=0.5)
 
     losses = [
         ConstantLoss(1 / 2, device),
-        ConstantLoss(1 / 3, device),
+        ConstantLoss2(1 / 3, device),
+    ]
+
+    eval_losses = [
```
jscanvic: I find it confusing to call them `eval_losses`, can you change to `metrics` instead?

deepinv/tests/test_trainer.py
```diff
@@ -757,11 +770,101 @@ def test_total_loss(dummy_dataset, imsize, device, dummy_model, tmpdir):
...
     trainer.train()
 
-    loss_history = trainer.loss_history
-    assert all(
-        [abs(value - sum([l.value for l in losses])) < 1e-6 for value in loss_history]
+    for k, (_, loss_history) in enumerate(trainer.train_loss_history.items()):
+        l = losses[k]
+        assert all([abs(value - l.value) < 1e-6 for value in loss_history])
+
+    for k, (_, loss_history) in enumerate(trainer.eval_metrics_history.items()):
```
jscanvic: `metrics_history` instead?

deepinv/tests/test_trainer.py
```diff
@@ -757,11 +770,101 @@ def test_total_loss(dummy_dataset, imsize, device, dummy_model, tmpdir):
 
     trainer.train()
 
-    loss_history = trainer.loss_history
-    assert all(
-        [abs(value - sum([l.value for l in losses])) < 1e-6 for value in loss_history]
+    for k, (_, loss_history) in enumerate(trainer.train_loss_history.items()):
+        l = losses[k]
```
jscanvic, line 814: I find it a bit brittle to rely on the order of the dict and the list being the same but happy to leave it as is

deepinv/tests/test_trainer.py
```diff
@@ -757,11 +770,101 @@ def test_total_loss(dummy_dataset, imsize, device, dummy_model, tmpdir):
...
+        assert all([abs(value - l.value) < 1e-6 for value in loss_history])
+
+
+class DummyModel(dinv.models.Reconstructor):
+    def __init__(self):
+        super().__init__()
+        self.eval_count = 0
+        self.train_count = 0
+        self.param = torch.nn.Parameter(torch.ones(1), requires_grad=True)
```
jscanvic, line 843: (It's the default value already)

```suggestion
        self.param = torch.nn.Parameter(torch.ones(1))
```

deepinv/tests/test_trainer.py
```diff
@@ -757,11 +770,101 @@ def test_total_loss(dummy_dataset, imsize, device, dummy_model, tmpdir):
...
+
+    def forward(self, y, physics, **kwargs):
+        x = physics.A_adjoint(y)
+        if self.training:
+            self.train_count += 1
```
jscanvic, line 848: I'd prefer a solution using `unittest.mock` instead of writing our custom mocks but happy to leave it as is if it's too much work

What I have in mind is wrapping forward in a mock with a `side_effect` along the lines of:

```python
def fn(self, *args):
   if self.training:
       train_count += 1
   else:
       eval_count += 1
```
   

Related: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect

deepinv/tests/test_trainer.py
```diff
@@ -757,11 +770,101 @@ def test_total_loss(dummy_dataset, imsize, device, dummy_model, tmpdir):
...
+        disable_train_metrics=disable_train_metrics,
     )
 
+    assert model.train_count == 0
+    assert model.eval_count == 0
+
+    trainer.train()
+
+    train_calls = len(dataloader) * epochs
+    eval_calls = len(eval_dataloader) * (epochs // eval_interval + 1)
```
jscanvic: Does the formula holds when `epochs` is not a multiple of `eval_interval`?

deepinv/tests/test_trainer.py
tachella: you are right - fixed

deepinv/tests/test_trainer.py
tachella: yes, you are right, abusing the copy pasting

deepinv/tests/test_trainer.py
tachella: I've added an assert for the name to make sure we're doing things right

deepinv/tests/test_trainer.py
tachella: I think its not bad to specify it for clarity

deepinv/tests/test_trainer.py
tachella: I prefer to keep it simple like this if you don't mind

deepinv/tests/test_trainer.py
tachella: yes, I've added some additional tests for epochs=5

Andrewwango:
- [x] TODO: fix #794

deepinv/loss/sup.py
tachella: I think it was a debugging artefact - I'll remove

deepinv/loss/sup.py
tachella: I think the additional message is not necessary anymore. The error happens when we try to compute metrics using the model in train() mode.

deepinv/models/gan.py
tachella: I agree! done

deepinv/tests/test_loss.py
tachella: yes, I've added to be safe

deepinv/tests/test_loss.py
tachella: done

Andrewwango:
@jscanvic  which points in #553 will be fixed by this PR?

deepinv/training/trainer.py
```diff
@@ -115,24 +116,32 @@ class Trainer:
...
-    :param bool disable_train_metrics: if `False`, metrics are computed both during training and testing on their respective data.
-        In this case, the metrics are computed using the model prediction during training (i.e in `model.train()` mode) to avoid an additional
-        forward pass. If `True` (default), do not compute metrics during training on train set.
+    :param bool disable_train_metrics: If `True` (default), do not compute metrics during training on train set.
+        If `False`, all metrics are computed both during training on the training dataloader.
         .. warning::
 
-            When using self-supervised losses for validation (i.e., with compute_losses_eval=True and metrics=[]),
-            we recommend setting disable_train_metrics=True to avoid computing metrics on train mode.
+            If `disable_train_metrics=False` the metrics are computed using the model prediction during training (i.e in `model.train()` mode) to avoid an additional
```
jscanvic: Minor typo: i.e. or i.e.,*

jscanvic:
> @jscanvic which points in #553 will be fixed by this PR?

* Saving the eval loss

Maybe this one? @tachella 
* Make it so we can use the eval loss as criterion for the best checkpoint

And it helps towards (but I expect even better, with 3 dataloaders ideally)
* Make the trainer easier to work with for a train/val/test workflow instead of the current train/val-test workflow

jscanvic (approved):
LGTM! Thanks Julián, it's a really important fix

Happy to merge after the typo is fixed, if CI passes and if @Andrewwango is happy with the UNSURE fix

tachella:
> > @jscanvic which points in #553 will be fixed by this PR?
> 
> * Saving the eval loss
> 
> Maybe this one? @tachella
> 
> * Make it so we can use the eval loss as criterion for the best checkpoint
> 
> And it helps towards (but I expect even better, with 3 dataloaders ideally)
> 
> * Make the trainer easier to work with for a train/val/test workflow instead of the current train/val-test workflow

also the point
- saving the eval loss

deepinv/training/trainer.py
```diff
@@ -224,16 +237,17 @@ class Trainer:
     losses: Loss | BaseLossScheduler | list[Loss] | list[BaseLossScheduler] = SupLoss()
     eval_dataloader: torch.utils.data.DataLoader = None
     early_stop: bool = False
+    early_stop_patience: int = 3
```
Andrewwango: should we just make this the same parameter as `early_stop` i.e. it can be bool or int?

Andrewwango (changes requested):
One comment for now, full review to come

deepinv/training/trainer.py
tachella: fixed

deepinv/training/trainer.py
tachella: I set None as default (no early stopping) and int otherwise

deepinv/tests/test_loss.py
```diff
@@ -365,15 +367,23 @@ def test_losses(
...
     )
 
     # test the untrained model
-    initial_test = trainer.test(test_dataloader=test_dataloader)
+    initial_test = trainer.test(
+        test_dataloader=test_dataloader, metrics=deepinv.metric.PSNR()
+    )
+
+    # in self-supervised cases, remove supervised metrics and compute self-sup losses on eval dataset
+    trainer.metrics = (
```
Andrewwango, line 385: It makes sense, but having to tell the user to remove the metric seems a bit hacky. Not sure what the solution is here. Note that [Metrics already return NaN](https://github.com/deepinv/deepinv/blob/main/deepinv/loss/metric/metric.py#L184) when the data is None, so the trainer could automatically use the self-sup loss if the metric returns NaN. This doesn't account for the case, on the other hand, when there is GT in the dataset but you simply don't want to cheat and look at it for self-sup eval

deepinv/tests/test_trainer.py
```diff
@@ -733,24 +733,36 @@ def forward(self, *args, **kwargs):
         )
 
 
-def test_total_loss(dummy_dataset, imsize, device, dummy_model, tmpdir):
+class ConstantLoss2(ConstantLoss):
+    pass
+
+
```
Andrewwango, line 762: I would like a trainer test using 4 dummy losses that cover the below possibilities:

1. Are evaluated differently at train vs test time
2. Adapt models that behave differently at train vs test time

The test should test that during various stages of training/val/eval, the losses are always evaluated correctly.

deepinv/training/trainer.py
```diff
@@ -116,19 +117,38 @@ class Trainer:
 
     :Evaluation:
 
+    .. note::
+
+        - **Supervised evaluation**: If ground-truth data is available for validation, use any
+          :ref:`full reference metric <full-reference-metrics>`, e.g. ``metrics=dinv.metric.PSNR()``.
+
+        - **Self-supervised evaluation**: If no ground-truth data is available for validation, it is
```
Andrewwango, line 116: Hmm, given my previous comment, I understand why this is good to discourage the user from "cheating" and using the PSNR for validation in the self-sup context. On the hand, sometimes researchers want to perform self-sup training but still compute full-reference metrics on GT (e.g. to compare to supervised learning with the exact same setup). I think we should explain that we can still allow the user to use PSNR for self-sup training in the following scenarios:

- You train using self-sup loss without GT (and we report loss only); you validate/early stop on a small dataset with GT with PSNR. While it's not the pinnacle of ML best practice, this would also prevent breaking current users' workflows.
- You train/validate completely without GT (and early stop on loss etc.); you test with PSNR (e.g. on a challenge dataset). (I think you already allow this, thanks)

The trainer functionality should reflect these possibilities. The examples should also show these possibilities.

In general I lean towards having options for user to continue to pass in metrics for metrics and losses for losses, and we have bools that allow the user to choose different behaviours, and we perform less hidden automatic selection. More bools would pollute more but hopefully all logging can be refactored soon anyway

deepinv/training/trainer.py
```diff
@@ -116,19 +117,38 @@ class Trainer:
...
+          still possible to validate using i) :ref:`no reference metrics
```
Andrewwango: I would add these as a nested list for clarity

deepinv/datasets/datagenerator.py
```diff
@@ -268,8 +268,10 @@ def generate_dataset(
         if os.path.exists(hf_path):
             if overwrite_existing:
                 warn(
-                    f"Dataset {hf_path} already exists, this will overwrite the previous dataset."
+                    f"Dataset {hf_path} already exists, this will close and overwrite the previous dataset."
                 )
+                # remove existing dataset to avoid open file error
+                os.remove(hf_path)
```
Andrewwango, line 274: @jscanvic  does this cause any problems/conflicts/breaking changes?

deepinv/models/gan.py
```diff
@@ -332,6 +333,7 @@ def random_latent(self, device, requires_grad=True):
             - 1
         )
 
+    @torch.enable_grad()
```
Andrewwango, line 336: nice, thanks

Andrewwango (changes requested):
Hi @tachella thanks for pushing these changes! I have some questions and general comments from having reviewed all files except for the meaty `training.py` and `test_trainer.py` files. Happy to discuss offline!

deepinv/datasets/datagenerator.py
jscanvic: Should be okay

deepinv/tests/test_trainer.py
```diff
@@ -733,24 +733,36 @@ def forward(self, *args, **kwargs):
...
     trainer = dinv.Trainer(
         model=dummy_model,
         losses=losses,
+        metrics=metrics,
         epochs=2,
         physics=physics,
+        device=device,
         train_dataloader=dataloader,
         eval_dataloader=eval_dataloader,
+        compute_losses_eval=True,
```
Andrewwango: test true vs false?

deepinv/tests/test_trainer.py
```diff
@@ -759,11 +771,112 @@ def test_total_loss(dummy_dataset, imsize, device, dummy_model, tmpdir):
 
     trainer.train()
 
-    loss_history = trainer.loss_history
-    assert all(
-        [abs(value - sum([l.value for l in losses])) < 1e-6 for value in loss_history]
+    for k, (loss_name, loss_history) in enumerate(trainer.train_loss_history.items()):
+        l = losses[k]
+        assert loss_name == l.__class__.__name__
+        assert all([abs(value - l.value) < 1e-6 for value in loss_history])
```
Andrewwango: torch.allclose instead of manual tolerance?

deepinv/tests/test_trainer.py
```diff
@@ -759,11 +771,112 @@ def test_total_loss(dummy_dataset, imsize, device, dummy_model, tmpdir):
...
+    eval_interval = 2
+    trainer = dinv.Trainer(
+        model=model,
+        device=device,
+        save_path=tmpdir,
+        verbose=False,
+        show_progress_bar=False,
+        physics=physics,
+        epochs=epochs,
+        eval_interval=eval_interval,
```
Andrewwango, line 884: parametrise with this as 1 or 2?

deepinv/tests/test_trainer.py
```diff
@@ -759,11 +771,112 @@ def test_total_loss(dummy_dataset, imsize, device, dummy_model, tmpdir):
...
+    eval_calls = len(eval_dataloader) * (epochs // eval_interval + 1)
+
+    # checking number of train calls
+    if compute_losses_eval:
+        assert model.train_count == train_calls + eval_calls
+    else:
+        assert model.train_count == train_calls
+
+    # checking number of eval calls
+    if not disable_train_metrics and compute_losses_eval:
```
Andrewwango: can you just expand this if into two sub-ifs for clarity?

deepinv/tests/test_trainer.py
```diff
@@ -879,6 +992,72 @@ def test_out_dir_collision_detection(
...
+
+    start = time.time()
+    trainer.train()
+    end = time.time()
+    time_trainer = end - start
+
+    # 30% overhead allowed
+    assert time_trainer / time_naive < 1.3
+
+    assert time_trainer / time_naive < 1.3
```
Andrewwango: repeated line

deepinv/tests/test_trainer.py
```diff
@@ -879,6 +992,72 @@ def test_out_dir_collision_detection(
...
+    end = time.time()
+    time_naive = end - start
+
+    start = time.time()
+    trainer.train()
+    end = time.time()
+    time_trainer = end - start
+
+    # 30% overhead allowed
+    assert time_trainer / time_naive < 1.3
```
Andrewwango: 30% seems like quite a lot! Where is the bulk of this coming from?
Maybe we could also add a metric call to the naive one and parametrize the test with disable_trian_metrics and compute_loss_eval to hopefully reduce the overhead?

deepinv/training/trainer.py
```diff
@@ -102,7 +102,8 @@ class Trainer:
     :param int max_batch_steps: Number of gradient steps per iteration.
         Default is `1e10`. The trainer will perform batch steps equal to the `min(epochs*n_batches, max_batch_steps)`.
     :param None, torch.optim.lr_scheduler.LRScheduler scheduler: Torch scheduler for changing the learning rate across iterations. Default is ``None``.
-    :param bool early_stop: If ``True``, the training stops when the evaluation metric is not improving. Default is ``False``.
+    :param None, int early_stop: If not ``None``, the training stops when the first evaluation metric is not improving
+        after `early_stop` validations. Default is ``None`` (no early stopping).
```
Andrewwango: "validations" = iterations of epochs?

deepinv/training/trainer.py
```diff
@@ -102,7 +102,8 @@ class Trainer:
     :param int max_batch_steps: Number of gradient steps per iteration.
         Default is `1e10`. The trainer will perform batch steps equal to the `min(epochs*n_batches, max_batch_steps)`.
     :param None, torch.optim.lr_scheduler.LRScheduler scheduler: Torch scheduler for changing the learning rate across iterations. Default is ``None``.
-    :param bool early_stop: If ``True``, the training stops when the evaluation metric is not improving. Default is ``False``.
+    :param None, int early_stop: If not ``None``, the training stops when the first evaluation metric is not improving
```
Andrewwango, line 82: I think we should either allow `early_stop = True` where we default to patience=5, or explicitly disallow bools (`True`, `False`) and raise an error - otherwise I'm worried that passing in `True` inadvertently gets casted to `1` epoch

deepinv/training/trainer.py
```diff
@@ -116,19 +117,38 @@ class Trainer:
...
+        - **Supervised evaluation**: If ground-truth data is available for validation, use any
+          :ref:`full reference metric <full-reference-metrics>`, e.g. ``metrics=dinv.metric.PSNR()``.
+
+        - **Self-supervised evaluation**: If no ground-truth data is available for validation, it is
+          still possible to validate using i) :ref:`no reference metrics
+          <no-reference-metrics>`, e.g. ``metrics=dinv.metric.NIQE()``, or ii)
+          :ref:`self-supervised losses <self-supervised-losses>` with
+          ``compute_losses_eval=True`` and ``metrics=None``. If self-supervised losses
+          are used we recommend setting ``disable_train_metrics=True`` to avoid computing
+          metrics in ``model.train()`` mode.
```
Andrewwango: Explain what is wrong with computing metrics in model.train mode, and which losses it might apply to

deepinv/training/trainer.py
tachella: Many thanks for the review!

Answers to some of the points (happy to discuss also some of them later)
- "sometimes researchers want to compute full-reference metrics on GT (e.g. to compare to supervised learning with the exact same setup)." -> we still provide this possibility doing `trainer.test()`
- "You train using self-sup loss without GT (and we report loss only); you validate/early stop on a small dataset with GT with PSNR. While it's not the pinnacle of ML best practice, this would also prevent breaking current users' workflows." I could make one of the examples to do that - not sure it makes sense to show both options in all examples. The current PR keeps the possibility to do supervised eval - just passing a full reference metric with a supervised eval dataset would do the job
- "we have bools that allow the user to choose different behaviours, and we perform less hidden automatic selection" I'm not sure what you mean by automatic hidden selection? Do you mean the choice of early stopping with a loss or a metric? I could add a boolean on that if you think it's necessary (although it will require the user to remember more stuff and apply more settings). Is there other automatic selection that I'm missing?

deepinv/training/trainer.py
```diff
@@ -116,19 +117,38 @@ class Trainer:
...
         see :ref:`datasets user guide <datasets>` for how we expect data to be provided.
-    :param Metric, list[Metric] metrics: Metric or list of metrics used for evaluating the model.
+    :param Metric, list[Metric], None metrics: Metric or list of metrics used for evaluating the model.
         They should have ``reduction=None`` as we perform the averaging using :class:`deepinv.utils.AverageMeter` to deal with uneven batch sizes.
         :ref:`See the libraries' evaluation metrics <metric>`. Default is :class:`PSNR <deepinv.loss.metric.PSNR>`.
-    :param bool disable_train_metrics: if `False` (default), metrics are computed both during training and testing on their respective data.
-        If `True`, do not compute metrics during training on train set. Useful if your model output during training
-        is in the wrong shape for metric computation.
+    :param bool disable_train_metrics: If `True` (default), do not compute metrics during training on train set.
+        If `False`, all metrics are computed both during training on the training dataloader.
```
Andrewwango: both what?

deepinv/training/trainer.py
tachella: do you mean the i) and ii) ? I thought it was taking too much space, and tried to compress it

deepinv/tests/test_loss.py
tachella: I could add a boolean deciding whether to early stop on metrics or losses

deepinv/training/trainer.py
```diff
@@ -116,19 +117,38 @@ class Trainer:
...
+
+        .. warning::
+
+            If `disable_train_metrics=False` the metrics are computed using the model prediction during training (i.e., in `model.train()` mode) to avoid an additional
```
Andrewwango: I think this necessarily begs the question: what if a user wants to have two forward passes per iteration, one in training mode and one in eval to mode to calculate metrics to ensure that training is going well on the train set? I think this is a nicer behaviour when `disable_train_metrics=False` rather than the old way which is technically a bug.

deepinv/training/trainer.py
```diff
@@ -116,19 +117,38 @@ class Trainer:
...
     :param None, torch.utils.data.DataLoader, list[torch.utils.data.DataLoader] eval_dataloader: Evaluation data loader(s),
         see :ref:`datasets user guide <datasets>` for how we expect data to be provided.
-    :param Metric, list[Metric] metrics: Metric or list of metrics used for evaluating the model.
+    :param Metric, list[Metric], None metrics: Metric or list of metrics used for evaluating the model.
         They should have ``reduction=None`` as we perform the averaging using :class:`deepinv.utils.AverageMeter` to deal with uneven batch sizes.
         :ref:`See the libraries' evaluation metrics <metric>`. Default is :class:`PSNR <deepinv.loss.metric.PSNR>`.
-    :param bool disable_train_metrics: if `False` (default), metrics are computed both during training and testing on their respective data.
-        If `True`, do not compute metrics during training on train set. Useful if your model output during training
-        is in the wrong shape for metric computation.
+    :param bool disable_train_metrics: If `True` (default), do not compute metrics during training on train set.
```
Andrewwango: Hmm I'm not sure about changing the default behaviour, our users might wonder why suddenly their training metrics have disappeared from their dashboard. I would say it's fairly standard practice to want to evaluate your model on your training data during training.

Andrewwango: I am also minded to change the name to `enable_train_metrics` to avoid double negatives being confusing

Andrewwango: Finally if we do change the default behaviour we should raise a warning to tell users how to revert back to how it was

deepinv/training/trainer.py
```diff
@@ -116,19 +117,38 @@ class Trainer:
...
+
+            If `disable_train_metrics=False` the metrics are computed using the model prediction during training (i.e., in `model.train()` mode) to avoid an additional
+            forward pass. This can lead to metrics that are different at test time when the model is in `model.eval()` mode.
+
     :param int eval_interval: Number of epochs (or train iters, if ``log_train_batch=True``) between each evaluation of
         the model on the evaluation set. Default is ``1``.
     :param bool log_train_batch: if ``True``, log train batch and eval-set metrics and losses for each train batch during training.
         This is useful for visualising train progress inside an epoch, not just over epochs.
         If ``False`` (default), log average over dataset per epoch (standard training).
+    :param bool compute_losses_eval: If ``True``, the losses are computed during evaluation. Default is ``False``.
```
Andrewwango: Explain why this might be useful and in what scenario the user might want to use this param

deepinv/training/trainer.py
```diff
@@ -223,17 +242,17 @@ class Trainer:
...
     online_measurements: bool = False
     physics_generator: PhysicsGenerator | list[PhysicsGenerator] = None
     loop_random_online_physics: bool = False
     optimizer_step_multi_dataset: bool = True
-    metrics: Metric | list[Metric] = field(default_factory=PSNR)
-    disable_train_metrics: bool = False
-    device: str | torch.device = "cuda" if torch.cuda.is_available() else "cpu"
-    ckpt_pretrained: str | None = None
-    save_path: str | Path = "."
+    metrics: Union[Metric, list[Metric], None] = field(default_factory=PSNR)
```
Andrewwango: in #789 we allowed nice python310 typing, so we can continue using pipe (`|`) notation instead of Union, as long as you import `from __future__ import annotations` at the top

deepinv/training/trainer.py
```diff
@@ -250,7 +269,7 @@ class Trainer:
     plot_measurements: bool = True
     plot_convergence_metrics: bool = False
     rescale_mode: str = "clip"
-    display_losses_eval: bool = False
```
Andrewwango, line 255: I think you would need to deprecate this argument properly i.e. keep it but it does nothing and if a user uses it at all in the init it throws a warning, e.g. using https://github.com/deepinv/deepinv/blob/main/deepinv/utils/decorators.py#L5

deepinv/training/trainer.py
```diff
@@ -280,8 +299,31 @@ def setup_train(self, train=True, **kwargs):
 
         self.save_path = Path(self.save_path) if self.save_path else None
 
+        if train:
+            self.compute_metrics_on_eval_mode = self.disable_train_metrics
+        else:
+            self.compute_metrics_on_eval_mode = True
```
```suggestion
self.compute_metrics_on_eval_mode = self.disable_train_metrics or not train
```
this could be made simpler if we use `enable_train_metrics`:
```suggestion
self.compute_metrics_on_eval_mode = not (self.enable_train_metrics and train)
```

deepinv/training/trainer.py
```diff
@@ -341,9 +383,15 @@ def setup_train(self, train=True, **kwargs):
         for l in self.losses:
             self.model = l.adapt_model(self.model)
 
+        if self.metrics is None:
+            self.metrics = []
+
         if not isinstance(self.metrics, (list, tuple)):
             self.metrics = [self.metrics]
 
+        if len(self.metrics) == 0 and not train:
```
Andrewwango, line 401: I would add a warning here rather than silently changing an attribute that the user might have purposefully specified in the init

deepinv/training/trainer.py
```diff
@@ -687,10 +744,11 @@ def compute_loss(self, physics, x, y, train=True, epoch: int = None, step=False)
         if train and step:
             self.optimizer.zero_grad()
 
-        # Evaluate reconstruction network
-        x_net = self.model_inference(y=y, physics=physics, x=x, train=train)
+        x_net = None
```
Andrewwango: Stylistic choice: I would add this as an else to the if below

deepinv/training/trainer.py
```diff
@@ -775,7 +839,7 @@ def compute_metrics(
                     logs[f"{l.__class__.__name__} no learning"] = (
                         self.logs_metrics_no_learning[k].avg
                     )
-        return logs
+        return x_net, logs
```
Andrewwango, line 896: update the docstring

deepinv/training/trainer.py
```diff
@@ -1288,6 +1383,7 @@ def test(
         save_path: str | Path = None,
         compare_no_learning: bool = True,
         log_raw_metrics: bool = False,
+        metrics: Union[Metric, list[Metric], None] = None,
```
Andrewwango: use new `|` typing

Andrewwango (changes requested):
Hi Julián, some more comments on trainer.py and test_trainer.py

deepinv/training/trainer.py
Andrewwango: I think it's worth being super clear on this nuance

deepinv/training/trainer.py
Andrewwango: > The current PR keeps the possibility to do supervised eval - just passing a full reference metric with a supervised eval dataset would do the job

Great, it would be nice to explain this as a possibility in the docs.

There seems to be many combos on how users might use our Trainer. Maybe it's worth having a clean enumeration of these in the trainer user guide instead of trying to compress it into the docstring

Andrewwango: > Do you mean the choice of early stopping with a loss or a metric? I could add a boolean on that if you think it's necessary (although it will require the user to remember more stuff and apply more settings).

Yes, I mean this. IMO (but I'm happy to be outvoted) a clear transparent interface is more important than having to remember more settings - the latter problem can be alleviated by having clear docs where the user can pick which "workflow" they want to use

deepinv/tests/test_loss.py
Andrewwango: Resolving this issue because the conversation continues in https://github.com/deepinv/deepinv/pull/777#discussion_r2404190832

deepinv/training/trainer.py
tachella: I'll disallow bools

deepinv/training/trainer.py
tachella: done

deepinv/tests/test_trainer.py
tachella: done

deepinv/tests/test_trainer.py
tachella: the issue is that there is a fixed overhead from trainer.setup() which becomes negligible for long train runs (eg several minutes), but then the test becomes too heavy.

deepinv/training/trainer.py
tachella: done

deepinv/training/trainer.py
tachella: done

deepinv/training/trainer.py
tachella: validation epochs

tachella: I tried to clarify writing "passes over the eval dataset"

deepinv/tests/test_trainer.py
tachella: done

deepinv/tests/test_trainer.py
tachella: done

deepinv/training/trainer.py
```diff
@@ -407,10 +450,39 @@ def setup_train(self, train=True, **kwargs):
         ):
             self.physics_generator = [self.physics_generator]
 
-        if train:
-            self.loss_history = []
```
Andrewwango, line 413: I think we need to deprecate this attribute, e.g. I (and I assume other users) will be using this attribute for logging so we should maybe make it a deprecated property and include a warning in its getter to use eval_loss_history

deepinv/training/trainer.py
tachella: I've added defaulting to patience=3 for backwards compatibility when early_stop=True

deepinv/tests/test_trainer.py
tachella: done

deepinv/training/trainer.py
tachella: I've changed my mind and kept the original naming, even if its suboptimal

deepinv/training/trainer.py
```diff
@@ -409,10 +465,38 @@ def setup_train(self, train=True, **kwargs):
...
+                assert (
+                    len(self.losses) > 0
+                ), "At least one loss should be provided for early stopping if early_stop_on_losses=False."
+                assert (
+                    self.compute_eval_losses
+                ), "compute_eval_losses should be True when early_stop_on_losses is True."
+            else:
+                assert (
+                    len(self.metrics) > 0
+                ), "At least one metric should be provided for early stopping if early_stop_on_losses=True."
```
```suggestion
                ), "At least one metric should be provided for early stopping if early_stop_on_losses=False."
```

deepinv/tests/test_trainer.py
```diff
@@ -737,44 +737,198 @@ def forward(self, *args, **kwargs):
...
         losses=losses,
+        metrics=metrics,
         epochs=2,
         physics=physics,
+        device=device,
         train_dataloader=dataloader,
         eval_dataloader=eval_dataloader,
-        optimizer=torch.optim.AdamW(dummy_model.parameters(), lr=1),
+        compute_eval_losses=compute_eval_losses,
+        compute_train_metrics=compute_train_metrics,
```
Andrewwango, line 804: compute_train_metrics behaviour isn't tested in the asserts below

deepinv/tests/test_trainer.py
```diff
@@ -737,44 +737,198 @@ def forward(self, *args, **kwargs):
...
+    eval_calls = len(eval_dataloader) * (
+        (epochs // eval_interval) + (eval_interval - 1)
     )
 
+    # checking number of eval calls
+    assert model.eval_count == eval_calls
+
+    # checking number of train calls
+    if compute_eval_losses:
+        assert model.train_count == train_calls + eval_calls
```
Andrewwango, line 909: nice!!

deepinv/training/trainer.py
```diff
@@ -112,23 +113,57 @@ class Trainer:
...
 
     :Evaluation:
 
+    .. note::
+
+        - **Supervised evaluation**: If ground-truth data is available for validation, use any
+          :ref:`full reference metric <full-reference-metrics>`, e.g. :class:`deepinv.loss.metric.PSNR`.
+
+        - **Self-supervised evaluation**: If no ground-truth data is available for validation, it is
+          still possible to validate using i) :ref:`no reference metrics
```
Andrewwango: I know you want to compress but I prefer clarifying this description of nuanced careful arguments by expanding this as a nested list

deepinv/training/trainer.py
```diff
@@ -112,23 +113,57 @@ class Trainer:
...
     :param None, torch.utils.data.DataLoader, list[torch.utils.data.DataLoader] eval_dataloader: Evaluation data loader(s),
         see :ref:`datasets user guide <datasets>` for how we expect data to be provided.
-    :param Metric, list[Metric] metrics: Metric or list of metrics used for evaluating the model.
+    :param Metric, list[Metric], None metrics: Metric or list of metrics used for evaluating the model.
         They should have ``reduction=None`` as we perform the averaging using :class:`deepinv.utils.AverageMeter` to deal with uneven batch sizes.
         :ref:`See the libraries' evaluation metrics <metric>`. Default is :class:`PSNR <deepinv.loss.metric.PSNR>`.
-    :param bool disable_train_metrics: if `False` (default), metrics are computed both during training and testing on their respective data.
-        If `True`, do not compute metrics during training on train set. Useful if your model output during training
-        is in the wrong shape for metric computation.
+    :param bool compute_train_metrics: If `False` (default), do not compute metrics during training on train set.
```
Andrewwango: - Hmm I would prefer keeping the default backward compatible behaviour of compute_train_metrics=True by default. Otherwise users may wonder why suddenly they cannot see train metrics anymore. Or, add a warning when compute_train_metrics=False.
- Also, explain briefly why you might use compute_train_metrics=False - i.e. it's necessary when model output during training is in the wrong shape for metric computation etc.

deepinv/training/trainer.py
```diff
@@ -112,23 +113,57 @@ class Trainer:
...
+
+            If `compute_train_metrics=True` the metrics are computed using the model prediction during training (i.e., in `model.train()` mode) to avoid an additional
+            forward pass. This can lead to metrics that are different at test time when the model is in `model.eval()` mode.
+
     :param int eval_interval: Number of epochs (or train iters, if ``log_train_batch=True``) between each evaluation of
         the model on the evaluation set. Default is ``1``.
     :param bool log_train_batch: if ``True``, log train batch and eval-set metrics and losses for each train batch during training.
         This is useful for visualising train progress inside an epoch, not just over epochs.
         If ``False`` (default), log average over dataset per epoch (standard training).
+    :param bool compute_eval_losses: If ``True``, the losses are computed during evaluation. Default is ``False``.
```
Andrewwango: Repeat briefly why this may be useful - i.e. useful for self-sup early stopping (any other useful reason?)

deepinv/training/trainer.py
Andrewwango: let's keep the old way

deepinv/training/trainer.py
Andrewwango: Thanks for addressing these points!

Andrewwango:
Note to self: one caveat we are unaddressing for now:
>  user wants to have two forward passes per iteration, one in training mode and one in eval to mode to calculate metrics to ensure that training is going well on the train set

deepinv/training/trainer.py
Andrewwango: readdressing new comment

deepinv/training/trainer.py
Andrewwango: readdressing in new comment

deepinv/training/trainer.py
```diff
@@ -252,7 +286,7 @@ class Trainer:
     plot_measurements: bool = True
     plot_convergence_metrics: bool = False
     rescale_mode: str = "clip"
-    display_losses_eval: bool = False
```
Andrewwango, line 255: I think you would need to deprecate this argument properly i.e. keep it but it does nothing and if a user uses it at all in the init it throws a warning, e.g. using https://github.com/deepinv/deepinv/blob/main/deepinv/utils/decorators.py#L5

deepinv/training/trainer.py
tachella: thanks for spotting this, fixed!

deepinv/training/trainer.py
tachella: yes, indeed. This is useful for self-sup early stopping - or for checking that losses are giving similar values on train set.

deepinv/training/trainer.py
tachella: ok, done

deepinv/training/trainer.py
tachella: I've added an explanation

deepinv/training/trainer.py
tachella: done

deepinv/tests/test_trainer.py
tachella: it's because it doesn't modify the number of train and eval calls, so it is being implicitly tested.

deepinv/training/trainer.py
tachella: done - it was a bit more fiddly than expected since the trainer doesn't have an explicit init

deepinv/training/trainer.py
Andrewwango: thanks

deepinv/training/trainer.py
Andrewwango: Thanks - you can also probably use `__post_init__` which is the standard way to customise dataclass inits

Andrewwango (changes requested):
LGTM thanks!
You could may try using `__post_init__` to tackle the deprecated argument:
```python
class Trainer:
    def __init_subclass__(cls, **kwargs):
        super().__init_subclass__(**kwargs)
        cls.__init__ = _deprecated_alias(display_losses_eval="compute_eval_losses")(cls.__init__)
```

PR #778, opened 2025-09-19
Allow flexible mask shapes in splitting loss + fix splitting losses for multicoil MRI

Andrewwango:
- Allow flexible mask shapes in splitting loss (e.g. for FastMRI data)
- Fix splitting losses for multicoil MRI (i.e. where there is an extra dim in y but not in mask)
- Fix backward pass for least squares when gamma is batched but y.ndim !=x.ndim

### DIFF
--- a/deepinv/loss/mri/measplit.py
+++ b/deepinv/loss/mri/measplit.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 from typing import Optional, Union
+from warnings import warn
 import torch
 from deepinv.physics.forward import Physics
 from deepinv.physics.noise import GaussianNoise
@@ -96,13 +97,20 @@ class WeightedSplittingLoss(SplittingLoss):
 
         def forward(self, y1, y2):
             """Weighted metric forward pass."""
-            
+
             if y1.shape[-2:] != y2.shape[-2]:
                 raise ValueError("Metric input tensors should be same image size.")
-            
+
             if self.weight.shape[1] != y1.shape[1]:
-                self.weight = WeightedSplittingLoss.compute_weight(mask_generator=self.mask_generator, physics_generator=self.physics_generator, img_size=y1.shape[-2:])
-            
+                warn(
+                    "WeightedSplittingLoss detected new y shape in forward pass. Recalculating weight..."
+                )
+                self.weight = WeightedSplittingLoss.compute_weight(
+                    mask_generator=self.mask_generator,
+                    physics_generator=self.physics_generator,
+                    img_size=y1.shape[-2:],
+                )
+
             return self.metric(
                 self.expand(self.weight, y1) * y1, self.expand(self.weight, y2) * y2
             )
@@ -120,13 +128,22 @@ class WeightedSplittingLoss(SplittingLoss):
         self.mask_generator = mask_generator
         self.physics_generator = physics_generator
         self.metric = self.WeightedMetric(
-            self.compute_weight(mask_generator=mask_generator, physics_generator=physics_generator, eps=eps),
-            metric
+            self.compute_weight(
+                mask_generator=mask_generator,
+                physics_generator=physics_generator,
+                eps=eps,
+            ),
+            metric,
         )
         self.normalize_loss = False
 
     @staticmethod
-    def compute_weight(mask_generator: BernoulliSplittingMaskGenerator, physics_generator: BaseMaskGenerator, eps: float = 1e-9, img_size: tuple = None) -> torch.Tensor:
+    def compute_weight(
+        mask_generator: BernoulliSplittingMaskGenerator,
+        physics_generator: BaseMaskGenerator,
+        eps: float = 1e-9,
+        img_size: tuple = None,
+    ) -> torch.Tensor:
         """
         Compute K for K-weighted splitting loss where K is a diagonal matrix of shape (H, W).
 
@@ -160,7 +177,7 @@ class WeightedSplittingLoss(SplittingLoss):
 
         # element-wise multiplication to get K
         k_weight = inv_diag_1_minus_PtP * diag_1_minus_P
-        k_weight = k_weight.unsqueeze(0) # (1, W)
+        k_weight = k_weight.unsqueeze(0)  # (1, W)
 
         # Calculate weight from K
         return (1 - k_weight).clamp(min=eps) ** (-0.5)
--- a/deepinv/loss/mri/measplit.py
+++ b/deepinv/loss/mri/measplit.py
@@ -52,7 +52,6 @@ class WeightedSplittingLoss(SplittingLoss):
 
     :param deepinv.physics.generator.BernoulliSplittingMaskGenerator mask_generator: splitting mask generator for further subsampling.
     :param deepinv.physics.generator.BaseMaskGenerator physics_generator: original mask generator used to generate the measurements.
-    :param float eps: small value to avoid division by zero.
     :param Metric, torch.nn.Module metric: metric used for computing data consistency, which is set as the mean squared error by default.
 
     |sep|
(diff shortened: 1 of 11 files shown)


deepinv/loss/mri/measplit.py
```diff
@@ -70,24 +71,48 @@ class WeightedSplittingLoss(SplittingLoss):
...
 
         def forward(self, y1, y2):
             """Weighted metric forward pass."""
+
+            if y1.shape[-2:] != y2.shape[-2:]:
+                raise ValueError(
+                    f"Metric input tensors should be same image size but got {y1.shape[-2:]}, {y2.shape[-2:]}."
+                )
+
+            if self.weight.shape[-1] != y1.shape[-1]:
```
Andrewwango: this is going to lead to some overhead at train time when you iterate over a dataset with different image sizes. A hacky solution would be to cache weights per width. I think it's ok for now to leave it. At least it's ok for single-image training.

matthieutrs: Agreed.
Another option is to add a shape parameter at init of the loss to force the user acknowledging that shape should always be the same.
(still not super elegant, I don't necessarily think it's better, current solution is good IMO)

deepinv/loss/mri/measplit.py
```diff
@@ -193,14 +237,34 @@ def __init__(
         self.noise_model = noise_model
         self.noise_model.update_parameters(sigma=noise_model.sigma * alpha)
 
+    @staticmethod
+    def expand_mask(mask, y):
+        """Expand mask intermediate dimensions to match those of y, where intermediate
+        dimensions are those (e.g. slice, coils, time) that are not the first two dims (batch, channel),
```
matthieutrs:
```suggestion
        dimensions are those (e.g. depth, coils, time) that are not the first two dims (batch, channel),
```

deepinv/loss/mri/measplit.py
```diff
@@ -193,14 +237,34 @@ def __init__(
...
         recon_loss = super().forward(x_net, y, physics, model, **kwargs)
 
         mask = model.get_mask() * getattr(physics, "mask", 1.0)  # M_\lambda\cap\omega
+
         n2n_metric = self.WeightedMetric(
-            (1 + 1 / (self.alpha**2)) * mask, self.metric.metric, expand=False
+            mask_generator=None,
+            physics_generator=None,
+            weight=(1 + 1 / (self.alpha**2)) * self.expand_mask(mask, y),
+            metric=self.metric.metric,
```
matthieutrs:
```suggestion
            metric=self.metric.metric,
```
unelated to your proposed changes but I am wondering if we should not think of a better naming to avoid `self.metric.metric`

deepinv/physics/generator/inpainting.py
```diff
@@ -91,13 +96,17 @@ def step(
 
         :param int batch_size: batch_size. If None, no batch dimension is created. If input_mask passed and has its own batch dimension > 1, batch_size is ignored.
         :param torch.Tensor, None input_mask: optional mask to be split. If None, all pixels are considered. If not None, only pixels where mask==1 are considered. input_mask shape can optionally include a batch dimension.
+        :param tuple img_size: optionally reset the 2D image size on-the-fly, must be of form (H, W).
```
matthieutrs: I think this sentence could be clarified: what exactly is meant by "reset the 2D image size on the fly"? Does it mean "reset the attribute self.img_size to the img_size value on the fly"? The way it is formulated now here suggests to me that `img_size` should be a bool.
Maybe it would be worth adding an example with this param used around line 40 above?

deepinv/physics/generator/inpainting.py
```diff
@@ -163,13 +174,20 @@ def check_pixelwise(self, input_mask=None) -> bool:
...
 
-    def batch_step(self, input_mask: torch.Tensor = None) -> dict:
+    def batch_step(
+        self, input_mask: torch.Tensor = None, img_size: Optional[tuple] = None
+    ) -> dict:
         r"""
         Create one batch of splitting mask.
 
         :param torch.Tensor, None input_mask: optional mask to be split. If ``None``, all pixels are considered. If not ``None``, only pixels where ``mask==1`` are considered. Batch dimension should not be included in shape.
+        :param tuple img_size: optionally reset the 2D image size on-the-fly, must be of form (H, W).
```
matthieutrs: same as above

deepinv/physics/generator/inpainting.py
```diff
@@ -262,16 +280,26 @@ def __init__(
         )
         self.split_generator = split_generator
 
-    def batch_step(self, input_mask: torch.Tensor = None) -> dict:
-        mask = self.split_generator.step(batch_size=1)["mask"].squeeze(0)
+    def batch_step(
+        self, input_mask: torch.Tensor = None, img_size: Optional[tuple] = None
+    ) -> dict:
+
```
matthieutrs, line 295: add docstrings?

deepinv/physics/generator/inpainting.py
```diff
@@ -474,15 +508,22 @@ def __init__(
...
+        self, input_mask: torch.Tensor = None, img_size: Optional[tuple] = None
+    ) -> dict:
         if len(self.img_size) != 4:
-            raise ValueError("img_size must be of shape (C, T, H, W)")
+            raise ValueError("Default img_size must be of shape (C, T, H, W)")
 
-        if not isinstance(input_mask, torch.Tensor) or input_mask.numel() <= 1:
-            input_mask = torch.ones(self.img_size, device=self.device)
+        if input_mask is not None and input_mask.shape != self.img_size:
+            raise ValueError("input_mask must be same shape as default img_size")
```
matthieutrs, line 543:
```suggestion
        if input_mask is not None:
           if input_mask.shape != self.img_size and img_size is None or img_size!=input_mask.shape:
            raise ValueError(f"input_mask must be same shape as default img_size. Got input_mask.shape {input_mask.shape:2e} and img_size {img_size:2e}.")
```
shouldnt there be a potential check on whether to use img_size or self.img_size?

deepinv/physics/generator/inpainting.py
```diff
@@ -522,7 +563,10 @@ def __init__(
         self.prev_split_size = None
 
     def batch_step(
-        self, input_mask: torch.Tensor = None, persist_prev: bool = False
+        self,
+        input_mask: torch.Tensor = None,
+        img_size: Optional[tuple] = None,
+        persist_prev: bool = False,
```
matthieutrs, line 592: add docstrings

deepinv/tests/test_loss.py
```diff
@@ -460,6 +492,11 @@ def test_measplit(device, loss_name, rng):
             eval_n_samples=eval_n_samples,
         )
     elif loss_name == "splitting_eval_split_input_output":
+        if physics_name == "MultiCoilMRI":
+            pytest.skip(
```
matthieutrs: Is it raising an error in this case? If yes, maybe it'd be good to check that it actually raises an error rather than just skipping (to avoid having something that works but shouldnt)

matthieutrs (changes requested):
Thanks a lot for this PR!! Just a few minor comments, mainly about docs

deepinv/loss/mri/measplit.py
Andrewwango: added some comments to make it clearer what is going on

deepinv/loss/mri/measplit.py
Andrewwango: Yeah to be honest it's not optimal but I can't think of a better solution

deepinv/physics/generator/inpainting.py
Andrewwango: good point, thanks

deepinv/tests/test_loss.py
Andrewwango: upon second thought, the test should be fine - so I removed the skip!

deepinv/physics/generator/inpainting.py
Andrewwango: thanks, I also added a img_size check

deepinv/physics/generator/inpainting.py
Andrewwango: done

deepinv/physics/generator/inpainting.py
Andrewwango: No, because only one of `img_size`  or `input_mask` should be specified, or None. If both are specified, then an error is raised in `step` (which calls `batch_step`)

deepinv/loss/mri/measplit.py
```diff
@@ -69,24 +72,49 @@ class WeightedSplittingLoss(SplittingLoss):
...
         def forward(self, y1, y2):
             """Weighted metric forward pass."""
+
+            if y1.shape[-2:] != y2.shape[-2:]:
+                raise ValueError(
+                    f"Metric input tensors should be same image size but got {y1.shape[-2:]}, {y2.shape[-2:]}."
+                )
+
+            if self.weight.shape[-1] != y1.shape[-1]:
+                # TODO these weights should be cached for a speed-up
```
tachella: this has been done?

deepinv/loss/mri/measplit.py
tachella: I would rename the `metric` method of  `WeightedMetric` as `forward` to avoid this

tachella (approved):
LGTM, minor modifs

deepinv/loss/mri/measplit.py
Andrewwango: renamed to pixel_metric

deepinv/loss/mri/measplit.py
Andrewwango: done

deepinv/optim/utils.py
```diff
@@ -900,11 +900,16 @@ def forward(
 
         # Save tensors only
         gamma_orig_shape = gamma.shape
-        # For broadcasting with other tensors
-        if gamma.ndim > 0:
-            gamma = gamma.view(
-                -1, *[1] * (y.ndim - 1)
-            )  # reshape for broadcasting if needed
+        # For broadcasting with other tensors. Note we already have checked gamma shapes
```
tachella, line 903: looks good to me

PR #781, opened 2025-09-19
Small MCMRI enhancements

Andrewwango:
- Coil map estimation in MCMRI now accelerated via cupy if available and desired
- RSS recon allows you to not reduce over channels (idk why this would be useful but it offers more generality)

### DIFF
--- a/deepinv/physics/mri.py
+++ b/deepinv/physics/mri.py
@@ -419,30 +419,43 @@ class MultiCoilMRI(MRIMixin, LinearPhysics):
                 "sigpy is required to estimate sens maps. Install it using pip install sigpy"
             )
 
-        device = sp.Device(-1)
         if use_cupy:  # pragma: no cover
             try:
-                import cupy
-
-                device = sp.Device(0) if cupy.cuda.is_available() else sp.Device(-1)
+                import cupy as cp
+                use_cupy = cp.cuda.is_available()
             except ImportError:
-                pass
+                warnings.warn(
+                    "cupy is not installed, using cpu for coil map estimation. Install cupy to speed up computation."
+                )
+                use_cupy = False
 
-        maps = np.stack(
-            [
-                EspiritCalib(yb, calib_size, show_pbar=False, device=device).run()
-                for yb in MRIMixin.to_torch_complex(y).numpy()
-            ]
-        )
+        complex_y = MRIMixin.to_torch_complex(y)
         if use_cupy:  # pragma: no cover
-            try:
-                import cupy
+            if y.device.type == "cuda":
+                cupy_y = cp.from_dlpack(complex_y)
+            else:
+                cupy_y = cp.from_dlpack(complex_y.to("cuda"))
 
-                maps = cupy.asnumpy(maps)
-            except (ImportError, ModuleNotFoundError):
-                pass
+            cupy_maps = cp.stack(
+                [
+                    EspiritCalib(yb, calib_size, show_pbar=False, device=cupy_y.device).run()
+                    for yb in cupy_y
+                ]
+            ) 
+            torch_maps = torch.from_dlpack(cupy_maps)
+        
+        else:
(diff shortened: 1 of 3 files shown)


deepinv/physics/mri.py
```diff
@@ -395,32 +395,54 @@ def simulate_birdcage_csm(self, n_coils: int) -> Tensor:
...
         )
+        if use_cupy:  # pragma: no cover
+            try:
+                import cupy
+
+                maps = cupy.asnumpy(maps)
+            except (ImportError, ModuleNotFoundError):
+                pass
+
+        return torch.from_numpy(maps)
```
romainvo: Thanks @Andrewwango for improving the coil map computation ! some comments on the torch/cupy tensor conversion, both libraries implement the `dl_pack` protocol, it allows zero-copy conversion of arrays already stored on a gpu:
- https://docs.cupy.dev/en/latest/reference/generated/cupy.from_dlpack.html#cupy.from_dlpack
- https://docs.pytorch.org/docs/stable/dlpack.html

Using the protocol would prevent unecessary cpu-gpu transfers, i.e. using `.numpy()` if `y` is already on a gpu. It should also preserve the GPU index during conversion, in case of multi-gpu computation. I suggest something like this, with `numpy(force=True)` which is necessary if `use_cupy=False` and `y` is on the gpu nonetheless. What do you think ?

```suggestion
        if use_cupy:  # pragma: no cover
            try:
                import cupy as cp
                use_cupy = cp.cuda.is_available()
            except ImportError:
                warnings.warn(
                    "cupy is not installed, using cpu for coil map estimation. Install cupy to speed up computation."
                )
                use_cupy = False

        complex_y = MRIMixin.to_torch_complex(y)
        if use_cupy:  # pragma: no cover
            if y.device.type == "cuda":
                cupy_y = cp.from_dlpack(complex_y)
            else:
                cupy_y = cp.from_dlpack(complex_y.to("cuda"))

            cupy_maps = cp.stack(
                [
                    EspiritCalib(yb, calib_size, show_pbar=False, device=cupy_y.device).run()
                    for yb in cupy_y
                ]
            ) 
            torch_maps = torch.from_dlpack(cupy_maps)
        
        else:
            device = sp.Device(-1)
            maps = np.stack(
                [
                    EspiritCalib(yb, calib_size, show_pbar=False, device=device).run()
                    for yb in complex_y.numpy(force=True)
                ]
            )
            
            torch_maps = torch.from_numpy(maps)

        return torch_maps
```

Andrewwango: Thanks @romainvo  so much for this suggestion, I like it! One question: why do we use `.numpy(force=True)` rather than `.detach().cpu().numpy()` which is the convention elsewhere in deepinverse?

romainvo: It's a shorthand for `.detach().cpu().resolve_conj().resolve_neg().numpy()`, basically it accounts for all the cases where you can't call `.numpy()` directly. In this case both `.numpy(force=True)` and `.detach().cpu().numpy()` do exactly the same, the former is just more handy to write !

deepinv/physics/mri.py
```diff
@@ -395,32 +395,70 @@ def simulate_birdcage_csm(self, n_coils: int) -> Tensor:
...
 
-        return torch.from_numpy(
-            np.stack(
+        if use_cupy:  # pragma: no cover
+            try:
+                import cupy as cp
+
+                use_cupy = cp.cuda.is_available()
+            except ImportError:
+                warnings.warn(
```
romainvo: I forgot to add the `import warnings`  in the suggested code, otherwise it throws a NameError when the warning is raised.

Andrewwango: done! thanks

jscanvic (approved):
LGTM! Thanks Andrew for the PR and @romainvo for the review!

Happy to merge once the merge conflict is solved and CI passes

PR #782, opened 2025-09-19
fix plot

### LINKED ISSUE #780: trainer plot order
### Environment

deepinv main

### Describe the bug

The behaviour of the deepinv plot function was changed in #684 with now rows and columns of the grid being swapped.

This is what we currently observe when training:
Image

Instead of 
Image

I suspect this being due to the unclear naming convention of torchvision wrt the `n_rows` vairable in the [make_grid](https://docs.pytorch.org/vision/main/generated/torchvision.utils.make_grid.html) function with final shape (batch_size/n_rows, n_rows)...

I'll propose a fix shortly.

### Steps to reproduce

```python3
import deepinv as dinv
a = torch.randn((32, 3, 10, 10))
from deepinv.utils.plotting import prepare_images
out = prepare_images(x=a, y=a, x_net=a, rescale_mode="min_max")
dinv.utils.plot([out[2].unsqueeze(0)])
print(out[2].shape)
> torch.Size([3, 386, 38])
```

### Expected behavior

```python3
print(out[2].shape)
> torch.Size([3, 38, 386])
```

_No response_

### Additional context

_No response_

matthieutrs:
Fixes #780 and adds a test to check that plots have correct shape

### DIFF
--- a/deepinv/tests/test_utils.py
+++ b/deepinv/tests/test_utils.py
@@ -900,6 +900,30 @@ def test_prepare_images(x, y, x_net, x_nl, rescale_mode):
         assert all(
             isinstance(title, str) for title in titles
         ), "All titles should be strings."
+        assert len(imgs) == len(
+            titles
+        ), "Number of images should match number of titles."
+        assert len(imgs) == sum(v is not None for v in [x, y, x_net, x_nl])
+
+
+@pytest.mark.parametrize(
+    "x",
+    [
+        torch.randn((32, 1, 10, 10)),
+        torch.randn((32, 2, 10, 10)),
+        torch.randn((32, 3, 10, 10)),
+    ],
+)
+def test_prepare_images_shapes(x):
+    imgs, titles, grid_image, caption = deepinv.utils.plotting.prepare_images(
+        x, y=x, x_net=x, x_nl=x, rescale_mode="min_max"
+    )
+
+    # Check that the grid image has the correct shape (4 below = number of non None inputs)
+    num_inputs = 4  # x, y, x_net, x_nl
+    assert grid_image.shape == torch.Size(
+        [3, num_inputs * (x.shape[-1] + 2) + 2, x.shape[0] * (x.shape[-2] + 2) + 2]
+    )
 
 
 # Module-level fixtures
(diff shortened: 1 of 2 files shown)


deepinv/tests/test_utils.py
```diff
@@ -900,6 +900,30 @@ def test_prepare_images(x, y, x_net, x_nl, rescale_mode):
...
             isinstance(title, str) for title in titles
         ), "All titles should be strings."
+        assert len(imgs) == len(
+            titles
+        ), "Number of images should match number of titles."
+        assert len(imgs) == sum(v is not None for v in [x, y, x_net, x_nl])
+
+
+@pytest.mark.parametrize(
+    "x",
```
jscanvic: Can you instead parametrize by seed (e.g., 0, 1, 2), and set `x = torch.randn(32, 1, 10, 10, generator=rng)` with `rng = torch.Generator().manual_seed(seed)`?

It makes the test deterministic & makes intent clearer imo!

deepinv/tests/test_utils.py
```diff
@@ -900,6 +900,30 @@ def test_prepare_images(x, y, x_net, x_nl, rescale_mode):
         assert all(
             isinstance(title, str) for title in titles
         ), "All titles should be strings."
+        assert len(imgs) == len(
+            titles
+        ), "Number of images should match number of titles."
+        assert len(imgs) == sum(v is not None for v in [x, y, x_net, x_nl])
```
jscanvic: It fails because the test to include `y` is not just `y is not None`, but: `y is not None and x is not None and y.shape == x.shape`.

An ugly but working fix (feel free to improve it!):

```suggestion
        assert len(imgs) == sum(v is not None for v in [x, x_net, x_nl]) + int(
            y is not None and x is not None and y.shape == x.shape
        )
```

jscanvic (changes requested):
Really neat fix, thanks Matthieu! Minor comments about the tests & I'll be happy to approve

deepinv/tests/test_utils.py
jscanvic: Also maybe add an assert message if you can think of a good one! Happy to leave it that way if not

jscanvic (approved):
LGTM! Thanks Matthieu

Issue #770, opened 2025-09-13
dirac_like device


tachella:
### Environment

any

### Describe the bug

The `dirac_like(x)` [function](https://deepinv.github.io/deepinv/api/stubs/deepinv.utils.dirac_like.html#deepinv.utils.dirac_like) should return a dirac in the same device as `x`, and currently it always returns a tensor on CPU.

### Steps to reproduce

import deepinv as dinv
import torch
x = torch.ones((1, 1, 32, 32), device='cuda')
print(dinv.utils.dirac_like(x).device)

returns 'cpu'

### Expected behavior

it should return 'cuda'

PR #785, opened 2025-09-21
Fix least squares solvers

tachella:
[least_squares](https://deepinv.github.io/deepinv/api/stubs/deepinv.optim.utils.least_squares.html#deepinv.optim.utils.least_squares) solvers have an optional regularization parameter `gamma` that can be either None, a scalar, or a tensor. The tensor version is used in cases where the regularization is different across elements in a batch. There are multiple checks on gamma, such as [this one](https://github.com/deepinv/deepinv/blob/6d99da8d90f6ab0216b36b460415faa7677044be/deepinv/optim/utils.py#L171), which assume that `gamma` is a scalar and generate an error if it is not. This is particularly problematic since it breaks `RAM` and other solvers when deployed on a batch.

Solves #786  and #770 

### DIFF
--- a/deepinv/models/ram.py
+++ b/deepinv/models/ram.py
@@ -247,8 +247,8 @@ class RAM(Reconstructor, Denoiser):
 
         .. note::
 
-            The noise levels `sigma` and `gain` can be kept as `None` if a noise model with :class:`deepinv.physics.GaussianNoise`,
-            :class:`deepinv.physics.PoissonNoise` or :class:`deepinv.physics.PoissonGaussianNoise`
+            The noise levels `sigma` and `gain` can be kept as `None` if a noise model with :class:`GaussianNoise <deepinv.physics.GaussianNoise>`,
+            :class:`PoissonNoise <deepinv.physics.PoissonNoise>` or :class:`PoissonGaussianNoise <deepinv.physics.PoissonGaussianNoise>`
             is specified in the physics. If both are provided, the `sigma` and `gain` values provided to the model will be used.
 
         :param torch.Tensor y: measurements
--- a/deepinv/models/ram.py
+++ b/deepinv/models/ram.py
@@ -263,10 +263,11 @@ class RAM(Reconstructor, Denoiser):
             )
 
         if isinstance(y, TensorList):
-            max_val = y[0].abs().amax(dim=(1, 2, 3), keepdim=False)
+            max_val = y[0].abs().reshape(y[0].size(0), -1).amax(dim=1, keepdim=False)
         else:
-            max_val = y.abs().amax(dim=(1, 2, 3), keepdim=False)
+            max_val = y.abs().reshape(y.size(0), -1).amax(dim=1, keepdim=False)
 
+        # rescale elements in the batch where max_val > 5 * sigma_threshold
         rescale_val = torch.where(
             max_val > 5 * self.sigma_threshold,
             torch.tensor(1.0, device=max_val.device, dtype=max_val.dtype),
@@ -285,7 +286,11 @@ class RAM(Reconstructor, Denoiser):
         x_temp = physics.A_adjoint(y)
 
         sigma, gain = self.obtain_sigma_gain(
-            physics, sigma, gain, rescale_val, device=y.device
+            physics=physics,
+            sigma=sigma,
+            gain=gain,
+            rescale_val=rescale_val,
+            device=y.device,
         )
 
         pad = (-x_temp.size(-2) % 8, -x_temp.size(-1) % 8)
--- a/deepinv/models/ram.py
+++ b/deepinv/models/ram.py
@@ -1,6 +1,6 @@
 from collections import OrderedDict
 from pathlib import Path
-
+from warnings import warn
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
@@ -340,6 +340,10 @@ class RAM(Reconstructor, Denoiser):
                 sigma = sigma.abs().max()
         else:
             sigma = sigma / rescale_val
+            if hasattr(physics.noise_model, "sigma"):
+                warn(
+                    "Both sigma provided to the model and a noise model in the physics. The sigma provided to the model will be used."
+                )
 
         if gain is None:
             gain = (
@@ -352,6 +356,10 @@ class RAM(Reconstructor, Denoiser):
                 gain = gain.abs().max()
         else:
             gain = gain / rescale_val
+            if hasattr(physics.noise_model, "gain"):
+                warn(
+                    "Both gain provided to the model and a noise model in the physics. The gain provided to the model will be used."
+                )
 
         if not isinstance(sigma, torch.Tensor) and not isinstance(sigma, TensorList):
             sigma = torch.tensor(sigma, device=device)
--- a/deepinv/models/ram.py
+++ b/deepinv/models/ram.py
@@ -175,6 +175,8 @@ class RAM(Reconstructor, Denoiser):
         snr = num / (sigma + 1e-4)  # SNR equivariant
         gamma = 1 / (1e-4 + 1 / (snr * f**2))
         gamma = gamma[(...,) + (None,) * (x.dim() - 1)]
+        gamma = gamma * self.fact_realign
+        gamma = gamma.clamp(min=1e-8)  # clamp to avoid negative or zero gamma
         model_input = physics.prox_l2(x, y, gamma=gamma * self.fact_realign)
 
         return model_input
--- a/deepinv/models/ram.py
+++ b/deepinv/models/ram.py
@@ -247,9 +247,9 @@ class RAM(Reconstructor, Denoiser):
 
         .. note::
 
-            The noise levels `(sigma and gain)` can be kept as None if a :class:`deepinv.physics.GaussianNoise`
+            The noise levels `sigma` and `gain` can be kept as `None` if a :class:`deepinv.physics.GaussianNoise`,
             :class:`deepinv.physics.PoissonNoise` or :class:`deepinv.physics.PoissonGaussianNoise` noise model with
-            is specified in the physics. If both are provided, the `(sigma, gain)` values provided to the model will be used.
+            is specified in the physics. If both are provided, the `sigma` and `gain` values provided to the model will be used.
 
         :param torch.Tensor y: measurements
         :param deepinv.physics.Physics physics: forward operator
--- a/deepinv/models/ram.py
+++ b/deepinv/models/ram.py
@@ -247,8 +247,8 @@ class RAM(Reconstructor, Denoiser):
 
         .. note::
 
-            The noise levels `sigma` and `gain` can be kept as `None` if a :class:`deepinv.physics.GaussianNoise`,
-            :class:`deepinv.physics.PoissonNoise` or :class:`deepinv.physics.PoissonGaussianNoise` noise model with
+            The noise levels `sigma` and `gain` can be kept as `None` if a noise model with :class:`deepinv.physics.GaussianNoise`,
+            :class:`deepinv.physics.PoissonNoise` or :class:`deepinv.physics.PoissonGaussianNoise`
             is specified in the physics. If both are provided, the `sigma` and `gain` values provided to the model will be used.
 
         :param torch.Tensor y: measurements
--- a/deepinv/models/ram.py
+++ b/deepinv/models/ram.py
@@ -262,7 +262,10 @@ class RAM(Reconstructor, Denoiser):
                 "Either physics, sigma or gain must be provided to the RAM model."
             )
 
-        max_val = y.abs().amax(dim=(1, 2, 3), keepdim=False)
+        if isinstance(y, TensorList):
+            max_val = y[0].abs().amax(dim=(1, 2, 3), keepdim=False)
+        else:
+            max_val = y.abs().amax(dim=(1, 2, 3), keepdim=False)
 
         rescale_val = torch.where(
             max_val > 5 * self.sigma_threshold,
@@ -270,11 +273,16 @@ class RAM(Reconstructor, Denoiser):
             max_val,
         )
 
+        if isinstance(y, TensorList):
+            for yi in y:
+                yi /= rescale_val.view([yi.shape[0]] + [1] * (yi.ndim - 1))
+        else:
+            y = y / rescale_val.view([y.shape[0]] + [1] * (y.ndim - 1))
+
         if physics is None:
             physics = dinv.physics.Denoising(noise_model=dinv.physics.ZeroNoise())
 
         x_temp = physics.A_adjoint(y)
-        y = y / rescale_val.view([y.shape[0]] + [1] * (y.ndim - 1))
 
         sigma, gain = self.obtain_sigma_gain(
             physics, sigma, gain, rescale_val, device=y.device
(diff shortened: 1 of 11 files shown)


deepinv/optim/utils.py
```diff
@@ -90,11 +91,24 @@ def least_squares(
...
         parallel_dim = [parallel_dim]
+
+    zero_gamma = False
     if gamma is None:
         gamma = 0.0
+        zero_gamma = True
+    elif isinstance(gamma, (int, float)):
+        zero_gamma = not (gamma > 0)
+    elif isinstance(gamma, Tensor):
+        zero_gamma = not torch.any(gamma > 0)
```
Andrewwango: I would use torch.all

deepinv/optim/utils.py
```diff
@@ -90,11 +91,24 @@ def least_squares(
...
+        if not zero_gamma:
+            gamma[gamma <= 0] = 1e8  # guardrails
```
Andrewwango: If I understand correctly, gamma is changed on the fly where it's not positive.

- Why doesn't this modification occur for a single gamma int/float too? should it be set to eps?
- Does it make sense that `gamma = [-1, 1e-10]` -> `[1e-8, 1e-10]`? surely the threshold should be set at eps

deepinv/optim/utils.py
```diff
@@ -90,11 +91,24 @@ def least_squares(
...
     if isinstance(parallel_dim, int):
         parallel_dim = [parallel_dim]
+
+    zero_gamma = False
     if gamma is None:
         gamma = 0.0
+        zero_gamma = True
+    elif isinstance(gamma, (int, float)):
+        zero_gamma = not (gamma > 0)
+    elif isinstance(gamma, Tensor):
```
Andrewwango: can we allow gamma to be a list too?

deepinv/optim/utils.py
```diff
@@ -90,11 +91,24 @@ def least_squares(
     """
     if isinstance(parallel_dim, int):
         parallel_dim = [parallel_dim]
+
+    zero_gamma = False
     if gamma is None:
         gamma = 0.0
+        zero_gamma = True
+    elif isinstance(gamma, (int, float)):
+        zero_gamma = not (gamma > 0)
```
Andrewwango: gamma <= 0 would be clearer I think

Andrewwango:
should we also add a test that RAM can work with batch size > 1?

deepinv/optim/utils.py
tachella: the solution is different if `gamma=0` or if `gamma != 0`. If gamma is a scalar, this is easy to check. If `gamma` is not a scalar, the solver can only work if all `gamma=0` or if all `gamma != 0`. So we shouldn't threshold by default, just handle the (rare) case of a batch of gamma where some are zero and some are not

deepinv/optim/utils.py
tachella: I don't see the application

tachella:
> should we also add a test that RAM can work with batch size > 1?

done!

deepinv/tests/test_optim.py
```diff
@@ -999,7 +1005,7 @@ def A_adjoint(self, y, **kwargs):
 @pytest.mark.parametrize("solver", solvers)
 def test_least_squares_implicit_backward(device, solver, physics_name):
     # Check that the backward gradient matches the finite difference gradient
-    batch_size = 1
+    batch_size = 2  # TODO: test batching
```
Andrewwango: add as param and do batch size 1 and 2?

deepinv/tests/test_optim.py
```diff
@@ -1016,7 +1022,9 @@ def test_least_squares_implicit_backward(device, solver, physics_name):
     # Check gradients w.r.t y, z, gamma
     y.requires_grad_(True)
     z = torch.randn_like(x).requires_grad_(True)
-    gamma = torch.tensor(0.4, dtype=dtype, requires_grad=True)
+    gamma = (
```
Andrewwango, line 1025: I think it would be good to test different gammas since the logic is quite nuanced + can change the function a lot!

deepinv/tests/test_models.py
```diff
@@ -906,7 +906,10 @@ def test_restoration_models(
         if physics is not None:
             physics.noise_model = dinv.physics.GaussianNoise(0.01, rng=rng)
 
-    x = DummyCircles(imsize=imsize, samples=1)[0].unsqueeze(0)
+    x = DummyCircles(imsize=imsize, samples=1)[0].unsqueeze(0).to(device)
+
+    # make batch with > 1 element to test batch processing
+    x = x.repeat(2, 1, 1, 1)
```
Andrewwango: can we test with two different images in a batch, and test that both individually are restored? to make sure the images in a batch don't affect each other

deepinv/models/ram.py
```diff
@@ -175,6 +187,8 @@ def realign_input(self, x, physics, y, sigma):
         snr = num / (sigma + 1e-4)  # SNR equivariant
         gamma = 1 / (1e-4 + 1 / (snr * f**2))
         gamma = gamma[(...,) + (None,) * (x.dim() - 1)]
+        gamma = gamma * self.fact_realign
```
Andrewwango, line 189: Given how brittle this was and how long it took you to find that Hai's PR had broken RAM, I think we should do more solid unit tests for RAM e.g. for different gammas, differently scaled inputs (e.g. butterfly / 1000, butterfly * 1000) to catch big things that might break it

deepinv/models/ram.py
```diff
@@ -37,6 +37,18 @@ class RAM(Reconstructor, Denoiser):
...
     :param bool, str pretrained: If `True`, the model will be initialized with pretrained weights. If `str`, load from file.
     :param float sigma_threshold: Threshold (minimum value) for the noise level. Default is 1e-3.
+
+    |sep|
+
+      >>> import deepinv as dinv
+      >>> x = dinv.utils.load_example("butterfly.png")
+      >>> physics = dinv.physics.Downsampling(filter="bicubic")
+      >>> y = physics(x)
+      >>> model = dinv.models.RAM(pretrained=True) # or any of the models listed below
```
Andrewwango:
```suggestion
      >>> model = dinv.models.RAM(pretrained=True)
```

deepinv/optim/utils.py
```diff
@@ -78,7 +87,7 @@ def least_squares(
     :param torch.Tensor y: input tensor of shape (B, ...)
     :param torch.Tensor z: input tensor of shape (B, ...) or scalar.
     :param torch.Tensor init: (Optional) initial guess for the solver. If None, it is set to a tensor of zeros.
-    :param None, float gamma: (Optional) inverse regularization parameter.
+    :param None, float, torch.Tensor gamma: (Optional) inverse regularization parameter. Can be batched (shape (B, ...)) or a scalar. If None, it is set to :math:`\infty` (no regularization).
     :param str solver: solver to be used.
```
Andrewwango: list different solvers or ref where these solves are listed?

deepinv/optim/utils.py
```diff
@@ -90,11 +99,39 @@ def least_squares(
     """
     if isinstance(parallel_dim, int):
         parallel_dim = [parallel_dim]
+
     if gamma is None:
-        gamma = 0.0
+        gamma = torch.tensor(0.0, device=y.device)
+        no_gamma = True
```
Andrewwango: could you rename this like `gamma_provided=False` or something to avoid double negatives?

deepinv/optim/utils.py
```diff
@@ -90,11 +99,39 @@ def least_squares(
...
+
+        if torch.any(gamma <= 0):
+            warnings.warn(
+                "Regularization parameter of least squares problem (gamma) should be positive."
+                "Otherwise, the problem can become non-convex and the solvers are not designed for that."
+                "Continuing anyway..."
+            )
+    if gamma.ndim > 0:  # if batched gamma
+        if isinstance(y, TensorList):
+            batch_size = y[0].size(0)
```
Andrewwango, line 120: maybe use the `shape` property of tensorlist? i.e. `y.shape[0][0]` - do you think in tensorlist we should enforce that all tensors inside the tensorlist have the same batch size?

deepinv/optim/utils.py
```diff
@@ -370,6 +407,7 @@ def _sym_ortho(a, b):
     ``1/eps`` in some important places.
 
     """
+    a, b = torch.broadcast_tensors(a, b)
```
Andrewwango, line 410: nice, I have never ever heard of broadcast_tensors before haha

deepinv/optim/utils.py
tachella: I think size is better as it returns a Tensor - and yes we could enforce this potentially, but I would leave it for other PR

deepinv/optim/utils.py
tachella: done

deepinv/optim/utils.py
tachella: done

deepinv/tests/test_optim.py
tachella: I'm randomizing to make sure that it works

deepinv/tests/test_models.py
tachella: done

deepinv/tests/test_optim.py
mh-nguyen712: This test takes quite long time to finish so we risk to increase the test time...

deepinv/models/ram.py
tachella: I've added `test_ram_scale` to check for this, then I found many bugs with RAM which I fixed

deepinv/tests/test_optim.py
tachella: I'm only testing batch size 2 for speed, but I left the option with a `parametrize`

deepinv/models/ram.py
```diff
@@ -246,33 +245,47 @@ def forward(self, y, physics=None, sigma=None, gain=None):
...
                 "Either physics, sigma or gain must be provided to the RAM model."
             )
 
-        if physics is None:
-            gain = self.gain_threshold if gain is None else gain
-            sigma = self.sigma_threshold if sigma is None else sigma
+        if isinstance(y, TensorList):
+            max_val = y[0].abs().amax(dim=(1, 2, 3), keepdim=False)
+        else:
+            max_val = y.abs().amax(dim=(1, 2, 3), keepdim=False)
```
Andrewwango: why `dim=(1,2,3)`? do you mean reducing over all dims that aren't the batch dim? Then you should do between 1 and ndim. `(1,2,3)` will fail for e.g. multicoil MRI data of shape `BCNHW`

Andrewwango: secondly, what is the reasoning behind switching from max_val calculated on ATy to calculated on y?

deepinv/models/ram.py
```diff
@@ -246,33 +245,47 @@ def forward(self, y, physics=None, sigma=None, gain=None):
...
 
-            physics = dinv.physics.Denoising(
-                noise_model=dinv.physics.PoissonGaussianNoise(sigma=sigma, gain=gain),
-            )
+        rescale_val = torch.where(
```
Andrewwango, line 271: I think you will need to add an inline comment here, as it is not immediately obvious what is going on

deepinv/models/ram.py
```diff
@@ -246,33 +245,47 @@ def forward(self, y, physics=None, sigma=None, gain=None):
...
+        else:
+            y = y / rescale_val.view([y.shape[0]] + [1] * (y.ndim - 1))
 
-        max_val = x_temp.abs().max()
-        rescale_val = 1.0 if max_val > 5 * self.sigma_threshold else max_val
-        y = y / rescale_val
+        if physics is None:
+            physics = dinv.physics.Denoising(noise_model=dinv.physics.ZeroNoise())
+
+        x_temp = physics.A_adjoint(y)
```
Andrewwango, line 286: What is the purpose of this x_temp? Can we remove this and use x_in below?

deepinv/models/ram.py
```diff
@@ -246,33 +245,47 @@ def forward(self, y, physics=None, sigma=None, gain=None):
...
 
         sigma, gain = self.obtain_sigma_gain(
-            physics=physics, y=y, sigma=sigma, gain=gain, rescale_val=rescale_val
+            physics, sigma, gain, rescale_val, device=y.device
```
Andrewwango: Why removing the kwargs? it's safer

deepinv/models/ram.py
```diff
@@ -292,54 +305,53 @@ def forward(self, y, physics=None, sigma=None, gain=None):
...
         :param float, torch.Tensor sigma: Gaussian noise level. If None, will be set to the threshold.
         :param float, torch.Tensor gain: Poisson noise level. If None, will be set to the threshold.
-        :param float rescale_val: Rescale value to apply to the sigma and gain values.
+        :param torch.Tensor rescale_val: Rescale value to apply to the sigma and gain values.
+        :param str, torch.device device: Device to which the sigma and gain values should be moved.
         """
-        if hasattr(physics, "noise_model"):
-            if sigma is not None or gain is not None:
-                warn(
-                    "noise_model specified in physics. Parameters passed to sigma or gain will be ignored."
```
Andrewwango, line 303: Surely we will need to keep this warning somewhere, when a user does `model(y, Denoising(GaussianNoise(0.1)), sigma=0.2)`?

deepinv/tests/test_models.py
```diff
@@ -856,6 +856,55 @@ def __len__(self):
     assert psnr(x_init, x) < psnr(x_hat, x)
 
 
+@pytest.mark.parametrize("use_physics", [True, False])
+@pytest.mark.parametrize("scale", [-5, 5])
```
Andrewwango: minor point, I would use 1e-5, 1e5 here to avoid any confusion

deepinv/models/ram.py
tachella: the idea is that noise levels are defined on the measurement space and not on the signal space, so scaling should happen on the measurement space

deepinv/models/ram.py
tachella: If the user specifies a sigma, I believe it is because they want to use that sigma (and not the sigma hidden in the physics). I think that the previous behavior, i.e. ignoring the user-specified sigma, was more confusing.

deepinv/models/ram.py
tachella: unfortunately we cannot get rid of it, since we need it to guess the size of the image to then use `PhysicsCropper`

deepinv/tests/test_models.py
tachella: done

deepinv/models/ram.py
tachella: I added the kwargs back

deepinv/models/ram.py
tachella: I've added a comment - the goal is to rescale only the batch elements that have very large/small values

deepinv/models/ram.py
tachella: fixed the dim issue

deepinv/models/ram.py
Andrewwango: Fair enough but I would add a warning in case

tachella: Ok, I've added a warning

PR #789, opened 2025-09-27
Move to Python 3.10+ type syntax (without breaking Python 3.9!)

jscanvic:
In #706 we got rid of the Python 3.10+ type syntax `T1 | T2 := Union[T1, T2]` in the code base after we realized it broke compatibility with Python 3.9 entirely (i.e., we couldn't even import the library).

As it turns out, it is not actually mandatory to avoid using this syntax to support Python 3.9 using `import __future__.annotations`. When enabled, the annotations are not evaluated at runtime which prevents failures for Python 3.9. This is a somewhat standard way of dealing with this problem as evidenced by available ruff rules to enable that (FA100, FA102).

In this PR, I make the new type syntax mandatory with additional ruff checks to make sure `import __future__.annotations` are present where necessary.

I also change the configuration for `vermin` to avoid false positives, for more details see: https://github.com/netromdk/vermin?tab=readme-ov-file#caveats

### DIFF
--- a/deepinv/training/trainer.py
+++ b/deepinv/training/trainer.py
@@ -220,9 +220,7 @@ class Trainer:
     train_dataloader: torch.utils.data.DataLoader
     epochs: int = 100
     max_batch_steps: int = 10**10
-    losses: Loss | BaseLossScheduler | list[Loss] | list[BaseLossScheduler] = (
-        SupLoss()
-    )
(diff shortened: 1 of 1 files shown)


Andrewwango (approved):
Looks good, thanks! Maybe drop a message in the discord contributing channel just to say that our typing follows this now? Just a thought - feel free to ignore! Thanks

PR #792, opened 2025-09-30
Fix least_squares gamma batching

Andrewwango:
Gamma in least_squares proposed in #785 was being reshaped to y rather than ATy. This breaks for when they are different shape e.g. in multicoil MRI. This fix forces least squares to not reshape when gamma is already shaped (e.g. as RAM already shapes it), otherwise reshape to ATy.


tachella (approved):
thanks for fixing this bug!

PR #797, opened 2025-09-30
Introduce mypy for static type checking

jscanvic:
Towards #536

A linter like `ruff` catches common programming mistakes statically using simple heuristics related to the way the code looks, it is however not able to make sense of the actual semantics of the code: what arguments does a function take? does it make sense to treat its output as a float? as a tensor? This is where a static type checker like `mypy` comes in: it helps catch more bugs and it also encourages clearer code through the use of type hints.

In this PR, I add mypy to the CI and enable a subset of the rules.


jscanvic:
Closing because it gives too many false positives - might reopen if I find a solution

Issue #804, opened 2025-10-01
Add 3D support from IO and model perspective


nucli-vicky:
### Describe the feature

Enable the current suite of methods in DeepInv to be seamlessly applied on 3D imaging data as well.

- [x] Models that do not rely on pretrained components have a `dim` keyword specifying whether the 2D or 3D variant is to be constructed #869 

- [x] Commonly used 3D imaging formats (nifti, npy, blosc2, ...) are supported from an IO standpoint. Partial loading and an associated dataset class to load smaller patches from volumes is introduced. #806 

- [ ] These tools are demonstrated on inverse problems / image restoration tasks where measurement and ground-truth space is shared (e.g. the problem can be viewed as image translation). For accessibility, this requires a dataset which can be freely downloaded (no filling in agreements / requesting access) and optionally hosted (partially) on the DeepInv HF page.

- [ ] The trainer supports logging of 3D data in the form of e.g. saving several slices from a patch.

- [ ] 3D volumes can often not be processed in a single network pass. For eval/test, a sliding window utility could be implemented.

- [ ] As a nice example for the DeepInv docs, a comparison is made between a method / model acting once on 2D data and once on 3D patches

### Motivation

DeepInv offers an extensive suite of methods, tools, and models. Currently applying them on 3D images is less out-of-the-box supported compared to 2D images. 

Improving the support for 3D data has three main benefits:

- It can increase the adoption of DeepInv by the medical imaging community (+ other communities working on 3D data)

- 3D imaging restoration and inverse problems have extensive real-world use cases and it would be good to increase its progress.

- Due to easier integration and more widespread adoption, papers proposing new methods often remain constrained to 2D data, and often just natural images. Improving 3D support can encourage/enable researchers to validate methods on more diverse modalities.

### Additional context

One dataset candidate is [the IXI MRI dataset](https://brain-development.org/ixi-dataset/). It is not specific to inverse problems or restoration, but can allow experimentation with synthetic noise, degradation, and downsampling models, similar to how things are often done in natural imaging. The license seems permissive?

Andrewwango:
Thanks for creating the detailed issue @nucli-vicky ! This is a great set of desired features for 3D support!

I would add:

- [ ] Demo for 3D visualisation with `plot_ortho3D`, `plot_videos`, and a new `plot_3D` function using `napari`:

```
>>> pip install "napari[all]"`
import napari
viewer, image_layer = napari.imshow(x.squeeze().numpy())
[...]

PR #805, opened 2025-10-01
Add Rician noise

nucli-vicky:
Implements the Rician noise model, used often in MRI. See e.g. https://users.phhp.ufl.edu/pqiu/research/rician.pdf

Rician noise is nice to have, because it respects the fact that pixel intensities cannot be negative. This is important for several fields where negative values are not physically possible.

Additionally, nice to have for the IXI dataset, #804 

### DIFF
--- a/deepinv/physics/phase_retrieval.py
+++ b/deepinv/physics/phase_retrieval.py
@@ -1,7 +1,9 @@
+from __future__ import annotations
 from functools import partial
 import math
 import torch
 import numpy as np
+from typing import Any
 from deepinv.optim.phase_retrieval import spectral_methods
 from deepinv.physics.compressed_sensing import CompressedSensing
 from deepinv.physics.forward import Physics, LinearPhysics
@@ -360,7 +362,7 @@ class PtychographyLinearOperator(LinearPhysics):
 
         if shifts is None:
             self.n_img = 25
-            shifts = torch.tensor(generate_shifts(img_size=img_size, n_img=self.n_img))
+            shifts = generate_shifts(img_size=img_size, n_img=self.n_img)
         else:
             self.n_img = len(shifts)
 
@@ -522,7 +524,9 @@ def build_probe(img_size, type="disk", probe_radius=10, device="cpu"):
     return probe
 
 
-def generate_shifts(img_size, n_img=25, fov=None):
+def generate_shifts(
+    img_size: Any, n_img: int = 25, fov: int | None = None
+) -> torch.Tensor:
     """
     Generates the array of probe shifts across the image.
     Based on probe radius and field of view.
@@ -530,7 +534,7 @@ def generate_shifts(img_size, n_img=25, fov=None):
     :param img_size: Size of the image.
     :param int n_img: Number of shifts (must be a perfect square).
     :param int fov: Field of view for shift computation.
-    :return np.ndarray: Array of (x, y) shifts.
+    :return: Array of (x, y) shifts.
     """
     if fov is None:
         fov = img_size[-1]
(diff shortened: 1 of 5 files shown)


nucli-vicky:
I do not fully understand the linting error. The error itself says:

"Oh no! 💥 💔 💥 1 file would be reformatted, 264 files would be left unchanged."

Isn't it expected that a file would be reformatted if a feature was added to it? Probably a basic question, or some part of the comments I added do give an error which is not displayed. I would appreciate any pointers!

Andrewwango:
> I do not fully understand the linting error. The error itself says:
> 
> "Oh no! 💥 💔 💥 1 file would be reformatted, 264 files would be left unchanged."
> 
> Isn't it expected that a file would be reformatted if a feature was added to it? Probably a basic question, or some part of the comments I added do give an error which is not displayed. I would appreciate any pointers!

Hi Vicky thanks for the comment! This just means you need to run 
```bash
black .
```
and commit again. This formats your code according to our code style!

You should also run our linter:
```bash
ruff check
```
You can find more info in our [contributing guide](https://deepinv.github.io/deepinv/contributing.html#how-to-contribute)!

deepinv/physics/__init__.py
```diff
@@ -34,6 +34,7 @@
     SaltPepperNoise,
     ZeroNoise,
     FisherTippettNoise,
+    RicianNoise,
```
Andrewwango, line 37: Thanks for contributing RicianNoise!

Please can you also add the noise to the User guide docs, in the [noise table](https://github.com/deepinv/deepinv/blob/main/docs/source/user_guide/physics/physics.rst?plain=1#L153)

And please can you also add a test to the [noise tests file](https://github.com/deepinv/deepinv/blob/main/deepinv/tests/test_noise_model.py)? They test simple things like the random generator working correctly, the noise can be composed correctly etc.

Thanks!

deepinv/physics/noise.py
```diff
@@ -891,37 +891,6 @@ def forward(self, x, p=None, s=None, seed: int = None, **kwargs):
         return y
 
 
-def _infer_device(
```
Andrewwango, line 894: Why has this been moved?

Andrewwango (changes requested):
Looks good, thank you for contributing Vicky! A couple of comments:

deepinv/physics/noise.py
nucli-vicky: I moved it back to the original order, `_infer_device` was always after defining all the noise models, but when `FisherTippettNoise` was added, it was defined after `_infer_device`, it seemed more logical to keep all noise models together (as was originally the case) and not have `_infer_device` added in between noise models randomly.

deepinv/physics/noise.py
```diff
@@ -952,3 +921,77 @@ def forward(self, x, l=None, **kwargs):
         x = torch.exp(x)
         gamma = GammaNoise(self.l)
         return torch.log(gamma(x))
+
+
+class RicianNoise(NoiseModel):
+    r"""
+    RicianNoise: :math:`y = \sqrt{(x + \sigma \cdot  \mathcal{N}_1)^2 + \sigma \cdot \mathcal{N}_2^2}`
```
tachella:
```suggestion
    RicianNoise: :math:`y = \sqrt{(x + \sigma \epsilon_1)^2 + (\sigma \epsilon_2)^2}`
    
    where :math:`\epsilon_1\sim\mathcal{N}(0,I)` and :math:`\epsilon_2\sim\mathcal{N}(0,I)`
```

tachella (approved):
thanks a lot @nucli-vicky, the PR looks good to me

Andrewwango (approved):
Thanks @nucli-vicky and 🎉for your first contribution to DeepInverse!

Issue #763, opened 2025-09-12
General nn/layers/op file to gracefully handle 3D vs 2D


nucli-vicky:
### Describe the feature

A collection of standard vision layers / components / ops (convs, pooling, upsampling, norms, ...) with a dim keyword that can return the 3D or 2D variant. Other option is using a dict e.g. `convnd = {2 : nn.Conv2d, 3 : nn.Conv3d}`

### Motivation

As the repo starts to tackle 3D data, various networks will need to have a 3D variant. The idea above can reduce overall code complexity and increase readability.

### Additional context

Related to my comment on PR #743

nucli-vicky:
See for instance diffunet.py which has this functionality internally. Still, a general file that is used across networks can be useful and reduce code redundancy.

jscanvic:
Thanks for suggesting that Vicky!

At the moment there's a single network supporting 3d convolutions but once Chaithya's PR gets merged I'll happily review a unifying PR.

PR #806, opened 2025-10-01
Support 3D patches datasets

nucli-vicky:
Rough draft for now. WIP

To do (at least):

- [x] Support nibabel and blosc2 for PatchesDataset as well
- [x] Support giving a 'measurements' and 'ground-truth' dir
- [x] Refactor (it's probably not in the correct place right now)
- [ ] Benchmark loading methods
- [x] Support 2D slicing and not just 3D patches

### DIFF
--- a/deepinv/datasets/random_sampler.py
+++ b/deepinv/datasets/random_sampler.py
@@ -102,7 +102,7 @@ class RandomPatchSampler(ImageDataset):
             )
         )
 
-        if len(self.imgs) == 0:
+        if len(self.imgs) == 0:  # pragma: no cover
             raise RuntimeError("No (shared) images available.")
 
         self.shapes = self._get_shapes()


Andrewwango:
Hey Vicky, if any of the functionality proposed here can be integrated with the structural changes proposed in #768  that would be great! in particular with e.g. `load_nifti` function

nucli-vicky:
indeed, I originally opened a PR on your fork for #768

nucli-vicky:
- [x] Add blosc2 to `dataset` optional dependencies
- [x] Maybe add a hint or warning when loading `.nii.gz` files without `indexed-gzip` available
- [x] Add blosc2 and nifti to `test_PatchDataset3D`
- [x] Add `PatchDataset3D` to userguide.
- [x] Robustness to images with channels.

nucli-vicky:
@Andrewwango the load functions added here were built from #768 . The functionality from that PR should remain since the default arguments would turn to what `load_np` and `load_nifti` did there, apart from
(i) I removed the `unsqueeze(0)`. I am not sure what is the best way to handle (un)squeezing to keep things as general as possible, but the `unsqueeze` assumes data is of a format without channels, so probably not ideal? 
(ii) How the dtypes are handled. See also Jérémy's comment. For now I handle this with the dtype argument (not added to `load_np` right now, will fix later), but for the future we could make use of headers when loading nifti files. Just letting users (or datasets) specify `dtype` to convert to seems fine for me as well.

Overall, what's the best way to proceed? #768 does a lot more than adding these 3 load functions; I don't want to give it too many merge conflicts..

nucli-vicky:
Does it make sense to add this to the base datasets in the user guide? It does allow users to work with their own dataset, they need to provide a directory with volumetric data in nifti, blosc2 or npy format.
Additionally, what would be the correct file to put the `PatchDataset3D` class in? I feel like putting it into `patch_dataset.py` is a bit misleading since its behaviour / usecase is different from `PatchDataset`.

Andrewwango:
Hey Vicky:

- I wouldn't worry about conflicts between this PR and #768 - the load_np and load_nifti functions there are small, and happy to be overridden by your contribution. Thanks for letting me know the difference though! I will update you when I progress with it, as I may rename some things
- The current PatchDataset could do with revamping to generalise it and make it more user friendly - do you think you could merge the functionality of it with PatchDataset3D, where it generalises patching to different dims? Happy to keep it in either file but we could definitely signpost better to it in the datasets user guide.
- Should get_loader go into the io funcs as an overall "dinv.io.load" that autodetects how to load? (this is a common theme in other io libraries)

nucli-vicky:
`PatchDataset` and `PatchDataset3D` (or nd?) have quite different use-cases and functionality.
(i) Sampling is different (predetermined patches over each image, each epoch all those patches are sampled vs. a single random patch from each image).
(ii) In `PatchDataset`, the entire dataset fits into memory and is passed as a single tensor. `PatchDataset3D` assumes this is not possible, and even moreso supports cases where even a single volume is too large.

Merging them is possible, but would require an argument controlling the sampling method and general `x` and `y` arguments which can be either a Path/str or a tensor. Optionally we could auto revert to the current `PatchDataset` sampling method if tensors are passed, which actually makes sense: if the dataset fits entirely into memory, that sampling method makes more sense than defining an epoch as merely sampling once from each image. So it is possible, without hurting backwards compatibility (although in that case the `stride` argument would be ignored if we get paths).

However, this would reduce readability and maintainability + silently changing the sampling method based on the types of `x`/`y` does not seem ideal. In essence it would be a single class with two types of behaviors depending on its arguments. It's possible, but may not be desired? Perhaps it's better to have two intuitive classes than a single class with non-straightforward behavior? 

Especially since, ideally `PatchDataset3/ND` should also be able to handle measurements with different resolutions for super-resolution datasets, so the current proposal will already gain some complexity.

nucli-vicky:
Good idea to move the `get_loader` functionality into io funcs! One potential issue that there exist at least two categories of formats: those supporting memory mapping vs. those that need to be loaded as whole. So these different formats would have different API arguments. We could however add the arguments for memory mapping to all loaders, which are simply ignored (with a warning?) if the format does not support memory mapping.

nucli-vicky:
Note: current `PatchDataset` has some resemblance to a sliding window set-up. It might make more sense to merge a sliding window loader and the current `PatchDataset` functionality.

Andrewwango:
Re. `PatchDataset` great points - maybe we overloaded the meaning of "patch" here. I agree that they should be two separate things. Maybe the following could clean it up:

1. Keep `PatchDataset`, and add a note to the docstring that this only handles 2D
2. Rename `PatchDataset3D` to `PatchRandomSamplerDataset` or something, because, as you said, the functionality is different to that of `PatchDataset`. This dataset can also generalise to `nd` - I assume also 2D? So that this dataset functionality is general and orthogonal to that of `PatchDataset`

Andrewwango:
Re. `get_loader`, I'm not an expert, but if you think that there are one or two arguments that would be common to many of the loader possibilities, I'm happy to add them explicitly and mention in the docstring that they are used for some data types and unused for others. Otherwise I think the most general way woud be just to have `**loader_kwargs` and pass that onto the specific loader, so that the user can pass whatever they want. I would say that the general loader is for users to do a quick demo and if they want more optimised functionality to speed up they should use the specific one instead (we should add warnings encouraging the user to do so)

nucli-vicky:
Re. `PatchDataset` Good suggestion, will indeed rename like this. Should I create a new file for it or put it in the `base.py` folder of `datasets`?

Re. `get_loader`, nice idea. I think however it's a bit outside the scope of this PR to already create this general util. I'll add a method `set_loader` to the class which for now does the same as `get_loader` in the current proposal. In the future we can change it to smth like
```python
def set_loader(self, format : str):
   assert format in self.allowed_formats # or some global var for files supporting memory mapping
   self.load = get_loader(format)
```

nucli-vicky:
Also, in the tests right now I use the `brainweb_dl` tool to download some example files. This adds yet another dependency to install during test + downloading takes time.. 
Maybe we could just create some zero/one arrays or 'phantoms' and save them instead of downloading. It does not test on "real" data but that shouldn't affect whether the test correctly evaluates the class functionality.

Andrewwango:
Re. brainweb: we have both a demo volume and a demo slice available in the deepinv huggingface https://huggingface.co/datasets/deepinv/images/tree/main which you can load with `dinv.utils.load_example("brainweb_t1_ICBM_1mm_subject_0.npy")` or `brainweb_t1_ICBM_1mm_subject_0_slice_0.npy". Saves you from having to use brainweb - we don't want to be spending overhead loading nii.gz files. Let me know if you want me to upload anything else to the huggingface so you can use it for tests!

PS: here is a full list of demo data download/load functionality we have in deepinv: https://deepinv.github.io/deepinv/user_guide/other/utils.html#other-utils

nucli-vicky:
Code-wise I think it is okay now, but believe it can be improved with some thoughts and suggestions from others :). Especially the docstrings are rough right now.
Also should add an entry to the userguide. 
While this functionality might also be nice to have in `/examples`, I think it's best to do this once we have the code for training 3D models as well. Then we can add an example with the PR which takes care of 3D nets & models (#763).

Also: we can support functionality for super-resolution datasets, but I think that might best be a new class which inherits from `RandomPatchSampler` to avoid adding too much complexity to a single class. Mainly the shape check needs to be changed and the `__getitem__`.

Andrewwango:
Thanks @nucli-vicky !

I also just came across the 3D patch sampler dataset in torchio. Are there any similarities? https://docs.torchio.org/patches/patch_training.html

deepinv/datasets/random_sampler.py
```diff
@@ -0,0 +1,208 @@
...
+
+    **Patch size handling:**
+    - Accepts either an integer (applied to all spatial dims) or a tuple.
+    - If ``patch_size`` is tuple, and ``patch_size[i] == 1``, this is equivalent to slicing across axis i (singleton at axis i will be squeezed). This can be used to e.g. extract 2D slices from a 3D volume
+    - If tuple length is one less than the image ndim, the channel axis is auto-filled with ``None``.
+
+    **Randomness & reproducibility:**
+    - Patch coordinates are drawn with Python’s ``random`` module.
+    - To ensure deterministic behavior across workers, set the DataLoader's
+      ``worker_init_fn`` or ``generator`` according to the PyTorch reproducibility guidelines.
```
Andrewwango: You seem to be getting an indent issue here (sorry sphinx is annoying!) ([see here](https://github.com/deepinv/deepinv/actions/runs/18326766013/job/52193017854?pr=806#step:5:624))

> `/home/runner/work/deepinv/deepinv/deepinv/datasets/random_sampler.py:docstring of deepinv.datasets.random_sampler.RandomPatchSampler:24: ERROR: Unexpected indentation. [docutils]`

docs/source/api/deepinv.datasets.rst
```diff
@@ -79,6 +79,7 @@ Other Datasets
    :nosignatures:
 
     deepinv.datasets.PatchDataset
+    deepinv.datasets.RandomPatchSampler
```
Andrewwango, line 82: Just adding a todo to add an entry to user guide as discussed

deepinv/utils/io_utils.py
```diff
@@ -0,0 +1,136 @@
...
+
+from pathlib import Path
+
+import numpy as np
+from numpy.lib.format import open_memmap
+
+import torch
+
+
+def load_np(
```
Andrewwango, line 17: Just a note to self that these should overwrite those proposed in #768 where the `__init__` is cleaned + API/User guide

deepinv/utils/io_utils.py
```diff
@@ -0,0 +1,136 @@
...
+    fname: str | Path,
+    as_memmap: bool = False,
+    start_coords: tuple | list = None,
+    patch_size: tuple | list = None,
+) -> torch.Tensor:
+    """Load numpy array from file as torch tensor.
+
+    :param str, pathlib.Path fname: file to load.
+    :param bool, as_memmap: open this file as a memmap and return
```
Andrewwango: add to doc why this may be useful to users

deepinv/utils/io_utils.py
```diff
@@ -0,0 +1,136 @@
...
+                tuple(
+                    slice(s, s + p if s is not None else None)
+                    for s, p in zip(start_coords, patch_size)
+                )
+            ]
+        )
+    elif as_memmap:
+        return open_memmap(fname)
+    else:
+        return torch.from_numpy(np.load(fname, allow_pickle=False))
```
Andrewwango: if we remove the unsqueeze here can you make sure it doesn't hurt any of the usages of load_np in the rest of the codebase/examples

deepinv/utils/io_utils.py
```diff
@@ -0,0 +1,136 @@
...
+) -> torch.Tensor:
+    """Load numpy array from file as torch tensor.
+
+    :param str, pathlib.Path fname: file to load.
+    :param bool, as_memmap: open this file as a memmap and return
+    :param tuple | list, start_coords: starting indices when patch is to be extracted
+    :param tuple | list | int, patch_size: patch size if start_coords is given. If sequence, length must match start_coords
+    :return: :class:`torch.Tensor` containing loaded numpy array.
+    """
+    if start_coords is not None:
```
Andrewwango: What is the purpose of having patch loading for numpy arrays only? i.e. why is this functionality not in load_torch; in addition it's slightly cumbersome to have repeated code across different loaders which are not extremely loader-specific (e.g. the array slicing). Could the patch-based loading be generalised in any way?

deepinv/utils/io_utils.py
```diff
@@ -0,0 +1,136 @@
...
+        arr = open_memmap(fname)
+        assert len(start_coords) == len(
+            arr.shape
+        )  # I am assuming here that the images are single channel / have no channel on disk. To fix so start_coords can be different length for other cases
+        if isinstance(patch_size, int):
+            patch_size = [patch_size for i in range(len(start_coords))]
+        assert len(start_coords) == len(patch_size)
```
Andrewwango: I would combine the two asserts and instead do one `if` and `raise ValueError` with an error message

deepinv/utils/io_utils.py
```diff
@@ -0,0 +1,136 @@
...
+    :param bool, as_memmap: open this file as a memmap and return
+    :param tuple | list, start_coords: starting indices when patch is to be extracted
+    :param tuple | list | int, patch_size: patch size if start_coords is given. If sequence, length must match start_coords
+    :return: :class:`torch.Tensor` containing loaded numpy array.
+    """
+    if start_coords is not None:
+        arr = open_memmap(fname)
+        assert len(start_coords) == len(
+            arr.shape
+        )  # I am assuming here that the images are single channel / have no channel on disk. To fix so start_coords can be different length for other cases
```
Andrewwango: Feels like quite a strong assumption - I would just add instead a hint or something in the docstring how users should pass in start coords/patch size depending on the data they want to load

deepinv/utils/io_utils.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    try:
+        import nibabel as nib
+    except ImportError:  # pragma: no cover
+        raise ImportError(
+            "load_nifti requires nibabel, which is not installed. Please install it with `pip install nibabel`."
+        )
+
+    if start_coords is not None:
+        arr = nib.load(fname).dataobj
```
Andrewwango: As mentioned above feels like quite a bit of duplicated code, which would be nice to generalise e.g. with a patching decorator or something that can wrap the loaders

deepinv/utils/io_utils.py
```diff
@@ -0,0 +1,136 @@
...
+        return torch.from_numpy(
+            arr[
+                tuple(
+                    slice(s, s + p if s is not None else None)
+                    for s, p in zip(start_coords, patch_size)
+                )
+            ].astype(dtype)
+        )
+    elif as_memmap:
+        return nib.load(fname).dataobj  # dtype argument is not used in this case...
```
Andrewwango: why is this not torch

deepinv/utils/io_utils.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def load_blosc2(
+    fname: str | Path,
+    as_memmap: bool = False,
+    start_coords: tuple | list = None,
+    patch_size: tuple | list = None,
+    dtype: np.typing.DTypeLike = np.float32,
+) -> torch.Tensor:
+    """Load volume from blosc2 file as torch tensor.
```
Andrewwango, line 67: Again lot of repeated code with above funcs

deepinv/datasets/random_sampler.py
```diff
@@ -0,0 +1,208 @@
...
+                ), f"patch_size arguments must be integers, got type {type(p)} at index {i}"
+        self.x_dir, self.y_dir = x_dir, y_dir
+        self.patch_size, self.ch_ax = patch_size, ch_axis
+        self._set_load(file_format)
+
+        x_imgs, y_imgs = None, None
+
+        if x_dir is not None:
+            assert os.path.exists(x_dir), f"Ground-truth dir {x_dir} does not exist."
+            x_imgs = [f for f in os.listdir(x_dir) if f.endswith(file_format)]
```
Andrewwango: The globbing functionality is already in `dinv.datasets.ImageFolder` - it would be great to inherit from that and keep this class clean

deepinv/datasets/random_sampler.py
```diff
@@ -0,0 +1,208 @@
...
+        r"""
+        :param str, optional x_dir: Path to folder of ground-truth images. Required if ``y_dir`` is not given.
+        :param str, optional y_dir: Path to folder of measurement images. Required if ``x_dir`` is not given.
+        :param int, tuple patch_size: Size of patches to extract. If int, applies the same size across all spatial dimensions.
+        :param str file_format : File format to load. Other files are ignored.
+        :param int ch_axis: Axis of the channel dimension. If None, a new singleton channel is added.
+        """
+        assert x_dir or y_dir, "Provide at least one of x_dir or y_dir."
+        if ch_axis is not None:
+            assert (
```
Andrewwango: in general we use asserts for logical tests (that should never appear to users in any circumstance) where we use exceptions/warnings for user input validation

deepinv/datasets/random_sampler.py
```diff
@@ -0,0 +1,208 @@
...
+
+        self._set_shapes()
+
+    def __len__(self):
+        return len(self.imgs)
+
+    def __getitem__(self, idx: int):
+
+        shape = self.shapes[idx]
+        # We use random here: need to ensure deterministic behaviour based on seed --> seed worker function, see torch reproducibility page
```
Andrewwango: Hmm, why can't we use a torch generator + `rng.manual_seed(seed)` to manually seed where we want to?

deepinv/datasets/random_sampler.py
```diff
@@ -0,0 +1,208 @@
...
+import random
+import torch
+from deepinv.datasets.base import ImageDataset
+
+
+class RandomPatchSampler(ImageDataset):
+    r"""
+    Dataset for nD images that samples one random patch per image.
+
+    This dataset builds from one or two directories of nD images (`.npy`, `.nii(.gz)`, or `.b2nd`).
```
Andrewwango: It would be cool to support any loader (like `loader` argument in `dinv.datasets.ImageFolder`) - the user might lose the nice memory mapping but at least they still get patches (e.g. a situation where they have a volume in memory and just want to patch it)

deepinv/tests/test_datasets.py
```diff
@@ -1090,3 +1091,147 @@ def test_SKMTEASliceDataset(download_SKMTEA, device):
         )
         == 1
     )
+
+
+@pytest.fixture
+def make_data(tmp_path):
```
Andrewwango: this is cool thanks!

deepinv/tests/test_datasets.py
```diff
@@ -1090,3 +1091,147 @@ def test_SKMTEASliceDataset(download_SKMTEA, device):
...
+    """Minimal synthetic datasets for 3D (.npy/.b2nd/.nii.gz), 2D+channels (C,H,W & H,W,C), and >3D no-channels (H,W,D,T)."""
```
Andrewwango, line 1098: Is (H,W,D,T) standard? In deepinv so far I have been using the convention (B,C,T,D,H,W)

deepinv/tests/test_datasets.py
```diff
@@ -1090,3 +1091,147 @@ def test_SKMTEASliceDataset(download_SKMTEA, device):
...
+    import nibabel as nib
+    import blosc2
+
+    root = tmp_path
+
+    cases = []
+
+    # 3D volumes all formats
+    shape3d = (40, 40, 40)
+    for fmt in (".npy", ".b2nd", ".nii.gz"):
```
Andrewwango: Maybe can we parametrize this function with the filetype? And then parametrize the pytest with the filetype too

Just to make this function more modular!

deepinv/datasets/random_sampler.py
nucli-vicky: Mainly added it as a reminder to myself this should be noted or implemented somewhere. If I remember correctly, when using multiple workers, python's `random` module needs to be seeded manually with a `seed_worker`.
But good point, I'll look to use a torch native generator for this. It's been a while since I looked into randomness & torch dataloading so I need to have a closer look again.

deepinv/tests/test_datasets.py
nucli-vicky: Not sure if there is a true standard across domains. This is simply an annotation in tests, shouldn't affect things (time & spatial domain are treated equally in terms of loading logic), but happy to change the annotation if it standardizes things

deepinv/tests/test_datasets.py
nucli-vicky: Good point, will take a look. I'm not that familiar with `pytest`

deepinv/utils/io_utils.py
nucli-vicky: Alright. Generally I think not doing unsqueeze makes more sense as the unsqueeze already places the strong assumption that the image has no channels

deepinv/utils/io_utils.py
nucli-vicky: Probably, will take a look to unify some code between the 3 functions! I think not having it in `load_torch` does somewhat make sense; the purpose of handling patch/slicing in the loading is so any memory map specific things can be abstracted away by the load utils. In case we are loading the entire thing into memory, might as well return it to the caller. I'll think about unifying things.

deepinv/datasets/random_sampler.py
nucli-vicky: Good to know, thanks!

deepinv/utils/io_utils.py
nucli-vicky: The purpose of `as_memmap` is to load a pure memmap, with no image data loaded into memory. For example, the random sampler uses it to infer shapes during `__init__`, which would be very slow if we needed to load each image into memory. I do not believe torch supports memory mapped tensors?

deepinv/datasets/random_sampler.py
nucli-vicky: > The globbing functionality is already in `dinv.datasets.ImageFolder` - it would be great to inherit from that and keep this class clean

Good point, but `ImageFolder`'s globbing enforces length to be equal, instead of just collecting all the matching files for `x` and `y`. This makes it such that in some scenario's, users get errors which force them to reorganize folders, so I think matching files is more user-friendly. Dataset length can still be logged using `__len__`.

nucli-vicky:
@Andrewwango I removed the duplicate code in the `io_utils.py` by placing the patching logic entirely inside the dataset. This is fine for now; as the `RandomPatchSampler` is the only functionality that requires it. If in the future more features require it, we can look to refactor, but right now this seems to me the "cleanest" way to do things. 

Also parametrized the tests, thanks for the suggestion! 
Tried to address and fix most of your comments.

deepinv/datasets/random_sampler.py
```diff
@@ -0,0 +1,223 @@
...
+from pathlib import Path
+
+import torch
+from deepinv.datasets.base import ImageDataset
+
+
+class RandomPatchSampler(ImageDataset):
+    r"""
+    Dataset for nD images that samples one random patch per image.
+
```
Andrewwango, line 14: I would make clear that this dataset for now only supports loading from npy nifti etc but if the user wants to implement something else (eg geospatial cubes) they can simply override the load function

deepinv/datasets/random_sampler.py
```diff
@@ -0,0 +1,223 @@
...
+        dtype: torch.dtype = torch.float32,
+    ):
+        r"""
+        :param str, optional x_dir: Path to folder of ground-truth images. Required if ``y_dir`` is not given.
+        :param str, optional y_dir: Path to folder of measurement images. Required if ``x_dir`` is not given.
+        :param int, tuple patch_size: Size of patches to extract. If int, applies the same size across all spatial dimensions.
+        :param str file_format : File format to load. Other files are ignored.
+        :param int ch_axis: Axis of the channel dimension. If None, a new singleton channel is added.
+        :param torch.dtype dtype: Data type to use when loading the images.
+        """
```
Andrewwango, line 66: Do you think it would be a good idea add an argument "loader" which allows the user to pass in a custom loader? That way the user can easily use this dataset with various data types

deepinv/datasets/random_sampler.py
```diff
@@ -0,0 +1,223 @@
...
+        if isinstance(patch_size, tuple) or isinstance(patch_size, list):
+            for i, p in enumerate(patch_size):
+                if not isinstance(p, int):
+                    raise TypeError(
+                        f"patch_size must be int or tuple of ints, got {type(patch_size)}"
+                    )
+        self.x_dir, self.y_dir = x_dir, y_dir
+        self.patch_size, self.ch_ax = patch_size, ch_axis
+        self.dtype = dtype
+        self._set_load(file_format)
```
Andrewwango: I generally prefer functions to return things rather than operate in place, and this could be combined with the previous comment like 
`self.loader = loader if loader is not None else self._get_loader(file_format)`

deepinv/datasets/random_sampler.py
```diff
@@ -0,0 +1,223 @@
...
+        self.imgs = list(
+            sorted(set(imgs[0]) & set(imgs[1]))
+            if (imgs[0] and imgs[1])
+            else (imgs[0] or imgs[1])
+        )
+
+        if len(self.imgs) == 0:
+            raise RuntimeError("No (shared) images available.")
+
+        self._set_shapes()
```
Andrewwango: Same here

deepinv/datasets/random_sampler.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def __len__(self) -> int:
+        return len(self.imgs)
+
+    def __getitem__(self, idx: int) -> tuple[torch.Tensor, torch.Tensor] | torch.Tensor:
+
+        shape = self.shapes[idx]
+
+        start_coords = [
+            torch.randint(0, s - p, (1,)).item() if p is not None else p
```
Andrewwango, line 118: Does this need a random generator here too?

deepinv/datasets/random_sampler.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def _set_shapes(self):
+        ndim = None
+        self.shapes = []
+        n_ch = None
+        for im in self.imgs:
+            if self.y_dir and self.x_dir:
+                s_x = self._load(os.path.join(self.x_dir, im), as_memmap=True).shape
+                s_y = self._load(os.path.join(self.y_dir, im), as_memmap=True).shape
+                assert (
```
Andrewwango: Please can you make this an if + raise ValueError/RuntimeError instead of assets for input validation?

deepinv/datasets/random_sampler.py
```diff
@@ -0,0 +1,223 @@
...
+                    s_x == s_y
+                ), f"Measurement and ground-truth image shapes must match, but mismatch for {im}"
+                shape = s_x
+            else:
+                shape = self._load(
+                    os.path.join(self.y_dir if self.y_dir else self.x_dir, im),
+                    as_memmap=True,
```
Andrewwango, line 169: I guess one caveat of my previous suggestion is that we can't necessarily assume that the loader has an as_mammal argument - not sure what the solution is - maybe only pass as_mental if file type is numpy etc ? Or inspect the function signature? Or require that user defined loader takes kwargs?

deepinv/utils/io_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        return im.dataobj
+    else:
+        return torch.from_numpy(im.get_fdata(dtype=dtype))
+
+
+def load_blosc2(
+    fname: str | Path,
+    as_memmap: bool = False,
+    dtype: np.typing.DTypeLike = np.float32,
+) -> torch.Tensor | blosc2.ndarray.NDArray:
```
Andrewwango, line 66: Just checking that this is ok for using deepinv if you don't have blosc2 installed?

nucli-vicky: Yup! Because of the `if TYPE_CHECKING` conditional import of `blosc2` & `nibabel` these should not give errors while still allowing correct typing :)

deepinv/datasets/random_sampler.py
nucli-vicky: I don't think so. Passing the generator at dataloader creation makes it such that the base seed is fixed. Seed per worker is then equal to base seed + worker_id; this ensures reproducibility (provided same # of workers is used). I think this is the usual way to do things when using multiple workers for dataloading?

deepinv/datasets/random_sampler.py
nucli-vicky: ✅

deepinv/datasets/random_sampler.py
nucli-vicky: I left it as saying users should ensure their `loader` also accepts the `as_memmap` argument. Simplifies code + it's no big issue if a wrapper of a few lines has to be written I think :)

nucli-vicky:
@Andrewwango decided to indeed add support for the custom loader! It might not be in the easiest way, but the flexibility is there now, and I think making the constraints on the custom loader less tight would greatly increase code complexity..

deepinv/datasets/random_sampler.py
```diff
@@ -0,0 +1,233 @@
...
+            )
+
+    def load(self, f: str | Path, start_coords: tuple) -> torch.Tensor:
+        arr = self._load(f, as_memmap=True)
+        slices = tuple(
+            slice(start, start + size) if size is not None else slice(None)
+            for start, size in zip(start_coords, self.patch_size)
+        )
+        arr = arr[slices]
+        return torch.from_numpy(arr).to(self.dtype)
```
Andrewwango: I would perform the numpy -> tensor conversion after having checked that arr isn't already a tensor

nucli-vicky: Thanks!

nucli-vicky:
Thanks @Andrewwango . It seems like some of my responses on the comments had some issues, [so I was unaware they were not visible](https://github.com/orgs/community/discussions/10369), which is a bit annoying. I just now made them visible finally.

nucli-vicky:
@Andrewwango I will have a look to improve the test coverage. The only functionality that should additionally be included in the tests now is the custom loader (maybe make it return a `torch.Tensor`? Also modify this bit in the docstring that tensor returns are allowed now). 
One downside of the codecov is that it counts `raise RuntimeError` as a line which needs to be covered in tests. I don't think this is something we want?

nucli-vicky:
@Andrewwango the codecov is now only because of the `raise RuntimeError`s not triggering. I think this is fine? 
Maybe more broadly, is it possible to exclude these things from codecov? I think if we want to target `raise`s being triggered during tests this would result in a combination of
(i) complexifying the tests, by a lot
(ii) discouraging the raising of errors, which leads to less informative errors being displayed to users, or silent bugs..

Andrewwango:
Yes use on the except # pragma: no cover (you can search for instances where it's used elsewhere in the library)

nucli-vicky:
@Andrewwango thanks!

PR #815, opened 2025-10-03
GPU tests

tachella:
Adding scheduled GPU tests and doctests


.github/workflows/test_gpu.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      # ✅ Install uv and project deps
+      - name: Install dependencies with uv
+        run: |
+          source "$(conda info --base)/etc/profile.d/conda.sh"
+          conda activate py311
+          pip install uv
+                
+          # Fix corrupted sympy install (remove & reinstall)
+          pip uninstall -y sympy || true
```
Andrewwango, line 36: Wow this seems unfortunately hacky...

.github/workflows/test_gpu.yml
```diff
@@ -0,0 +1,87 @@
...
+        run: |
+          source "$(conda info --base)/etc/profile.d/conda.sh"
+          conda activate py311
+          python -m pytest -n 2 --dist=loadgroup
+
+      - name: Sphinx build
+        run: |
+          source "$(conda info --base)/etc/profile.d/conda.sh"
+          conda activate py311
+          sphinx-build docs/source _build
```
Andrewwango, line 69: does this run the examples?

Andrewwango (commented):
Nice, two questions

.github/workflows/test_gpu.yml
tachella: Yes unfortunately it doesn't work otherwise. And using uv speeds up 35 min approx

.github/workflows/test_gpu.yml
tachella: If it passes the tests before yes

tachella:
Really cool! Let's go with that solution 

> We should be able to have install times below the minute by using uv instead of conda (it has its own faster isolated environments)
> 
> See the proof of concept: https://github.com/deepinv/deepinv/actions/runs/18244634459/job/51950965466?pr=819

tachella:
I'll close this PR, since we can continue on #819

PR #823, opened 2025-10-06
rm einops from banned module

mh-nguyen712:
`einops` is now a dependency of `deepinv` but it's banned from the ruff check so tests are not passing.


jscanvic:
Hey Hai! We've decided to disallow top-level imports of not just optional dependencies but non-crucial dependencies as well (pretty much everything except torch and torchvision). They used to cause significant slow down at startup

CI should normally pass as long as einops is imported in the classes/functions it's used in

Let me know if you think we should still allow them

(For reference: #681)

mh-nguyen712:
Thanks for your reply @jscanvic , I should ask you before submitting this :D 
No worries I can import it inside my function. Sorry for this I'll close

PR #825, opened 2025-10-07
Implement / improve convolution functional ops

mh-nguyen712:
This PR implements:
- `conv3d`, `conv_transpose3d` with all `padding` mode 
- Extend `conv3d_fft`, `conv_transpose3d_fft` to all `padding` mode 
- Extend `conv2d_fft`, `conv_transpose2d_fft` to all `padding` mode 
- Make sure that spatial convolution and FFT-based convolution give the same output (both 2D and 3D) for all padding (pytested)
- Better pytest. 
- Fix some error in docstring and better warnings / assertion 

### DIFF
--- a/deepinv/physics/functional/convolution.py
+++ b/deepinv/physics/functional/convolution.py
@@ -186,7 +186,9 @@ def conv_transpose2d(
     return out
 
 
-def conv2d_fft(x: Tensor, filter: Tensor, real_fft: bool = True) -> torch.Tensor:
+def conv2d_fft(
+    x: Tensor, filter: Tensor, real_fft: bool = True, padding: str = "circular"
+) -> torch.Tensor:
     r"""
     A helper function performing the 2d convolution of images ``x`` and ``filter`` using FFT.
 
@@ -196,14 +198,16 @@ def conv2d_fft(x: Tensor, filter: Tensor, real_fft: bool = True) -> torch.Tensor
     `numpy <https://numpy.org/doc/stable/user/basics.broadcasting.html>`_.
     Otherwise, each channel of each image is convolved with the corresponding kernel.
 
-    For convolution using FFT consider only ``'circular'`` padding (i.e., circular convolution).
+    .. note::
 
-    .. note:
-
-        The convolution here is a convolution, not a correlation as in conv2d.
+        The convolution here is a convolution, not a correlation as in :func:`torch.nn.functional.conv2d`.
 
     :param torch.Tensor x: Image of size ``(B, C, W, H)``.
     :param torch.Tensor filter: Filter of size ``(b, c, w, h)`` where ``b`` can be either ``1`` or ``B`` and ``c`` can be either ``1`` or ``C``.
+    :param bool real_fft: for real filters and images choose `True` (default) to accelerate computation.
+    :param str padding: can be ``'valid'``, ``'circular'``, ``'replicate'``, ``'reflect'``, ``'constant'``.
+        If ``padding = 'valid'`` the output is smaller than the image (no padding),
+        otherwise the output has the same size as the image. Default is ``'circular'``.
     :return: torch.Tensor : the output of the convolution of the shape size as :math:`x`
     """
     assert x.dim() == filter.dim() == 4, "Input and filter must be 4D tensors"
@@ -220,24 +224,64 @@ def conv2d_fft(x: Tensor, filter: Tensor, real_fft: bool = True) -> torch.Tensor
     if b != B:
         assert b == 1, "Batch size of the kernel is not matched for broadcasting"
 
-    filter_f = filter_fft_2d(filter, img_size, real_fft)
-    x_f = fft.rfft2(x) if real_fft else fft.fft2(x)
+    ph, pw = h // 2, w // 2
+    ih, iw = (h - 1) % 2, (w - 1) % 2
 
-    return fft.irfft2(x_f * filter_f).real
+    def fft2(t, s=None):
+        return fft.rfft2(t, s=s) if real_fft else fft.fft2(t, s=s)
+
+    def ifft2(t, s):
+        return fft.irfft2(t, s=s).real if real_fft else fft.ifft2(t, s=s).real
+
+    if padding == "circular":
+        img_size = (H, W)
+        fx = fft2(x)
+        ff = fft2(filter, s=img_size)  # no pre-shift
+        y = ifft2(fx * ff, s=img_size)
+        # Align kernel center (equivalent to pre-shifting the filter)
+        return torch.roll(y, shifts=(-ph, -pw), dims=(-2, -1))
+
+    elif padding == "valid":
+        # Full linear conv via zero-padding in FFT, then crop to valid
+        sH, sW = H + h - 1, W + w - 1
+        img_size = (sH, sW)
+        fx = fft2(x, s=img_size)
+        ff = fft2(filter, s=img_size)
+        full = ifft2(fx * ff, s=img_size)
+        return full[:, :, h - 1 : H, w - 1 : W]
+
+    elif padding in ("constant", "reflect", "replicate"):
+        # Pad in spatial domain, do circular FFT-conv on padded grid, center-crop back
+        pad = (pw, pw - iw, ph, ph - ih)  # (W_left, W_right, H_top, H_bottom)
+        x_pad = F.pad(x, pad, mode=padding, value=0)
+        Hp, Wp = x_pad.shape[-2:]
+        img_size = (Hp, Wp)
+        fx = fft2(x_pad)
+        ff = fft2(filter, s=img_size)
+        y_pad = ifft2(fx * ff, s=img_size)
+        y_pad = torch.roll(y_pad, shifts=(-ph, -pw), dims=(-2, -1))
+        return y_pad[:, :, ph : -ph + ih, pw : -pw + iw]
+
+    else:
+        raise ValueError(
+            f"padding = '{padding}' not implemented. Please use one of 'valid', 'circular', 'replicate', 'reflect' or 'constant'."
+        )
 
 
 def conv_transpose2d_fft(
-    y: Tensor, filter: Tensor, real_fft: bool = True
+    y: Tensor, filter: Tensor, real_fft: bool = True, padding: str = "circular"
 ) -> torch.Tensor:
     r"""
-    A helper function performing the 2d transposed convolution 2d of ``x`` and ``filter`` using FFT. The adjoint of this operation is :func:`deepinv.physics.functional.conv2d_fft`.
+    A helper function performing the 2d transposed convolution 2d of ``x`` and ``filter`` using FFT.
+    The adjoint of this operation is :func:`deepinv.physics.functional.conv2d_fft`.
 
     :param torch.Tensor y: Image of size ``(B, C, W, H)``.
     :param torch.Tensor filter: Filter of size ``(b, c, w, h)`` ) where ``b`` can be either ``1`` or ``B`` and ``c`` can be either ``1`` or ``C``.
 
     If ``b = 1`` or ``c = 1``, then this function supports broadcasting as the same as `numpy <https://numpy.org/doc/stable/user/basics.broadcasting.html>`_. Otherwise, each channel of each image is convolved with the corresponding kernel.
 
-    For convolution using FFT consider only ``'circular'`` padding (i.e., circular convolution).
+    :param bool real_fft: for real filters and images choose `True` (default) to accelerate computation.
+    :param str padding: can be ``'valid'``, ``'circular'``, ``'replicate'``, ``'reflect'``, ``'constant'``. If ``padding = 'valid'`` the output is larger than the image (padding), otherwise the output has the same size as the image. Default is ``'circular'``.
 
     :return: torch.Tensor : the output of the convolution, which has the same shape as :math:`y`
     """
@@ -255,12 +299,95 @@ def conv_transpose2d_fft(
     if b != B:
         assert b == 1
 
-    filter_f = filter_fft_2d(filter, img_size, real_fft)
-    y_f = fft.rfft2(y) if real_fft else fft.fft2(y)
+    ph, pw = h // 2, w // 2
+    ih, iw = (h - 1) % 2, (w - 1) % 2
 
-    return fft.irfft2(y_f * torch.conj(filter_f), s=(H, W)).real
+    def fft2(t, s=None):
+        return fft.rfft2(t, s=s) if real_fft else fft.fft2(t, s=s)
+
+    def ifft2(t, s):
+        return fft.irfft2(t, s=s).real if real_fft else fft.ifft2(t, s=s).real
+
+    if padding == "circular":
+        img_size = (H, W)
+        # Adjoint of conv2d_fft circular: roll input by +center before multiplication
+        y_roll = torch.roll(y, shifts=(ph, pw), dims=(-2, -1))
+        fy = fft2(y_roll, s=img_size)
+        ff = fft2(filter, s=img_size)
+        return ifft2(fy * torch.conj(ff), s=img_size)
+
+    elif padding == "valid":
+        # Adjoint of: full linear conv then center-crop [h-1:H, w-1:W]
+        sH, sW = H + h - 1, W + w - 1
+        img_size = (sH, sW)
+        y_full = torch.zeros((B, C, sH, sW), device=y.device, dtype=y.dtype)
+        y_full[:, :, h - 1 : h - 1 + H, w - 1 : w - 1 + W] = y
+        fy = fft2(y_full, s=img_size)
+        ff = fft2(filter, s=img_size)
+        return ifft2(fy * torch.conj(ff), s=img_size)
+
+    elif padding in ("constant", "reflect", "replicate"):
+        if ph == 0 or pw == 0:
+            raise ValueError(
+                "Both dimensions of the filter must be strictly greater than 2 for this padding mode."
+            )
+        # Forward was: pad (P) -> conv (C) -> roll (R) -> crop (S)
+        # Adjoint should be: (S*) -> roll (R*) -> conv (C*) -> pad (P*))
+        Hp = H + ph + (ph - ih)
+        Wp = W + pw + (pw - iw)
+        img_size = (Hp, Wp)
+
+        # S*: embed y into center of padded grid
+        y_big = torch.zeros((B, C, Hp, Wp), device=y.device, dtype=y.dtype)
+        y_big[:, :, ph : -ph + ih, pw : -pw + iw] = y
+        # R*: roll by +center
+        y_big = torch.roll(y_big, shifts=(ph, pw), dims=(-2, -1))
+
+        # C*: circular transpose conv on padded grid
+        fy = fft2(y_big, s=img_size)
+        ff = fft2(filter, s=img_size)
+        z_big = ifft2(fy * torch.conj(ff), s=img_size)
+
+        # P*: adjoint of padding -> fold to original HxW
+        if padding == "constant":
+            out = z_big[:, :, ph : -ph + ih, pw : -pw + iw]
+        elif padding == "reflect":
+            out = z_big[:, :, ph : -ph + ih, pw : -pw + iw].clone()
+            for sy in (-1, 0, 1):
+                for sx in (-1, 0, 1):
+                    if sy == 0 and sx == 0:
+                        continue
+                    ty, ysrc = _tgt_src_for_axis_reflect(sy, ph, ih)
+                    tx, xsrc = _tgt_src_for_axis_reflect(sx, pw, iw)
+                    flip_dims = tuple(dim for s, dim in zip((sy, sx), (2, 3)) if s != 0)
+                    chunk = z_big[:, :, ysrc, xsrc]
+                    if flip_dims:
+                        chunk = chunk.flip(dims=flip_dims)
+                    out[:, :, ty, tx].add_(chunk)
+        else:  # replicate
+            out = z_big[:, :, ph : -ph + ih, pw : -pw + iw].clone()
+            for sy in (-1, 0, 1):
+                for sx in (-1, 0, 1):
+                    if sy == 0 and sx == 0:
+                        continue
+                    ty, ysrc, yred = _tgt_src_for_axis_replicate(sy, ph, ih)
+                    tx, xsrc, xred = _tgt_src_for_axis_replicate(sx, pw, iw)
+                    reduce_dims = tuple(
+                        dim for red, dim in zip((yred, xred), (2, 3)) if red
+                    )
+                    chunk = z_big[:, :, ysrc, xsrc]
+                    if reduce_dims:
+                        chunk = chunk.sum(dim=reduce_dims)
+                    out[:, :, ty, tx].add_(chunk)
+        return out
+
+    else:
+        raise ValueError(
+            f"padding = '{padding}' not implemented. Please use one of 'valid', 'circular', 'replicate', 'reflect' or 'constant'."
+        )
 
 
+# This function is no longer used but kept for reference
 def filter_fft_2d(filter, img_size, real_fft=True):
     ph = int((filter.shape[-2] - 1) / 2)
     pw = int((filter.shape[-1] - 1) / 2)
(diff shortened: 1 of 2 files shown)


deepinv/physics/functional/convolution.py
```diff
@@ -354,48 +772,222 @@ def conv_transpose3d_fft(
...
+        return -1, slice(-p + i, None), True
+
+
+def _center_crop_slice_1d(p: int, i: int) -> slice:
+    # When p == i == 0, return the whole axis; otherwise replicate p : -p + i
+    # This prevent issues with 0:-0 slices
+    return slice(None) if (p == 0 and i == 0) else slice(p, -p + i)
+
+
+def _center_filter_2d(filter, img_size):
```
nucli-vicky: I believe `_center_filter_2d` `_center_filter_3d` can be a single function if we assume the format B, C, H, W, (D) by using `2:`. Could reduce code complexity + increase maintainability :)

mh-nguyen712: Thanks. I just updated the code to do the FFT inside so I removed this function. Can you pull again please. Many thanks!

deepinv/physics/functional/convolution.py
```diff
@@ -354,48 +738,218 @@ def conv_transpose3d_fft(
...
+    pd = int((fd - 1) / 2)
+    ph = int((fh - 1) / 2)
+    pw = int((fw - 1) / 2)
+    filter = F.pad(filter, (0, iw - fw, 0, ih - fh, 0, id - fd))
+    filter = torch.roll(filter, shifts=(-pd, -ph, -pw), dims=(-3, -2, -1))
+    return (
+        fft.rfftn(filter, dim=(-1, -2, -3))
+        if real_fft
+        else fft.fftn(filter, dim=(-1, -2, -3))
+    )
```
nucli-vicky: Something like:
```python
def filter_fft_nd(filter, img_size, real_fft=True):
    r"""
    A helper function to compute the centered FFT of a filter zero-padded to img_size.
    """
    f = filter.shape[2:]
    i = img_size[2:]
    p = list(int( (f_i - 1) / 2) for f_i in f)
    padding = list(
    filter = F.pad(filter, [x for p_k, i_k in zip(p, i) for x in (0, i_k - p_k)])
    d = tuple(range(2, filter.ndim))
    filter = torch.roll(filter, shifts=*p, dims=d)
    return (fft.rfftn(filter, dim=d[::-1]) if real_fft else fft.fftn(filter, dim=d[::-1])
```
(I wrote this in the GitHub webclient, could be there are some syntax errors, did not check.. Just to show the general idea)

deepinv/physics/functional/convolution.py
```diff
@@ -40,18 +62,23 @@ def conv2d(
...
         ph = h // 2
         ih = (h - 1) % 2
         pw = w // 2
         iw = (w - 1) % 2
+        if ph == 0 and pw == 0:
+            _warn_once_padding([ph, pw], padding)
+
         pad = (pw, pw - iw, ph, ph - ih)  # because functional.pad is w,h instead of h,w
 
         x = F.pad(x, pad, mode=padding, value=0)
```
nucli-vicky, line 94: `F.pad` has as value default 0. I think it's fine to pass it explicitly but then we should be consistent (not everywhere in `convolution.py` do we pass `value=0`)

deepinv/physics/functional/convolution.py
mh-nguyen712: Thanks that's a good idea I can adapt the code

deepinv/physics/functional/convolution.py
```diff
@@ -17,17 +36,20 @@ def conv2d(
...
         ``constant`` corresponds to zero padding or ``same`` in :func:`torch.nn.functional.conv2d`
-    :return: (:class:`torch.Tensor`) : the output
+    :return: :class:`torch.Tensor`: the blurry output.
+
+    .. note::
+
+        Contrarily to Pytorch :func:`torch.nn.functional.conv2d`, which performs a cross-correlation, this function performs a convolution.
+
+        This function gives the same result as :func:`deepinv.physics.functional.conv2d_fft`. However, for small kernels, this function is faster.
+        For large kernels, :func:`deepinv.physics.functional.conv2d_fft` is usually more efficient.
```
nucli-vicky: Should we also give a note about the fact we return a view, not a contiguous tensor? In some cases I believe this may cause silent performance decreases downstream in other ops, so better to let the user know so they can call `.contiguous` if needed.

mh-nguyen712: or what do you think if we return a `contiguous` tensor directly?

nucli-vicky: Good point, don't see any immediate downside to it while it certainly avoids potential issues users may encounter.

mh-nguyen712:
Some preliminary benchmark between spatial convolution and FFT on my GPU with `padding='valid'`. 
We can switch between them in physics based on input and kernel size. 


conv2d_vs_conv2d_fft_img32_cuda:0
conv2d_vs_conv2d_fft_img64_cuda:0
conv2d_vs_conv2d_fft_img128_cuda:0
conv2d_vs_conv2d_fft_img256_cuda:0
conv3d_vs_conv3d_fft_img32_cuda:0
conv3d_vs_conv3d_fft_img64_cuda:0
conv3d_vs_conv3d_fft_img128_cuda:0
conv3d_vs_conv3d_fft_img256_cuda:0

deepinv/physics/functional/convolution.py
mh-nguyen712: Agree. I prefer to pass it explicitly. I'll make it be consistent

deepinv/physics/forward.py
```diff
@@ -1178,10 +1179,11 @@ def A_dagger(self, y, mask=None, **kwargs):
 
         # avoid division by singular value = 0
         if not isinstance(self.mask, float):
-            mask = torch.zeros_like(self.mask)
-            mask[self.mask > 1e-5] = 1 / self.mask[self.mask > 1e-5]
+            mask = torch.where(
+                self.mask > 1e-5, self.mask.reciprocal(), torch.zeros_like(self.mask)
+            )
```
nucli-vicky: `torch.zeros_like(self.mask)` constructs a full tensor of same shape as `self.mask`. I think we can make it cheaper by replacing it with `torch.tensor(0, dtype=self.mask.dtype, device=self.mask.device)`?

deepinv/utils/decorators.py
```diff
@@ -63,3 +63,64 @@ def new_init(self, *args, **kwargs):
     cls.__init__ = new_init
 
     return cls
+
+
+def _deprecated_func_replaced_by(
+    replacement, *, redirect=False, since=None, remove_in=None, extra=None
+):
```
nucli-vicky, line 70: Nice :)

deepinv/physics/forward.py
```diff
@@ -1162,6 +1162,7 @@ def prox_l2(self, z, y, gamma, **kwargs):
                 isinstance(gamma, torch.Tensor) and gamma.dim() < self.mask.dim()
             ):  # may be the case when mask is fft related
                 gamma = gamma[(...,) + (None,) * (self.mask.dim() - gamma.dim())]
+                gamma = gamma.to(self.mask.device)
```
nucli-vicky: Good that we set the device now! Should we also control the `dtype`?

deepinv/physics/functional/convolution.py
```diff
@@ -17,43 +47,46 @@ def conv2d(
...
     If ``b = 1`` or ``c = 1``, then this function supports broadcasting as the same as `numpy <https://numpy.org/doc/stable/user/basics.broadcasting.html>`_. Otherwise, each channel of each image is convolved with the corresponding kernel.
 
     :param padding: (options = ``valid``, ``circular``, ``replicate``, ``reflect``, ``constant``) If ``padding = 'valid'`` the output is smaller than the image (no padding), otherwise the output has the same size as the image.
         ``constant`` corresponds to zero padding or ``same`` in :func:`torch.nn.functional.conv2d`
-    :return: (:class:`torch.Tensor`) : the output
+    :return: :class:`torch.Tensor`: the blurry output.
+
+    .. note::
+
+        Contrarily to Pytorch :func:`torch.nn.functional.conv2d`, which performs a cross-correlation, this function performs a convolution.
```
nucli-vicky: "Contrary to PyTorch's :func:`torch.nn.functional.conv2d`, which performs a cross-correlation, this function performs a convolution."

deepinv/physics/functional/convolution.py
```diff
@@ -17,43 +47,46 @@ def conv2d(
...
+
+        This function gives the same result as :func:`deepinv.physics.functional.conv2d_fft`. However, for small kernels, this function is faster.
+        For large kernels, :func:`deepinv.physics.functional.conv2d_fft` is usually faster but requires more memory.
```
nucli-vicky, line 67: It would be cool to refer in some way to the small benchmark you did (or some other material discussing and showing the differences in performance and their reason). This may not be obvious to some users. 
I don't think this is a priority, feel free to ignore!

deepinv/physics/functional/convolution.py
```diff
@@ -77,46 +110,42 @@ def conv_transpose2d(
...
     assert y.dim() == filter.dim() == 4, "Input and filter must be 4D tensors"
-
-    if not correlation:
-        filter = torch.flip(filter, [-2, -1])
+    _raise_value_error_padding_messages(padding)
+    filter = _flip_filter_if_needed(filter, correlation, dims=(-2, -1))
 
     # Get dimensions of the input and the filter
     B, C, H, W = y.size()
     b, c, h, w = filter.size()
```
nucli-vicky, line 140: Using upper and lower case to distinguish these variables seems a bit tricky from a readability standpoint + the fact that upper case variables are often used for global constants. 
Maybe we can do like `b_y`, `b_f`, ... Adding the y and f also feels a bit clunky.. Not sure what the best way to go is.

deepinv/physics/functional/convolution.py
```diff
@@ -192,99 +171,252 @@ def conv2d_fft(x: Tensor, filter: Tensor, real_fft: bool = True) -> torch.Tensor
...
+        if pd == 0 and pw == 0 and ph == 0:
+            _warn_once_padding([ph, pw, pd], padding)
 
+        B, C, D, H, W = x.shape
 
-def conv_transpose3d(y: Tensor, filter: Tensor, padding: str = "valid"):
+    # Grouped convolution trick for per-batch filters and channels
+    x = x.reshape(1, B * C, D, H, W)
+    filter = filter.reshape(B * C, -1, d, h, w)
+    out = F.conv3d(x, filter, padding="valid", groups=B * C)
```
nucli-vicky, line 384: Just a general note we should keep in mind: there are some performance issues with the grouped conv. Sometimes it may actually be slower than manually looping. However I have insufficient knowledge to propose some way to determine which to use or to say that one is generally a safer option over the other.
See [pytorch#70954](https://github.com/pytorch/pytorch/issues/70954) [pytorch#73764](https://github.com/pytorch/pytorch/issues/73764) and finally [pytorch#37406](https://github.com/pytorch/pytorch/issues/37406) which has a recent comment regarding vmap which may be useful.

Overall I think this should not really be dealt with in this PR? Just something we should not forget about

deepinv/tests/test_physics_functional.py
```diff
@@ -1,139 +1,236 @@
...
-# -*- coding: utf-8 -*-
-"""
-Created on Thu Jul 11 14:48:05 2024
-
-@author: fsarron
-"""
 import pytest
 import torch
+import deepinv.physics.functional as dF
+from functools import partial
```
nucli-vicky, line 4: nice usage!

deepinv/physics/functional/convolution.py
mh-nguyen712: Many thanks for the comment with very useful reference. I'll create an isuse/discussion about this. From my experience, when passing to 3D spatial is too slow when filter is large (very common in inverse problem). I think I'll can look into this in a future PR. Thanks!

deepinv/physics/functional/convolution.py
mh-nguyen712: I agree that it's bit not very clean too use upper case. However I think it's fine since B, C, H, W are very common in pytorch...

deepinv/physics/functional/convolution.py
mh-nguyen712: Thanks for the suggestion but I think it's quite complex, since the performance will depend on the device, torch version, ... so I'll ignore this. thanks for the remark anyway

deepinv/physics/forward.py
mh-nguyen712: thanks for the remark! Good catch!

deepinv/physics/forward.py
mh-nguyen712: Good point. Done!

mh-nguyen712:
Thanks for you review @nucli-vicky! I made the changes and addressed most of your comments. Very helpful your comments and your references!

deepinv/physics/functional/convolution.py
```diff
@@ -192,99 +171,252 @@ def conv2d_fft(x: Tensor, filter: Tensor, real_fft: bool = True) -> torch.Tensor
...
-    A helper function to perform 3D convolution of images :math:``x`` and ``filter``.
+    A helper function to perform 3D convolution of images `x` and `filter`.
 
-    The transposed of this operation is :func:`deepinv.physics.functional.conv_transpose3d`
+    The transposed of this operation is :func:`deepinv.physics.functional.conv_transpose3d`.
+
+    :param torch.Tensor x: Image of size ``(B, C, D, H, W)``.
+    :param torch.Tensor filter: Filter of size ``(b, c, d, h, w)`` where ``b`` can be either ``1`` or ``B`` and ``c`` can be either ``1`` or ``C``.
+    :param str padding: can be ``'valid'`` (default), ``'circular'``, ``'replicate'``, ``'reflect'``, ``'constant'``. If ``padding = 'valid'`` the output is smaller than the image (no padding), otherwise the output has the same size as the image.
+
```
nucli-vicky: Add the same note about "contrary to PyTorch" that we gave in conv2d

nucli-vicky (approved):
LGTM apart from minor comment. Thanks for the changes! One nitpick: in `convolution.py`, we use `Tensor` for typing the function arguments but `torch.Tensor` for typing the function return. Best to have everything as `Tensor`?

deepinv/physics/functional/convolution.py
nucli-vicky: Okay, happy to keep as is!

deepinv/physics/functional/convolution.py
```diff
@@ -192,99 +171,252 @@ def conv2d_fft(x: Tensor, filter: Tensor, real_fft: bool = True) -> torch.Tensor
...
+    The adjoint of this operation is :func:`deepinv.physics.functional.conv2d_fft`.
 
     :param torch.Tensor y: Image of size ``(B, C, W, H)``.
     :param torch.Tensor filter: Filter of size ``(b, c, w, h)`` ) where ``b`` can be either ``1`` or ``B`` and ``c`` can be either ``1`` or ``C``.
 
     If ``b = 1`` or ``c = 1``, then this function supports broadcasting as the same as `numpy <https://numpy.org/doc/stable/user/basics.broadcasting.html>`_. Otherwise, each channel of each image is convolved with the corresponding kernel.
 
-    For convolution using FFT consider only ``'circular'`` padding (i.e., circular convolution).
+    :param bool real_fft: for real filters and images choose `True` (default) to accelerate computation.
+    :param str padding: can be ``'valid'``, ``'circular'``, ``'replicate'``, ``'reflect'``, ``'constant'``. If ``padding = 'valid'`` the output is larger than the image (padding), otherwise the output has the same size as the image. Default is ``'valid'``.
```
tachella: are we saying what padding is done under `constant`? why not `zeros` which is compatible with `torch.nn.Conv2d`?

deepinv/physics/functional/convolution.py
```diff
@@ -192,99 +171,252 @@ def conv2d_fft(x: Tensor, filter: Tensor, real_fft: bool = True) -> torch.Tensor
...
 ) -> torch.Tensor:
     r"""
-    A helper function performing the 2d transposed convolution 2d of ``x`` and ``filter`` using FFT. The adjoint of this operation is :func:`deepinv.physics.functional.conv2d_fft`.
+    A helper function performing the 2d transposed convolution 2d of ``x`` and ``filter`` using FFT.
+    The adjoint of this operation is :func:`deepinv.physics.functional.conv2d_fft`.
 
     :param torch.Tensor y: Image of size ``(B, C, W, H)``.
     :param torch.Tensor filter: Filter of size ``(b, c, w, h)`` ) where ``b`` can be either ``1`` or ``B`` and ``c`` can be either ``1`` or ``C``.
 
     If ``b = 1`` or ``c = 1``, then this function supports broadcasting as the same as `numpy <https://numpy.org/doc/stable/user/basics.broadcasting.html>`_. Otherwise, each channel of each image is convolved with the corresponding kernel.
```
tachella: this lies in the middle of the params and cuts the Params section into two:
image

can we move it up?

docs/source/user_guide/physics/functional.rst
```diff
@@ -48,6 +48,12 @@ Similar to the PyTorch structure, they are available within ``deepinv.physics.fu
    * - :func:`deepinv.physics.functional.conv_transpose2d_fft`
      - Computes the 2D transposed convolution with FFT, efficiently implementing upsampling or deconvolution.
 
+   * - :func:`deepinv.physics.functional.conv3d`
+     - Performs 3D convolution, typically used in volumetric data processing for filtering and feature extraction.
+
+   * - :func:`deepinv.physics.functional.conv_transpose3d`
+     - Computes the 3D transposed convolution (deconvolution), used for volumetric data upsampling or reversing convolutional operations.
```
tachella:
```suggestion
     - Computes the 3D transposed convolution
```
(it is not really deconvolving nor reverting in my opinion)

docs/source/user_guide/physics/functional.rst
```diff
@@ -48,6 +48,12 @@ Similar to the PyTorch structure, they are available within ``deepinv.physics.fu
    * - :func:`deepinv.physics.functional.conv_transpose2d_fft`
      - Computes the 2D transposed convolution with FFT, efficiently implementing upsampling or deconvolution.
 
+   * - :func:`deepinv.physics.functional.conv3d`
+     - Performs 3D convolution, typically used in volumetric data processing for filtering and feature extraction.
```
tachella:
```suggestion
     - Performs 3D convolution.
```

docs/source/changelog.rst
```diff
@@ -38,6 +39,7 @@ New Features
 - Add manual physics to reduced resolution loss (:gh:`808` by `Andrew Wang`_)
 - Multi-coil MRI coil-map estimation acceleration via CuPy (:gh:`781` by `Andrew Wang`_)
 - Add SpatialUnwrapping forward model and ItohFidelity data fidelity (:gh:`723` by `Brayan Monroy`_)
+- Add conv3d, conv_transpose3d, support all conv functional to all padding modes, make FFT and spatial equivalent (:gh:`825` by `Minh Hai Nguyen`_)
```
tachella:
```suggestion
- make FFT and spatial convolution equivalent (:gh:`825` by `Minh Hai Nguyen`_)
```

docs/source/changelog.rst
```diff
@@ -12,6 +12,7 @@ New Features
 - Add support for 3D CNNs (:gh:`869` by `Vicky De Ridder`)
 - Add support for complex dtypes in WaveletDenoiser, WaveletDictDenoiser and WaveletPrior (:gh:`738` by `Chaithya G R`_)
 - dinv.io functions for loading DICOM, NIFTI, COS, GEOTIFF etc. (:gh:`768` by `Andrew Wang`_)
+- Implement/extend functional for 2D/3D convolution with spatial and FFT, support all padding modes (:gh:`825` by `Minh Hai Nguyen`_)
```
tachella:
```suggestion
- Implement/extend functional for 2D/3D convolution with spatial and FFT (:func:`deepinv.physics.functional.conv3d` and  :func:`deepinv.physics.functional.conv_transpose3d`), support all padding modes (:gh:`825` by `Minh Hai Nguyen`_)
```

tachella (changes requested):
the PR looks good, many thanks for this work!

aside from some minor fixes, I'm wondering if we shouldn't change the naming of the `constant` padding for `zeros` or `same`? If I'm not wrong, the standard in pytorch is `zeros`?

deepinv/physics/functional/convolution.py
mh-nguyen712: thanks! I'll change to `zeros`

jscanvic:
(For future reference)

Another discrepancy with asymmetric kernels

**Code**

```python
import torch
import deepinv as dinv

# dirac off-centered to the right
h = torch.zeros(1, 1, 3, 3)
h[:, :, 2, 2] = 1.0

# dirac input image
x = torch.zeros(1, 1, 8, 8)
x[:, :, 0, 0] = 1.0

print(f"Kernel:")
print(h)

print("Input image:")
print(x)

physics_conv = dinv.physics.Blur(filter=h, padding="circular")
physics_fft = dinv.physics.BlurFFT(img_size=x.shape[-3:], filter=h)

y_conv = physics_conv(x)
y_fft = physics_fft(x)

print("BlurConv:")
print(y_conv.round().abs())
print("BlurFFT:")
print(y_fft.round().abs())
```

**Output**

```python
Kernel:
tensor([[[[0., 0., 0.],
          [0., 0., 0.],
          [0., 0., 1.]]]])
Input image:
tensor([[[[1., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.]]]])
BlurConv:
tensor([[[[0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 1., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.]]]])
BlurFFT:
tensor([[[[0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 1.]]]])
```

mh-nguyen712:
Thanks @jscanvic ! This is already fixed in this PR I think.

jscanvic:
Thanks Hai! Will review asap

deepinv/models/ram.py
```diff
@@ -39,12 +39,12 @@ class RAM(Reconstructor, Denoiser):
 
       >>> import deepinv as dinv
       >>> x = dinv.utils.load_example("butterfly.png")
-      >>> physics = dinv.physics.Downsampling(filter="bicubic")
+      >>> physics = dinv.physics.Downsampling(filter="bicubic", noise_model=dinv.physics.GaussianNoise(0.01))
```
jscanvic, line 46: Why?

deepinv/physics/functional/convolution.py
```diff
@@ -2,6 +2,41 @@
...
 from torch import Tensor
 import torch.fft as fft
+import warnings
+from itertools import chain, product
+from deepinv.utils.decorators import _deprecated_func_replaced_by
+
+_warned_messages = set()
+
+
+def _warn_once_padding(pad_value: list[int], padding: str, category=UserWarning):
```
jscanvic, line 12: Why once per value of padding size and mode?

deepinv/physics/functional/convolution.py
```diff
@@ -2,6 +2,41 @@
...
+    """Emit a warning only once per unique message."""
+    unit_kernel = "x".join(["1"] * len(pad_value))
+    message = (
+        f"You're using padding = '{padding}' with a {unit_kernel} kernel. This is equivalent to no padding. "
+        f"Consider using padding = 'valid' instead."
+    )
+    if message not in _warned_messages:
+        warnings.warn(message, category)
```
jscanvic:
```suggestion
        warnings.warn(message, category, stacklevel=2)
```

It's a warning helper so the stack level is increased

deepinv/physics/functional/convolution.py
```diff
@@ -17,43 +52,45 @@ def conv2d(
...
+    :param padding: (options = ``'valid'``, ``'circular'``, ``'replicate'``, ``'reflect'``, ``'constant'`` or ``'zeros'``). If ``padding = 'valid'`` the output is smaller than the image (no padding), otherwise the output has the same size as the image. Note that ``'constant'`` and ``'zeros'`` are equivalent. Default is ``'valid'``.
+    :return: :class:`torch.Tensor`: the blurry output.
 
-    If ``b = 1`` or ``c = 1``, then this function supports broadcasting as the same as `numpy <https://numpy.org/doc/stable/user/basics.broadcasting.html>`_. Otherwise, each channel of each image is convolved with the corresponding kernel.
+    .. note::
 
-    :param padding: (options = ``valid``, ``circular``, ``replicate``, ``reflect``, ``constant``) If ``padding = 'valid'`` the output is smaller than the image (no padding), otherwise the output has the same size as the image.
-        ``constant`` corresponds to zero padding or ``same`` in :func:`torch.nn.functional.conv2d`
-    :return: (:class:`torch.Tensor`) : the output
+        Contrarily to Pytorch's :func:`torch.nn.functional.conv2d`, which performs a cross-correlation, this function performs a convolution when ``correlation=False``.
```
jscanvic:
```suggestion
        Contrarily to PyTorch's :func:`torch.nn.functional.conv2d`, which performs a cross-correlation, this function performs a convolution by default unless ``correlation=True``.
```

jscanvic (changes requested):
Minor comments happy to approve once it's done

deepinv/models/ram.py
mh-nguyen712: Just make it more realistic. In absence of noise, A_dagger can gives better PSNR than RAM

deepinv/physics/functional/convolution.py
mh-nguyen712: To avoid repetitive warnings for example in training (as these functions are usually called inside Physics)

deepinv/models/ram.py
jscanvic: Thanks

deepinv/physics/functional/convolution.py
jscanvic: Sorry I wasn't clear my point is not it's not warning sufficiently often but that it's warning too often

deepinv/physics/functional/convolution.py
jscanvic: Sorry I just noticed another typo, it should be Contrary*

mh-nguyen712: Done!

docs/source/changelog.rst
```diff
@@ -72,6 +73,7 @@ New Features
 - Add manual physics to reduced resolution loss (:gh:`808` by `Andrew Wang`_)
 - Multi-coil MRI coil-map estimation acceleration via CuPy (:gh:`781` by `Andrew Wang`_)
 - Add SpatialUnwrapping forward model and ItohFidelity data fidelity (:gh:`723` by `Brayan Monroy`_)
+- [Backward non-compatible] Make :class:`deepinv.physics.BlurFFT` compute a true convolution (now) instead of cross-correlation (before). It is now equivalent to :class:`deepinv.physics.Blur` with `padding="circular"` (:gh:`825` by `Minh Hai Nguyen`_)
```
jscanvic:
```suggestion
- (Breaking) Make :class:`deepinv.physics.BlurFFT` compute a true convolution (now) instead of cross-correlation (before). It is now equivalent to :class:`deepinv.physics.Blur` with `padding="circular"` (:gh:`825` by `Minh Hai Nguyen`_)
```

The previous entries use `(Breaking)` let's stick to it to make the changelog easier to search through

jscanvic: Also explain the change for `deepinv.physics.Blur` with even filters, that the outputs are shifted by 1px up and 1px left compared to previously

deepinv/physics/functional/convolution.py
mh-nguyen712: It only raises warning of padding != valid now and kernel_size = 1 now.

docs/source/changelog.rst
mh-nguyen712: Thanks! I missed this one!

mh-nguyen712:
@tachella can you take a look again?

examples/unfolded/demo_DEQ.py
```diff
@@ -75,8 +75,8 @@
 # Generate a motion blur operator.
 kernel_index = 1  # which kernel to chose among the 8 motion kernels from 'Levin09.mat'
 kernel_torch = load_degradation("Levin09.npy", DEG_DIR / "kernels", index=kernel_index)
-kernel_torch = kernel_torch.unsqueeze(0).unsqueeze(
-    0
+kernel_torch = (
+    kernel_torch.unsqueeze(0).unsqueeze(0).to(torch.float32)
```
tachella, line 79: why this?

docs/source/changelog.rst
```diff
@@ -72,6 +73,7 @@ New Features
 - Add manual physics to reduced resolution loss (:gh:`808` by `Andrew Wang`_)
 - Multi-coil MRI coil-map estimation acceleration via CuPy (:gh:`781` by `Andrew Wang`_)
 - Add SpatialUnwrapping forward model and ItohFidelity data fidelity (:gh:`723` by `Brayan Monroy`_)
+- [Breaking] Make :class:`deepinv.physics.BlurFFT` compute a true convolution (now) instead of cross-correlation (before). It is now equivalent to :class:`deepinv.physics.Blur` with `padding="circular"` (:gh:`825` by `Minh Hai Nguyen`_). For even kernel sizes, the output is now shifted by one pixel to the top-left compared to before.
```
tachella:
```suggestion
- (Breaking) Make :class:`deepinv.physics.BlurFFT` compute a true convolution (now) instead of cross-correlation (before). It is now equivalent to :class:`deepinv.physics.Blur` with `padding="circular"` (:gh:`825` by `Minh Hai Nguyen`_). For even kernel sizes, the output is now shifted by one pixel to the top-left compared to before.
```

docs/source/changelog.rst
```diff
@@ -40,6 +40,7 @@ New Features
 - dinv.io functions for loading DICOM, NIFTI, COS, GEOTIFF etc. (:gh:`768` by `Andrew Wang`_)
 - Add `Open in Colab` button to examples (:gh:`907` by `Minh Hai Nguyen`_)
 - Integration with HuggingFace Diffusers library to use pretrained diffusion models as denoisers and for posterior sampling (:gh:`893` by `Minh Hai Nguyen`_)
+- Implement/extend functional for 2D/3D convolution with spatial and FFT (:func:`deepinv.physics.functional.conv3d` and  :func:`deepinv.physics.functional.conv_transpose3d`), support all padding modes with equivalent outputs (:gh:`825` by `Minh Hai Nguyen`_)
```
tachella: can you move the changelog entries to the latest version?

deepinv/tests/test_physics_functional.py
```diff
@@ -1,139 +1,237 @@
...
+    )
+
+    if transposed:
+        spatial_fn = dF.conv_transpose3d
+        fft_fn = partial(dF.conv_transpose3d_fft, real_fft=True)  # Only test real_fft
+    else:
+        spatial_fn = dF.conv3d
+        fft_fn = partial(dF.conv3d_fft, real_fft=True)
+
+    for sim in size_im:
```
tachella: can you move these to mark.parametrize? I think it will be more readable if we add

```
@pytest.mark.parametrize("im_x", [5,6])
@pytest.mark.parametrize("im_y", [5,6])
@pytest.mark.parametrize("im_z", [5,6])
@pytest.mark.parametrize("filt_x", [3,4])
@pytest.mark.parametrize("filt_y", [3,4])
@pytest.mark.parametrize("filt_z", [3,4])
```

deepinv/tests/test_physics_functional.py
```diff
@@ -1,139 +1,237 @@
...
+        for sfil in size_filt:
+            for bf in (1, B):
```
tachella: I would remove this for loop, since it should be taken into account by
`@pytest.mark.parametrize("B", [1, 2])`

deepinv/tests/test_physics_functional.py
```diff
@@ -1,139 +1,237 @@
...
+@pytest.mark.parametrize("nchan_im,nchan_filt", [(1, 1), (3, 1), (3, 3)])
+@pytest.mark.parametrize("padding", ALL_CONV_PADDING)  # safe set
+@pytest.mark.parametrize("real_fft", [True, False])
+@pytest.mark.parametrize("use_fft", [True, False])
+def test_conv3d_adjointness(
+    device, B, nchan_im, nchan_filt, padding, real_fft, use_fft
+):
+    torch.manual_seed(0)
 
+    size_im = (
```
tachella: same as below, move to pytest.mark.parametrize

deepinv/tests/test_physics_functional.py
```diff
@@ -1,139 +1,237 @@
...
+@pytest.mark.parametrize("transposed", [True, False])
+def test_conv2d_spatial_and_fft_equivalence(
+    device, B, nchan_im, nchan_filt, padding, transposed
+):
+    torch.manual_seed(0)
 
-                    Axy = torch.sum(Ax * y)
-                    Atyx = torch.sum(Aty * x)
+    size_im = ([nchan_im, 5, 5], [nchan_im, 6, 6], [nchan_im, 5, 6], [nchan_im, 6, 5])
+    size_filt = (
```
tachella: same here, move to pytest.mark.parametrize

deepinv/tests/test_physics_functional.py
```diff
@@ -1,139 +1,237 @@
...
+@pytest.mark.parametrize("use_fft", [False, True])
+@pytest.mark.parametrize("correlation", [True, False])
+def test_conv2d_adjointness(
+    device, B, nchan_im, nchan_filt, padding, real_fft, use_fft, correlation
+):
+    torch.manual_seed(0)
 
-        paddings = ("valid", "constant", "circular", "reflect", "replicate")
+    size_im = ([nchan_im, 5, 5], [nchan_im, 6, 6], [nchan_im, 5, 6], [nchan_im, 6, 5])
+    size_filt = (
```
tachella: same here

deepinv/tests/test_physics.py
```diff
@@ -1251,13 +1251,19 @@ def test_noise_domain(device):
...
+@pytest.mark.parametrize("filter_size", [(1, 5, 5), (1, 6, 6), (1, 6, 5)])
+def test_blur(img_size, filter_size, device):
     r"""
     Test that :class:`deepinv.physics.Blur` with `padding="circular"` and :class:`deepinv.physics.BlurFFT` compute the same circular blur.
     """
     torch.manual_seed(0)
-    x = torch.randn((3, 128, 128), device=device).unsqueeze(0)
-    h = torch.ones((1, 1, 5, 5)) / 25.0
+    x = torch.randn(*img_size, device=device).unsqueeze(0)
+    h = torch.ones(*filter_size, device=device).unsqueeze(0) / (
```
tachella: should we check with a very directional blur just to be very sure?

why not rand? with normalization

deepinv/tests/test_models.py
```diff
@@ -1052,16 +1053,25 @@ def test_restoration_models(
...
+            if hasattr(physics.noise_model, "sigma"):
+                physics.noise_model.sigma = torch.tensor(
+                    [max(physics.noise_model.sigma, sigma)]
+                )
+            else:
+                physics.noise_model = dinv.physics.GaussianNoise(sigma)
+
         if physics is not None:
-            physics.noise_model = dinv.physics.GaussianNoise(0.01, rng=rng)
+            for attr in dir(physics):
```
tachella, line 1068: why this? we could add some comment explaining

deepinv/physics/blur.py
```diff
@@ -454,10 +456,11 @@ class BlurFFT(DecomposablePhysics):
 
 
 
-    :param tuple img_size: Input image size in the form (C, H, W).
-    :param torch.Tensor filter: torch.Tensor of size (1, c, h, w) containing the blur filter with h<=H, w<=W and c=1 or c=C e.g.,
+    :param tuple img_size: Input image size in the form `(C, H, W)`.
+    :param torch.Tensor filter: torch.Tensor of size `(1, c, h, w)` containing the blur filter with h<=H, w<=W and c=1 or c=C e.g.,
         :func:`deepinv.physics.blur.gaussian_blur`.
-    :param torch.device, str device: Device this physics lives on. If filter is updated, it will be cast to BlurFFT's device.
+    :param torch.device, str device: Device this physics lives on. If `filter` is updated, it will be cast to BlurFFT's device.
```
tachella:
```suggestion
    :param torch.device, str device: Device this physics lives on. If `filter` is modified via `physics.update`, it will be automatically casted to BlurFFT's device.
```

examples/plug-and-play/demo_PnP_DPIR_deblur.py
```diff
@@ -50,7 +50,9 @@
 
 # Generate a motion blur operator.
 kernel_index = 1  # which kernel to chose among the 8 motion kernels from 'Levin09.mat'
-kernel_torch = load_degradation("Levin09.npy", DEG_DIR / "kernels", index=kernel_index)
+kernel_torch = load_degradation(
+    "Levin09.npy", DEG_DIR / "kernels", index=kernel_index
```
tachella, line 54: why the float? does it crash without it?

deepinv/tests/test_models.py
mh-nguyen712: The previous code did not change correctly the sigma of the noise_model for StackedPhysics. I'll add comments

deepinv/tests/test_physics.py
mh-nguyen712: I'll change to rand and a directional blur

deepinv/tests/test_physics_functional.py
mh-nguyen712: Done

mh-nguyen712: Note that I'm not testing all possible because it's too much (~7k tests)

docs/source/changelog.rst
mh-nguyen712: done

examples/plug-and-play/demo_PnP_DPIR_deblur.py
mh-nguyen712: Yes it crashes on my local: 

```
RuntimeError: Input type (torch.cuda.DoubleTensor) and weight type (torch.cuda.FloatTensor) should be the same
```

examples/unfolded/demo_DEQ.py
mh-nguyen712: don't know why it passes on the CI but on my local machine it fails

PR #826, opened 2025-10-07
Fix diffraction

mh-nguyen712:
This PR:
- Fix a bug in Zernike polynomial formula 
- Extend to higher radial order of Zernike polynomial (upto 7 now comparing to 4 previously)
- More flexible and easier way to convert between Noll, ANSI and standard indexing convention
- Improve 3D generator
- Add: random rotate, apodize to the `DiffractionBlurGenerator`

### DIFF
--- a/deepinv/physics/generator/blur.py
+++ b/deepinv/physics/generator/blur.py
@@ -6,7 +6,7 @@ from deepinv.physics.generator import PhysicsGenerator
 from deepinv.physics.functional import histogramdd, conv2d
 from deepinv.physics.functional.interp import ThinPlateSpline
 from deepinv.utils.decorators import _deprecated_alias
-
+from deepinv.transform.rotate import rotate_via_shear
 from .zernike import Zernike
 
 
@@ -156,7 +156,9 @@ class MotionBlurGenerator(PSFGenerator):
         :param float l: the length scale of the trajectory
         :param int seed: the seed for the random number generator.
 
-        :return: dictionary with key `filter`: the generated PSF of shape `(batch_size, 1, psf_size[0], psf_size[1])`
+        :return: dictionary with keys
+
+            - `filter`: the generated PSF of shape `(batch_size, 1, psf_size[0], psf_size[1])`
         """
         self.rng_manual_seed(seed)
         f_x = self.f_matern(batch_size, sigma, l)[..., None]
@@ -196,8 +198,6 @@ class DiffractionBlurGenerator(PSFGenerator):
     Zernike polynomials are a sequence of orthogonal polynomials defined on the unit disk.
     They are commonly used in optical systems to describe wavefront aberrations.
 
-
-
     The PSF :math:`h(\theta)` is defined as the squared magnitude of the Fourier transform of the pupil function :math:`p_{\theta} = \exp(- i 2 \pi \phi_{\theta})`:
 
     .. math::
@@ -211,8 +211,9 @@ class DiffractionBlurGenerator(PSFGenerator):
         \phi_\theta= \sum_{k \in K} \theta_k z_k
 
     where :math:`K` is set of indices of Zernike polynomials :math:`z_k` used in the decomposition (equal to `zernike_index`).
-    See :footcite:t:`lakshminarayanan2011zernike` for more details
-    (`or this link <https://e-l.unifi.it/pluginfile.php/1055875/mod_resource/content/1/Appunti_2020_Lezione%2014_4_Zernikepolynomialsaguidefinal.pdf>`_).
+    See :footcite:t:`lakshminarayanan2011zernike`
+    `or this link <https://e-l.unifi.it/pluginfile.php/1055875/mod_resource/content/1/Appunti_2020_Lezione%2014_4_Zernikepolynomialsaguidefinal.pdf>`_
+    or :class:`deepinv.physics.generator.Zernike` for more details.
 
     In the ideal diffraction-limited case (i.e., no aberrations), the PSF corresponds to the Airy pattern.
 
@@ -221,7 +222,12 @@ class DiffractionBlurGenerator(PSFGenerator):
 
     :param tuple psf_size: the shape ``H x W`` of the generated PSF in 2D
     :param int num_channels: number of images channels. Defaults to 1.
-    :param tuple[int] zernike_index: activated Zernike coefficients in the following `index_convention` convention.
+    :param tuple[int, ...], tuple[tuple[int, int], ...] zernike_index: activated Zernike coefficients in the following `index_convention` convention.
+        It can be either:
+
+            - a tuple of `int` corresponding to the Noll or ANSI indices, in which case the `index_convention` parameter is required to interpret them correctly.
+            - a tuple of `tuple[int, int]` corresponding to the standard radial-angular indexing :math:`(n,m)`. In this case, the `index_convention` parameter is ignored.
+
         Defaults to ``(4, 5, 6, 7, 8, 9, 10, 11)``, correspond to radial order `n` from 2 to 3 (included) and the spherical aberration.
         These correspond to the following aberrations: defocus, astigmatism, coma, trefoil and spherical aberration.
     :param float fc: cutoff frequency ``(NA/emission_wavelength) * pixel_size``. Should be in ``[0, 0.25]``
@@ -244,6 +250,15 @@ class DiffractionBlurGenerator(PSFGenerator):
 
     >>> from deepinv.physics.generator import DiffractionBlurGenerator
     >>> generator = DiffractionBlurGenerator((5, 5), num_channels=3)
+    >>> print("\n".join(generator.zernike_polynomials)) # list of Zernike polynomials used
+    Zernike(n = 2, m = 0) -- Defocus
+    Zernike(n = 2, m = -2) -- Oblique Astigmatism
+    Zernike(n = 2, m = 2) -- Vertical Astigmatism
+    Zernike(n = 3, m = -1) -- Vertical Coma
+    Zernike(n = 3, m = 1) -- Horizontal Coma
+    Zernike(n = 3, m = -3) -- Vertical Trefoil
+    Zernike(n = 3, m = 3) -- Oblique Trefoil
+    Zernike(n = 4, m = 0) -- Primary Spherical
     >>> blur = generator.step()  # dict_keys(['filter', 'coeff', 'pupil'])
     >>> print(blur['filter'].shape)
     torch.Size([1, 3, 5, 5])
@@ -256,7 +271,9 @@ class DiffractionBlurGenerator(PSFGenerator):
         self,
         psf_size: tuple,
         num_channels: int = 1,
-        zernike_index: tuple[int, ...] | tuple[str, ...] = tuple(range(4, 12)),
+        zernike_index: tuple[int, ...] | tuple[tuple[int, int], ...] = tuple(
+            range(4, 12)
+        ),
         fc: float = 0.2,
         max_zernike_amplitude: float = 0.15,
         pupil_size: tuple[int, ...] = (256, 256),
@@ -275,6 +292,7 @@ class DiffractionBlurGenerator(PSFGenerator):
             rng=rng,
         )
         # For backward compatibility if a list / tuple of a string is given
+        # Should be removed in future versions
         zernike_index = list(zernike_index)
         for i, index in enumerate(zernike_index):
             if isinstance(index, str):
@@ -350,7 +368,14 @@ class DiffractionBlurGenerator(PSFGenerator):
             ),
         )
         for k, index in enumerate(self.zernike_index):
-            n, m = Zernike.index_conversion(index, convention=index_convention)
+            if isinstance(index, int):
+                n, m = Zernike.index_conversion(index, convention=index_convention)
+            elif isinstance(index, tuple) and len(index) == 2:
+                n, m = index
+            else:
+                raise ValueError(
+                    f"Zernike index must be either int or tuple of (n,m), got {index!r}"
+                )
             self.Z[:, :, k] = Zernike.cartesian_evaluate(
                 n, m, XX, YY
             )  # defining the k-th Zernike polynomial
@@ -374,6 +399,7 @@ class DiffractionBlurGenerator(PSFGenerator):
         :param int seed: the seed for the random number generator.
 
         :return: dictionary with keys
+
             - `filter`: tensor of size `(batch_size x num_channels x psf_size[0] x psf_size[1])` batch of PSFs,
             - `coeff`: list of sampled Zernike coefficients in this realization,
             - `pupil`: the pupil function,
@@ -399,7 +425,7 @@ class DiffractionBlurGenerator(PSFGenerator):
         if self.random_rotate:
             if angle is None:
                 angle = self.generate_angles(batch_size)
-            psf = rotate_image_via_shear(psf, angle)
+            psf = rotate_via_shear(psf, angle)
 
         if self.apodize:
             psf = self.apodize_mask * psf
@@ -414,7 +440,25 @@ class DiffractionBlurGenerator(PSFGenerator):
             params["angle"] = angle
         return params
 
-    def generate_coeff(self, batch_size):
+    @property
+    def zernike_polynomials(self) -> list[str]:
+        r"""
+        List of Zernike polynomials used in the decomposition, with the corresponding aberration if available.
+        """
+        zernike_polynomials = []
+        for index in self.zernike_index:
+            if isinstance(index, int):
+                n, m = Zernike.index_conversion(index, convention=self.index_convention)
+            elif isinstance(index, tuple) and len(index) == 2:
+                n, m = index
+            else:
+                raise ValueError(
+                    f"Zernike index must be either int or tuple of (n,m), got {index!r}"
+                )
+            zernike_polynomials.append(Zernike.get_name(n, m))
+        return zernike_polynomials
+
+    def generate_coeff(self, batch_size) -> torch.Tensor:
         r"""Generates random coefficients of the decomposition in the Zernike polynomials.
 
         :param int batch_size: batch_size.
@@ -430,7 +474,7 @@ class DiffractionBlurGenerator(PSFGenerator):
         coeff = (coeff - 0.5) * self.max_zernike_amplitude
         return coeff
 
-    def generate_angles(self, batch_size):
+    def generate_angles(self, batch_size) -> torch.Tensor:
         r"""
         Generates random angles for the rotation of the PSF.
 
@@ -439,64 +483,6 @@ class DiffractionBlurGenerator(PSFGenerator):
         :return: a tensor of shape `(batch_size,)` angles in degrees.
         """
         return torch.rand(batch_size, generator=self.rng, **self.factory_kwargs) * 360
-
-
-def rotate_image_via_shear(image: torch.Tensor, angle_deg: torch.Tensor, center=None):
-    r"""
-    2D rotation of image by angle via shear composition through FFT.
-
-    :param torch.Tensor image: input image of shape (B,C,H,W)
-    :param torch.Tensor angle_deg: input rotation angles in degrees of shape (B,)
-    :return: torch.Tensor containing the rotated images of shape (B, C, H, W )
-    """
-    # Convert angle to radians
-    angle = torch.deg2rad(angle_deg)
-    N0, N1 = image.shape[-2:]
-    if center is None:
-        center = (N0 // 2, N1 // 2)
-
-    mask_angles = (angle > torch.pi / 2.0) & (angle <= 3 * torch.pi / 2)
-
-    angle[angle > 3 * torch.pi / 2] -= 2 * torch.pi
-
-    transformed_image = torch.zeros_like(image).expand(mask_angles.shape[0], -1, -1, -1)
-    expanded_image = image.clone().expand(mask_angles.shape[0], -1, -1, -1)
-    transformed_image[~mask_angles] = expanded_image[~mask_angles]
-    transformed_image[mask_angles] = torch.rot90(
-        expanded_image[mask_angles], k=-2, dims=(-2, -1)
-    )
-
-    angle[mask_angles] -= torch.pi
-
-    tant2 = -torch.tan(-angle / 2)
-    st = torch.sin(-angle)
-
-    def shearx(image, shear):
-        fft1 = torch.fft.fft2(image, dim=(-1))
-        freq_1 = torch.fft.fftfreq(N1, d=1.0, device=image.device)
-        freq_0 = (
-            shear[:, None] * (torch.arange(N0, device=image.device) - center[0])[None]
-        )
-        phase_shift = torch.exp(
-            -2j * torch.pi * freq_0[..., None] * freq_1[None, None, :]
-        )
-        image_shear = fft1 * phase_shift[:, None]
-        return torch.abs(torch.fft.ifft2(image_shear, dim=(-1)))
-
-    def sheary(image, shear):
-        fft0 = torch.fft.fft2(image, dim=(-2))
-        freq_0 = torch.fft.fftfreq(N0, d=1.0, device=image.device)
-        freq_1 = (
-            shear[:, None] * (torch.arange(N1, device=image.device) - center[1])[None]
-        )
-        phase_shift = torch.exp(
-            -2j * torch.pi * freq_0[None, :, None] * freq_1[:, None, :]
-        )
-        image_shear = fft0 * phase_shift[:, None]
-        return torch.abs(torch.fft.ifft2(image_shear, dim=(-2)))
-
-    rot = shearx(sheary(shearx(transformed_image, tant2), st), tant2)
-    return rot
 
 
 def cart2pol(x, y):
@@ -708,7 +694,14 @@ class DiffractionBlurGenerator3D(PSFGenerator):
 
     :param tuple psf_size: give in the order `(depth, height, width)` the size of the PSF to generate.
     :param int num_channels: number of channels. Default to `1`.
-    :param tuple[int] zernike_index: list of activated Zernike coefficients.
+    :param tuple[int, ...], tuple[tuple[int, int], ...] zernike_index: activated Zernike coefficients in the following `index_convention` convention.
+        It can be either:
+
+            - a tuple of `int` corresponding to the Noll or ANSI indices, in which case the `index_convention` parameter is required to interpret them correctly.
+            - a tuple of `tuple[int, int]` corresponding to the standard radial-angular indexing :math:`(n,m)`. In this case, the `index_convention` parameter is ignored.
+
+        Defaults to ``(4, 5, 6, 7, 8, 9, 10, 11)``, correspond to radial order `n` from 2 to 3 (included) and the spherical aberration.
+        These correspond to the following aberrations: defocus, astigmatism, coma, trefoil and spherical aberration.
     :param float fc: cutoff frequency `(NA/emission_wavelength) * pixel_size`. Should be in `[0, 1/4]` to respect Shannon, defaults to `0.2`.
     :param float kb: wave number `(NI/emission_wavelength) * pixel_size` or `(NA/NI) * fc`. Must be greater than `fc`. Defaults to `0.3`.
     :param float max_zernike_amplitude: maximum amplitude of Zernike coefficients. Defaults to 0.15.
@@ -726,10 +719,10 @@ class DiffractionBlurGenerator3D(PSFGenerator):
 
     .. note::
 
-        `NA`: numerical aperture, `NI`: refraction index of the immersion medium,
-        `emission_wavelength`: wavelength of the light,
-        `pixel_size`: physical size of the pixels in the xy plane
-        in the same unit as `emission_wavelength`
+        - `NA`: numerical aperture,
+        - `NI`: refraction index of the immersion medium,
+        - `emission_wavelength`: wavelength of the light,
+        - `pixel_size`: physical size of the pixels in the :math:`xy` plane in the same unit as `emission_wavelength`.
 
     |sep|
 
@@ -737,7 +730,9 @@ class DiffractionBlurGenerator3D(PSFGenerator):
 
     >>> import torch
     >>> from deepinv.physics.generator import DiffractionBlurGenerator3D
-    >>> generator = DiffractionBlurGenerator3D((21, 51, 51), stepz_pixel = 2, zernike_index=(3,))
+    >>> generator = DiffractionBlurGenerator3D((21, 51, 51), stepz_pixel = 2, zernike_index=(3,), index_convention='ansi')
+    >>> print(generator.zernike_polynomials) # list of Zernike polynomials used
+    ['Zernike(n = 2, m = -2) -- Oblique Astigmatism']
     >>> dict = generator.step()
     >>> filter = dict['filter']
     >>> print(filter.shape)
@@ -756,7 +751,9 @@ class DiffractionBlurGenerator3D(PSFGenerator):
         self,
         psf_size: tuple,
         num_channels: int = 1,
-        zernike_index: tuple[int | str, ...] = tuple(range(4, 12)),
+        zernike_index: tuple[int, ...] | tuple[tuple[int, int], ...] = tuple(
+            range(4, 12)
+        ),
         fc: float = 0.2,
         kb: float = 0.25,
         max_zernike_amplitude: float = 0.15,
@@ -831,6 +828,7 @@ class DiffractionBlurGenerator3D(PSFGenerator):
         :param int seed: the seed for the random number generator.
 
         :return: dictionary with keys
+
             - `filter`: tensor of size `(batch_size x num_channels x psf_size[0] x psf_size[1])` batch of PSFs,
             - `pupil`: the pupil function,
             - `coeff`: list of sampled Zernike coefficients in this realization,
@@ -865,9 +863,7 @@ class DiffractionBlurGenerator3D(PSFGenerator):
             if angle is None:
                 angle = self.generator2d.generate_angles(batch_size)
 
-            psf = rotate_image_via_shear(
-                rearrange(psf, "b c d h w -> b (c d) h w"), angle
-            )
+            psf = rotate_via_shear(rearrange(psf, "b c d h w -> b (c d) h w"), angle)
             psf = rearrange(psf, "b (c d) h w -> b c d h w", d=self.psf_size[0])
 
         if self.apodize:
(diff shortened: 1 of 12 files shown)


deepinv/physics/generator/blur.py
```diff
@@ -213,47 +242,63 @@ class DiffractionBlurGenerator(PSFGenerator):
...
         super().__init__(
             psf_size=psf_size,
             num_channels=num_channels,
             device=device,
             dtype=dtype,
             rng=rng,
-            **kwargs,
         )
-
-        self.list_param = list_param  # list of parameters to provide
```
jscanvic, line 255: We should deprecate it more gently - I think you can add a property `list_param` that computes the right value based on `self.zernike_index` and yields a warning kinda like here: https://github.com/deepinv/deepinv/pull/764/files#diff-754282ca2703dcddeeb8bc58472eaed96feb0adf6f02170ea5e6e18ef04d16e8R306

deepinv/physics/generator/blur.py
```diff
@@ -213,47 +242,63 @@ class DiffractionBlurGenerator(PSFGenerator):
...
-            "Z4",
-            "Z5",
-            "Z6",
-            "Z7",
-            "Z8",
-            "Z9",
-            "Z10",
-            "Z11",
-        ),
+        zernike_index: tuple[int | str, ...] = tuple(range(4, 12)),
```
jscanvic:
```suggestion
        zernike_index: tuple[int, ...] | tuple[str, ...] = tuple(range(4, 12)),
```

deepinv/physics/generator/blur.py
```diff
@@ -280,246 +325,449 @@ def __init__(
             floor((self.pupil_size[1] - self.psf_size[1]) / 2),
         )
 
-        # a list of indices of the parameters
-        self.index_params = np.sort([int(param[1:]) for param in list_param])
```
jscanvic, line 287: Same, should be deprecated more gently

deepinv/physics/generator/blur.py
```diff
@@ -213,47 +242,63 @@ class DiffractionBlurGenerator(PSFGenerator):
...
             rng=rng,
-            **kwargs,
         )
-
-        self.list_param = list_param  # list of parameters to provide
+        # For backward compatibility if a list / tuple of a string is given
+        zernike_index = list(zernike_index)
+        for i, index in enumerate(zernike_index):
+            if isinstance(index, str):
+                index = int(index[1:])  # Convert "Z4" to 4
```
jscanvic, line 301: Can you raise a `ValueError` if `index[0].upper() != "Z"` while we're at it?

deepinv/physics/generator/blur.py
```diff
@@ -280,246 +325,449 @@ def __init__(
...
+        m = n % 2 + 2 * floor((index - n * (n + 1) / 2 - 1 + (n + 1) % 2) / 2)
+        # Correct sign of m
+        m = m * (-1) ** index
+
+        return n, m
+    else:
+        raise NotImplementedError("Only 'ansi' and 'noll' conventions are implemented.")
+
+
+def rotate_image_via_shear(image: torch.Tensor, angle_deg: torch.Tensor, center=None):
```
jscanvic: The good ol rotation by shear huh!...

Is there a reason we use it instead of bilinear/bicubic rotation?

Does it work well beyond 90° rotations?

deepinv/physics/generator/blur.py
```diff
@@ -280,246 +325,449 @@ def __init__(
...
     )
-    return Z
+
+    angle[mask_angles] -= torch.pi
+
+    tant2 = -torch.tan(-angle / 2)
+    st = torch.sin(-angle)
+
+    def shearx(image, shear):
+        fft1 = torch.fft.fft2(image, dim=(-1))
```
jscanvic: I think FFTs are not computationally efficient for those, could you use torch.gather instead? Here's a reference implementation I wrote a while back:

```
import torch
from math import tan, sin

class RotationByShearing:
    def __call__(self, image, angle):
        assert image.ndim >= 2, "The input tensor image must have at least two dimensions"
        H, W = image.shape[-2:]
        t = -tan(angle / 2)
        s = sin(angle)
        image = self.shear(image, t, -1)
        image = self.shear(image, s, -2)
        image = self.shear(image, t, -1)
        return image

    def shear(self, x, factor, dim):
        dim = (dim + x.ndim) % x.ndim - x.ndim
        assert dim in [-2, -1], "Only last two dimensions are supported"
        tau = torch.arange(x.shape[-dim - 3], device=x.device)
        tau = tau - x.shape[-dim - 3] // 2
        tau = factor * tau
        tau = tau.round().long()
        return self.varying_roll(x, tau, dim)

    def varying_roll(self, x, tau, dim):
        assert x.ndim >= 2, "The input tensor x must have at least two dimensions"
        assert tau.ndim <= 2, "The tensor tau must have at most two dimensions"
        assert tau.ndim != 2, "Batched roll is not supported yet"
        dim = (dim + x.ndim) % x.ndim - x.ndim
        assert dim in [-2, -1], "Only last two dimensions are supported"
        N = x.shape[dim]
        indices = torch.arange(N, device=x.device)
        tau = tau.unsqueeze(2 + dim)
        indices = indices.unsqueeze(-dim - 1)
        indices = indices - tau
        indices = indices % N
        indices = indices.expand(x.shape)
        return torch.gather(x, dim, indices)
```

deepinv/physics/generator/blur.py
```diff
@@ -280,246 +325,449 @@ def __init__(
...
+
+    else:
+        # TODO: For higher order polynomials, explicit formulas are not implemented here.
+        # We fall back to the general definition using recursion.
+        raise NotImplementedError(
+            "Zernike polynomials of radial order > 7 are not implemented yet."
+        )
+
+
+def zernike_index_conversion(index, convention="ansi"):
```
jscanvic: Can you add type hints and make convention keyword only?

```suggestion
def zernike_index_conversion(index: int, *, convention: str = "ansi") -> tuple[int, int]:
```

deepinv/physics/generator/blur.py
```diff
@@ -726,114 +980,152 @@ class DiffractionBlurGenerator3D(PSFGenerator):
...
+        if self.random_rotate:
+            from einops import rearrange
 
-        psf = psf3 / torch.sum(psf3, dim=(-3, -2, -1), keepdim=True)
+            if angle is None:
+                angle = self.generator2d.generate_angles(batch_size)
 
-        return {
+            psf = rotate_image_via_shear(
+                rearrange(psf, "b c d h w -> b (c d) h w"), angle
```
jscanvic: The rearrange simply does `psf.flatten(start_dim=1, end_dim=2)` right? Might be clearer with flatten? Happy with either options!

deepinv/physics/generator/blur.py
```diff
@@ -726,114 +980,152 @@ class DiffractionBlurGenerator3D(PSFGenerator):
...
+            )
+            psf = rearrange(psf, "b (c d) h w -> b c d h w", d=self.psf_size[0])
```
jscanvic, line 867: And this one would be `psf.unflatten(dim=1, sizes=(-1, self.psf_size[0]))` I believe, right?

deepinv/physics/generator/blur.py
```diff
@@ -726,114 +980,152 @@ class DiffractionBlurGenerator3D(PSFGenerator):
...
-    def __update__(self):
-        # self.factory_kwargs = {"device": self.params.device, "dtype": self.params.dtype}
-        self.generator2d.rho = self.generator2d.rho.to(**self.factory_kwargs)
-        self.generator2d.Z = self.generator2d.Z.to(**self.factory_kwargs)
+        self.psf_size = psf_size
+        self.nzs = psf_size[0]
+        self.fc = fc
+        self.zernike_index = zernike_index
+        self.n_zernike = len(self.zernike_index)
+        self.defocus = (
```
jscanvic: If you're happy with the attribute being private (less maintenance!)
```suggestion
        self._defocus = (
```

deepinv/physics/generator/blur.py
```diff
@@ -726,114 +980,152 @@ class DiffractionBlurGenerator3D(PSFGenerator):
...
+            pupil_size=pupil_size,
+            apodize=apodize,
             device=device,
             dtype=dtype,
+            rng=rng,
+            index_convention=index_convention,
             **kwargs,
         )
-
+        self.apodize = apodize
```
jscanvic, line 797: Same here, private?

deepinv/physics/generator/blur.py
```diff
@@ -726,114 +980,152 @@ class DiffractionBlurGenerator3D(PSFGenerator):
...
+        self.random_rotate = random_rotate
```
jscanvic, line 798: Private?

deepinv/physics/generator/blur.py
mh-nguyen712: Thanks!

deepinv/physics/generator/blur.py
mh-nguyen712: thanks! good point I didn't know that

deepinv/physics/generator/blur.py
mh-nguyen712: Yes since bilinear / bicubic will create artefact that we don't want to have in the PSF. I found that shear is good, even beyond 90 degrees

deepinv/physics/generator/blur.py
mh-nguyen712: Thanks for your remark. But the specific rotation of diffraction PSF, we really need precision so FFT is best adaptive. As it does interpolation in freq domain so subpixel accurate. torch.gather will be discontinuous (due to .round()). And this will be differentiable as well

deepinv/physics/generator/blur.py
mh-nguyen712: I think it can be public, just like other attributes

deepinv/physics/generator/blur.py
mh-nguyen712: yes indeed they are equivalent but rearrange is easy to understand. I'm ok to change to flatten as well but it can be helpful for debugging I think

deepinv/physics/generator/blur.py
mh-nguyen712: I totally understand you point. This parameter is very specific and I think there is not much users using it and it can be misleading. In fact, `list_param` can be confusing with the coefficient, but they are not parameters, they are only index of zernike polynomials. And `list_param` is just only used once to construct the basis so I think we could change the name.

deepinv/physics/generator/blur.py
jscanvic: Ooh, really interesting! thanks

mh-nguyen712:
Thanks Jeremy for your review. I made some changes but leave some other points to discuss

deepinv/physics/generator/blur.py
```diff
@@ -280,246 +327,451 @@ def __init__(
...
+    transformed_image = (
+        torch.zeros_like(image).expand(mask_angles.shape[0], -1, -1, -1).clone()
     )
-    Z[37] = lambda x, y: 3 * (
-        70 * r2(x, y) ** 4
-        - 140 * r2(x, y) ** 3
-        + 90 * r2(x, y) ** 2
-        - 20 * r2(x, y)
-        + 1
+    expanded_image = image.clone().expand(mask_angles.shape[0], -1, -1, -1).clone()
```
jscanvic: Why 2 clones?

deepinv/physics/generator/base.py
```diff
@@ -161,7 +161,8 @@ def average(self, n: int = 2000, batch_size: int = 1, **kwargs) -> dict:
             params = self.step(batch_size=n_batch, **kwargs)
             n_processed += n_batch
             params_partial_sum = {
-                k: v.sum(0, keepdim=True) for (k, v) in params.items()
+                k: v.sum(0, keepdim=True) if v is not None else None
```
jscanvic: Why do we need to take care of possibly None values here?

deepinv/physics/generator/blur.py
```diff
@@ -184,21 +186,48 @@ class DiffractionBlurGenerator(PSFGenerator):
...
+    .. math::
+
+        h(\theta) = \left| \mathcal{F} \left( \exp \left( - i 2 \pi \phi_\theta \right) \right) \right|^2,
+
+    where :math:`\phi_\theta : \mathbb{R}^2 \to \mathbb{R}` is defined as:
+
+    .. math::
+        \phi_\theta= \sum_{k \in K} \theta_k z_k
+
+    is the phase aberration expressed in the Zernike basis, :math:`K` is set of indexes of Zernike polynomials used in the decomposition (equal to `zernike_index`).
```
jscanvic: indices*

deepinv/physics/generator/blur.py
```diff
@@ -280,246 +327,451 @@ def __init__(
...
         coeff = torch.rand(
-            (batch_size, len(self.list_param)),
+            (batch_size, len(self.zernike_index)),
             generator=self.rng,
             **self.factory_kwargs,
         )
         coeff = (coeff - 0.5) * self.max_zernike_amplitude
         return coeff
 
+    def generate_angles(self, batch_size):
```
jscanvic: Is it meant to be called from outside the class? If not, could you make it private? Having it listed in the documentation looks confusing to me

deepinv/physics/generator/blur.py
```diff
@@ -280,246 +327,451 @@ def __init__(
...
-    )
-    Z[35] = lambda x, y: (
-        8 * x**2 * y * (3 * r2(x, y) ** 2 - 16 * x**2 * y**2)
-        + 4 * y * (x**2 - y**2) * (r2(x, y) ** 2 - 16 * x**2 * y**2)
-    )
-    Z[36] = lambda x, y: (
-        4 * x * (x**2 - y**2) * (r2(x, y) ** 2 - 16 * x**2 * y**2)
-        - 8 * x * y**2 * (3 * r2(x, y) ** 2 - 16 * x**2 * y**2)
+    out_of_range_error_message = f"The value of m={m} is not valid for n={n}. It should be in the range -n <= m <= n and n - m should be even."
+    if n == 0:
```
jscanvic: Let's hope nothing bad slipped in here!

deepinv/physics/generator/blur.py
jscanvic: Okay I guess we can let it slip then

deepinv/physics/generator/base.py
mh-nguyen712: Before I return None for `angle` in `DiffractionGenerator` when `random_rotate=False`. But then I decided to not return it and forgot to change this one. Thanks

mh-nguyen712: reverted it!

deepinv/physics/generator/blur.py
mh-nguyen712: It's similar to `generate_coeff` so I just put it public. For me it's good to know how random angles are generated in the docs

deepinv/physics/generator/blur.py
mh-nguyen712: It's quite annoying to check all of these formulas I agree :D Trust me we check it multiple time with Florian :D

deepinv/physics/generator/blur.py
mh-nguyen712: Nice catch!

deepinv/physics/generator/blur.py
```diff
@@ -184,21 +186,48 @@ class DiffractionBlurGenerator(PSFGenerator):
...
+    Zernike polynomials are a sequence of orthogonal polynomials defined on the unit disk.
+    They are commonly used in optical systems to describe wavefront aberrations.
+
+    The PSF :math:`h(\theta)` is defined as the squared magnitude of the Fourier transform of the pupil function :math:`\phi_{\theta}` as:
+
+    .. math::
+
+        h(\theta) = \left| \mathcal{F} \left( \exp \left( - i 2 \pi \phi_\theta \right) \right) \right|^2,
+
+    where :math:`\phi_\theta : \mathbb{R}^2 \to \mathbb{R}` is defined as:
```
tachella:
```suggestion
    where :math:`\phi_\theta : \mathbb{R}^2 \to \mathbb{R}`  is the phase aberration expressed in the Zernike basis:
```

deepinv/physics/generator/blur.py
```diff
@@ -184,21 +186,48 @@ class DiffractionBlurGenerator(PSFGenerator):
...
+
+    .. math::
+        \phi_\theta= \sum_{k \in K} \theta_k z_k
+
+    is the phase aberration expressed in the Zernike basis, :math:`K` is set of indices of Zernike polynomials used in the decomposition (equal to `zernike_index`).
```
tachella:
```suggestion
    where :math:`K` is set of indices of Zernike polynomials :math:`z_k` used in the decomposition (equal to `zernike_index`).
```

deepinv/physics/generator/blur.py
```diff
@@ -184,21 +186,48 @@ class DiffractionBlurGenerator(PSFGenerator):
...
+    See :footcite:t:`lakshminarayanan2011zernike` for more details.
```
tachella:
```suggestion
    See :footcite:t:`lakshminarayanan2011zernike` for more details 
```

deepinv/physics/generator/blur.py
```diff
@@ -184,21 +186,48 @@ class DiffractionBlurGenerator(PSFGenerator):
...
+    Conversion from the two conventions to the standard radial-angular indexing is done internally (see `wikipedia page <https://en.wikipedia.org/wiki/Zernike_polynomials>`_).
 
     :param tuple psf_size: the shape ``H x W`` of the generated PSF in 2D
     :param int num_channels: number of images channels. Defaults to 1.
-    :param list[str] list_param: list of activated Zernike coefficients in Noll's convention,
-        defaults to ``["Z4", "Z5", "Z6","Z7", "Z8", "Z9", "Z10", "Z11"]``.
+    :param tuple[int] zernike_index: tuple of activated Zernike coefficients in the following `index_convention` convention.
+        Defaults to ``(4, 5, 6, 7, 8, 9, 10, 11)``, correspond to radial order `n` from 2 to 3 (included) and the spherical aberration.
+        These correspond to the following aberrations: defocus, astigmatism, coma, trefoil and spherical aberration.
     :param float fc: cutoff frequency (NA/emission_wavelength) * pixel_size. Should be in ``[0, 0.25]``
```
tachella:
```suggestion
    :param float fc: cutoff frequency ``(NA/emission_wavelength) * pixel_size``. Should be in ``[0, 0.25]``
```

deepinv/physics/generator/blur.py
```diff
@@ -184,21 +186,48 @@ class DiffractionBlurGenerator(PSFGenerator):
...
         to respect the Shannon-Nyquist sampling theorem, defaults to ``0.2``.
+    :param float max_zernike_amplitude: maximum amplitude of the Zernike coefficients, defaults to ``0.15``.
+        The amplitude of each Zernike coefficient is sampled uniformly in ``[-max_zernike_amplitude/2, max_zernike_amplitude/2]``.
     :param tuple[int] pupil_size: this is used to synthesize the super-resolved pupil.
```
tachella: This is not fully clear to me. Do you mean 
```suggestion
    :param tuple[int] pupil_size: pixel size used to synthesize the super-resolved pupil.
```

deepinv/physics/generator/blur.py
```diff
@@ -184,21 +186,48 @@ class DiffractionBlurGenerator(PSFGenerator):
...
+    (`or this link <https://e-l.unifi.it/pluginfile.php/1055875/mod_resource/content/1/Appunti_2020_Lezione%2014_4_Zernikepolynomialsaguidefinal.pdf>`_).
+
+    The Zernike polynomials :math:`z_k` are indexed using the ``'noll'`` or ``'ansi'`` convention (defined by `index_convention` parameter).
+    Conversion from the two conventions to the standard radial-angular indexing is done internally (see `wikipedia page <https://en.wikipedia.org/wiki/Zernike_polynomials>`_).
 
     :param tuple psf_size: the shape ``H x W`` of the generated PSF in 2D
     :param int num_channels: number of images channels. Defaults to 1.
-    :param list[str] list_param: list of activated Zernike coefficients in Noll's convention,
-        defaults to ``["Z4", "Z5", "Z6","Z7", "Z8", "Z9", "Z10", "Z11"]``.
+    :param tuple[int] zernike_index: tuple of activated Zernike coefficients in the following `index_convention` convention.
```
tachella:
```suggestion
    :param tuple[int] zernike_index: activated Zernike coefficients in the following `index_convention` convention.
```

deepinv/physics/generator/blur.py
```diff
@@ -280,246 +327,449 @@ def __init__(
...
+        angle: torch.Tensor = None,
         seed: int = None,
         **kwargs,
     ):
         r"""
         Generate a batch of PFS with a batch of Zernike coefficients
 
         :param int batch_size: batch_size.
-        :param torch.Tensor coeff: batch_size x len(list_param) coefficients of the Zernike decomposition (defaults is None)
+        :param torch.Tensor coeff: batch_size x len(zernike_index) coefficients of the Zernike decomposition (defaults is `None`)
```
tachella:
```suggestion
        :param torch.Tensor coeff: `batch_size x len(zernike_index)` coefficients of the Zernike decomposition (default is `None`)
```

deepinv/physics/generator/blur.py
```diff
@@ -699,19 +949,25 @@ class DiffractionBlurGenerator3D(PSFGenerator):
...
     :param int num_channels: number of channels. Default to 1.
-    :param list[str] list_param: list of activated Zernike coefficients, defaults to ``["Z4", "Z5", "Z6","Z7", "Z8", "Z9", "Z10", "Z11"]``
+    :param tuple[int] zernike_index: list of activated Zernike coefficients.
     :param float fc: cutoff frequency (NA/emission_wavelength) * pixel_size. Should be in `[0, 1/4]` to respect Shannon, defaults to `0.2`
-    :param float kb: wave number (NI/emission_wavelength) * pixel_size or (NA/NI) * fc. `Must be greater than fc`. Defaults to `0.3`.
+    :param float kb: wave number (NI/emission_wavelength) * pixel_size or (NA/NI) * fc. Must be greater than `fc`. Defaults to `0.3`.
     :param float max_zernike_amplitude: maximum amplitude of Zernike coefficients. Defaults to 0.15.
-    :param tuple[int] pupil_size: this is used to synthesize the super-resolved pupil. The higher the more precise, defaults to (512, 512).
-        If an int is given, a square pupil is considered.
+    :param tuple[int] pupil_size: this is used to synthesize the super-resolved pupil. The higher the more precise, defaults to `(512, 512)`.
```
tachella: same here

deepinv/physics/generator/blur.py
```diff
@@ -699,19 +949,25 @@ class DiffractionBlurGenerator3D(PSFGenerator):
...
-    Fresnel/Fraunhoffer diffraction theory, see :class:`deepinv.physics.generator.DiffractionBlurGenerator` for more details.
+    Fresnel/Fraunhoffer diffraction theory, this class uses :class:`deepinv.physics.generator.DiffractionBlurGenerator` under the hood to generate the pupil function. Refer to its documentation for more details.
 
     :param tuple psf_size: give in the order (depth, height, width)
     :param int num_channels: number of channels. Default to 1.
-    :param list[str] list_param: list of activated Zernike coefficients, defaults to ``["Z4", "Z5", "Z6","Z7", "Z8", "Z9", "Z10", "Z11"]``
+    :param tuple[int] zernike_index: list of activated Zernike coefficients.
     :param float fc: cutoff frequency (NA/emission_wavelength) * pixel_size. Should be in `[0, 1/4]` to respect Shannon, defaults to `0.2`
-    :param float kb: wave number (NI/emission_wavelength) * pixel_size or (NA/NI) * fc. `Must be greater than fc`. Defaults to `0.3`.
+    :param float kb: wave number (NI/emission_wavelength) * pixel_size or (NA/NI) * fc. Must be greater than `fc`. Defaults to `0.3`.
```
tachella:
```suggestion
    :param float kb: wave number `(NI/emission_wavelength) * pixel_size` or `(NA/NI) * fc`. Must be greater than `fc`. Defaults to `0.3`.
```

deepinv/physics/generator/blur.py
```diff
@@ -974,14 +1260,19 @@ def step(
         Generate a batch of 3D confocal PSF with a batch of Zernike coefficients
         for illumination and collection
 
-        :return: dictionary with keys **'filter'**: tensor of size (batch_size x num_channels x psf_size[0] x psf_size[1]) batch of psfs,
-              **'coeff_ill'**: list of sampled Zernike coefficients in this realization of illumination,
-              **'coeff_coll'**: list of sampled Zernike coefficients in this realization of collection,
+        :param int batch_size: number of PSFs to generate.
+        :param torch.Tensor coeff_ill: tensor of size (batch_size x len(zernike_index)) containing the Zernike coefficients for illumination.
```
tachella:
```suggestion
        :param torch.Tensor coeff_ill: tensor of size `batch_size x len(zernike_index)` containing the Zernike coefficients for illumination.
```

deepinv/physics/generator/blur.py
```diff
@@ -974,14 +1260,19 @@ def step(
...
+            If `None`, random coefficients are generated.
+        :param torch.Tensor coeff_coll: tensor of size (batch_size x len(zernike_index)) containing the Zernike coefficients for collection.
```
tachella:
```suggestion
        :param torch.Tensor coeff_coll: tensor of size `batch_size x len(zernike_index)` containing the Zernike coefficients for collection.
```

deepinv/physics/generator/blur.py
```diff
@@ -974,14 +1260,19 @@ def step(
...
+            If `None`, random coefficients are generated.
+
+        :return: dictionary with keys
+            `filter`: tensor of size `(batch_size x num_channels x psf_size[0] x psf_size[1])` batch of PSFs,
```
tachella:
```suggestion
            `filter`: tensor of size `batch_size x num_channels x psf_size[0] x psf_size[1]` batch of PSFs,
```

tachella:
testing that no bot posts

tachella:
@mh-nguyen712 can you merge main to see if the GPU docs error is solved?

deepinv/physics/generator/blur.py
```diff
@@ -187,21 +189,48 @@ class DiffractionBlurGenerator(PSFGenerator):
...
+    The Zernike polynomials :math:`z_k` are indexed using the ``'noll'`` or ``'ansi'`` convention (defined by `index_convention` parameter).
+    Conversion from the two conventions to the standard radial-angular indexing is done internally (see `wikipedia page <https://en.wikipedia.org/wiki/Zernike_polynomials>`_).
 
     :param tuple psf_size: the shape ``H x W`` of the generated PSF in 2D
     :param int num_channels: number of images channels. Defaults to 1.
-    :param list[str] list_param: list of activated Zernike coefficients in Noll's convention,
-        defaults to ``["Z4", "Z5", "Z6","Z7", "Z8", "Z9", "Z10", "Z11"]``.
-    :param float fc: cutoff frequency (NA/emission_wavelength) * pixel_size. Should be in ``[0, 0.25]``
+    :param tuple[int] zernike_index: activated Zernike coefficients in the following `index_convention` convention.
+        Defaults to ``(4, 5, 6, 7, 8, 9, 10, 11)``, correspond to radial order `n` from 2 to 3 (included) and the spherical aberration.
```
tachella, line 231: counting the 11th is not adding radial order n=4?

image

deepinv/physics/generator/blur.py
```diff
@@ -216,47 +245,65 @@ class DiffractionBlurGenerator(PSFGenerator):
...
+            lin_0 = torch.linspace(
+                -psf_size[0] // 2, psf_size[0] // 2, psf_size[0], **self.factory_kwargs
+            )
+            lin_1 = torch.linspace(
+                -psf_size[1] // 2, psf_size[1] // 2, psf_size[1], **self.factory_kwargs
+            )
+            XX0, XX1 = torch.meshgrid(lin_0, lin_1, indexing="ij")
+            dist = (XX0**2 + XX1**2) ** 0.5
+            radius = min(psf_size) / 2
+            apodize_length = 10
```
tachella: should we add an assert here checking that radius > apodize_length?

deepinv/physics/generator/blur.py
```diff
@@ -283,246 +330,449 @@ def __init__(
...
 
         :param int batch_size: batch_size.
-        :param torch.Tensor coeff: batch_size x len(list_param) coefficients of the Zernike decomposition (defaults is None)
+        :param torch.Tensor coeff: `batch_size x len(zernike_index)` coefficients of the Zernike decomposition (default is `None`)
+        :param torch.Tensor angle: `batch_size` angles in degree to rotate the PSF (defaults is `None`)
         :param int seed: the seed for the random number generator.
 
-        :return: dictionary with keys **'filter'**: tensor of size (batch_size x num_channels x psf_size[0] x psf_size[1]) batch of psfs,
-            **'coeff'**: list of sampled Zernike coefficients in this realization, **'pupil'**: the pupil function
+        :return: dictionary with keys
```
tachella, line 401: add bullet points to display better the dictionary:
```
- `filter`: tensor of size `(batch_size x num_channels x psf_size[0] x psf_size[1])` batch of PSFs,
- `coeff`: list of sampled Zernike coefficients in this realization,
- `pupil`: the pupil function,
- `angle`: the random rotation angle in degrees if `random_rotate` is `True`, nothing otherwise.
```

examples/physics/demo_microscopy_3d.py
```diff
@@ -132,8 +132,8 @@
 # %%
 # It is also possible to directly specify the Zernike decomposition.
```
tachella, line 138: I've spotted a small typo in this example, 
image


Can you add code style to `lambda`?

deepinv/physics/generator/blur.py
```diff
@@ -283,246 +330,449 @@ def __init__(
...
+        m = n % 2 + 2 * floor((index - n * (n + 1) / 2 - 1 + (n + 1) % 2) / 2)
+        # Correct sign of m
+        m = m * (-1) ** index
+
+        return n, m
+    else:
+        raise NotImplementedError("Only 'ansi' and 'noll' conventions are implemented.")
+
+
+def rotate_image_via_shear(image: torch.Tensor, angle_deg: torch.Tensor, center=None):
```
tachella: this is a quite cool function, should we move it to the functional and add it to the API?

deepinv/physics/generator/blur.py
```diff
@@ -702,175 +952,221 @@ class DiffractionBlurGenerator3D(PSFGenerator):
     r"""
     3D diffraction limited kernels using Zernike decomposition of the phase mask.
```
tachella: I would add here a more detailed explanation of the PSF definition as you have added to DiffractionBlurGenerator

deepinv/physics/generator/blur.py
```diff
@@ -702,175 +952,221 @@ class DiffractionBlurGenerator3D(PSFGenerator):
...
 
-        :return: dictionary with keys **'filter'**: tensor of size (batch_size x num_channels x psf_size[0] x psf_size[1]) batch of psfs,
-             **'pupil'**: the pupil function, **'coeff'**: list of sampled Zernike coefficients in this realization,
+        :param int batch_size: number of PSFs to generate.
+        :param torch.Tensor coeff: tensor of size (batch_size x len(zernike_index)) containing the Zernike coefficients.
+            If `None`, random coefficients are generated.
+        :param int seed: the seed for the random number generator.
+
+        :return: dictionary with keys
+            `filter`: tensor of size `(batch_size x num_channels x psf_size[0] x psf_size[1])` batch of PSFs,
```
tachella: write with bulletpoints

examples/physics/demo_microscopy_3d.py
```diff
@@ -132,8 +132,8 @@
...
 # For instance, if the pupil is null, the PSF is the Airy pattern.
-n_zernike = len(
-    diffraction_generator.list_param
+n_zernike = (
```
tachella: this property is a bit hidden - we should somewhere tell the user in the doc that they can access this info like this. Or otherwise add a get_zernike_list() method?

deepinv/physics/generator/blur.py
mh-nguyen712: the 11th polynomial corresponds to the spherical (the central one when n = 4) and not other ones.

deepinv/physics/generator/blur.py
mh-nguyen712: done

deepinv/physics/generator/blur.py
mh-nguyen712: Agree but I think it makes more sense to put in deepinv.transform.functional but for now we don't have the transform functional yet. It's a bit weird to put it in physics functional...

examples/physics/demo_microscopy_3d.py
mh-nguyen712: thanks!

deepinv/physics/generator/blur.py
mh-nguyen712: added

examples/physics/demo_microscopy_3d.py
mh-nguyen712: it's just the `len(zernike_index)`. However it could be cool to have a function to return the name of the aberrations. But it requires a bit of more works

deepinv/physics/generator/blur.py
tachella: ahh I see, thanks!

deepinv/physics/generator/blur.py
tachella: maybe we could add it in [here](https://github.com/deepinv/deepinv/blob/ea9296e875ef65bc44276dde4df1d43e86fd7587/deepinv/transform/rotate.py#L8), add it as an option in `Rotate`, and import it from there? 
Is it faster than `torchvision`'s rotate?

examples/physics/demo_microscopy_3d.py
tachella: my main concern is having a function defined somewhere in the generator that tells how to access this information, this is why maybe a `generator.get_zernike_n()` or similar would be enough (otherwise I feel that the users need to dig inside the function or go over this specific example.

deepinv/physics/generator/blur.py
mh-nguyen712: Added. 
`torchvision`'s rotate use `grid_sample` under the hood, which is faster than this I think. But this rotate_by_shear is better for subpixel details (by keeping all frequencies rather than interpolation in grid sample) 

Example: 
image

examples/physics/demo_microscopy_3d.py
mh-nguyen712: I added a class property to retrieve Zernike polynomials (and the corresponding aberration if available)

PR #827, opened 2025-10-07
Typing optim

nucli-vicky:
Some notes:
- I did not type `Bregman_ICNN` init because it was not annotated. I assume `forward_model` has to be `nn.Module` but wanted to check.
- `ZeroDistance` returns a float (0.0) for `fn`. Is this intended? We could replace by `torch.zeros(x.size(0), device=x.device, dtype=x.dtype)` like done in `ZeroFidelity` or a singleton zero tensor with device and dtype still correct.
- I did not type a lot of the dicts used in `optimizers.py` as it seems quite variable.
- `utils.py` still partially to do.


deepinv/optim/bregman.py
```diff
@@ -10,10 +12,10 @@ class Bregman(Potential):
     :param Callable h: Potential function :math:`\phi(x)` to be used in the Bregman framework.
     """
 
-    def __init__(self, phi=None):
+    def __init__(self, phi: Callable[torch.Tensor, torch.Tensor] = None):
```
mh-nguyen712:
```suggestion
    def __init__(self, phi: Callable[[torch.Tensor], torch.Tensor] = None):
```
if you meant a fn take torch.Tensor as argument as return torch.Tensor

deepinv/optim/data_fidelity.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
+from typing import Callable
```
mh-nguyen712: Can you combine this with the line below? Thanks

deepinv/optim/data_fidelity.py
```diff
@@ -122,11 +135,13 @@ class StackedPhysicsDataFidelity(DataFidelity):
 
     """
 
-    def __init__(self, data_fidelity_list):
+    def __init__(self, data_fidelity_list: list[DataFidelity]):
```
mh-nguyen712, line 138:
```suggestion
    def __init__(self, data_fidelity_list: list[DataFidelity, ...]):
```
I think it has varying length

deepinv/optim/data_fidelity.py
```diff
@@ -253,12 +272,20 @@ class L2(DataFidelity):
...
         self.norm = 1 / (sigma**2)
 
-    def prox(self, x, y, physics, *args, gamma=1.0, **kwargs):
+    def prox(
+        self,
+        x: torch.Tensor,
+        y: torch.Tensor,
+        physics: Physics,
+        *args,
+        gamma: float = 1.0,
```
mh-nguyen712: gamma can be a float or a tensor (batched or not)

deepinv/optim/data_fidelity.py
```diff
@@ -483,42 +510,13 @@ class LogPoissonLikelihood(DataFidelity):
     :param float mu: normalization constant
     """
 
-    def __init__(self, N0=1024.0, mu=1 / 50.0):
+    def __init__(self, N0: float = 1024.0, mu: float = 1 / 50.0):
         super().__init__()
         self.d = LogPoissonLikelihoodDistance(N0=N0, mu=mu)
         self.mu = mu
         self.N0 = N0
```
mh-nguyen712, line 754: Can you remove this code in `if __name__ == "__main__"` as well please.

deepinv/optim/dpir.py
```diff
@@ -58,7 +59,7 @@ def __init__(
         self,
         sigma: float | Tensor = 0.1,
         denoiser: Denoiser = None,
-        device="cuda",
+        device: str | torch.device = "cuda",
```
mh-nguyen712:
```suggestion
        device: str | torch.device = "cpu",
```

deepinv/optim/epll.py
```diff
@@ -85,7 +90,7 @@ def __init__(
 
     def forward(
         self,
-        y,
+        y: torch.Tensor,
```
mh-nguyen712, line 93: Can you add type for the physics as well please

deepinv/optim/phase_retrieval.py
```diff
@@ -102,10 +107,10 @@ def cosine_similarity(a: torch.Tensor, b: torch.Tensor):
 def spectral_methods(
     y: torch.Tensor,
     physics,
-    x=None,
-    n_iter=50,
+    x: torch.Tensor = None,
+    n_iter: int = 50,
     preprocessing=default_preprocessing,
-    lamb=10.0,
+    lamb: float = 10.0,
```
mh-nguyen712, line 115: can you add type for `processing: Callable` as well?

mh-nguyen712 (commented):
Thank a lot for this PR. I have some minor comments. 
- For ZeroDistance, I think 0.0 is fined as it will work for all tensor, but it's cleaner to return zeros tensor as you suggested

nucli-vicky:
Thanks for the suggestions @mh-nguyen712 ! Incorporated them. I also did some fixes with regard to docstring <--> code mismatches (argument did not exist in code but was included in docstring etc.)

deepinv/optim/optimizers.py
```diff
@@ -146,27 +152,27 @@ class BaseOptim(Reconstructor):
 
     def __init__(
         self,
-        iterator,
+        iterator: OptimIterator,
         params_algo=MappingProxyType({"lambda": 1.0, "stepsize": 1.0}),
         data_fidelity=None,
```
samuro95: Could you add Prior and data_fidelity types ?

deepinv/optim/optimizers.py
```diff
@@ -146,27 +152,27 @@ class BaseOptim(Reconstructor):
...
         prior=None,
-        max_iter=100,
-        crit_conv="residual",
+        max_iter: int = 100,
+        crit_conv: str = "residual",
         thres_conv=1e-5,
```
samuro95: float type here

deepinv/optim/optimizers.py
```diff
@@ -146,27 +152,27 @@ class BaseOptim(Reconstructor):
...
-        backtracking=False,
-        gamma_backtracking=0.1,
-        eta_backtracking=0.9,
+        early_stop: bool = False,
+        has_cost: bool = False,
+        backtracking: bool = False,
+        gamma_backtracking: float = 0.1,
+        eta_backtracking: float = 0.9,
         custom_metrics=None,
         custom_init=None,
```
samuro95: This is a Callable that takes y and physics and outputs a dictionary (see docstring).

deepinv/optim/optimizers.py
```diff
@@ -146,27 +152,27 @@ class BaseOptim(Reconstructor):
...
         get_output=lambda X: X["est"][0],
```
samuro95: This is a Callable again

deepinv/optim/optimizers.py
```diff
@@ -270,7 +276,7 @@ def __init__(
 
         self.psnr = PSNR()
 
-    def update_params_fn(self, it):
+    def update_params_fn(self, it: int):
```
samuro95: Could you add the output type in these functions ?

deepinv/optim/optimizers.py
samuro95, line 1: It is missing some typings here. If you need, I can help when it is not clear.

samuro95 (changes requested):
Thanks for the amazing work ! The optimizers.py file is missing some small type elements. Do not hesitate to contact me (on discord ?) you need help on this file.

nucli-vicky:
Thanks @samuro95 , I tried my best to finalize the typing in `optimizers.py`. 
(i) `X` and `X_prev` are always `dict`, but their typing is not always the same, so I left it simply as `dict` without specifying other things.
(ii) `F_fn`'s arguments are extremely long. Do we revert to just typing it as `Callable` ?

samuro95 (approved):
Maybe we can let F_fn to be sumply Callable indeed. Sorry for the delay on this. I 'll detail the F_fn arguments in #572.

nucli-vicky:
@mh-nguyen712 @samuro95 I think we can leave `F_fn` as is now (for correctness), and wait until some `F_fn` base exists so we can use that for typing and reduce the verbosity.
Good to go?

samuro95:
Yes good to me.

PR #592, opened 2025-07-08
Reformat the way optimization algorithms are defined 

samuro95:
In this PR, we give a simpler way to call optimization and unfolded algorithms. 

Now you can create a model for each algorithm of the library by just calling it by its own name. Moreover, no need to create a dictionary of parameters anymore, you can just call the main parameters (stepsize, lambda_reg, g_param, beta...) of each algorithm in the function call. 

Also, to make the algorithm unfolded or DEQ, one can just set unfolded and/or DEQ to True in the function call 

For example 

`
model = ProximalGradientDescent(
    unfold=True,
    trainable_params=trainable_params,
    data_fidelity=data_fidelity,
    max_iter=max_iter,
    prior=prior,
    stepsize=1.,
    g_param=0.1,
    lambda_reg=0.1
)
`

### Checks to be done before submitting your PR

- [ ] `python3 -m pytest deepinv/tests` runs successfully.
- [ ] `black .` runs successfully.
- [ ] `make html` runs successfully (in the `docs/` directory).
- [ ] Updated docstrings related to the changes (as applicable).
- [ ] Added an entry to the [CHANGELOG.rst](../CHANGELOG.rst).

PR #829, opened 2025-10-08
Learned Regularizers and nmAPG

johertrich:
Adds some learned regularizers ((weakly) convex ridge regularizers and least squares residual) and an efficient algorithm to solve the corresponding variational problem (nonmonotonic accelerated proximal gradient algorithm, basically gradient descent + momentum + line search). For the regularizers, there are pretrained weights available for grayscale and color images trained on BSDS500.


deepinv/optim/optimizers.py
```diff
@@ -628,3 +630,143 @@ def optim_builder(
 
 def str_to_class(classname):
     return getattr(sys.modules[__name__], classname)
+
+
+class NMAPG(Reconstructor):
```
mh-nguyen712: Personally I prefer a more explicit name (eg `NonmonotonicAcceleratedPGD`) but I'm happy if it was chosen like this in the paper

deepinv/optim/optimizers.py
```diff
@@ -628,3 +630,143 @@ def optim_builder(
...
+        lambda_reg=1.0,
+        max_iter=2000,
+        custom_init=None,
+        L_init=1.0,
+        thres_conv=1e-4,
+        rho=0.9,
+        delta=0.1,
+        eta=0.8,
+        gradient_for_both=True,
+        verbose=False,
```
mh-nguyen712:
```suggestion
        data_fidelity: DataFidelity=None,
        prior: Prior=None,
        lambda_reg: Tensor | float=1.0,
        max_iter: int=2000,
        custom_init: Callable=None,
        L_init: float=1.0,
        thres_conv:float=1e-4,
        rho:float=0.9,
        delta:float=0.1,
        eta:float=0.8,
        gradient_for_both:bool=True,
        verbose:bool=False,
```

We're trying to add typing / type hint everywhere in the library. could you please add the typing?

deepinv/optim/optimizers.py
```diff
@@ -628,3 +630,143 @@ def optim_builder(
...
+        self.L_init = L_init
+        self.tol = thres_conv
+        self.rho = rho
+        self.delta = delta
+        self.eta = eta
+        self.gradient_for_both = gradient_for_both
+        self.verbose = verbose
+        self.custom_init = custom_init
+
+    def forward(self, y, physics, compute_stats=False, x_init=None):
```
mh-nguyen712, line 768:
```suggestion
    @torch.no_grad()
    def forward(self, y, physics, compute_stats=False, x_init=None):
```
you can add `@torch.no_grad()` decorators and remove all `.detach()`. It make sure that no gradients will be recorded

deepinv/optim/prior.py
```diff
@@ -705,3 +709,387 @@ def prox(self, x, *args, gamma=1.0, **kwargs):
         # 1 - gamma/max(z, gamma) = relu(z - gamma) / z, adding 1e-12 to avoid division by 0
         z = torch.nn.functional.relu(z - gamma) / (z + 1e-12)
         return z * x
+
+
+class WCRR(Prior):
```
mh-nguyen712: Same here I would prefer a more explicit name

deepinv/optim/optimizers.py
```diff
@@ -628,3 +630,143 @@ def optim_builder(
...
+            prox_g = None
+        else:
+            f = lambda x, y: self.lambda_reg * self.prior(x).detach()
+            nabla_f = lambda x, y: self.lambda_reg * self.prior.grad(x).detach()
+            f_and_nabla = lambda x, y: (f(x, y), nabla_f(x, y))
+            g = lambda x, y: self.data_fidelity(x, y, physics).detach()
+            prox_g = lambda x, y, gamma: self.data_fidelity.prox(
+                x, y, physics, gamma=gamma
+            ).detach()
+        if self.custom_init is None:
```
mh-nguyen712: `x_init` is not used here

deepinv/optim/optimizers.py
mh-nguyen712: And why we cannot inherit it from BaseOptim? maybe I'm missing something

deepinv/optim/prior.py
```diff
@@ -705,3 +709,387 @@ def prox(self, x, *args, gamma=1.0, **kwargs):
...
+    r"""
+    (Weakly) Convex Ridge Regularizer :math:`\reg{x}=\sum_{c} \psi_c(W_c x)`
+
+    for filters :math:`W_c` and potentials :math:`\psi_c`. The filters :math:`W_c` are realized by a concatination multiple convolution
```
mh-nguyen712:
```suggestion
    for filters :math:`W_c` and potentials :math:`\psi_c`. The filters :math:`W_c` are realized by a concatenation multiple convolution
```

deepinv/optim/prior.py
```diff
@@ -705,3 +709,387 @@ def prox(self, x, *args, gamma=1.0, **kwargs):
...
+        self.output_scaling = nn.Parameter(
+            torch.tensor(0.0, device=device)
+        ).requires_grad_(weak_convexity == 0.0)
+        self.weak_cvx = weak_convexity
+
+        if pretrained is not None:
+            if pretrained == "download":
+                if in_channels == 1 and weak_convexity == 0.0:
+                    file_name = "CRR_gray.pt"
+                    url = "https://drive.google.com/uc?export=download&id=1Yz2eSCM85EaGQTDviPnmqMY1ySqti3hr"
```
mh-nguyen712, line 807: I would be great if you can upload these weight to the deepinv huggingface hub, or just need to make sure that the google drive will not be erased

deepinv/optim/prior.py
```diff
@@ -705,3 +709,387 @@ def prox(self, x, *args, gamma=1.0, **kwargs):
...
+                        "Weights are only available for weak_convexity in [0.0, 1.0] and in_channels in [1, 3]!"
+                    )
+                weights = torch.hub.load_state_dict_from_url(
+                    url, map_location=lambda storage, loc: storage, file_name=file_name
+                )
+                self.load_state_dict(weights, strict=True)
+            else:
+                self.load_state_dict(torch.load(pretrained, map_location=device))
+
+    def smooth_l1(self, x):
```
mh-nguyen712: I feel that it can be a private method. Otherwise, can you add docstring please

deepinv/optim/prior.py
```diff
@@ -705,3 +709,387 @@ def prox(self, x, *args, gamma=1.0, **kwargs):
...
+        return torch.clip(x**2, 0.0, 1.0) / 2 + torch.clip(torch.abs(x), 1.0) - 1.0
+
+    def grad_smooth_l1(self, x):
```
mh-nguyen712: same here and some functions below

deepinv/optim/prior.py
```diff
@@ -705,3 +709,387 @@ def prox(self, x, *args, gamma=1.0, **kwargs):
...
+        return torch.clip(x, -1.0, 1.0)
+
+    def get_conv_lip(self):
+        impulse = self.filters(self.dirac)
+        for filt in reversed(self.filters):
+            impulse = F.conv_transpose2d(impulse, filt.weight, padding=filt.padding)
+        return torch.fft.fft2(impulse, s=[256, 256]).abs().max()
```
mh-nguyen712: maybe use `rfft2` for real tensor

deepinv/optim/prior.py
```diff
@@ -705,3 +709,387 @@ def prox(self, x, *args, gamma=1.0, **kwargs):
...
+            self.output_scaling - 2 * self.scaling - 2 * self.input_scaling
+        )
+        reg = reg.sum(dim=(1, 2, 3))
+        return reg
+
+    def _apply(self, fn):
+        self.dirac = fn(self.dirac)
+        return super()._apply(fn)
+
+    def prox(self, x, *args, gamma=1.0, **kwargs):
```
mh-nguyen712, line 914: You can use the decorator `@torch.no_grad()` and remove `.detach()`. And can you add a warning or a note saying that this prox is not differentiable?

deepinv/optim/prior.py
```diff
@@ -705,3 +709,387 @@ def prox(self, x, *args, gamma=1.0, **kwargs):
...
+                0.5 * torch.sum((z - y) ** 2, (1, 2, 3)) + out_f.detach(),
+                z - y + out_grad.detach(),
+            )
+
+        return nonmonotone_accelerated_proximal_gradient(
+            x, f, nabla_f=nabla_f, f_and_nabla=f_and_nabla, y=x
+        )[0]
+
+
+class LSR(Prior):
```
mh-nguyen712: If it's specific to GSDRUNet, can you give it another name?

deepinv/optim/prior.py
```diff
@@ -705,3 +709,387 @@ def prox(self, x, *args, gamma=1.0, **kwargs):
...
+    The specific implementation wraps the :class:`GSDRUNet<deepinv.models.GSPnP.GSDRUNet>`.
+
+    :param int in_channels: Number of input channels (`1` for gray valued images, `3` for color images). Default: `3`
+    :param str device: Device for the weights. Default: `"cpu"`
+    :param str, None pretrained: use pretrained weights. If ``pretrained=None``, the weights will be initialized at random
+        using Pytorch's default initialization. If ``pretrained='download'``, the weights will be downloaded from an
+        online repository (only available for the default architecture with 3 or 1 input/output channels).
+        Finally, ``pretrained`` can also be set as a path to the user's own pretrained weights.
+        See :ref:`pretrained-weights <pretrained-learned-reg>` for more details.
+    :param list of int nc: number of channels of the DRUNet, cf. :class:`deepinv.models.DRUNet`. Default: `[32, 64, 128, 256]`
```
mh-nguyen712:
```suggestion
    :param list of int nc: number of channels of the DRUNet, cf. :class:`deepinv.models.GSDRUNet`. Default: `[32, 64, 128, 256]`
```
I would use GSDRUNet here otherwise it's confusing, even they have the same parameters

deepinv/optim/prior.py
```diff
@@ -705,3 +709,387 @@ def prox(self, x, *args, gamma=1.0, **kwargs):
...
+        :param torch.Tensor x: Variable :math:`x` at which the prior is computed.
+        :return: (:class:`torch.Tensor`) prior :math:`\reg{x}`.
+        """
+        return torch.exp(
+            self.output_scaling + self.input_scaling
+        ) * self.model.potential(
+            torch.exp(self.input_scaling) * x.contiguous(), self.sigma
+        )
+
+    def prox(self, x, *args, gamma=1.0, **kwargs):
```
mh-nguyen712, line 1052: same here use decorator

mh-nguyen712 (changes requested):
Many thanks for this PR, the example look really good! I can not review all changes in one go. Here is my first comments, mostly related to docs and typing.

deepinv/optim/optimizers.py
johertrich: Inheriting from BaseOptim would mean that we inherit all the submethods which don't make sense for the nmAPG since it does not rely on an Iterator object. Theoretically, one could define the nmAPG via such an Iterator, but I would consider this as a very "dirty" solution since many arguments of the BaseOptim don't make sense in this context and we would have to maintain a large dict with parameters from previous iterations. Moreover, one would have to define an exception for the first iteration as the Barzilai-Borwein step sizes are only applicable from the second iteration onward.

So short: Fusing the nmAPG with the BaseOptim would lead to something which is unintuitive to use and  to maintain. Therefore, I orientated a bit on the definition of optimization algorithms in #592.

deepinv/optim/prior.py
johertrich: We can do either way, I am not sure how uploading them to huggingface works though. I guess I don't have permissions to upload them?

deepinv/optim/prior.py
johertrich: I am a bit hesitant to name it after the GSDRUNet, as it hasn't to do much with a gradient step or a denoiser. It is just using the same NN architecture as the GS denoiser... The naming comes from [this paper](https://arxiv.org/abs/2303.05386). If you prefer, we can use the full name `LeastSquaresResidual`, but I wouldn't use the term GSDRUNet.

mh-nguyen712: Agree. So I think you can just pass a general argument `denoiser` for `D` and by default it's the GSDRUNet and remove other arguments related to GSDRUNet in the `__init__`, as you said that it hasn't much to do with DRUNet. And then I would prefer the explicit name `LeastSquaresResidual`

mh-nguyen712: And can you also add typing please. thanks!

deepinv/optim/optimizers.py
mh-nguyen712: Thanks for the clarification!

deepinv/optim/prior.py
mh-nguyen712: ah sorry. I'm happy to leave them in the drive. I'll ask one of the maintainers to upload them to the hub later

deepinv/optim/utils.py
```diff
@@ -1396,3 +1397,276 @@ def _EM_step(self, dataloader, verbose):
...
+    The algorithm admits convergence guarantees towards a stationary point also in the case of non-convex functions
+    :math:`f` and :math:`g`.
+    It combines the proximal gradient algorithm with momentum and a line search.
+
+    In our implementation, the functions :math:`F`, :math:`f` and :math:`g` depends on some additional parameter :math:`y\in\mathbb{R}^d`.
+    Moreover, the implementation uses batching over the parameters :math:`y` such that several problems of the form
+
+    .. math::
+        \begin{equation*}
+        \min_y F(x,y_i)=f(x,y_i)+\lambda g(x,y_i),\quad i=1,...,\mathrm{batch size}
```
mh-nguyen712:
```suggestion
        \min_{y_i} F(x,y_i)=f(x,y_i)+\lambda g(x,y_i),\quad i=1,...,\mathrm{batch\_size}
```

deepinv/optim/utils.py
```diff
@@ -1396,3 +1397,276 @@ def _EM_step(self, dataloader, verbose):
...
+
+    # initialize variables
+    x = x0.clone()  # Noation of the paper: x1
+    x_old = x.clone()  # x0
+    z = x0.clone()  # z1
+    t = 1.0  # t1
+    t_old = 0.0  # t0
+    q = 1.0  # q1
+    c = f(x, y) + weighting * g(x, y)  # c1
+    L = torch.full((x.shape[0],), L_init, dtype=torch.float32, device=x.device)
```
mh-nguyen712:
```suggestion
    L = torch.full((x.shape[0],), L_init, dtype=x.dtype, device=x.device)
```

deepinv/optim/utils.py
```diff
@@ -1396,3 +1397,276 @@ def _EM_step(self, dataloader, verbose):
...
+        Note that `prox_g` must not be `None` if `g is not None`. Default: `None`.
+    :param float weighting: Parameter :math:`\lambda` from the objective value. Default: `1.0`
+    :param int max_iter: Maximal number of iterations. Default: `2000`
+    :param float L_init: Initial guess of the (local) Lipschitz constant of :math:`\nabla_x f(x,y)`. Default: `1.0`
+    :param float tol: Convergence crieterion. The algorithm is stopped if the relative residual between two iterates is smaller than `tol`. Default: `1e-4`
+    :param float rho: Decrease ratio of the step size if the line search fails. Default: `0.9`
+    :param float delta: Quadratic decay parameter for the line search. Should be :math:`>0`. Default: `0.1`
+    :param float eta: Non-monotonicity parameter for the line search. Should be in :math:`[0,1)`. Default: `0.8`
+    :param bool verbose: Set to `True` to print the number of iterations used in the algorithm. Default: `False`.
+    :return: Tuple out of the approximated minimizer `x` of :math:`F`, the estimated local Lipschitz constant `L`, the number `i` of used iterations and
```
mh-nguyen712: It's missing a return value (function returns 5 but docstring only has 4)

deepinv/optim/utils.py
```diff
@@ -1396,3 +1397,276 @@ def _EM_step(self, dataloader, verbose):
...
+        f = lambda x, y: f_(x)
+        if nabla_f is not None:
+            nabla_f_ = nabla_f
+            nabla_f = lambda x, y: nabla_f_(x, y)
+        if g is not None:
+            g_ = g
+            g = lambda x, y: g_(x)
+        if prox_g is not None:
+            prox_g_ = prox_g
+            prox_g = lambda x, y, param: prox_g(x, param)
```
mh-nguyen712:
```suggestion
            prox_g = lambda x, y, param: prox_g_(x, param)
```

deepinv/optim/utils.py
```diff
@@ -1396,3 +1397,276 @@ def _EM_step(self, dataloader, verbose):
...
+    :param float L_init: Initial guess of the (local) Lipschitz constant of :math:`\nabla_x f(x,y)`. Default: `1.0`
+    :param float tol: Convergence crieterion. The algorithm is stopped if the relative residual between two iterates is smaller than `tol`. Default: `1e-4`
+    :param float rho: Decrease ratio of the step size if the line search fails. Default: `0.9`
+    :param float delta: Quadratic decay parameter for the line search. Should be :math:`>0`. Default: `0.1`
+    :param float eta: Non-monotonicity parameter for the line search. Should be in :math:`[0,1)`. Default: `0.8`
+    :param bool verbose: Set to `True` to print the number of iterations used in the algorithm. Default: `False`.
+    :return: Tuple out of the approximated minimizer `x` of :math:`F`, the estimated local Lipschitz constant `L`, the number `i` of used iterations and
+        a bool `converged` indicating whether the algorithm reached the convergence criterion or not.
+    """
+    if y is None:
```
mh-nguyen712, line 1484: Can you clarify in the docs the signature of the functions `f, g, nabla_f, prox_g`? I have the feeling that if `y = None`, `f` can only take `x` as input, but it's not consistent for other functions. Correct me if I'm wrong.

examples/optimization/demo_learned_regularization.py
```diff
@@ -0,0 +1,198 @@
...
+For the reconstructions we sovle the variational problem
+
+    .. math::
+        \begin{equation}
+        \label{eq:min_prob}
+        \tag{1}
+        \underset{x}{\arg\min} \quad  \datafid{Ax}{y} + \lambda \reg{x},
+        \end{equation}
+
+for a data-fidelity term :math:`\datafid{x}{y}`, forward operator :math:`A` and a learned regularizer :math:`reg` using the
```
mh-nguyen712: why `\datafid{Ax}{y}` in the equation and `\datafid{x}{y}` here?

mh-nguyen712:
```suggestion
for a data-fidelity term :math:`\datafid{x}{y}`, forward operator :math:`A` and a learned regularizer :math:`\reg` using the
```

samuro95:
Thanks @johertrich for the very nice feature. Concerning the optimization algorithm, the current BaseOptim class is coming with a major refactoring with #592. I guess the best way to handle it is first to merge #592 and then to merge it within this PR. I can help to merge it within the new BaseOptim format.

johertrich:
Sure, I am happy to wait for the optim refactoring. However, I am still not convinced that forcing it into the BaseOptim format is the right way to do it here (since the first iteration is different, the additional combination with backtracking and the Anderson acceleration doesn't make sense and features like DEQ and unfolding won't work; in addition one looses the possibility to apply the function outside the `data_fidelity`/`prior` setting).

samuro95:
After analyzing the PR, we think that it could be nice to split the PR in two : a first PR for the explicit priors, and a second PR for the optimization algorithm. It will make the integration smoother. @johertrich what do you think ?

johertrich:
Sure we can do that. I am currently traveling, I will take care of it when I am back...

PR #837, opened 2025-10-08
Allow str in rotate interp mode

Andrewwango:
Tiny PR to allow str in rotate interp mode for a friendlier interface (i.e. without having to import torchvision just to use bicubic).

Also, when multiples are multiples of 90 degrees (we already test that this is perfect), we don't need to warn about any interpolation.

### DIFF
--- a/deepinv/transform/rotate.py
+++ b/deepinv/transform/rotate.py
@@ -43,13 +43,14 @@ class Rotate(Transform):
         self.limits = limits
         self.multiples = multiples
         self.positive = positive
-        if interpolation_mode is None and multiples % 90 != 0:
+        if interpolation_mode is None:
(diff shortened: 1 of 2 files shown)


deepinv/transform/rotate.py
```diff
@@ -35,21 +36,25 @@ def __init__(
...
-        interpolation_mode: InterpolationMode | None = None,
+        interpolation_mode: str | InterpolationMode | None = None,
         **kwargs,
     ):
         super().__init__(*args, **kwargs)
         self.limits = limits
         self.multiples = multiples
         self.positive = positive
-        if interpolation_mode is None:
+        if interpolation_mode is None and multiples % 90 != 0:
```
jscanvic: Nice

PR #845, opened 2025-10-13
Fix broken download Set14

### LINKED ISSUE #729: Set14 download link broken
### Environment

All

### Describe the bug

The Set14 link is broken from the source.

We should either find a difference source, or remove the download link. Either way, we should make clear in the dataset docs (all datasets that download things) where the original data is, so people can clearly know where to find the original data if they want to.

Alternative: host set14 ourselves on HF - it's not very big.

### Steps to reproduce

`dataset = dinv.datasets.Set14HR(".", download=True)`

### Expected behavior

Dataset is downloaded.

Note that this is not a deepinverse problem, as we test mocked set14 in our tests.

_No response_

### Additional context

_No response_

nucli-vicky:
To fix #729 
I can make it such that we reproduce exactly what was there before in terms of structure. Not sure if the checksum needs to be changed anyway.

Now, it would download as 
- `Set14/Set14_HR` --> all HR versions of the images
- `Set14/Set14_LR_x2` --> all images downsampled by factor 2
...

So the folder naming will need some cleaning + in the original behavior (from my understanding) had each HR image inside each folder (basically duplicates?).

Additionally, file names in this downloading are not integers, but image names (e.g. `baboon.png`).

TBD if we 
(i) Make the format indeed like this (with some minor change so we do not repeat `Set14` twice (once top level, one per resolution folder). 
(ii) Try to change the extracted folder structure such that we have an exact replica (requires most code).
(iii) Host the version we want on HuggingFace as suggested by @Andrewwango 

From my understanding, the purpose anyway is to simply have access to the HR versions. This could thus be greatly simplified by only downloading `Set14_HR.tar.gz`. 
I think thus it is the easiest approach to only download the HR images. Thoughts?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ Changed
 Fixed
 ^^^^^
 - Blur physics objects now put new filters to physics device regardless of input filter device (:gh:`844` by `Vicky De Ridder`_)
+- Set14HR dataset now downloads from a different source (and has slightly different folderstructure), since old link broke. (:gh:`845` by `Vicky De Ridder`_)
 
 v0.3.5
 ------
--- a/deepinv/datasets/set14.py
+++ b/deepinv/datasets/set14.py
@@ -100,7 +100,6 @@ class Set14HR(ImageFolder):
                 )
 
         # Initialize ImageFolder
-
         super().__init__(self.img_dir, transform=transform)
 
     def check_dataset_exists(self) -> bool:


Andrewwango:
Downloading HR images is sufficient, to match the behaviour of other datasets e.g. Div2k! The common usage pattern is to use it as a HR dataset to simulate a dataset of your own.

nucli-vicky:
Thanks for the info @Andrewwango . Indeed went with just downloading HR. The different (and simplified) folder structure should be okay since the `Set14HR`'s behavior remains same. 

Only thing which might be relevant is that the dataloading order will be different compared to the previous version at the same seed.
Due to the size of the dataset, I don't think this will have any tangible influence on users, but added it to the changelog just in case.

nucli-vicky:
@Andrewwango it was a random indeed! we just need to rerun the no optional dependencies benchmark, as it was canceled in the very first test run. I do not have permissions to do such a rerun.

PR #866, opened 2025-10-17
Fix LsdirHR download

### LINKED ISSUE #421: LsdirHR dataset broken download link
There are new links given by: https://ofsoundof.github.io/lsdir-data/

Need to change the old link.

nucli-vicky:
Fix #421 
This restores the exact functionality from the previous LsdirHR; except for a single bugfix: previously, `mode='val'` would actually capture the `X2`, `X3`, `X4` images too, resulting in `len(dataset)==1000, != 250`. 

Additionally, after extracting a shard, it deletes the associated `tar.gz` to save space, because LsdirHR in its entirety is relatively huge; if we do not delete the `.tar.gz` files, some users might not be able to download without running out of space, despite their system having comfortable amount of available space relative to the final dataset size (if the `.tar.gz` files are not present anymore).

Note: I tested the checksum only for val, and it remained the same. I did not explicitly test the one for train due to disk space / time constraints.

Note 2: HF uses Xet and [it's quite hard to disable or change the caching limits at runtime ](https://github.com/huggingface/huggingface_hub/issues/3266). It can still consume up to 10GB of cache without us being able to delete (safely). While possible if the user does not import huggingface hub related components, and the first import is performed inside `LsdirHR`, I do not think we should complexify things (what if, for example, the user later uses HF hub related components in their own code, and all of a sudden these settings were changed by us?).
Especially considering that 10GB is relatively small compared to the entire dataset, I think things are fine as they are on this front.

Note 3: While the `ImageFolder` init is now robust for `mode="val"`, for `mode="train"` it is only initialized properly if `self.root` only contains `.png` files coming from the `LsdirHR` dataset. This is silent to the user right now. 
Two options:
1. Simply check post super init that the `__len__` returns the expected values.
2. `x_path="shard-[0-1][0-9]/**/*.png"` in the case of `mode="train"`. This fails only in the special case where e.g. `shard-17` would somehow be present, so it avoids most problems except for a (I would argue) synthetic edge case.

Also fixed some syntactic errors in the docstrings of `RandomPatchSampler`.

### DIFF
--- a/docs/source/user_guide/reconstruction/denoisers.rst
+++ b/docs/source/user_guide/reconstruction/denoisers.rst
@@ -77,6 +77,7 @@ See :ref:`pretrained-weights` for more information on pretrained denoisers.
      - Any C, H, W
      - No
      - No
+     - No
    * - :class:`deepinv.models.SwinIR`
      - CNN-Transformer
      - Any C, H, W


nucli-vicky:
Note 4: I thought about making the dataset such that users can specify to only download / use a subset of shards. However, shards are not drawn randomly from the entire training set. If you go through a collection of images, you will see very often there is a whole range of image id's that have a lot similarities. One extreme case of this is `0001919.png` and `0001918.png`:




It might still be good for initial experimentation purposes; however I do not have the time / patience to download the entire dataset and compute the hashes for each shard. (Aside from the fact that shards are not equivalent to randomly sampling that amount of images from the entire set.)
Additionally, it would further complicate the globbing stuff, in the case the user specifies certain shards while others were already downloaded. Globbing that correctly would basically require `ImageFolder`'s initialization to take a list of patterns. While adding that to `ImageFolder` **may be** useful, I think this is outside the scope of this PR (which is a simple bugfix after all) + not sure how much adding that functionality to `ImageFolder` is useful outside this specific case.

nucli-vicky:
Note 5: `LsdirHR` would now require `huggingface_hub`. However this is in turn a dependency of `datasets` which is already present in the dataset dependencies, so I think not specifying it in `pyproject.toml` is fine.

nucli-vicky:
Better solution for train and ensuring correct behavior of `ImageFolder`:
Download all shards into and extract them in `os.path.join(self.root, "train_lsdir")`

But this would break backward compatibility for users who previously downloaded the dataset..

deepinv/datasets/lsdir.py
```diff
@@ -141,47 +135,64 @@ def __init__(
...
+                # download tar file from HuggingFace and save it locally
+                hf_hub_download(
+                    repo_id="ofsoundof/LSDIR",
+                    filename=filename,
+                    local_dir=os.path.join(self.root),
+                    cache_dir=os.path.join(self.root, ".cache/huggingface"),
+                    local_dir_use_symlinks=False,
                 )
+                # Since LSDIR is relatively large, we want to avoid taking too much redundant disk space.
+                shutil.rmtree(os.path.join(self.root, ".cache/huggingface"))
```
Andrewwango, line 148: Does this risk deleting other non-LSDIR things the user may have already in their HF cache?

Andrewwango (changes requested):
Thanks @nucli-vicky for this, and for closing an old issue! Just one minor coment. 

Thanks for the details - I think a large portion of users will want to anyway download the dataset themselves via GUI / have it already downloaded, since it's so big. The auto download is more necessary for the smaller, quick-debug datasets.

deepinv/datasets/lsdir.py
nucli-vicky: In principle yes, in practice mostly no: the code directs the cache to `join(root, '.cache/huggingface')`, which is usually not where the rest of HuggingFace cache is located (most often `/home/.cache/huggingface`). In that sense it's quite safe.

nucli-vicky:
This should now also fix the broken doc build from #869

PR #867, opened 2025-10-18
Preventive fix: Tensor non-tuple indexing deprecated

nucli-vicky:
In the testing logs right now there is already a warning:

"UserWarning: Using a non-tuple sequence for multidimensional indexing is deprecated and will be changed in pytorch 2.9; use x[tuple(seq)] instead of x[seq]. In pytorch 2.9 this will be interpreted as tensor index, x[torch.tensor(seq)], which will result either in an error or a different result"

This PR is a preventive fix for PyTorch 2.9

### DIFF
--- a/deepinv/loss/metric/distortion.py
+++ b/deepinv/loss/metric/distortion.py
@@ -425,8 +425,8 @@ class QNR(Metric):
             for c in range(n_bands):
                 out += (
                     abs(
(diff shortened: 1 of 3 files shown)


nucli-vicky:
Hmm, I might have misinterpreted the warning; it's probably still safe to use lists for indices across a single axis (as is the case for all changes outside of `tensorlist.py`. This likely explains why the warning is only raised for the function `dirac`; apologies..

PR #871, opened 2025-10-21
Gentler loader check

Andrewwango:
This softens the dataloader check inside trainer.

The problem before is if you use a custom dataloader, whose internal `dataset` does not return tensors (maybe returns list of patches for eaxmple, and the custom loader does the tensor loading), then that will raise an error in the trainer, even though the loader can be used perfectly fine. So now the check only happens if you use the normal torch utils data dataloader.

### DIFF
--- a/deepinv/training/trainer.py
+++ b/deepinv/training/trainer.py
@@ -419,7 +419,7 @@ class Trainer:
             and self.show_progress_bar
         ):
             warnings.warn(
-                "Update progress bar frequency of 1 may slow down training on GPU. Consider increasing this."
+                "Update progress bar frequency of 1 may slow down training on GPU. Consider setting freq_update_progress_bar > 1."
             )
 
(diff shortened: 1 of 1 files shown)


deepinv/training/trainer.py
```diff
@@ -277,7 +277,7 @@ def setup_train(self, train=True, **kwargs):
         for loader in self.train_dataloader + (
             self.eval_dataloader if self.eval_dataloader is not None else []
         ):
-            if loader is not None:
+            if loader is not None and isinstance(loader, torch.utils.data.DataLoader):
```
matthieutrs, line 280: Maybe add a warning if this is not triggered?

matthieutrs (approved):
Thanks for the PR! I just have a minor suggestion but happy to reconsider if you think it's just adding clutter

PR #883, opened 2025-11-02
Update attention with native torch function 

mh-nguyen712:
This PR updates the attention computation in some models (DiffUNet, NCSNpp, SCUNet, ADMUnet).

Previously the computations were in older pytorch version, using manual `torch.einsum` and scaling. This PR leverages the use of [scaled_dot_product_attention]( https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html), which is supported in torch 2.x

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
   - name: DeepInverse Discord community support
     url: https://discord.gg/qBqY5jKw3p
--- a/.github/scripts/convert_to_notebook.py
+++ b/.github/scripts/convert_to_notebook.py
@@ -1,0 +1,640 @@
+from sphinx_gallery.py_source_parser import split_code_and_text_blocks
+from sphinx_gallery.notebook import jupyter_notebook, save_notebook
+from sphinx_gallery import gen_gallery
+from pathlib import Path
+import copy
+import nbformat
+import ast
+import re
+
+###############################################################################
+# Citation handling
+###############################################################################
+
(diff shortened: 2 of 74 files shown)


nucli-vicky:
Good idea, @mh-nguyen712 ! I was just quickly wondering: would this further decrease support for older PyTorch versions? Or are the PyTorch versions which don't have `scaled_dot_product_attention` already incompatible with deepinv?

mh-nguyen712:
Thanks for the remark! This functions is supported in pytorch 2.x already so I think it's should be ok.

mh-nguyen712:
Thanks for the review @jscanvic! I remove some unused code (defined but not used anywhere) to clean the codebase and hopefully increase a bit codecov. The main part that is uncovered is `forward_patch` in diffunet, only triggered when image > 512x512 so a bit hard to test.

nucli-vicky:
Hmm I just checked `patch_forward`. Shouldn't `if x.shape[-2] < 520 and x.shape[-1] < 520:` be `<=512` then (in the main `forward`)?

Also: It would still be valuable to test `patch_forward`. Maybe just for a single network pass? Or would it lead to memory issues in CI/CD? 

(*Technically out of scope of this PR, just stating to take a note for future, or we can do it here anyway.*) Otherwise we can instantiate a very lightweight version of the network; this would simply require support for specifying e.g. the network width at every stage and the number of stages. Seems like a good feature to have in general?

mh-nguyen712:
I just update the tests: As the DiffUNet denoising perf is already tested in `test_models/test_denoiser_perf`, I changed the `test_diffunet` to test:
- patch_forward
- use_fp16
- a smaller model (no checkpoint, only test output shape) 
Thanks for suggestion @jscanvic @nucli-vicky !

PR #886, opened 2025-11-04
Trainer improvement

mh-nguyen712:
This PR updates some minor points of the Trainer:
- allows `non_blocking` transfers, optimizer `set_to_none` 
- some changes for better readability and more pythonic
- re-use grad_norm from `clip_grad_norm` for logging, avoid recomputing

### DIFF
--- a/deepinv/training/trainer.py
+++ b/deepinv/training/trainer.py
@@ -1089,7 +1089,12 @@ class Trainer:
 
     def plot(self, epoch, physics, x, y, x_net, train=True):
         r"""
-        Plot and optinally save the reconstructions.
+        Plot ground truths, measurements and reconstructions and at test time, optionally save them.
+
+        .. note::
+
+            Images can be saved to disk at test time by providing a value for the parameter ``save_folder_im``
+            when calling the method :func:`deepinv.Trainer.test`. Note that in that case, every test sample is saved and not only the first ones.
 
         :param int epoch: Current epoch.
         :param deepinv.physics.Physics physics: Current physics operator.
@@ -1142,15 +1147,14 @@ class Trainer:
                 mlflow.log_image(image, key=f"{post_str} samples", step=epoch)
 
         if save_images:
-            # save images
-            for k, img in enumerate(imgs):
+            for k, img in enumerate(imgs):  # ground truths, reconstructions, etc.
+                img_name = f"{self.save_folder_im}/{titles[k]}/"
+                Path(img_name).mkdir(parents=True, exist_ok=True)
                 for i in range(img.size(0)):
-                    img_name = f"{self.save_folder_im}/{titles[k]}/"
-                    # make dir
-                    Path(img_name).mkdir(parents=True, exist_ok=True)
-                    save_image(img, img_name + f"{self.img_counter + i}.png")
+                    entry = img[i].unsqueeze(0)
+                    save_image(entry, img_name + f"{self.img_counter + i}.png")
 
-                self.img_counter += len(imgs[0])
+            self.img_counter += imgs[0].size(0)
 
         if self.conv_metrics is not None:
             plot_curves(
@@ -1312,8 +1316,8 @@ class Trainer:
         stop = epoch > self.early_stop * self.eval_interval + best_epoch
         if stop and self.verbose:
             print(
-                "Early stopping triggered as validation metrics have not improved in "
-                f"the last {self.early_stop} validation steps, disable it with early_stop=None, or"
+                f"Early stopping triggered at epoch {epoch} as validation metrics have not improved in "
+                f"the last {self.early_stop} validation steps. Disable it with early_stop=None, or "
                 f"modify early_stop>0 to wait for more validation steps."
             )
 
@@ -1492,9 +1496,13 @@ class Trainer:
         r"""
         Test the model, compute metrics and plot images.
 
+        .. note::
+
+            It is possible to save the reconstructed images along with the ground truths and measurements by specifying a value for the parameter ``save_path``. Note that in this case, every test sample is saved and not only the first ones.
+
         :param torch.utils.data.DataLoader, list[torch.utils.data.DataLoader] test_dataloader: Test data loader(s), see :ref:`datasets user guide <datasets>`
             for how we expect data to be provided.
-        :param str save_path: Directory in which to save the plotted images.
+        :param str save_path: Path to the directory where to save the plotted images if desired (optional).
         :param bool compare_no_learning: If ``True``, the linear reconstruction is compared to the network reconstruction.
         :param bool log_raw_metrics: if `True`, also return non-aggregated metrics as a list.
         :param Metric, list[Metric], None metrics: Metric or list of metrics used for evaluation. If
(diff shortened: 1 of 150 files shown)


deepinv/training/trainer.py
```diff
@@ -296,7 +296,10 @@ class Trainer:
     verbose: bool = True
     verbose_individual_losses: bool = True
     show_progress_bar: bool = True
-    freq_update_progress_bar: int = 1
+    freq_update_progress_bar: int = 10
```
Andrewwango: Can you explain why this is needed? I changed it to default 1 in #839, otherwise it risks giving the user a bad experience (those who use progress bar already risk being confused why they no longer see metrics updated, silently). I also already had a warning to encourage users to increase it anyway

deepinv/training/trainer.py
```diff
@@ -296,7 +296,10 @@ class Trainer:
...
+    non_blocking_transfers: bool = (
```
Andrewwango, line 308: Does this have to be an extra argument? if absolutely needed it would need a docstring

deepinv/training/trainer.py
```diff
@@ -467,12 +486,11 @@ def setup_train(self, train=True, **kwargs):
...
         # make physics and data_loaders of list type
-        if type(self.physics) is not list:
+        if not isinstance(self.physics, (list, tuple)):
             self.physics = [self.physics]
 
-        if (
-            self.physics_generator is not None
-            and type(self.physics_generator) is not list
+        if self.physics_generator is not None and not isinstance(
+            self.physics_generator, (list, tuple)
```
Andrewwango, line 501: All others should have tuple too, or this should only be list

deepinv/training/trainer.py
```diff
@@ -518,6 +536,14 @@ def setup_train(self, train=True, **kwargs):
             )
 
         _ = self.load_model()
+        # Cache whether model.forward accepts 'update_parameters' to avoid per-call inspect
```
Andrewwango, line 547: the risk with a global flag is that it persists after training. what if the user loads a model later that isn't updated?

deepinv/training/trainer.py
```diff
@@ -518,6 +536,14 @@ def setup_train(self, train=True, **kwargs):
...
+        try:
+            sig = inspect.signature(self.model.forward)
+            self._model_accepts_update_parameters = (
+                "update_parameters" in sig.parameters
+            )
+        except Exception:
```
Andrewwango: can we be more specific?

deepinv/training/trainer.py
```diff
@@ -594,23 +620,26 @@ def check_clip_grad(self):
...
-            # from https://discuss.pytorch.org/t/check-the-norm-of-gradients/27961/7
-            grads = [
-                param.grad.detach().flatten()
-                for param in self.model.parameters()
-                if param.grad is not None
-            ]
-            norm_grads = torch.cat(grads).norm()
-            out = norm_grads.item()
-            self.check_grad_val.update(norm_grads.item())
+            if grad_norm is not None:
```
Andrewwango, line 639: for both of these branches, it would be good to have a pytest

deepinv/training/trainer.py
```diff
@@ -652,7 +681,7 @@ def get_samples_online(self, iterators, g):
                 )
             params = self.physics_generator[g].step(batch_size=x.size(0))
 
-        # Update parameters both via update and, if implemented in physics, via forward pass
+        # Update parameters via update
```
Andrewwango: why remove the clarification? we do **kwargs in both update and forward pass (line below)

deepinv/training/trainer.py
```diff
@@ -752,31 +785,29 @@ def model_inference(self, y, physics, x=None, train=True, **kwargs):
...
-
-        if not train:
             with torch.no_grad():
                 if self.plot_convergence_metrics:
                     x_net, self.conv_metrics = self.model(
-                        y, physics, x_gt=x, compute_metrics=True, **kwargs
+                        y,
+                        physics,
+                        x_gt=x,
+                        compute_metrics=self.plot_convergence_metrics,
```
Andrewwango: but this is necessarily always true in this branch?

deepinv/training/trainer.py
```diff
@@ -970,9 +1002,9 @@ def step(
         :returns: The current physics operator, the ground truth, the measurement, and the network reconstruction.
         """
         if train and self.optimizer_step_multi_dataset:
-            self.optimizer.zero_grad()  # Clear stored gradients
+            self.optimizer.zero_grad(set_to_none=True)  # Clear stored gradients
```
Andrewwango, line 1013: Can you also fix this in the adversarial trainer?

examples/sampling/demo_diffusion_sde.py
```diff
@@ -99,6 +99,7 @@
 # When the data fidelity is not given, the posterior diffusion is equivalent to the unconditional diffusion.
 # Sampling is performed by solving the reverse-time SDE. To do so, we generate a reverse-time trajectory.
 
+denoiser = dinv.models.DiffUNet(pretrained="download").to(device)
```
Andrewwango: explain why we are using a new denoiser?

mh-nguyen712: Sorry I committed on the wrong branch. Reverted it. Thanks!

deepinv/training/trainer.py
mh-nguyen712: I though that we only sync the GPU every `freq_update_progress_bar` so increasing it would reduce the synchronization time. But in fact it doesn't help here. I can revert it to 1. But I think we should reduce the device sync.

deepinv/training/trainer.py
mh-nguyen712: added docstring

deepinv/training/trainer.py
mh-nguyen712: I think both list and tuple could work here. Just a safe check.

deepinv/training/trainer.py
mh-nguyen712: If I understood well, users need to initialize the Trainer with a model before using it? If yes, I think it's safe to cache the flag.

deepinv/training/trainer.py
mh-nguyen712: I don't know what it returns when the inspection fails :D

deepinv/training/trainer.py
mh-nguyen712: My mistake. I though that physics.update(**kwargs) is enough not in fact not.

deepinv/training/trainer.py
mh-nguyen712: thanks!

deepinv/training/trainer.py
mh-nguyen712: Sure!

deepinv/training/trainer.py
mh-nguyen712: Done! Always check (list, tuple) now

deepinv/training/trainer.py
mh-nguyen712: tested

Andrewwango:
@mh-nguyen712  can you merge from main and run GPU tests?

PR #892, opened 2025-11-05
Enable ruff rule N805

jscanvic:
Ruff rule N805 flags uses of other names than `self` for the 1st argument in class instance methods. It is error-prone given it might lead a reader to assume a method is static when it isn't, e.g., if `cls` is used instead of `self` (there are 2 such cases in the code base at the moment). I suggest we enable it, following Andrew enabling it in commit 1eaa5ce5669c47e9d428f6bda55e1e60a37f6769 for the draft PR #705.

The 2 cases that don't currently pass the check are easily solved by renaming the 1st argument to `self`, which we do automatically using `ruff check --fix --unsafe-fixes`.

### DIFF
--- a/deepinv/tests/test_external_libraries.py
+++ b/deepinv/tests/test_external_libraries.py
@@ -6,10 +6,10 @@ import unittest.mock as mock
 
 
 class TestTomographyWithAstra:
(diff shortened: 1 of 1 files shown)


mh-nguyen712 (approved):
Nice! Thanks Jérémy!

PR #812, opened 2025-10-02
Adding Latent Diffusion Model, DDIM and PSLD algorithm

alexspagn:
Hi all!

This is a PR that implements mainly 3 things:
- A text-conditioned Latent Denoiser (using Stable Diffusion 1.5)
- A DDIM algorithm to sample using the Latent Denoiser
- The [PSLD](https://arxiv.org/pdf/2307.00619) algorithm that uses this model and sampling scheme to perform posterior sampling

To do so, a new latentden.py model and diffusion_discrete.py sampling are introduced, along with a demo_SD.py demo and a test_latent_diff.py to test these new classes.

There are two new dependencies, Diffusers and Transformers, which are required to handle the CLIP embeddings, the Unet, and the VAE. They have been added as optional dependencies in the denoiser category.

As this is my first PR, I hope I am not missing anything important.

Example obtained running the demo, PSLD is applied on an image generated by the Latent Diffusion Model implemented:

sphx_glr_demo_SD_003

### Checks to be done before submitting your PR

- [ ] `python3 -m pytest deepinv/tests` runs successfully.
- [x] `black .` runs successfully.
- [x] `make html` runs successfully (in the `docs/` directory).
- [x] Updated docstrings related to the changes (as applicable).
- [ ] Added an entry to the [CHANGELOG.rst](../CHANGELOG.rst).

PR #893, opened 2025-11-05
Diffuser integration

mh-nguyen712:
This PR:
- makes the integration with `diffusers` library: wrapper a diffusion model from diffusers to a deepinv denoiser, from only repo_id 
- fully compatible with deepinv, only one more line of code is needed for PnP or PosteriorSampling
- shows a demo: how to use the wrapper for image generation and posterior sampling with our SDE formulution (VE or VP)
- Rescale the measurement `y` (to -1, 1) inside the posterior sampling and the sample (to 0, 1) generated by `PosteriorDiffusion`

### DIFF
--- a/deepinv/models/base.py
+++ b/deepinv/models/base.py
@@ -28,7 +28,7 @@ class Denoiser(torch.nn.Module):
     def forward(self, x, sigma, **kwargs):
         r"""
         Applies denoiser :math:`\denoiser{x}{\sigma}`.
-        The input `x` is expected to be in `[0, 1]` range and the output is also in `[0, 1]` range.
+        The input `x` is expected to be with pixel values in `[0, 1]` range, up to random noise. The output is also expected to be in `[0, 1]` range.
 
         :param torch.Tensor x: noisy input, of shape `[B, C, H, W]`.
         :param torch.Tensor, float sigma: noise level. Can be a `float` or a :class:`torch.Tensor` of shape `[B]`.
--- a/deepinv/models/wrapper.py
+++ b/deepinv/models/wrapper.py
@@ -5,7 +5,7 @@ from deepinv.models import Denoiser
 
 class DiffusersDenoiserWrapper(Denoiser):
     """
-    Wraps a diffusers model as a DeepInv Denoiser.
+    Wraps a `HuggingFace diffusers <https://huggingface.co/docs/diffusers/index>`_ model as a DeepInv Denoiser.
 
     :param str mode_id: Diffusers model id or HuggingFace hub repository id. For example, 'google/ddpm-cat-256'.
         The id must work with `DiffusionPipeline`.
@@ -90,7 +90,7 @@ class DiffusersDenoiserWrapper(Denoiser):
 
     def _nearest_t_from_sigma(self, sigma: torch.Tensor) -> torch.Tensor:
         """
-        Map a sigma to the nearest training time-step index.
+        Maps a sigma to the nearest training time-step index.
         Supports scalar or per-batch tensor sigma.
         """
         sigmas = self.sigma_schedule  # [T]
@@ -112,7 +112,7 @@ class DiffusersDenoiserWrapper(Denoiser):
     ) -> torch.Tensor:
         r"""
         Applies denoiser :math:`\denoiser{x}{\sigma}`.
-        The input `x` is expected to be in `[0, 1]` range and the output is also in `[0, 1]` range.
+        The input `x` is expected to be in `[0, 1]` range (up to random noise) and the output is also in `[0, 1]` range.
 
         :param torch.Tensor x: noisy input, of shape `[B, C, H, W]`.
         :param torch.Tensor, float sigma: noise level. Can be a `float` or a :class:`torch.Tensor` of shape `[B]`.
--- a/deepinv/tests/test_models.py
+++ b/deepinv/tests/test_models.py
@@ -1409,6 +1409,11 @@ def test_diffuser_wrapper(batch_size, clip_output, device):
         reason="This test requires diffusers. It should be "
         "installed with `pip install diffusers`",
     )
+    pytest.importorskip(
+        "transformers",
+        reason="This test requires transformers. It should be "
+        "installed with `pip install transformers`",
+    )
     model = dinv.models.DiffusersDenoiserWrapper(
         mode_id="google/ddpm-cifar10-32", clip_output=clip_output, device=device
     )


mh-nguyen712:
Related to #812 cc @samuro95 @alexspagn

mh-nguyen712:
Images generated in the example:

- For denoising:
image

- For unconditional generation (150 Euler steps, VE-SDE):
image

- For posterior sampling (150 Euler steps, VP-SDE, DPS):
image

deepinv/models/wrapper.py
```diff
@@ -0,0 +1,195 @@
...
+        ), "Currently, only DDPMScheduler is supported."
+
+        self.clip_output = clip_output
+
+        # Precompute sigma(t) over training timeline
+        ac = self.scheduler.alphas_cumprod
+
+        self.register_buffer("sqrt_alpha_cumprod", ac.sqrt())  # alpha_t
+        self.register_buffer(
+            "sqrt_one_minus_alpha_cumprod", (1.0 - ac).clamp(min=0).sqrt()
```
samuro95: maybe to make it simpler we could call it something like sigma_schedule  instead of sqrt_one_minus_alpha_cumprod ?

samuro95: And then we could keep only this sigma_schedule in the self no ? because we can infer sqrt_alpha_cumprod from it ?

deepinv/models/wrapper.py
```diff
@@ -0,0 +1,195 @@
...
+
+        # Resolve timestep
+        if sigma is not None:
+            if timestep is not None:
+                warnings.warn(
+                    "Both sigma and timestep are provided. Ignoring timestep and using sigma."
+                )
+            timestep = self._nearest_t_from_sigma(sigma.squeeze())
+        else:
+            timestep = self._handle_sigma(
```
samuro95: can we keep a single call to _handle_sigma  ?

deepinv/models/wrapper.py
mh-nguyen712: I agree about the name. I change to `scale_schedule` and `sigma_schedule`, similar to `scale_t` and `sigma_t` in the continuous SDE. 
However I would keep both in the `self`, since we only need to compute it once.

deepinv/models/wrapper.py
mh-nguyen712: I don't see any way to do it. We need 2 calls to `_handle_sigma` to make sure that the `sigma` and `scale` are in the correct form. An additional call is triggered if users call the denoiser with the timestep.

deepinv/models/wrapper.py
nucli-vicky: On the name change: I would like to add that `sqrt_one_minus_alpha_cumprod` has by now been used in a lot of diffusion code (and the original papers), so quite some people are quite familiar with the notation. Given the rename, could we maybe mention the original name as a `#` comment?
This is mainly just for users who are familiar with it so they have a little pointer / "recognition point"

mh-nguyen712: Thanks @nucli-vicky, I'll add a comment about these names!

examples/sampling/demo_diffusers.py
```diff
@@ -0,0 +1,151 @@
...
+
+model = PosteriorDiffusion(
+    data_fidelity=ZeroFidelity(),
+    sde=sde,
+    denoiser=denoiser,
+    solver=solver,
+    dtype=dtype,
+    device=device,
+    verbose=True,
+)
```
chaozg, line 91: I'm just curious if we need `minus_one_one=False` here as 1) I think ddpm is trained with range [-1, 1] and 2) `PosteriorDiffusion`'s documentation states that https://github.com/deepinv/deepinv/blob/2bc07427272ac56d20f501bb8e266e0d3aeff9b2/deepinv/sampling/diffusion_sde.py#L508

mh-nguyen712: Hi! All denoisers in deepinv works with data on [0,1]. That's why we need this option

chaozg: So `DiffusersDenoiserWrapper` works on [0,1] though its _internal_ `Diffusers` model works on [-1,1]? I'm sorry for this very naive question... I'm still learning to use diffusion models 😃, and thanks very much for the PR and the response!

mh-nguyen712: Yes exactly

deepinv/models/wrapper.py
```diff
@@ -0,0 +1,175 @@
...
+        >>> x_noisy = x + sigma * torch.randn_like(x)
+        >>> with torch.no_grad():
+        ...     x_denoised = denoiser(x_noisy, sigma=sigma)
+
+    """
+
+    def __init__(
+        self,
+        mode_id: str = None,
+        clip_output: bool = True,
```
samuro95, line 49: I am wondring if we should clip by default, it means that we change the default diffusers model behavior no ? Is it what is done in the forward of the diffusers diffusion sampling ?

mh-nguyen712: Yes in diffusers sampling they clip the output, either between -1 1 or dynamic clipping between 0.05 and 0.95 quantiles.

samuro95: ok thanks

mh-nguyen712: But I think we can leave the clipping option in the SDESolver as well, because in diffusers they do have some other options (mostly for numerical stability rather than theoretical) in the sampling, eg `clip_sample`, `thresholding` or `rescale_betas_zero_snr`. What do you think?

deepinv/models/wrapper.py
```diff
@@ -0,0 +1,175 @@
...
+            self.scheduler, DDPMScheduler
+        ), "Currently, only DDPMScheduler is supported."
+
+        self.clip_output = clip_output
+
+        # Precompute sigma(t) over training timeline
+        ac = self.scheduler.alphas_cumprod
+
+        self.register_buffer(
+            "scale_schedule", ac.sqrt()
```
samuro95, line 78: I think scale_schedule should be 1 / ac.sqrt()

deepinv/models/wrapper.py
```diff
@@ -0,0 +1,175 @@
...
+        )  # the scaling factor -- sqrt_alphas_cumprod
+        self.register_buffer(
+            "sigma_schedule", (1.0 - ac).clamp(min=0).sqrt()
```
samuro95: For variance preserving, sigma_schedule = sqrt(1/scale_schedule**2 - 1). So there is something wierd in the sign here ... no ?

mh-nguyen712:
I made some changes about scaling and sigma to match the EDM formulation. Should be good now @samuro95!

deepinv/models/wrapper.py
samuro95: yes in both it's ok

deepinv/models/base.py
```diff
@@ -28,9 +28,13 @@ def __init__(self, device="cpu"):
     def forward(self, x, sigma, **kwargs):
         r"""
         Applies denoiser :math:`\denoiser{x}{\sigma}`.
+        The input `x` is expected to be in `[0, 1]` range and the output is also in `[0, 1]` range.
```
matthieutrs: I am not sure about this formulation. While I understand what you mean, we want to avoid people forcing a renormalization in exact [0, 1] range for denoising data with target already in [0, 1]. 

```suggestion
        The input `x` is expected to be with pixel values in `[0, 1]` range, up to random noise. The output is also expected to be in `[0, 1]` range.
```

deepinv/models/wrapper.py
```diff
@@ -0,0 +1,171 @@
+from __future__ import annotations
+import torch
+from deepinv.models import Denoiser
+
+
+class DiffusersDenoiserWrapper(Denoiser):
+    """
+    Wraps a diffusers model as a DeepInv Denoiser.
```
matthieutrs:
```suggestion
    Wraps a [HuggingFace diffusers](https://huggingface.co/docs/diffusers/index) model as a DeepInv Denoiser.
```

deepinv/models/wrapper.py
```diff
@@ -0,0 +1,171 @@
...
+        The id must work with `DiffusionPipeline`.
+        See `Diffusers Documentation <https://huggingface.co/docs/diffusers/v0.35.1/en/api/pipelines/overview#diffusers.DiffusionPipeline>`_.
+    :param bool clip_output: Whether to clip the output to the model range. Default is `True`.
+    :param device: Device to load the model on. Default is 'cpu'.
+
+    .. note::
+        Currently, only models trained with `DDPMScheduler` are supported.
+
+    .. warning::
+        This wrapper requires the `diffusers` and `transformers` packages.
```
matthieutrs, line 20: Is `transformers` a required dependency, or an optional one? You are only testing `diffusers` below.

deepinv/models/wrapper.py
```diff
@@ -0,0 +1,171 @@
...
+    def forward(
+        self,
+        x: torch.Tensor,
+        sigma: float | torch.Tensor = None,
+        *args,
+        **kwargs,
+    ) -> torch.Tensor:
+        r"""
+        Applies denoiser :math:`\denoiser{x}{\sigma}`.
+        The input `x` is expected to be in `[0, 1]` range and the output is also in `[0, 1]` range.
```
matthieutrs:
```suggestion
        The input `x` is expected to be in `[0, 1]` range (up to random noise) and the output is also in `[0, 1]` range.
```

deepinv/models/wrapper.py
```diff
@@ -0,0 +1,171 @@
...
+        # prediction_type: "epsilon", "v_prediction", or "sample"
+        self.prediction_type = getattr(
+            self.scheduler.config, "prediction_type", "epsilon"
+        )
+
+        self.to(device)
+
+    def _nearest_t_from_sigma(self, sigma: torch.Tensor) -> torch.Tensor:
+        """
+        Map a sigma to the nearest training time-step index.
```
matthieutrs:
```suggestion
        Maps a sigma to the nearest training time-step index.
```

deepinv/tests/test_models.py
```diff
@@ -1393,3 +1399,25 @@ def test_swinir_upsample_without_upsampler(upscale, upsampler):
 
     with ctx:
         _ = dinv.models.SwinIR(in_chans=3, upsample=1, pretrained=None, **kwargs)
+
+
+@pytest.mark.parametrize("clip_output", [False, True])
+@pytest.mark.parametrize("batch_size", [1, 2])
+def test_diffuser_wrapper(batch_size, clip_output, device):
```
matthieutrs, line 1406: if transformers is also required (cf above) then we need to add it too

pyproject.toml
```diff
@@ -91,6 +91,8 @@ denoisers = [
     "sigpy",
     "astra-toolbox>=2.2.0,!=2.3.0; platform_system=='Linux'",
     "spyrit>=3.0.0; platform_system=='Linux'",
+    "diffusers", 
+    "transformers",
```
matthieutrs, line 95: is `transformers` required (cf above)?

matthieutrs (commented):
This is awesome, great addition @mh-nguyen712!! Just a few minor comments below

deepinv/models/wrapper.py
mh-nguyen712: Sometimes it requires `transformers` to load model from the hub. I suppose that they use models defined in `transformer`, but not always

PR #904, opened 2025-11-07
Make image saving in Trainer.test more conventional

jscanvic:
As previously suspected in #553, the image saving behavior in `Trainer.test` is very quirky:
1. There is a mismatch between filenames across directories, e.g., results for the 1st dataset entry might be saved to `Ground truth/0.png` and `Measurement/2.png` (sic!)
2. The same images are saved to different files, e.g., `Ground truth/0.png` and `Ground truth/1.png` might contain the same image
3. The code clearly intends for saving individual images separately but what actually gets saved is batches merged into a single file

**Reproducing**

```python
# Adapted from examples/models/demo_training.py
import deepinv as dinv
import torch
from torchvision.transforms import Compose, ToTensor, Resize, CenterCrop, Grayscale
import os
import shutil

rng = torch.Generator(device="cpu").manual_seed(0)
physics = dinv.physics.Denoising(
    dinv.physics.GaussianNoise(sigma=0.1, rng=rng)
)

dataset = dinv.datasets.Urban100HR(
    ".",
    download=True,
    transform=Compose([ToTensor(), Grayscale(), Resize(256), CenterCrop(64)]),
)

train_dataset, test_dataset = torch.utils.data.random_split(
    torch.utils.data.Subset(dataset, range(10)), (0.8, 0.2)
)

dataset_path = dinv.datasets.generate_dataset(
    train_dataset=train_dataset,
    test_dataset=test_dataset,
    physics=physics,
    device="cpu",
    save_dir=".",
    batch_size=1,
)

train_dataloader = torch.utils.data.DataLoader(
    dinv.datasets.HDF5Dataset(dataset_path, train=True), shuffle=True, batch_size=2
)
test_dataloader = torch.utils.data.DataLoader(
    dinv.datasets.HDF5Dataset(dataset_path, train=False), shuffle=False, batch_size=2
)

model = dinv.models.ArtifactRemoval(
    dinv.models.UNet(1, 1, scales=2, batch_norm=False).to("cpu")
)

trainer = dinv.Trainer(
    model=model,
    physics=physics,
    optimizer=torch.optim.Adam(model.parameters(), lr=1e-3),
    eval_dataloader=test_dataloader,
    epochs=1,
    losses=dinv.loss.SupLoss(metric=dinv.metric.MSE()),
    metrics=dinv.metric.PSNR(),
    device="cpu",
    plot_images=False,
    show_progress_bar=False,
)

_ = trainer.train()

out_dir = "./test_results"

shutil.rmtree(out_dir, ignore_errors=True)

trainer.test(test_dataloader, save_path=out_dir)

# Check the files produced by Trainer.test

def run_cmd(cmd: str) -> None:
    print(f"$ {cmd}")
    os.system(cmd)

# Bug 1. The indices in the filenames are wrong.
run_cmd("find ./test_results")

# Bug 2. The same images are saved in multiple files.
run_cmd('cmp "./test_results/Ground truth/0.png" "./test_results/Ground truth/1.png" && echo "The two files are the same! (Error)" || echo "The two files are different (OK)"')
```

Current outcome

```
$ find ./test_results
./test_results
./test_results/No learning
./test_results/No learning/4.png
./test_results/No learning/5.png
./test_results/Measurement
./test_results/Measurement/2.png
./test_results/Measurement/3.png
./test_results/Ground truth
./test_results/Ground truth/0.png
./test_results/Ground truth/1.png
./test_results/Reconstruction
./test_results/Reconstruction/7.png
./test_results/Reconstruction/6.png
$ cmp "./test_results/Ground truth/0.png" "./test_results/Ground truth/1.png" && echo "The two files are the same! (Error)" || echo "The two files are different (OK)"
The two files are the same! (Error)
```

Corrected outcome

```
$ find ./test_results
./test_results
./test_results/No learning
./test_results/No learning/0.png
./test_results/No learning/1.png
./test_results/Measurement
./test_results/Measurement/0.png
./test_results/Measurement/1.png
./test_results/Ground truth
./test_results/Ground truth/0.png
./test_results/Ground truth/1.png
./test_results/Reconstruction
./test_results/Reconstruction/0.png
./test_results/Reconstruction/1.png
$ cmp "./test_results/Ground truth/0.png" "./test_results/Ground truth/1.png" && echo "The two files are the same! (Error)" || echo "The two files are different (OK)"
./test_results/Ground truth/0.png ./test_results/Ground truth/1.png differ: byte 36, line 3
The two files are different (OK)
```

**Cause**

https://github.com/deepinv/deepinv/blob/90184628f5829a46767afba4d5b37ced74b856e4/deepinv/training/trainer.py#L1110-L1119

* Quirk 1 is due to an additional indent: the counter gets mistakenly incremented in the inner loop instead of the outer
* Quirk 2 and 3 is due to the code being written partly for saving batches and partly for saving individual images

**Solution**

* Get rid of the extra indent
* In the inner loop, save the batch entry at the current index instead of the whole batch

It's technically a breaking change but @Andrewwango and I agree that it's unlikely to have negative consequences - I still document it in the changelog.

### DIFF
--- a/deepinv/training/trainer.py
+++ b/deepinv/training/trainer.py
@@ -1055,7 +1055,11 @@ class Trainer:
 
     def plot(self, epoch, physics, x, y, x_net, train=True):
         r"""
-        Plot and optinally save the reconstructions.
+        Plot ground truths, measurements and reconstructions and at test time, optionally save them.
+
+        .. note::
+
+            Images can be saved to disk at test time by providing a value for the parameter ``save_folder_im`` when calling the method :meth:`deepinv.training.Trainer.test`. Note that in that case, every test sample is saved and not only the first ones.
 
         :param int epoch: Current epoch.
         :param deepinv.physics.Physics physics: Current physics operator.
@@ -1457,9 +1461,13 @@ class Trainer:
         r"""
         Test the model, compute metrics and plot images.
 
+        .. note::
+
+            It is possible to save the reconstructed images along with the ground truths and measurements by specifying a value for the parameter ``save_path``. Note that in this case, every test sample is saved and not only the first ones.
+
         :param torch.utils.data.DataLoader, list[torch.utils.data.DataLoader] test_dataloader: Test data loader(s), see :ref:`datasets user guide <datasets>`
             for how we expect data to be provided.
-        :param str save_path: Directory in which to save the plotted images.
+        :param str save_path: Path to the directory where to save the plotted images if desired (optional).
         :param bool compare_no_learning: If ``True``, the linear reconstruction is compared to the network reconstruction.
         :param bool log_raw_metrics: if `True`, also return non-aggregated metrics as a list.
         :param Metric, list[Metric], None metrics: Metric or list of metrics used for evaluation. If


deepinv/training/trainer.py
```diff
@@ -1114,9 +1114,10 @@ def plot(self, epoch, physics, x, y, x_net, train=True):
                     img_name = f"{self.save_folder_im}/{titles[k]}/"
                     # make dir
                     Path(img_name).mkdir(parents=True, exist_ok=True)
-                    save_image(img, img_name + f"{self.img_counter + i}.png")
+                    entry = img[i].unsqueeze(0)
+                    save_image(entry, img_name + f"{self.img_counter + i}.png")
 
-                self.img_counter += len(imgs[0])
+            self.img_counter += len(imgs[0])
```
Andrewwango: what's the point of img_counter? It's always reset to 0 every epoch anyway. 

Surely it would be nice to take advantage of this somehow and save images with the epoch in the filename so one can visualise the images throughout training?

Or am I missing something?

deepinv/training/trainer.py
Andrewwango, line 1118: Either make this len(img) or make the counter below add imgs[0].size
```suggestion
                for i in range(len(img)): # per element in batch
```

```suggestion
            for k, img in enumerate(imgs): # ground truth, recon, etc.
```

Andrewwango: remove

deepinv/training/trainer.py
```diff
@@ -1114,9 +1114,10 @@ def plot(self, epoch, physics, x, y, x_net, train=True):
                     img_name = f"{self.save_folder_im}/{titles[k]}/"
                     # make dir
```
Andrewwango: remove

deepinv/training/trainer.py
```diff
@@ -1114,9 +1114,10 @@ def plot(self, epoch, physics, x, y, x_net, train=True):
                     img_name = f"{self.save_folder_im}/{titles[k]}/"
```
Andrewwango: these can go in the outer loop?

Andrewwango (changes requested):
Thanks for fixing this Jeremy!

The comments are primarily not to do with your code, but trying to clean up the code further. sorry you've had to deal with this mess!

deepinv/training/trainer.py
jscanvic: Afaict the point is to save the index of the image in the dataset at test time but the function the snippet is in doesn't outlive a single batch (it's called once per batch) so you have to store the count outside of its scope (here, in the attributes)

It's not the prettiest thing but it serves its purpose

deepinv/training/trainer.py
jscanvic: I think len(tensor) is a bit confusing, I'll go with tensor.size(0)

jscanvic:
Thanks for the review! Should be good now

deepinv/training/trainer.py
jscanvic: (This is never reached at train/eval I think)

Andrewwango: I see thanks!

Andrewwango: Maybe we should clarify in the docstring concisely which/how many images are expected to be saved during train/val/test? 

I still don't know what it's meant to be (I just take what I am given) and it makes it difficult for me to help other users debug.

jscanvic: I double checked and indeed either all of the test samples get saved to disk at test time, i.e., when calling `Trainer.test`, or nothing gets saved at all.

It's fairly easy to verify

Here's the outline of the method:
```
    def plot(self, epoch, physics, x, y, x_net, train=True):
        ...
        save_images = self.save_folder_im is not None
        ...
        if save_images:
            for k, img in enumerate(imgs):  # ground truths, reconstructions, etc.
                img_name = f"{self.save_folder_im}/{titles[k]}/"
                Path(img_name).mkdir(parents=True, exist_ok=True)
                for i in range(img.size(0)):
                    entry = img[i].unsqueeze(0)
                    save_image(entry, img_name + f"{self.img_counter + i}.png")
        ...
```

Either everything gets saved or nothing and it does if and only if `self.save_folder_im` is set. The value of this attribute is changed at exactly two places in the trainer module. 1) At the beginning of the `Trainer.test` function, setting it to the user-provided value passed in as a parameter to the method, and 2) In `Trainer.setup_train` which is automatically called at the beginning of `Trainer.train`, where the attribute is set to `None` thereby disabling saving any image to disk.

It might not be the best behavior but that's the way it is!

jscanvic: (I clarified that in ca19599)

PR #911, opened 2025-11-08
NIQE without pyiqa

### LINKED ISSUE #891: Consider alternatives to `pyiqa` for metrics
`pyiqa` has an increasing number of [required dependencies](https://github.com/chaofengc/IQA-PyTorch/blob/main/requirements.txt), many of which are heavy yet not needed for specific metrics which we use in deepinv.metric. Furthermore, its import is bloated (mostly due to required imports of pandas, scipy and huggingface_hub). We should prefer standalone options (e.g. `lpips`) or reimplement.

Tasks:
- [ ] Remove pyiqa from op deps
- [ ] Remove pyiqa from NIQE #911
- [ ] Remove pyiqa from LPIPS (should be easy)
- [ ] Add optional pyiqa metric adapter for other pyiqa metrics

nucli-vicky:
Following #891 , this PR aims to implement NIQE directly in deepinv, based on the original NIQE matlab code. 

In principle, we could copy paste the `PyIQA` implementation, but this could lead to some licensing issues (CC-NC clashes with BSD 3-Clause  because of the differences in allowing commercial use).

In the future, it would also be nice if we could allow users to fit NIQE on their own groundtruth datasets :)

**Note:** Right now it requires scipy to load the weights which are only available online with `.m` extension. This adds `scipy` to the requirements of denoisers. I would propose to convert them to `.pt` and add them to deepinv's HF page to avoid the need for scipy. Weight files are only 9 KB!

I will review the math and implementation again myself, but would appreciate a cross-check! Unfortunately I don't have a Matlab license, so it's a bit difficult to check the outcome of the original Matlab implementation

The `PyIQA` implementation and this one do not match very well. However, it does return a more logical result for the images used in https://quality.nfdi4ing.de/en/latest/image_quality/NIQE.html:
`PyIQA` returns a better score for the 'dark' (distorted image), which is not the case in the matlab implementation. Small note: the values themselves do depend a lot on which weight versions are used. If we use https://huggingface.co/chaofengc/IQA-PyTorch-Weights/resolve/main/niqe_matlab_params.mat , the values for the current implementation get quite close to the ones shown on the site.

To test:

```python
from deepinv.loss.metric.perceptual import NIQE
import torch

import requests
import numpy as np
from PIL import Image
from io import BytesIO
from deepinv.loss.metric.metric import Metric, import_pyiqa


class OldNIQE(Metric):
    r"""
    Natural Image Quality Evaluator (NIQE) metric.

    Calculates the NIQE :math:`\text{NIQE}(\hat{x})` where :math:`\hat{x}=\inverse{y}`.
    It is a no-reference image quality metric that estimates the quality of images.
    Uses implementation from `pyiqa <https://pypi.org/project/pyiqa/>`_.

    .. note::

        By default, no reduction is performed in the batch dimension.

    :Example:

    >>> from deepinv.utils.demo import load_example
    >>> from deepinv.loss.metric import NIQE
    >>> m = NIQE() # doctest: +IGNORE_RESULT
    (...)
    >>> x_net = load_example("celeba_example.jpg", img_size=128)
    >>> m(x_net) # doctest: +ELLIPSIS
    tensor([...])

    :param str device: device to use for the metric computation. Default: 'cpu'.
    :param bool complex_abs: perform complex magnitude before passing data to metric function. If ``True``,
        the data must either be of complex dtype or have size 2 in the channel dimension (usually the second dimension after batch).
    :param str reduction: a method to reduce metric score over individual batch scores. ``mean``: takes the mean, ``sum`` takes the sum, ``none`` or None no reduction will be applied (default).
    :param str norm_inputs: normalize images before passing to metric. ``l2``normalizes by L2 spatial norm, ``min_max`` normalizes by min and max of each input.
    :param bool check_input_range: if True, ``pyiqa`` will raise error if inputs aren't in the appropriate range ``[0, 1]``.
    """

    def __init__(self, device="cpu", check_input_range=False, **kwargs):
        super().__init__(**kwargs)
        pyiqa = import_pyiqa()
        self.niqe = pyiqa.create_metric(
            "niqe", check_input_range=check_input_range, device=device
        ).to(device)
        self.lower_better = self.niqe.lower_better

    def metric(self, x_net, *args, **kwargs):
        n = self.niqe(x_net).float()
        return n.unsqueeze(0) if n.dim() == 0 else n


if __name__ == "__main__":
    # https://quality.nfdi4ing.de/en/latest/image_quality/NIQE.html

    n = NIQE()
    old_n = OldNIQE()
    url = "https://quality.nfdi4ing.de/en/latest/_images/Reference_Image.png"

    resp = requests.get(url, timeout=15)
    resp.raise_for_status()

    img = Image.open(BytesIO(resp.content))
    arr = np.asarray(img).transpose(2, 1, 0)
    t = torch.tensor(arr).unsqueeze(0).float()
    print(n.metric(t))
    t = torch.tensor(arr).unsqueeze(0).float()
    print(old_n.metric(t))
    url = "https://quality.nfdi4ing.de/en/latest/_images/Image_Dark.png"

    resp = requests.get(url, timeout=15)
    resp.raise_for_status()

    img = Image.open(BytesIO(resp.content))
    arr = np.asarray(img).transpose(2, 1, 0)
    t = torch.tensor(arr).unsqueeze(0).float()
    print(n.metric(t))
    t = torch.tensor(arr).unsqueeze(0).float()
    print(old_n.metric(t))

    url = "https://quality.nfdi4ing.de/en/latest/_images/Image_Sunshine.png"

    resp = requests.get(url, timeout=15)
    resp.raise_for_status()

    img = Image.open(BytesIO(resp.content))
    arr = np.asarray(img).transpose(2, 1, 0)
    t = torch.tensor(arr).unsqueeze(0).float()
    print(n.metric(t))
    t = torch.tensor(arr).unsqueeze(0).float()
    print(old_n.metric(t))
```


nucli-vicky:
On the small test: 
Best to check behavior is to set 
```python
self.niqe = pyiqa.create_metric(
            "niqe", check_input_range=check_input_range, version="matlab", device=device, 
        ).to(device)
```
so PyIQA also uses the original matlab weights. Additionally setting `round_tensor=True` for the new `NIQE`, which is needed to match original matlab behavior and is also what PyIQA does, we get:

---
Standard image:
(PyIQA): 4.2061
(New): 3.7427
(Website): 3.75

Dark image:
(PyIQA): 4.2463
(New): 4.6494
(Website): 4.64

Sunshine image:
(PyIQA): 4.0809
(New): 3.8854
(Website): 3.93

---

So:

(i) PyIQA does not lead to the same relative (which image is better) than the website, new implementation does

(ii) New implementation and website correspond well, but not perfectly. Note here the image resizing is different between matlab and our proposed implementation. 
First, we use bilinear instead of bicubic. I tested bicubic too but this gave worse numerical correspondence (still way closer than PyIQA!). 
Secondly, it's likely the antialiasing between matlab's imresize and torch's interpolate are different. 
(we also use fp32 whereas matlab uses fp64, but I think this should lead only to negligible differences)
So I think some numerical differences are fine and to be expected. 

---

The main concerns are simply:
- To check everything is indeed correctly implemented
-  The fact this would break compatibility between previous versions and the new one in terms of results.

nucli-vicky:
**Notes**

- We should probably do (at least) the inverse in float64 for stability
- Even if we use bicubic, the interpolation does not match. [`pytorch` uses a `-0.75` coefficient ](https://github.com/pytorch/pytorch/blob/5135ace3a3ca836338201a08404ac36c96b58b7c/aten/src/ATen/native/UpSample.h#L409), while [matlab uses `-0.5`](https://stackoverflow.com/a/22093004).
- Right now, eq.1 from the paper is used exactly. So we perform (im - mu) / (sigma + 1). This `+1` works fine for range [0,255], but likely leads to issues if instead we have normalized inputs, or range [0,1]. To be checked. We can have this stabilizer be a parameter maybe, or make clear inputs need to be in the [0,255] intensity setting
- `F.unfold` will not include patches which extent beyond the image, e.g. for a 128,128 image, `F.unfold(im, kernel_size=96, stride=96 will only yield a single patch. By contrast, the original matlab code uses blkproc, [which does zero-padding when necessary](http://www.ece.northwestern.edu/local-apps/matlabhelp/toolbox/images/blkproc.html). Therefore in the matlab code they perform a crop so the image shape is already multiples of the blocksize. This influences the resize / interpolation result. --> We should perform the same thing; original matlab code segment:
```matlab
[row col]        = size(im);
block_rownum     = floor(row/blocksizerow);
block_colnum     = floor(col/blocksizecol);
im               = im(1:block_rownum*blocksizerow, ...
                   1:block_colnum*blocksizecol);      
```

nucli-vicky:
Technically, the input to pseudoinv should be symmetric:
```python
        invcov = torch.linalg.pinv(
            0.5 * (cov_d.to(torch.float64) + cov_p.to(torch.float64))
        ).to(
```
so we can remove fp rounding errors using `0.5*(cov + cov.T)` and set `hermitian=True` in `pinv`. This would give a speedup but more importantly would it lead to better stability?

nucli-vicky:
Will enable fitting NIQE on a dataset too. Should not be too much additional work / code, but I guess does introduce a type of functionality / API which was not yet present? In that sense it is a significant change, but to me it's fine to do this one-off for now and formalize / make code for "building metrics from datasets" more systematic later?

Andrewwango:
@nucli-vicky 
> Should not be too much additional work / code, but I guess does introduce a type of functionality / API which was not yet present?

Yes, it would be great to include this even if it's not general, not sure what the best interface would be, maybe we could simply include it as an example rather than making a feature (because it's not clear how we can have code inside the metric class that imports losses)

nucli-vicky:
@Andrewwango we actually do not need to import loss in this case. The current draft already has code to calculate custom NIQE weights from a dataset

Andrewwango (changes requested):
Hi @nucli-vicky , thanks for this contribution, it's great to go in this direction!

General comments:

- Clarify that we follow original MATLAB + cite for trustworthiness
- Cite pyiqa if we are using their weights, otherwise we should download from original source or host ourselves
- can we copy the images you used for the MATLAB+vs our implementation into a test, to have a test on the output of the NIQE values?
- maybe mention that pyiqa niqe isn’t close to the MATLAB implementation whereas ours is
- (for the interpolation maybe we can use `dinv.physics.DownsamplingMatlab` ([here](https://deepinv.github.io/deepinv/api/stubs/deepinv.physics.DownsamplingMatlab.html)) which does the bicubic interpolation + antialiasing MATLAB-style)
- for backward compatibility maybe we can keep the pyiqa NIQE as an option, throw a warning in the new implementation (we can remove the pyiqa dependency from the optional deps in pyproject and remove the test since it’s just kept as legacy)
- an example for NIQE fitting could be nice e.g. medical images

nucli-vicky:
I have made some adjustments to more closely match original MatLab.

Since I do not have a MatLab license, I tested the proposed implementation against `PyIQA` and `BasicSR` implementation.

Two notes:
- I only based myself on the original MatLab code, not on either of the other two implementations.
- `PyIQA` has confusing naming. On HF, `niqe_modelparameters.mat` and `niqe_matlab_params.mat` are available. I falsely assumed the latter are the weights provided by the authors of NIQE [1]. `niqe_modelparameters.mat` actually matches original weights, so my test does not demonstrate `PyIQA` to be incorrect, as setting `version='matlab'` leads to wrong params being used, while `version='original'` will use the correct params. Additionally I originally did not rescale inputs from [0,255] to [0,1] when using `PyIQA`, which is required as `PyIQA` always applies a multiplication by 255 (always assuming input is [0,1]). The comparison below properly rescales to [0,1] for the `PyIQA` implementations. 

That being said, graphs below show the absolute and relative (normalized to `BasicSR` implementation) NIQE score for different Python-based implementations on Set14. 

Proposed denotes the implementation proposed in this PR.

niqe
normalized_niqe

Note here, `BasicSR` docstring says 
```
    This implementation could produce almost the same results as the official
    MATLAB codes: http://live.ece.utexas.edu/research/quality/niqe_release.zip

    > MATLAB R2021a result for tests/data/baboon.png: 5.72957338 (5.7296)
    > Our re-implementation result for tests/data/baboon.png: 5.7295763 (5.7296)
```
the results displayed in the graphs above actually have for the baboon image:
```
baboon:
  Proposed: 5.714389324188232
  PyIQA_matlab: 6.4498515129089355
  PyIQA_original: 5.738537788391113
  basicsr: 5.90808476428397
```
Both `PyIQA_original` and `Proposed` closely match the value denoted in the docstring, but `BasicSR` strangely enough does not match closely what is described in their docstring. Perhaps earlier testing of `basicsr` used `float64` (live code uses `float32`, also for pseudoinverse). It is ofcourse also possible I made some error in processing.



[1] Original NIQE paper and implementation: https://github.com/utlive/niqe Mittal, Anish, Rajiv Soundararajan, and Alan C. Bovik. "Making a “completely blind” image quality analyzer." IEEE Signal processing letters 20.3 (2012): 209-212.

nucli-vicky:
Based on the above comment I'd say it's likely we match both `PyIQA` and the authors' implementation quite well, so I'd consider the `NIQE` calculation logic in this PR relatively correct. Still a comparison to the original implementation may be desired.

I'll test the logic for making new weights for another dataset (likely MRI) in the upcoming days.

Andrewwango:
Thanks for the in-depth analysis @nucli-vicky , it's really good to have this so we can confidently implement it! May I suggest we document this behaviour in the docstring? I think it would attract users, to know that we match the original version on set14, even better than other python versions out there!

nucli-vicky:
Yes, it should indeed be documented in the docstring, although I myself would not yet be sure to claim we match the original more closely than other Python-based implementations.

deepinv/loss/metric/perceptual.py
```diff
@@ -89,56 +93,426 @@ class NIQE(Metric):
...
+    :param torch.device, str device: device to use for the metric computation. Default: 'cpu'.
     :param bool complex_abs: perform complex magnitude before passing data to metric function. If ``True``,
         the data must either be of complex dtype or have size 2 in the channel dimension (usually the second dimension after batch).
     :param str reduction: a method to reduce metric score over individual batch scores. ``mean``: takes the mean, ``sum`` takes the sum, ``none`` or None no reduction will be applied (default).
-    :param str norm_inputs: normalize images before passing to metric. ``l2`` normalizes by :math:`\ell_2` spatial norm, ``min_max`` normalizes by min and max of each input.
-    :param bool check_input_range: if True, ``pyiqa`` will raise error if inputs aren't in the appropriate range ``[0, 1]``.
-    :param int, tuple[int], None center_crop: If not `None` (default), center crop the tensor(s) before computing the metrics.
```
Andrewwango, line 142: center_crop is still compatible with this, no?

deepinv/loss/metric/perceptual.py
```diff
@@ -89,56 +93,426 @@ class NIQE(Metric):
...
+        self.round = round_tensor
+        self.lower_better = True
+        self.patch_size = patch_size
+        self.device = device
+        self.n_scales = 2
+        self.patch_overlap = patch_overlap
+        self.denominator = denominator
+        self.dtype = dtype
+        if weights_path == "download":
+            resp = requests.get(
```
Andrewwango: https://github.com/deepinv/deepinv/blob/b9b60ac64b1ac04df092c6ab6db1c11dd5e80a6d/deepinv/models/drunet.py#L174-L177

Do you think it would be better to use an existing function to load the model? Might be cleaner

deepinv/loss/metric/perceptual.py
```diff
@@ -89,56 +93,426 @@ class NIQE(Metric):
...
+                "https://huggingface.co/deepinv/demo/resolve/main/niqe_weights.pt",
+                timeout=2.5,
+            )
+            resp.raise_for_status()
+
+            params = torch.load(io.BytesIO(resp.content), weights_only=False)
```
Andrewwango: should generally use strict=True for torch load

examples/evaluation/README.rst
```diff
@@ -0,0 +1,3 @@
+Evaluation
```
Andrewwango: I would call it "Metrics"

examples/evaluation/demo_custom_niqe.py
```diff
@@ -0,0 +1,230 @@
...
+to evaluate denoiser performance.
+
+NIQE is a no-reference image quality metric that compares local image statistics
+against those of pristine (distortion-free) images. Fitting NIQE on a domain-specific
+dataset can better capture the expected image characteristics.
+
+While DIV2K is also natural imaging data, the image quality and sharpness is better compared
+to the dataset NIQE was originally fitted on. The fitting in this example can be done with
+any dataset returning RGB or single channel tensors. Do set a reasonable denominator argument
+relative to your data's pixel intensity range (e.g. for [0, 1], set denominator to 1/255)
```
Andrewwango: This paragraph is slightly weirdly worded - it's unclear what this example does vs. what the user should do to generalise it themselves. For what this example does: we fit it to div2k which is not the same as the original, but we still obtain good weights?

examples/evaluation/demo_custom_niqe.py
```diff
@@ -0,0 +1,230 @@
...
+all_indices = list(range(n_images))
+
+# %%
+# Define denoisers
+# ----------------
+# We wrap each denoiser to handle device placement and precision.
+# We compare DRUNet against MedianFilter at two kernel sizes.
+
+
+class Denoise:
```
Andrewwango: this is just to speed up the example no? not essential code

deepinv/loss/metric/perceptual.py
```diff
@@ -89,56 +93,426 @@ class NIQE(Metric):
...
+        sharpness_threshold: float = 0.75,
+        save_path: str | Path | None = None,
+    ):
+        r"""
+        Fit NIQE model parameters (mu_prisparam, cov_prisparam) from a dataset of 'pristine' images,
+        following the original MATLAB pipeline with two scales and sharpness-based patch selection.
+        ``patch_size``, ``patch_overlap``, and ``denominator`` used are those passed at init (unless modified post-init by user).
+
+        ``dataset`` should yield a (C, H, W) ``Tensor``, where C=1 and C=3 are allowed. If C=3, RGB is assumed and will be converted
+        to greyscale using 0.299*R + 0.587*G + 0.114*B.
```
Andrewwango, line 425: we could use https://deepinv.github.io/deepinv/api/stubs/deepinv.physics.Decolorize.html? it uses the same convention

deepinv/loss/metric/perceptual.py
```diff
@@ -89,56 +93,426 @@ class NIQE(Metric):
...
+
+        all_feats = []
+
+        for x in dataset:
+
+            if x.ndim == 2:
+                x = x.unsqueeze(0)
+            if x.ndim == 3 and x.shape[0] in (1, 3):
+                pass
+            elif x.ndim == 3:
```
Andrewwango: you've asked for CHW in the docstring so this gives now WCH?

deepinv/loss/metric/perceptual.py
```diff
@@ -89,56 +93,426 @@ class NIQE(Metric):
...
+                raise RuntimeError(
+                    f"NIQE only operates on single channel images. Got {x.shape[1]} channels."
+                )
+
+            if self.round:
+                x = x.round()
+
+            _, _, H, W = x.shape
+            if H < self.patch_size or W < self.patch_size:
+                continue  # too small -> should we raise a warning here?
```
Andrewwango: could just print a warning here

deepinv/loss/metric/perceptual.py
```diff
@@ -89,56 +93,426 @@ class NIQE(Metric):
...
+                feats_scales.append(feats)
+
+                if scale == 1:
+                    U = F.unfold(
+                        sigma, kernel_size=(k, k), stride=(stride, stride)
+                    )  # (1, k*k, L)
+                    sharpness = U.mean(dim=1).squeeze(0)  # (L,)
+
+                if scale < self.n_scales:
+                    x_scale = imresize_matlab(
```
Andrewwango: finally, a use for legacy matlab code! more legacy matlab code!

deepinv/loss/metric/perceptual.py
```diff
@@ -89,56 +93,426 @@ class NIQE(Metric):
...
+        block_wnum = math.floor(W / self.patch_size)
+        x_net = x_net[
+            :, :, : block_hnum * self.patch_size, : block_wnum * self.patch_size
+        ]
 
-    def metric(self, x_net, *args, **kwargs):
         n = self.niqe(x_net).float()
         return n.unsqueeze(0) if n.dim() == 0 else n
 
+    def create_weights(
```
Andrewwango, line 413: either noqa this whole function or add a very dummy pytest that tests it end to end (not super important because it's not a core method)

deepinv/loss/metric/perceptual.py
```diff
@@ -89,56 +93,426 @@ class NIQE(Metric):
...
+
+            if Lv < 2:
+                continue  # covariance undefined -> NaN (match MATLAB behavior)
+            Xc = Xv - mu_b
+            cov[b] = (Xc.t() @ Xc) / (Lv - 1)
+
+        return mu, cov
+
+    def metric(self, x_net: torch.Tensor, *args, **kwargs) -> torch.Tensor:
+        """We base ourselves on the original Matlab code (available at http://live.ece.utexas.edu/research/quality/niqe_release.zip), but allow some exceptions:
```
Andrewwango: put this key info in the docstring instead. Let this docstring inherit that of the base class

deepinv/loss/metric/perceptual.py
```diff
@@ -89,56 +93,426 @@ class NIQE(Metric):
...
+        stride = self.patch_size - self.patch_overlap
+        n_patches_h = (H - self.patch_size) // stride + 1
+        n_patches_w = (W - self.patch_size) // stride + 1
+        if n_patches_h * n_patches_w < 2:  # pragma: no cover
+            raise RuntimeError(
+                f"NIQE requires more than 1 patch to compute covariance, but got only {n_patches_h * n_patches_w} patches "
+                f"for batch of shape {x_net.shape} with patch_size={self.patch_size} and patch_overlap={self.patch_overlap}. "
+            )
+        if C == 3:
+            luminance_weights = torch.tensor(
```
Andrewwango, line 390: decolorize

Andrewwango (changes requested):
Thanks @nucli-vicky for this awesome PR! Mostly small changes now + one question: should we have at least one pytest that at least checks that niqe gives some expected hardcoded values for an image that you know well? that way we don't need the old niqe, but still can make sure that everything works well in pytest. Thanks!

deepinv/loss/metric/perceptual.py
nucli-vicky: Good catch, thanks!

deepinv/loss/metric/perceptual.py
nucli-vicky: I made use of `load_state_dict_from_url` now. This does cause the downloading to be verbose by default.

deepinv/loss/metric/perceptual.py
nucli-vicky: I'd personally prefer to do this in pure PyTorch, as it doesn't add much extra code and adds readability + reduces dependency

deepinv/loss/metric/perceptual.py
nucli-vicky: This is some legacy code as I was initially working with a dataset returning PIL images. 

In essence this makes the code still works if we get HWC, will remove to be more strict on the CHW requirement!

deepinv/loss/metric/perceptual.py
nucli-vicky: agree this needs a test, done!

examples/evaluation/demo_custom_niqe.py
nucli-vicky: (i) Speedup
(ii) Handle placing img on correct device somewhere outside the relevant loop.

examples/evaluation/demo_custom_niqe.py
nucli-vicky: agree. Tried addressing now

examples/evaluation/demo_custom_niqe.py
nucli-vicky: edit: I removed the Denoise class from the example

nucli-vicky:
@Andrewwango thanks for the review! I tried to address all of your comments

Regarding the general question on testing:

I now reintroduced NIQE to the no reference metric tests, where it passes the sanity checks. Is this sufficient, or would an additional test specifically for NIQE `.metric` be required?

deepinv/tests/test_metric.py
```diff
@@ -78,14 +72,14 @@ def choose_full_reference_metric(metric_name, device, **kwargs) -> metric.Metric
...
 def choose_no_reference_metric(metric_name, device, **kwargs) -> metric.Metric:
-    if metric_name in ("NIQE",):
-        pytest.importorskip(
-            "pyiqa",
-            reason="This test requires pyiqa. It should be "
-            "installed with `pip install pyiqa`",
-        )
     if metric_name == "NIQE":
-        return metric.NIQE(**kwargs, device=device)
+        return metric.NIQE(
```
Andrewwango, line 82: Do you think we could add a test that basically reassures the user that our NIQE implementation matches other implementations? i.e. the plot you made, comparing on the set3 images, our NIQE vs pyqia vs matlab vs basicsr.

Image

Andrewwango (approved):
Thanks @nucli-vicky ! Approving for now, but if you have time it would be great to have one more test basically matching the nice existing plot you made, to futureproof the class!

deepinv/tests/test_metric.py
nucli-vicky: In principle we could, but then we'd have to hardcode a list of values for the other implementations.

It also seems a bit difficult to have  an appropriate 'closeness' threshold, which is not handpicked? As seen in the graph, there are ofcourse images where there is some difference between values given by implementations.

Andrewwango: I guess a threshold of 2 points will cover the graph. Just to make sure it's sensible I guess? But I don't mind - whatever you think is best. Happy to merge once GPU tests pass

nucli-vicky:
@Andrewwango added the test against the other two Python NIQE implementations. It's a bit ugly since we make use of the filename to map to results from other implementations..

Andrewwango (approved):
LGTM, thanks a lot! Will merge after GPU tests/docs

nucli-vicky:
@Andrewwango 

Seems the GPU failure is related to something else,

```
FAILED deepinv/tests/test_models.py::test_gaussian_noise_estimators[device1-0.1-1-image-pca]@main - AssertionError: assert False
 +  where False = <built-in method allclose of type object at 0x7f7e7639a800>(tensor([0.1014, 0.2043, 0.3019], device='cuda:0'), tensor([0.1014, 0.2043, 0.3010], device='cuda:0'), rtol=0.001, atol=0.0001)
 +    where <built-in method allclose of type object at 0x7f7e7639a800> = torch.allclose
 +    and   tensor([0.1014, 0.2043, 0.3019], device='cuda:0') = PatchCovarianceNoiseEstimator()(tensor([[[[ 0.1828,  0.1722,  0.4257,  ...,  0.3157,  0.1120,  0.3098],\n          [ 0.1983,  0.1375,  0.1953,  ...,  0...,  0.2661, -0.0294],\n          [ 0.5931,  0.6235,  0.4967,  ...,  0.2852,  0.8504,  0.2376]]]],\n       device='cuda:0'))
 +    and   tensor([0.1014, 0.2043, 0.3010], device='cuda:0') = <built-in method cat of type object at 0x7f7e7639a800>([tensor([0.1014], device='cuda:0'), tensor([0.2043], device='cuda:0'), tensor([0.3010], device='cuda:0')])
 +      where <built-in method cat of type object at 0x7f7e7639a800> = torch.cat
```

nucli-vicky:
@Andrewwango Apologies for the mistake in the example. I addressed and reconfirmed the demo runs without warnings locally.

Andrewwango:
@nucli-vicky seems like one more tiny mistake to make GPU docs pass:
`/local/jtachell/deepinv/deepinv/docs/source/auto_examples/metrics/demo_custom_niqe.rst:445: WARNING: Title underline too short.`

PR #921, opened 2025-11-10
Add Laplace noise model

bemc22:
This pull request introduces a new Laplace noise model to the physics module, allowing users to simulate Laplace-distributed noise in measurements. The changes include implementation, documentation, and testing support for the new noise model.

**New Laplace noise model:**

* Added the `LaplaceNoise` class to `deepinv/physics/noise.py`, which implements Laplace-distributed noise with configurable scale parameter `b` and random generator support. The class includes docstrings and an example for user guidance.

**Testing updates:**

* Added a new test `test_laplace_noise_moments` in `test_noise_model.py` to verify their mean and variance against expected statistics.  
* Included "Laplace" in the list of noise models tested in `test_noise_model.py`.

### DIFF
--- a/deepinv/physics/noise.py
+++ b/deepinv/physics/noise.py
@@ -5,7 +5,6 @@ from typing import Callable
 from collections.abc import Iterable
 import warnings
 import deepinv as dinv
-from deepinv.physics.generator import noise
 
 
 class NoiseModel(nn.Module):


jscanvic:
Thanks! Small pointer, sorry if it's not relevant, we should be able to seed `torch.distributions.Laplace` using one of at least two tricks:
* Using the inverse cdf trick: `return x + d.icdf(torch.rand_like(x.shape, generator=self.rng))`
* Setting the global seed using `torch.manual_seed` temporarily and reverting back to the previous RNG state

Hopefully that helps!

bemc22:
> Thanks! Small pointer, sorry if it's not relevant, we should be able to seed `torch.distributions.Laplace` using one of at least two tricks:
> 
> * Using the inverse cdf trick: `return x + d.icdf(torch.rand_like(x.shape, generator=self.rng))`
> * Setting the global seed using `torch.manual_seed` temporarily and reverting back to the previous RNG state
> 
> Hopefully that helps!

Thanks, I think the inverse trick might introduce some computational overhead. I'll try the manual seed test to see if it works.

deepinv/physics/__init__.py
```diff
@@ -32,9 +32,9 @@
     LogPoissonNoise,
     GammaNoise,
     SaltPepperNoise,
-    ZeroNoise,
```
nucli-vicky, line 35: Good catch!

deepinv/physics/noise.py
```diff
@@ -997,3 +997,53 @@ def _infer_device(
         )
 
     return input_devices.pop() if input_devices else default
+
+
+class LaplaceNoise(NoiseModel):
```
nucli-vicky: Can you please put `LaplacianNoise` above `_infer_device`, instead of at the end of the file? It makes more sense for all noise operators to be grouped together in the code and not have `_infer_device` appear between them randomly :).

deepinv/tests/test_noise_model.py
```diff
@@ -50,6 +56,16 @@ def choose_noise(noise_type, rng):
     return noise_model
 
 
+def choose_noise_stats(noise_type):
+    b = 0.1
+    if noise_type == "Laplace":
+        noise_model = (dinv.physics.LaplaceNoise(b=b), b)
+    else:
+        raise Exception("Noise model not found")
```
nucli-vicky, line 53: Could you please put `else: # pragma: no cover`? Otherwise, the exception raise is seen as not covered by the tests by coverage.

deepinv/physics/noise.py
bemc22: thanks, solved

deepinv/tests/test_noise_model.py
bemc22: thanks, solved

bemc22:
> Thanks! Small pointer, sorry if it's not relevant, we should be able to seed `torch.distributions.Laplace` using one of at least two tricks:
> 
> * Using the inverse cdf trick: `return x + d.icdf(torch.rand_like(x.shape, generator=self.rng))`
> * Setting the global seed using `torch.manual_seed` temporarily and reverting back to the previous RNG state
> 
> Hopefully that helps!

Hi, I included `torch.manual_seed` here in [5bd368e] and also added two functions, `get_rng_state` and `restore_rng`, to manage the RNG state. Let me know if this correctly addresses your suggestion, thanks.

deepinv/tests/test_noise_model.py
```diff
@@ -78,6 +97,30 @@ def test_rng(name, device, rng, dtype):
     assert not torch.allclose(y_1, y_2)
 
 
+@pytest.mark.parametrize("name", NO_RNG_NOISES)
+@pytest.mark.parametrize("device", DEVICES)
+@pytest.mark.parametrize("dtype", DTYPES)
+def test_stats(name, device, dtype):
```
jscanvic:
```suggestion
def test_noise_moments(name, device, dtype):
```

Slightly longer/descriptive

deepinv/tests/test_noise_model.py
```diff
@@ -78,6 +97,30 @@ def test_rng(name, device, rng, dtype):
     assert not torch.allclose(y_1, y_2)
 
 
+@pytest.mark.parametrize("name", NO_RNG_NOISES)
```
jscanvic: I understand the test is meant to be used for other distributions as well but for now it only supports Laplace distributions so I'd prefer if we got rid of `name` and let the next person adding a new test make the appropriate generalization

deepinv/tests/test_noise_model.py
```diff
@@ -78,6 +97,30 @@ def test_rng(name, device, rng, dtype):
...
+
+    imsize = (1, 3, 7, 16)
+    x = torch.zeros(imsize, device=device, dtype=dtype)
+    noise_model, stats = choose_noise_stats(name)
+    y = noise_model(x, seed=0)
+    noise = y - x
+
+    true_mean = x.mean().item()
+
+    if name == "Laplace":
```
jscanvic: Don't put an if here (see above)

deepinv/tests/test_noise_model.py
```diff
@@ -44,12 +51,24 @@ def choose_noise(noise_type, rng):
...
         noise_model = dinv.physics.RicianNoise(sigma=sigma, rng=rng)
+    elif noise_type == "Laplace":
+        noise_model = dinv.physics.LaplaceNoise(b=sigma, rng=rng)
     else:
         raise Exception("Noise model not found")
 
     return noise_model
 
 
+def choose_noise_stats(noise_type):
```
jscanvic: Inline this function (see above)

deepinv/tests/test_noise_model.py
```diff
@@ -14,7 +14,14 @@
     "LogPoisson",
     "SaltPepper",
     "RicianNoise",
+    "Laplace",
 ]
+
+# Noise model which do not have a `rng` attribute
+NO_RNG_NOISES = [
```
jscanvic: Remove this variable (see above)

deepinv/physics/noise.py
```diff
@@ -968,6 +968,78 @@ def forward(
...
+        :param torch.Tensor x: measurements
+        :param float, torch.Tensor b: scale of the noise. If not None, it will overwrite the current noise level.
+        :param int seed: the seed for the random number generator, if `rng` is provided.
+        :returns: noisy measurements
+        """
+        self.update_parameters(b=b, **kwargs)
+        self.to(x.device)
+
+        # Save current RNG state if seed is provided
+        if seed is not None:
```
jscanvic: * Seed using `self.rng_manual_seed` then get the state of `self.rng` instead to be more consistent with what we do everywhere else

* Use `torch.random.fork_rng` instead (sorry I should have suggested it earlier). It handles restoring the previous state and also with the multiple rng states for the cpu and cuda

jscanvic (changes requested):
Thanks Brayan it looks really good! Mostly minor comments about the way tests are written

Can you run a quick benchmark testing if the `icdf` trick is any slower for reproducible sampling? I believe it might be what they do internally so we might be able to use it. I also think it's cleaner than tweaking global rng state. Happy to keep things as is if the benchmark says it's slower to change

mh-nguyen712:
Thanks @bemc22 for the nice contribution!

I think we can just use the inverse CDF, as suggested by  @jscanvic ! It has a closed form and isn’t too expensive.
See the [wiki](https://en.wikipedia.org/wiki/Laplace_distribution#Cumulative_distribution_function).
In fact, this is what’s done under the hood in PyTorch: https://github.com/pytorch/pytorch/blob/v2.9.1/torch/distributions/laplace.py#L73-L85

We can take inspiration from their approach to clipping for numerical stability.

bemc22:
Thanks! You’re absolutely right @mh-nguyen712 @jscanvic. There is no perceptible computational overhead from using `icdf`, and I’ll address your comments accordingly. Here’s the benchmark.

code:
```python
import time,torch,matplotlib.pyplot as plt
sizes=[2**i for i in range(1,20)]
t_s,t_i=[],[]
for s in sizes:
    x=torch.zeros(s); d=torch.distributions.Laplace(0.,torch.ones_like(x)*0.1)
    start=time.perf_counter()
    for _ in range(1000): _=x+d.sample()
    t_s.append(time.perf_counter()-start)
    start=time.perf_counter()
    for _ in range(1000): _=x+d.icdf(torch.rand_like(x))
    t_i.append(time.perf_counter()-start)
plt.plot(sizes,t_s,label='sample');plt.plot(sizes,t_i,label='icdf')
plt.xscale('log',base=2);plt.yscale('log');plt.legend();plt.show()
````

output:
image

File: deepinv/tests/test_noise_model.py (hunk 5 above)
Comment (bemc22) [reply deepinv/tests/test_noise_model.py]: Thanks, I’ve set the test specifically for Laplace now

File: deepinv/physics/noise.py (hunk 9 above)
Comment (bemc22) [reply deepinv/physics/noise.py]: Thanks, now that we’re using the ICDF for Laplace, I’m using the `self.rand_link` method already built into `NoiseModel`, which internally calls `self.rng_manual_seed`. 

I’m not sure if it’s necessary to use `torch.random.fork_rng` with the current code, since it’s not used in the other noise models, let me know if it’s needed. 

The code is now much easier to read and simpler :)

File: deepinv/physics/noise.py [hunk 10]
```diff
@@ -968,6 +968,55 @@ def forward(
...
+        b: float | torch.Tensor = 0.1,
+        rng: torch.Generator | None = None,
+    ):
+        device = _infer_device([b, rng])
+        super().__init__(rng=rng)
+        b = self._float_to_tensor(b)
+        b = b.to(device)
+        self.register_buffer("b", b)
+
+    def forward(self, x, b=None, seed=None, **kwargs):
```
mh-nguyen712: can you add type hint please

```suggestion
    def forward(self, x: Tensor, b: float | Tensor = None, seed: int = None, **kwargs):
```

deepinv/physics/noise.py
```diff
@@ -968,6 +968,55 @@ def forward(
...
+
+        :param torch.Tensor x: measurements
+        :param float, torch.Tensor b: scale of the noise. If not None, it will overwrite the current noise level.
+        :param int seed: the seed for the random number generator, if `rng` is provided.
+        :returns: noisy measurements
+        """
+        self.update_parameters(b=b, **kwargs)
+        self.to(x.device)
+
+        d = torch.distributions.Laplace(0.0, torch.ones_like(x) * self.b)
```
mh-nguyen712: I would implement the icdf directly here to get rid of creating an instance of torch distribution.

mh-nguyen712 (approved):

PR #922, opened 2025-11-10
Hotfix for optional missing dependency breaking CI

jscanvic:


deepinv/tests/test_models.py
```diff
@@ -1364,6 +1364,12 @@ def test_client_mocked(return_metadata):
 @pytest.mark.parametrize("upscale", [None, 1, 2])
 @pytest.mark.parametrize("upsampler", [None, "pixelshuffle"])
 def test_swinir_upsample_without_upsampler(upscale, upsampler):
+    pytest.importorskip(
```
mh-nguyen712, line 1367: It means the test will be skipped? And also, in denoiser dependencies we already have timm.

jscanvic: Only when the `timm` is not installed, it will still run in the pipeline with optional dependencies installed

Issue #725, opened 2025-09-08
Mistake in the error computation for test_tomography


jscanvic:
### Environment

/

### Describe the bug

The error is not computed properly in the tests for the tomography operator, it does the average before taking the absolute value instead of doing the reverse:

https://github.com/deepinv/deepinv/blob/cd41f68adeb5f83ebaef98957461e221076765a1/deepinv/tests/test_physics.py#L1323

Related: #715 

### Steps to reproduce

/

### Expected behavior

/

PR #930, opened 2025-11-14
Clean Tomography operator and fix related tests

romainvo:
In this PR, I am modifying the default behavior of the Tomography operator to ensure it passes all the required tests of a linear `OPERATOR`, as all the necessary bricks have already been implemented. I have added a `2DParallelBeamCT` and `2DFanBeamCT` to the list of tested operators.
- [x] `test_operators_adjointness` --> now the call to A_adjoint defaults to `adjoint_via_backprop=True`
- [x] `test_operators_norm` --> already handled in #715 
- [x] `test_pseudo_inverse` --> now A_dagger calls the super() implementation by default and computes the pseudo inverse with the `CG` solver. The user has the option to use filtered back-projection with `physics.A_dagger(y, fbp=True)`
- [x] `test_physics_state_dict`  / `test_clone` --> I cleaned the Radon, IRadon modules in `physics/functional/radon.py`. I have slightly altered the behaviour of `get_all_tensor_attrs` in the test to account for non-persistent buffers (i.e., buffers that should not be saved). For example, in Radon and IRadon, some tensors required for the forward are pre-computed. Because they should match the device of the Module, they are set as buffers, but saving them would be pointless as they're always built the same way at startup.

I also fix the separate pseudo-inverse test implemented in `test_tomography` (raised in #725). The relative threshold `r_tol` is set to the minimal value that passes the test when `fbp=True`. 

~~I did not apply these changes to `TomographyWithAstra`, especially what A_dagger calls~~. From my experience, lsqr/CG solvers are very sensitive to having the true adjoint. When investigating numerically, computing `Tomography.WithAstra.A_dagger` with the `lsqr` solver consistently yields worse results than `fbp` computation. What do you think @askorikov ?

### DIFF
--- a/deepinv/physics/functional/radon.py
+++ b/deepinv/physics/functional/radon.py
@@ -284,7 +284,9 @@ class Radon(nn.Module):
         N, C, W, _ = x.shape
 
         if self.parallel_computation:
-            rotated_par = grid_sample(x, self.all_grids.expand(N, -1, -1, -1).to(x.device))
+            rotated_par = grid_sample(
+                x, self.all_grids.expand(N, -1, -1, -1).to(x.device)
+            )
             out = (
                 rotated_par.sum(2).reshape(N, C, len(self.theta), -1).transpose(-2, -1)
             )


askorikov:
Hi @romainvo! Interesting. I would expect iterative solvers to be close or better than FBP, but they can explode (because of overfitting to noise) or produce artifact patterns (because of mismatched adjoint) if you go for very high number of iterations. Can I run some specific code to test?

romainvo:
> Hi @romainvo! Interesting. I would expect iterative solvers to be close or better than FBP, but they can explode (because of overfitting to noise) or produce artifact patterns (because of mismatched adjoint) if you go for very high number of iterations. Can I run some specific code to test?

To test what I mention you can modify `TomographyWithAstra` like this, by default super() runs a least squares solver for maximum 50 iterations. With the kwarg `solver` you can choose the iterative scheme, e.g. `'CG'` or  `'lsqr'`

```python
    def fbp(self, y: torch.Tensor, **kwargs) -> torch.Tensor:
        filtered_y = self.filter(y, dim=-1)
        out = self.A_adjoint(self.fbp_weighting(filtered_y))
        if self.normalize:
            out *= self.operator_norm**2

        return out
    
    def A_dagger(self, y: torch.Tensor, fbp: bool = False, **kwargs) -> torch.Tensor:
        if fbp:
            return self.fbp(y, **kwargs)
        else:
            return super(TomographyWithAstra, self).A_dagger(y, **kwargs)
```

On the noise-free tests `python -m pytest test_external_libraries.py`, it does appear like running A_dagger with 'CG' yields lower error.

My experience was with the Walnut dataset, where I would have better results with the method (accelerated gradient descent ) released with the dataset (https://github.com/cicwi/WalnutReconstructionCodes) than a CG solver. But I don't have code ready right now to show you.

askorikov:
Right, I see. I do suspect that subjectively FBP quality may be better than iterative solver results in many cases. For a general guess though (in `A_dagger`), it feels like the optimization-based approach should be more robust.

romainvo:
> Right, I see. I do suspect that subjectively FBP quality may be better than iterative solver results in many cases. For a general guess though (in `A_dagger`), it feels like the optimization-based approach should be more robust.

I have updated the `TomographyWithAstra` accordingly, now FDK/FBP can be used optionally by calling `A_dagger(y, fbp=True)`

deepinv/physics/functional/radon.py
```diff
@@ -286,9 +284,7 @@ def forward(self, x):
         N, C, W, _ = x.shape
 
         if self.parallel_computation:
-            rotated_par = grid_sample(
-                x, self.all_grids_par.repeat(N, 1, 1, 1).to(x.device)
-            )
+            rotated_par = grid_sample(x, self.all_grids.repeat(N, 1, 1, 1).to(x.device))
```
mh-nguyen712:
```suggestion
            rotated_par = grid_sample(x, self.all_grids.expand(N, -1, -1, -1).to(x.device))
```

deepinv/physics/tomography.py
```diff
@@ -122,7 +122,7 @@ def __init__(
         img_width: int,
         circle: bool = False,
         parallel_computation: bool = True,
-        adjoint_via_backprop: bool = False,
+        adjoint_via_backprop: bool = True,
```
mh-nguyen712, line 125: This change is not backward compatible... but I'm happy to keep it. Can you just add a note to the changelog please?

romainvo: I think it's done already, please tell me if the comment is sufficiently explicit ?

mh-nguyen712: I meant a note saying that the default behavior is changed. So that it will appear here: https://deepinv.github.io/deepinv/changelog.html

romainvo: In `docs/source/changelog.rst` you mean ? The "(Breaking)" indication is already added with a comment, is it okay like this ? (https://github.com/deepinv/deepinv/pull/930/files#diff-ef86237172bb2f91c585e845b38d6fb28bb6ac128ff8efa541cdd86f81751bf2R18)

mh-nguyen712: Sorry I missed that information. I'll re-run the failing test and prove once It've passed

Issue #898, opened 2025-11-06
Liu-Jia Padding for Real-World Deblurring


jscanvic:
### Describe the feature

Spectral deblurring algorithms (Wiener deconvolution, Richardson-Lucy) result in severe ringing artifacts when applied to real-world blurry images. Liu and Jia (2008) overcome that using a padding and cropping scheme that smoothes out image borders before deconvolution is performed. I suggest we add support for it.

The method is a bit old but it has been used successfully in a recent paper for real-world super-resolution:

Deep Unfolding Network for Image Super-Resolution, Zhang et al. (CVPR 2020)
https://arxiv.org/abs/2003.10428

### Motivation

/

PR #934, opened 2025-11-17
Liu-Jia Padding

jscanvic:
Towards #898

Liu-Jia padding reduces ringing artefacts introduced by spectral deconvolution methods that generally assume a periodic blur model that is unsuitable for real-world deblurring where opposite edges are decorrelated.

**TODO**

- [x] Improve the the explanations in the example
- [x] Document `deepinv.utils.liu_jia_pad`

**Simple experiment**

I blur a 256x256 RGB image using a Gaussian filter with a standard deviation of 1px and compare the effect of using no padding vs using Liu-Jia padding before using one of two deconvolution method (simple inverse filtering and Wiener deconvolution). We see that Liu-Jia padding prevents the ringing artifacts in Wiener deconvolution.

_Inverse filtering_
image

_Wiener deconvolution_
image

**Other changes**
- I added support for using the standard (non-signed flipped orthogonal) DST-I
- I made it n-dimensional


Tmodrzyk:
Hi @jscanvic is this ready for review ? Do you need any help for the last TODOs ? :smile:

jscanvic:
Almost I'll wrap it up shortly but thanks for offering your help!

deepinv/utils/_spectral.py
```diff
@@ -0,0 +1,179 @@
+"""Tools from spectral analysis"""
+
+from __future__ import annotations
+
+import torch
+
+
+def _dst1(x: torch.Tensor, *, dim: int = -1, inverse: bool) -> torch.Tensor:
+    r"""
+    Compute the one-dimensional discrete sine transform of type I (DST-I) or its inverse (IDST-I)
```
Tmodrzyk: I didn't know about DST ! Maybe we could add a link to wikipedia https://en.wikipedia.org/wiki/Discrete_sine_transform ?

deepinv/utils/_spectral.py
```diff
@@ -0,0 +1,179 @@
...
+    if not inverse:
+        c = -1 / 2
+    else:
+        c = -1 / (N + 1)
+
+    # Compute the transform
+    return c * y.imag
+
+
+def liu_jia_pad(
```
Tmodrzyk: I’m a bit concerned that this functionality may remain hidden in ``deepinv/utils/``.
It might be more discoverable if moved to ``deepinv/physics/blur``, next to functions such as gaussian_blur.
It could also just become a padding option in the Blur physics.
Keepin DST here is fine though.

examples/physics/demo_liu_jia_padding.py
```diff
@@ -0,0 +1,164 @@
...
+
+# Define blur kernel and physics
+gaussian_std = 1.0
+ksize = 6 * math.ceil(gaussian_std) + 1
+u = torch.arange(ksize, device=device) - ksize // 2
+v = torch.arange(ksize, device=device) - ksize // 2
+u, v = torch.meshgrid(u, v, indexing="ij")
+kernel = torch.exp(-(u**2 + v**2) / (2 * gaussian_std**2))
+kernel /= kernel.sum()
+kernel = kernel.unsqueeze(0).unsqueeze(0)
```
Tmodrzyk: I think we should use ``gaussian_blur`` from `deepinv/physics/blur`

examples/physics/demo_liu_jia_padding.py
```diff
@@ -0,0 +1,164 @@
...
+    )
+
+
+def oetf(x: torch.Tensor) -> torch.Tensor:
+    # Map from linear sRGB to sRGB
+    return torch.where(
+        x <= 0.0031308,
+        x * 12.92,
+        1.055 * (x ** (1 / 2.4)) - 0.055,
+    )
```
Tmodrzyk: is this step required to apply Liu-Jia padding ? I'm not familiar with the linear sRGB color space

examples/physics/demo_liu_jia_padding.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+y = physics(eotf(x))
+
+# Crop for comparison
+if kernel.shape[-2] % 2 != 1 or kernel.shape[-1] % 2 != 1:
+    raise ValueError("Kernel size is expected to be odd")
```
Tmodrzyk: Same comment here, is it just for this demo or is it required for Liu-Jia padding to work properly ?

examples/physics/demo_liu_jia_padding.py
```diff
@@ -0,0 +1,164 @@
...
+
+dinv.utils.plot(
+    [x, oetf(y), x_hat_liu_jia_padding, x_hat_no_padding],
+    [
+        f"GT",
+        f"Blurry {base_psnr:.1f} dB",
+        f"Liu-Jia Padding {psnr_liu_jia_padding:.1f} dB",
+        f"No Padding {psnr_no_padding:.1f} dB",
+    ],
+)
```
Tmodrzyk, line 165: We could add a title to the figure or to the cell with the method tested (inverse filtering, wiener...)

examples/physics/demo_liu_jia_padding.py
```diff
@@ -0,0 +1,164 @@
+"""
```
Tmodrzyk: Here it would be better to have roughly the same introduction as the ``liu_jia_pad`` method with links to bibliography and web-pages.

Tmodrzyk (commented):
Thanks a lot for this nice feature @jscanvic ! 
My main concern is where the padding is implemented + some comments on the demo file.

I fear that users might not know about this new feature if it stays in ``deepinv/utils/_spectral``. 
I think it would be better to have it as a padding option (if the requirements for it to work are not too restrictive) or as a helper function in ``deepinv/physics/blur``.

jscanvic:
Thanks Thibaut! Will get back to you asap

deepinv/utils/_spectral.py
jscanvic: I did & noticed that there's already a DST-I implementation in the code base so I merged the two and added a regression test to make sure it doesn't break backwards compatibility 36f0da2294908cfab27fe8db37cee127dad7ffec

examples/physics/demo_liu_jia_padding.py
jscanvic: It's a bit orthogonal it's preprocessing to make the problem more realistic (blur happens in a linearized color space)

I don't mind getting rid of it if you think it makes the example confusing (and/or opening a separate issue to support it in Blur/BlurFFT)

Tmodrzyk: Ok I was not familiar with this Blur model, this is probably worth a separate PR with a new feature ! We can remove it from the example for now.

examples/physics/demo_liu_jia_padding.py
jscanvic: Not needed I got rid of it

examples/physics/demo_liu_jia_padding.py
jscanvic: Sounds good to me, I got rid of it and made an issue for it: #1301

jscanvic:
Thanks for the review Thibaut and sorry for the delay! It should be good now

deepinv/physics/functional/blur.py
```diff
@@ -363,6 +365,136 @@ def bilinear_filter(
...
+    B[..., :, :alpha] = x[..., :, -alpha:]
+    B[..., :, -alpha:] = x[..., :, :alpha]
+    b = torch.arange(padding_w) / (padding_w - 1)
+    b = b.view((1,) * len(BC) + b.shape)
+    B[..., 0, alpha:-alpha] = (1 - b) * B[..., 0, alpha - 1, None] + b * B[
+        ..., 0, -alpha, None
+    ]
+    B[..., -1, alpha:-alpha] = (1 - b) * B[..., -1, alpha - 1, None] + b * B[
+        ..., -1, -alpha, None
+    ]
```
Tmodrzyk: This code block is a bit hard to read, could we add one or two comments to explain what's going on ?

deepinv/tests/test_physics_functional.py
```diff
@@ -347,3 +347,52 @@ def gaussian_blur(
...
+
+    torch.manual_seed(0)
+    x = torch.randn(*shape, device=device)
+
+    ref = dst1_legacy(x)
+    new = dF.dst1(x, dim=-1, inverse=False, orthosf=True)
+
+    assert ref.shape == new.shape
+    assert torch.allclose(ref, new, rtol=1e-4, atol=1e-5)
+
```
Tmodrzyk: Thanks for testing this ! I'm wondering if we should still keep it, because we will probably forget that another implementation of DST existed at some point.
Maybe we should specify a version beyond which this test could be removed ?

examples/physics/demo_liu_jia_padding.py
```diff
@@ -0,0 +1,165 @@
...
+                raise ValueError(f"Unknown deconvolution kind: {self.kind}")
+
+        # Clipping to attenuate artifacts
+        x_hat = x_hat.clip(0, 1)
+
+        # Cropping if Liu-Jia padding was used
+        if margin is not None:
+            x_hat = x_hat[..., margin[0] : -margin[0], margin[1] : -margin[1]]
+
+        return x_hat
```
Tmodrzyk, line 124: It's awesome to show how to build a reconstructor integrating the new padding.
At the same time, it doesnt seem very conveninent to have to wrap everything.
Couldn't we integrate the padding inside the ``BlurFFT`` physics somehow ? 
So that we could just use the calls to ``physics_circular.A_dagger(y)`` and ``physics_circular.prox_l2(y=y, z=0, gamma=self.gamma)`` with the liu-jia padding.

deepinv/physics/functional/blur.py
```diff
@@ -363,6 +365,136 @@ def bilinear_filter(
     return w.unsqueeze(0).unsqueeze(0)
 
 
+def liu_jia_pad(
+    x: torch.Tensor, *, padding: tuple[int, int], alpha: int = 1
+) -> torch.Tensor:
```
Tmodrzyk: Do we support different devices ? It seems everything is meant for CPU here

deepinv/physics/functional/blur.py
```diff
@@ -363,6 +365,136 @@ def bilinear_filter(
...
+    else:
+        C[..., alpha - 1 : -alpha + 1, alpha - 1 : -alpha + 1] = _compute_padding(
+            C[..., alpha - 1 : -alpha + 1, alpha - 1 : -alpha + 1]
+        )
+
+    # Combine the original image and the padding images to form the final padded image
+    A = A[..., alpha - 1 : -alpha - 1, :]
+    B = B[..., :, alpha:-alpha]
+    C = C[..., alpha:-alpha, alpha:-alpha]
+    zB = torch.cat((x, B), dim=-1)
```
Tmodrzyk: Wouldn't this fail if `x` and `B` have different devices ?

examples/physics/demo_liu_jia_padding.py
```diff
@@ -0,0 +1,165 @@
...
+
+This demo compares deconvolution with and without Liu-Jia padding, for both inverse filtering
+and Wiener filtering, on a realistic blurred image obtained using valid (cropped) convolution instead of circular convolution.
+"""
+
+import torch
+import deepinv as dinv
+import math
+
+device = "cpu"
```
Tmodrzyk: We should select a cuda device if available

Tmodrzyk (changes requested):
Hi @jscanvic thank you for the updates, we're almost there !
The demo looks awesome :)

My main concern is how devices are handled inside the functional, we should make sure everything is sent to the same device as the input.
The demo only tested on `cpu`, when I swapped to `cuda` I got device-related errors.

An LLM also pointed towards possible crashes if one padding direction is zero, maybe we should double check that.

Otherwise just minor comments / discussions for the integration, but I would be ok merging as is (with the device bug fixed).

examples/physics/demo_liu_jia_padding.py
jscanvic: Good catch thanks

deepinv/physics/functional/blur.py
jscanvic: Indeed I added the missing device and dtype!

deepinv/physics/functional/blur.py
jscanvic: I made it compatible with cuda

examples/physics/demo_liu_jia_padding.py
jscanvic: I'm not sure it makes sense to put it in A_dagger (it wouldn't be the true pseudo-inverse) or in prox_l2 (it wouldn't be the true Wiener filter)

Also, I agree that it's inconvenient to have to wrap things but it's a tangential issue, if you want to use inverse filtering and/or Wiener filtering for deblurring you already have to write a Reconstructor wrapper anyway (that's what I've been doing)

Some options I can see are:
1. Have a LiuJiaReconstructor that does the wrapping of an existing reconstructor with the padding preprocessing step and the cropping postprocessing step
2. Possibly export the no learning reconstructor currently used in the trainer in deepinv.models
3. If we do, possibly add an option for "prox_l2" reconstruction with a specific z and gamma

Note that option 1. is a bit similar to eg EquivariantReconstructor

If we do changes 1-3 we could get rid of this class and have:

```python
model_inv_nopad = NoLearningModel(mode="pinv")
model_inv_pad = LiuJiaReconstructor(backbone=model_inv_nopad)
model_wiener_nopad = NoLearningModel(mode="prox_l2", z=0, gamma=1e4)
model_wiener_pad = LiuJiaReconstructor(backbone=model_wiener_nopad)
```

(It could be used with other reconstructors eg UNet artifact removal with circular convs too)

I'm also happy with leaving the pre/post-processing logic in the example and let users integrate it to their own setup as they want (if they're using this kind of tricks they're probably using custom logic on top of deepinv anyway, our deblurring setup is a bit barebones)

deepinv/tests/test_physics_functional.py
jscanvic: Something like faa0211533beeed334261de0d3d838e01995b2ec?

I'm happy to delete the test now we know it passes too if you prefer

examples/physics/demo_liu_jia_padding.py
Tmodrzyk: Thanks for all the details, we can keep the demo as is for now.

The issue seems to be more general than I thought, maybe we have to modify the design of `Reconstructors` in general. I don't really like the idea of multiplying pre-defined wrappers for each and every pre and post processing steps. I feel like this should just be an option somewhere.

deepinv/physics/functional/blur.py
jscanvic: It should be clearer now - basically

- `liu_jia_pad` computes padding blocks (horizontal, vertical, diagonal) to concatenate with the input
- It starts by computing the boundaries of each block using copies of the values of x at its boundaries
- Then it fills the inside of each block using harmonic inpainting (smooth extension with zero Laplacian)

Let me know if I need to explain it differently in the docs / code

jscanvic:
Thanks for the review Thibaut! I made most of the changes

Good catch for the bug with zero vertical / horizontal padding - I think we have a couple of options let me know what you think is best
1. I can rewrite the code so that there's 3 cases: 1) vertical+horizontal padding: I compute A, B and C and concatenate the three, 2) vertical padding: I compute A and concatenate it, 3) horizontal padding: I compute B and concatenate it
2. I add an error saying Liu Jia padding doesn't support padding on a single side

The issue with option 1 is it's going to make the function much more complex and it's not entirely clear in which case you might want to force your image to be circular horizontally but not vertically for instance. The typical use case is you pad all sides so I'm worried this extra complexity is not going to be very useful

PS not sure why coverage fails

deepinv/tests/test_physics_functional.py
Tmodrzyk: Yes, love this feature ! In this case we can remove the test right now :smile:

Tmodrzyk:
Hi @jscanvic it's fine to just output a `ValueError` or something in the case of padding on a single side.
We can always improve this later if it becomes a priority. 
After these last few changes, I'll give it a final look and we're good to go. :smile:

jscanvic:
Thanks Thibaut! It should be good now

Tmodrzyk:
@jscanvic I think there's a bug related to this PR, see the CPU test https://github.com/deepinv/deepinv/actions/runs/30442098786/job/90544463563?pr=934

jscanvic:
It should be good now!

Tmodrzyk (approved):
LGTM !

jscanvic:
Thanks Thibaut!

Merging - CI failures are unrelated

Issue #941, opened 2025-11-19
DPS not working when physics changes img size


tachella:
### Environment

all

### Describe the bug

[dinv.physics.DPS](https://deepinv.github.io/deepinv/api/stubs/deepinv.sampling.DPS.html#deepinv.sampling.DPS) gives an error when the physics operator has measurements that do not have the same shape as the signal. This is due to this line https://github.com/deepinv/deepinv/blob/b7a4ab4ac39225bb8ddda12b748aac3bdaf9b962/deepinv/sampling/diffusion.py#L626

which initializes `x` to have the same shape as measurements `y`. A simple fix would be to init with `physics.A_dagger(y)`

### Steps to reproduce

```
import deepinv as dinv
import torch

device = 'cuda'
x = torch.randn((1, 3, 128, 128), device=device)

physics = dinv.physics.BlurFFT(img_size=(3, 128, 128), filter=dinv.physics.blur.gaussian_blur(5.), device=device)

model = dinv.sampling.DPS(model=dinv.models.DRUNet(device=device), data_fidelity=dinv.optim.L2(), device=device)

y = physics(x)

x = model(y, physics)
```
gives an error

### Expected behavior

does not give an error. A test checking this behaviour should be added.

Issue #940, opened 2025-11-19
spectral_methods and Ptychography


tachella:
### Environment

all

### Describe the bug

The [Ptychography ](https://deepinv.github.io/deepinv/api/stubs/deepinv.physics.Ptychography.html#deepinv.physics.Ptychography)operator is not compatible with the [spectral_methods](https://deepinv.github.io/deepinv/api/stubs/deepinv.optim.phase_retrieval.spectral_methods.html#deepinv.optim.phase_retrieval.spectral_methods) reconstructor.

### Steps to reproduce

```
physics = dinv.physics.Ptychography(...)
physics.A_dagger(x) # runs spectral_methods

```
raises an error 

### Expected behavior

no error

Issue #938, opened 2025-11-19
MSE metric input args description


tachella:
### Documentation location

https://deepinv.github.io/deepinv/api/stubs/deepinv.loss.metric.MSE.html#deepinv.loss.metric.MSE

### Describe the issue or suggestion

The MSE metric inputs [in the docs](https://deepinv.github.io/deepinv/api/stubs/deepinv.loss.metric.MSE.html#deepinv.loss.metric.MSE) are described as 

```
deepinv.loss.metric.MSE(metric=None, complex_abs=False, train_loss=False, reduction=None, norm_inputs=None, center_crop=None)
```

The arguments `metric` and `train_loss` should be removed since they do not apply to this case.

PR #946, opened 2025-11-21
Small phase retrieval fixes, minor doc fixes, dirac combs, and faster RAM

tachella:
- Solves #938 #940 #941 
- Adds some efficiency fixes to `deepinv.models.RAM` which speed up approximately x2
- Adds `deepinv.utils.dirac_comb` (nice util for checking the response of spatially varying blur kernels)

### DIFF
--- a/deepinv/physics/forward.py
+++ b/deepinv/physics/forward.py
@@ -706,7 +706,7 @@ class LinearPhysics(Physics):
         :param torch.Tensor y: measurements tensor
         :param torch.Tensor z: signal tensor
         :param float gamma: hyperparameter of the proximal operator
-        :param str solver: solver to use for the proximal operator, see :func:`deepinv.utils.least_squares` for details
+        :param str solver: solver to use for the proximal operator, see :func:`deepinv.optim.utils.least_squares` for details
         :param int max_iter: maximum number of iterations for iterative solvers
         :param float tol: tolerance for iterative solvers
         :param bool verbose: whether to print information during the solver execution
--- a/deepinv/physics/wrappers.py
+++ b/deepinv/physics/wrappers.py
@@ -169,7 +169,7 @@ class LinearPhysicsMultiScaler(PhysicsMultiScaler, LinearPhysics):
         :param torch.Tensor y: measurements tensor
         :param torch.Tensor z: signal tensor
         :param float gamma: hyperparameter of the proximal operator
-        :param str solver: solver to use for the proximal operator, see :func:`deepinv.utils.least_squares` for details
+        :param str solver: solver to use for the proximal operator, see :func:`deepinv.optim.utils.least_squares` for details
         :param int max_iter: maximum number of iterations for iterative solvers
         :param float tol: tolerance for iterative solvers
         :param bool verbose: whether to print information during the solver execution


deepinv/loss/metric/distortion.py
```diff
@@ -14,7 +14,7 @@
 
 
 class MAE(Metric):
-    r"""
+    r"""deepinv.metric.MAE(complex_abs, reduction, norm_inputs, center_crop, \*\*kwargs)
```
mh-nguyen712: Why adding this? To give hint on how to use the function? If yes suggest adding :class:`...` and a new line after this.

mh-nguyen712:
```suggestion
    r""":class:`deepinv.metric.MAE(complex_abs, reduction, norm_inputs, center_crop, \*\*kwargs)`
    
```

deepinv/loss/metric/distortion.py
```diff
@@ -52,7 +52,7 @@ def metric(self, x_net, x, *args, **kwargs):
 
 
 class MSE(Metric):
-    r"""
+    r"""deepinv.metric.MSE(complex_abs, reduction, norm_inputs, center_crop, \*\*kwargs)
```
mh-nguyen712: Same here

mh-nguyen712: image
add a new line

deepinv/loss/metric/distortion.py
```diff
@@ -522,7 +522,7 @@ def metric(
 
 
 class SpectralAngleMapper(Metric):
-    r"""
+    r"""deepinv.metric.SpectralAngleMapper(train_loss, reduction, norm_inputs, center_crop, \*\*kwargs)
```
mh-nguyen712: same here

deepinv/models/ram.py
```diff
@@ -188,7 +187,9 @@ def realign_input(self, x, physics, y, sigma):
         gamma = gamma[(...,) + (None,) * (x.dim() - 1)]
         gamma = gamma * self.fact_realign
         gamma = gamma.clamp(min=1e-8)  # clamp to avoid negative or zero gamma
-        model_input = physics.prox_l2(x, y, gamma=gamma)
+        model_input = physics.base.prox_l2(
```
mh-nguyen712: I think it's safer to just call `physics.prox_l2` here. Since if it's a DecomposablePhysics, `physics.prox_l2` should use the fast implementation. If it's not (I think that it's not the case), I would change the behavior in physics class, not here.

deepinv/models/ram.py
```diff
@@ -437,18 +476,18 @@ def krylov_embeddings(y, p, factor, v=None, N=4, x_init=None):
     """
 
     if x_init is None:
-        x = p.A_adjoint(y)
+        x = y
     else:
-        x = x_init.clone()
+        x = x_init  # .clone()
```
mh-nguyen712:
```suggestion
        x = x_init
```

deepinv/models/ram.py
```diff
@@ -437,18 +476,18 @@ def krylov_embeddings(y, p, factor, v=None, N=4, x_init=None):
...
-    AtA = lambda u: p.A_adjoint(p.A(u)) * norm  # Define the linear operator
+    AtA = lambda u: p.A_adjoint_A(u) * norm  # Define the linear operator
 
     v = v if v is not None else torch.zeros_like(x)
 
-    out = x.clone()
+    out = x  # .clone()
     # Compute Krylov basis
-    x_k = x.clone()
+    x_k = x  # .clone()
```
mh-nguyen712: Just a remark that It could be unsafe if AtA has in-place operation.

deepinv/sampling/diffusion.py
```diff
@@ -568,7 +569,7 @@ class DPS(Reconstructor):
     def __init__(
         self,
         model,
-        data_fidelity,
+        data_fidelity=L2(),
```
mh-nguyen712: I think it's safer to use: 
```suggestion
        data_fidelity=None,
```
then 
``` if data_fidelity is None:
        data_fidelity = L2()
```
This can prevent bugs, since default arguments are evaluated only once, at definition time

deepinv/loss/metric/distortion.py
tachella: @mh-nguyen712 It is for overwriting the way the function signature is displayed in the docs (see #938 for an example). The issue is that these classes do not have custom `__init__`, so they inherit the init from the parent `deepinv.metric.Metric` which has an undesired input `metric` (this is already defined as the MAE in this case). 

In order to overwrite, we need to add the string at the very first line with a new description of input params as I did in the PR.

deepinv/loss/metric/distortion.py
tachella: (see my comment above)

deepinv/loss/metric/distortion.py
tachella: same, see comment above

deepinv/models/ram.py
tachella: agreed, I move this to the parent class `LinearPhysicsMultiScaler`, which now uses the base class `A_dagger` and `prox_l2` if it is run at `scale=0` (full res)

deepinv/models/ram.py
tachella: done

deepinv/models/ram.py
tachella: I kept it just in case

deepinv/sampling/diffusion.py
tachella: done

deepinv/loss/metric/distortion.py
mh-nguyen712: Thanks for the clarification. Can you just add a newline after the signature if possible, because when using type hint (eg in vscode) it gives the docstring right after the signature (see photo in my previous comment).
(Not really important, you can ignore it if you feel it's not necessary)

deepinv/models/ram.py
mh-nguyen712:
```suggestion
        model_input = physics.prox_l2(
```

deepinv/physics/wrappers.py
```diff
@@ -129,6 +129,54 @@ def A_adjoint(self, y, scale=None, **kwargs):
...
+        .. math::
+
+            \underset{x}{\arg\min} \; \frac{\gamma}{2}\|Ax-y\|^2 + \frac{1}{2}\|x-z\|^2
+
+        If the scale is set to 0, it uses the base physics proximal operator, which might have a more efficient implementation.
+
+        :param torch.Tensor y: measurements tensor
+        :param torch.Tensor z: signal tensor
+        :param float gamma: hyperparameter of the proximal operator
+        :param str solver: solver to use for the proximal operator, see :meth:`deepinv.utils.least_squares` for details
```
mh-nguyen712:
```suggestion
        :param str solver: solver to use for the proximal operator, see :func:`deepinv.utils.least_squares` for details
```

deepinv/physics/wrappers.py
```diff
@@ -129,6 +129,78 @@ def A_adjoint(self, y, scale=None, **kwargs):
...
+        .. math::
+
+            \underset{x}{\arg\min} \; \frac{\gamma}{2}\|Ax-y\|^2 + \frac{1}{2}\|x-z\|^2
+
+        If the scale is set to 0, it uses the base physics proximal operator, which might have a more efficient implementation.
+
+        :param torch.Tensor y: measurements tensor
+        :param torch.Tensor z: signal tensor
+        :param float gamma: hyperparameter of the proximal operator
+        :param str solver: solver to use for the proximal operator, see :func:`deepinv.utils.least_squares` for details
```
mh-nguyen712:
```suggestion
        :param str solver: solver to use for the proximal operator, see :func:`deepinv.optim.utils.least_squares` for details
```

deepinv/physics/forward.py
```diff
@@ -706,6 +706,11 @@ def prox_l2(
         :param torch.Tensor y: measurements tensor
         :param torch.Tensor z: signal tensor
         :param float gamma: hyperparameter of the proximal operator
+        :param str solver: solver to use for the proximal operator, see :func:`deepinv.utils.least_squares` for details
```
mh-nguyen712:
```suggestion
        :param str solver: solver to use for the proximal operator, see :func:`deepinv.optim.utils.least_squares` for details
```

mh-nguyen712 (changes requested):
LGTM!
I only have to comments related to function ref to make the docs pass.

tachella:
all tests seem to pass now, @mh-nguyen712 can we merge?

PR #948, opened 2025-11-23
Refactor generate_dataset + fix pillow error

### LINKED ISSUE #894: Handle datasets without transforms more gracefully in generate_dataset
### Environment

Latest

### Describe the bug

Certain datasets like `DIV2K` default to using no transform and returning PIL images instead of tensors. This behavior is fine and follows the conventions from torchvision but it interracts poorly with `deepinv.datasets.generate_dataset`. Indeed, it fails with a cryptic error when passed in a dataset like that. I suggest to handle things more gracefully by detecting PIL images and warn the user that it is not what the function expects and also possibly by converting the images to tensors automatically.

### Steps to reproduce

```python
import deepinv as dinv

dataset = dinv.datasets.DIV2K(root="DIV2K", mode="val", download=True)

physics = dinv.physics.Denoising(
    noise_model=dinv.physics.GaussianNoise(sigma=0.1)
)

_ = dinv.datasets.generate_dataset(
    train_dataset=dataset,
    physics=physics,
    device="cpu",
    save_dir="./measurements",
)
```

**Error**

```bash
  Line 9, in <module>
    _ = dinv.datasets.generate_dataset(
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "deepinv/datasets/datagenerator.py", line 294, in generate_dataset
    x0 = x0.to(device).unsqueeze(0)
         ^^^^^
AttributeError: 'Image' object has no attribute 'to'
```

### Expected behavior

The error message should at least be clearer.

_No response_

### Additional context

_No response_

nucli-vicky:
This PR:

- Refactors `generate_dataset` by removing a lot of duplicate code between train/val/test split creation, hopefully improving maintainability and clarity.
- Functionality should be equivalent aside from some discrepancies being fixed, which I would argue were unintended. Detailed below.
- Fixes #894 ; however in most cases I think the dataset will not contain images all with equal shape and users still get an error. I thought about including an argument which makes a center crop of specified shape, but whether something like this is appropriate is too data / task dependent so I believe it's best to not provide such a utility out-of-the-box to not force / default to such a choice; the error we raise now should be quite informative.

*What does change due to the refactor:*
- For train split, we would only do `pin_memory` when working on `cuda`, while val & test split would always put `pin_memory` to `True`, now for all splits we check if working on `cuda`.
- For val split, `TensorList` measurements were not supported; for test & train splits they were. Now `TensorList` is supported in all splits.

### DIFF
--- a/deepinv/tests/test_datasets.py
+++ b/deepinv/tests/test_datasets.py
@@ -268,6 +268,7 @@ def test_generate_dataset():
         assert sample_hdf[0].equal(
             sample
         ), "Ground-truth from HDF5 does not match original dataset, despite going through the same preprocessing."
+    hdf_ds.hd5.close()
     shutil.rmtree(tmp_data_dir)
     os.remove(hdf_path)
 


nucli-vicky:
So basically now

```python
import deepinv as dinv

dataset = dinv.datasets.DIV2K(root="DIV2K", mode="val", download=True)

physics = dinv.physics.Denoising(
    noise_model=dinv.physics.GaussianNoise(sigma=0.1)
)

_ = dinv.datasets.generate_dataset(
    train_dataset=dataset,
    physics=physics,
    device="cpu",
    save_dir="./measurements",
)
```
Will give an error like `RuntimeError: generate_dataset expects dataset to return elements of same shape, but received at least two different shapes: torch.Size([3, 1356, 2040]) and torch.Size([3, 1536, 2040]). Please add a crop/pad or other shape handling to your dataset.`

While 
```python
import deepinv as dinv
import torchvision as tv

dataset = dinv.datasets.DIV2K(
    root="DIV2K", mode="val", download=True, transform=tv.transforms.RandomCrop(64)
)

physics = dinv.physics.Denoising(noise_model=dinv.physics.GaussianNoise(sigma=0.1))

_ = dinv.datasets.generate_dataset(
    train_dataset=dataset,
    physics=physics,
    device="cpu",
    save_dir="./measurements",
)
```
will work.

Prior to the changes, both would fail with the error as documented in #894

jscanvic:
Thanks a lot Vicky!

I'll do the review shortly.

Can you also add a short regression test to make sure the new error message will pop off instead of the old cryptic one?

deepinv/tests/test_datasets.py
```diff
@@ -232,6 +233,36 @@ def test_hdfdataset(physgen):
     assert dataset.hd5 is None
 
 
+def test_generate_dataset():
```
Andrewwango, line 236: nice!!

deepinv/datasets/datagenerator.py
```diff
@@ -152,9 +153,65 @@ def close(self):
...
+                        arr = np.array(img, dtype=np.float32)
+                        if arr.ndim == 2:
+                            arr = arr[:, :, None]
+                        t = torch.from_numpy(arr.transpose(2, 0, 1) / 255.0)
+
+                        tensors.append(t)
+                        shapes.add(t.shape)
+                    if len(shapes) != 1:  # pragma: no cover
+                        raise RuntimeError(
+                            f"generate_dataset expects dataset to return elements of same shape, but received at least two different shapes: {list(shapes)[0]} and {list(shapes)[1]}. Please add a crop/pad or other shape handling to your dataset."
```
jscanvic: I'd log the whole set here

deepinv/datasets/datagenerator.py
```diff
@@ -152,9 +153,65 @@ def close(self):
...
+                    batch: list[Image.Image | list[Image.Image]],
+                ) -> torch.Tensor:
+                    tensors = []
+                    shapes = set()
+                    for sample in batch:
+                        if isinstance(sample, Image.Image):
+                            img = sample
+                        elif isinstance(sample, (list, tuple)):
+                            # only keeping the first element is same behavior as when dataset returns list of tensors!
+                            img = sample[0]
```
jscanvic: ... otherwise raise value error unexpected type

deepinv/datasets/datagenerator.py
```diff
@@ -152,9 +153,65 @@ def close(self):
...
+                        arr = np.array(img, dtype=np.float32)
+                        if arr.ndim == 2:
+                            arr = arr[:, :, None]
+                        t = torch.from_numpy(arr.transpose(2, 0, 1) / 255.0)
+
+                        tensors.append(t)
```
jscanvic: It's a bit redundant to track the tensors plus their shapes. Another way is to do something like:

```python
if tensors and t.shape != tensors[-1].shape:
  raise error
```
Happy to leave things as is as well if you prefer

deepinv/datasets/datagenerator.py
nucli-vicky: Fair, I didn't because otherwise the error message blows up for large batch sizes

deepinv/datasets/datagenerator.py
nucli-vicky: Sounds good, this also avoids the blowup issue I mentioned in your other comment

nucli-vicky:
Thanks for the suggestions! @jscanvic

deepinv/datasets/datagenerator.py
```diff
@@ -152,9 +153,66 @@ def close(self):
             self.hd5 = None
 
 
+def collate(dataset: Dataset):
+    example_output = dataset[0]
+    example_output = (
+        example_output[0]
+        if (isinstance(example_output, list) or isinstance(example_output, tuple))
```
Andrewwango:
```suggestion
        if isinstance(example_output, (list, tuple))
```

deepinv/datasets/datagenerator.py
```diff
@@ -152,9 +153,66 @@ def close(self):
             self.hd5 = None
 
 
+def collate(dataset: Dataset):
```
Andrewwango: Please can you add a docstring and return type to explain what this does? e.g. why it returns none, or a func...

deepinv/datasets/datagenerator.py
```diff
@@ -152,9 +153,66 @@ def close(self):
...
+                            # only keeping the first element is same behavior as when dataset returns list of tensors!
+                            img = sample[0]
+                        else:  # pragma: no cover
+                            raise ValueError(
+                                f"generate_dataset expects datasets to consistently return a (list of) Tensor, Array, or PIL images. Detected use of PIL in a sample, but received a new item of type {type(sample)}."
+                            )
+                        arr = np.array(img, dtype=np.float32)
+                        if arr.ndim == 2:
+                            arr = arr[:, :, None]
+                        t = torch.from_numpy(arr.transpose(2, 0, 1) / 255.0)
```
Andrewwango: Can we not replace this all with the `torchvision.transforms.ToTensor` like we do in datasets? or the `PILToTensor` - I think it checks the PIL dtypes too before doing the normalisation?

deepinv/datasets/datagenerator.py
```diff
@@ -261,6 +319,11 @@ def generate_dataset(
 
     hf_paths = []
 
+    def _is_cpu(dev: torch.device | str) -> bool:
```
Andrewwango: Surely `_is_cpu(dev)` could just be replaced equivalently with `torch.device(device).type == "cpu"` where `torch.device` noops when already a device

deepinv/tests/test_datasets.py
```diff
@@ -232,6 +233,36 @@ def test_hdfdataset(physgen):
...
+        )
+    # Test that no error is raised when we add crop
+    ds = Set14HR(tmp_data_dir, transform=CenterCrop(32))
+    _ = generate_dataset(
+        train_dataset=ds,
+        batch_size=1,
+        physics=physics,
+        device="cpu",
+        save_dir="./measurements",
+    )
```
Andrewwango, line 262: Maybe add a quick test where we load this generated dataset with `HDFDataset` and check that every tensor is exactly equal to every tensor in the original dataset

Andrewwango (changes requested):
Thanks @nucli-vicky , the refactor is nice! I added a few minor comments, but looks good!

deepinv/datasets/datagenerator.py
nucli-vicky: Good point, thanks for catching!

deepinv/datasets/datagenerator.py
nucli-vicky: Good point! I originally did it like this because `torchvision` has some plans for changing the transforms API, so seemed like a safer option long-term? But happy to use `ToTensor` too.

deepinv/tests/test_datasets.py
nucli-vicky: Thanks! 
For now I made this part very strict (exact equality). I'm always a bit skeptical when it comes to such strict checks on float data, but in principal this should always pass, so being stricter than `isclose` can be argued for.

nucli-vicky:
Thanks for the comments @Andrewwango , catched a few overcomplications :)

deepinv/datasets/datagenerator.py
jscanvic: ToTensor is being slowly deprecated as far as I understand https://github.com/pytorch/vision/pull/5640

For now it's only deprecated in v2 transforms while v1 transforms are not deprecated (yet?)

This code path is fairly critical so I'd prefer if we used `PILToTensor` with manual casting/scaling or `v2.ToDtype` (in that case we need to check there's no backwards compatibility issue with using v2 transforms)

nucli-vicky: Good to get an actual reference to these plans. To me it makes sense to not rely on torchvision at all here, since the lines to do it within `deepinv` is so small while having the upside of removing potential future issues due to torchvision's API changing (are we completely sure `PILToTensor` will not be changed / renamed / ... ?

jscanvic: (I tend to agree)

nucli-vicky: I took a quick look at `PILToTensor`'s [source code](https://github.com/pytorch/vision/blob/96e779759a883651e6ec2b394bf89de8beb5b709/torchvision/transforms/functional.py#L181). 
It always makes a copy of the data and does not change the `dtype`. In our use-case the copy is not required, so the 'manual way' of doing things has runtime advantages by avoiding an extra copy. 

Ofcourse this is a tiny optimization and its unlikely the extra copy adds a true bottleneck; the `physics` and disk transfers are surely dominating factors, but it might have some influence in specific scenarios (huge .tiff images, ...)

Andrewwango: Agreed both, I appreciate the healthy torchvision scepticism :)

Andrewwango (approved):
LGTM once tests pass

nucli-vicky:
Thanks @Andrewwango , the Windows test was failing as the os does not allow a file to be deleted if it is open in some process. This is now fixed

PR #950, opened 2025-11-23
Clean deepinv model module / complex denoiser docstring / typing

mh-nguyen712:
This PR cleans some old code of the `model` module and add some docstring / typing

### DIFF
--- a/deepinv/models/dsccp.py
+++ b/deepinv/models/dsccp.py
@@ -97,6 +97,12 @@ class DScCP(Denoiser):
             self.to(device)
 
     def forward(self, x: Tensor, sigma: float = 0.03) -> Tensor:
+        r"""
+        Run the denoiser on noisy image.
+
+        :param torch.Tensor x: noisy image.
+        :param float sigma: noise level.
+        """
         x_prev = x
         x_curr = x
         u = self.conv[0](x)
--- a/deepinv/models/wrapper.py
+++ b/deepinv/models/wrapper.py
@@ -219,6 +219,12 @@ class ComplexDenoiserWrapper(Denoiser):
 
     Note that the phase estimate :math:`\denoisername(\phi,\sigma)` is **clipped** back to :math:`(-\pi,\pi]`.
 
+    .. note::
+
+        This wrapper can only process complex inputs that are compatible with the underlying real-valued denoiser.
+        For example, if the wrapped ``denoiser`` supports only single-channel (grayscale) real images, then the
+        corresponding complex input must also be single-channel.
+
     |sep|
 
     :Examples:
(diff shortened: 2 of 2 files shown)


deepinv/models/dsccp.py
```diff
@@ -96,13 +96,7 @@ def __init__(
         if device is not None:
             self.to(device)
 
-    def forward(self, x, sigma=0.03):
-        r"""
```
Andrewwango, line 100: Why  removing the docstring

deepinv/models/wrapper.py
```diff
@@ -169,3 +169,115 @@ def forward(
...
+
+    The denoiser is applied on the magnitude and phase parts (same :math:`\sigma` broadcast across both halves).
+    The reconstructed complex output is
+
+    .. math::
+
+        \hat x = \denoisername(m, \sigma) \exp \big(i\, \denoisername(\phi, \sigma)\big).
+
+    Note that the phase estimate :math:`\denoisername(\phi,\sigma)` is **clipped** back to :math:`(-\pi,\pi]`.
+
```
Andrewwango, line 221: Maybe we need to explain that here we mean torch-complex which allows arbitrary number of dims (eg RGB complex ie 6 numerical channels) rather than channel-complex which we use for eg MRI (2 channel)

deepinv/models/wrapper.py
```diff
@@ -169,3 +169,115 @@ def forward(
         # Rescale to [0, 1]
         x0 = (x0 + 1) / 2
         return x0
+
+
+class ComplexDenoiserWrapper(Denoiser):
+    r"""
```
Andrewwango, line 175: I think we should add a small test for this

Andrewwango (changes requested):
Looks good, just some small comments then good to go once fixed

deepinv/models/dsccp.py
mh-nguyen712: nothing special here so it should reuse the docstring from base Denoiser class

deepinv/models/wrapper.py
mh-nguyen712: I'm not sure understand what you mean here

deepinv/models/wrapper.py
mh-nguyen712: Done

deepinv/models/wrapper.py
Andrewwango: I.e. can this denoiser be used with 2-channel "complex" MRI data?

Sorry if I am confused

deepinv/models/dsccp.py
Andrewwango: Just to check - does this inherit - can you check?

I thought you removed the docstring inheritance when you were fixing the examples?

Andrewwango:
Thanks @mh-nguyen712 , two replies to your comments

deepinv/models/dsccp.py
mh-nguyen712: Thanks Andrew I forgot this. Reverted the docstring. Sorry!

deepinv/models/wrapper.py
mh-nguyen712: I'm not familiar with MRI data, but what do you mean by "2 channel complex"? For me, the wrapper only work for complex data of shape (B, C, H, W) of dtype complex

mh-nguyen712: I added a note to clarify this

mh-nguyen712:
I think it should be good now @Andrewwango. Let's me know if the note is clear for you.

PR #957, opened 2025-11-25
Pytomography binding

CoConstan:
Pytomography physics binding with deepinv for emission CT. Add a demo for the use of the physics class.

### DIFF
--- a/deepinv/physics/emission_tomography.py
+++ b/deepinv/physics/emission_tomography.py
@@ -1,8 +1,10 @@
-import deepinv as dinv
 import torch
+import pytomography
+from pytomography.projectors.SPECT import SPECTSystemMatrix
+from pytomography.transforms.SPECT import SPECTAttenuationTransform, SPECTPSFTransform
+from deepinv.physics.forward import LinearPhysics
 
-
-class Emission_Tomography(dinv.physics.LinearPhysics):
+class SPECT(LinearPhysics):
     r"""
     Emission Tomography forward operator :math:`A` using the `pytomography` package (see : :footcite:t:`polson2025pytomography`).
 
@@ -26,51 +28,45 @@ class Emission_Tomography(dinv.physics.LinearPhysics):
     :param att_transform: Attenuation transform (see `pytomography` documentation for more informations).
     :param psf_transform: Point Spread Function (PSF) transform (see `pytomography` documentation for more informations).
     """
+    def __init__(self, object_meta, proj_meta, psf_meta = None, attenuation = None, CT_file = None, device="cpu", verbose: bool = True, **kwargs):
 
-class SPECT(dinv.physics.LinearPhysics):
-    def __init__(self, attenuation, object_meta, proj_meta, psf_meta, device="cpu", verbose: bool = True, **kwargs):
-
-        super().__init__(**kwargs)
-        
+        super().__init__()
+    
         pytomography.set_device(device)
         pytomography.set_verbose(verbose)
-
-        self.update_parameters(attenuation=attenuation, psf_meta=psf_meta)
-
-        self.system = SPECTSystemMatrix(
-            obj2obj_transforms=[self.att_transform, self.psf_transform],
+        
+        self.update_parameters(object_meta = object_meta, proj_meta = proj_meta, psf_meta=psf_meta, attenuation=attenuation, CT_file = CT_file)
+        self.system_matrix = SPECTSystemMatrix(
+            obj2obj_transforms=self.transforms,
             proj2proj_transforms=[],
             object_meta=object_meta,
             proj_meta=proj_meta,                
         )
 
-    def update_parameters(self, attenuation=None, psf_meta=None, **kwargs):
+    def update_parameters(self, object_meta, proj_meta, psf_meta=None, attenuation=None, CT_file=None, **kwargs):
         """Update physics parameters.
 
         :param torch.Tensor attenuation: attenuation map in image domain
         :param SPECTPSFMeta psf_meta: PSF meta object. In an ideal implementation this would be the PSF as a tensor but it would require modifying pytomography
         """
-        self.register_buffer("attenuation", attenuation)
+        transform_list = []
+        if attenuation is not None:
+            att_transform = SPECTAttenuationTransform(attenuation_map=attenuation.squeeze())
+            transform_list.append(att_transform)
+            self.register_buffer("attenuation", att_transform.attenuation_map)
 
-        self.att_transform = SPECTAttenuationTransform(attenuation_map=attenuation.squeeze())
-        self.psf_transform = SPECTPSFTransform(psf_meta)
+        if CT_file is not None:
+            att_transform = SPECTAttenuationTransform(filepath=CT_file)
+            att_transform.configure(object_meta, proj_meta)
+            transform_list.append(att_transform)
+            self.register_buffer("attenuation", att_transform.attenuation_map)
+
+        if psf_meta is not None:
+            psf_transform = SPECTPSFTransform(psf_meta)
+            transform_list.append(psf_transform)
         
+        self.transforms = transform_list
         super().update_parameters(**kwargs)
-
-    def A...
-        super().__init__(**kwargs)
-        from pytomography.projectors.SPECT import SPECTSystemMatrix
-
-        att_transform.configure(object_meta, proj_meta)
-
-        system_matrix = SPECTSystemMatrix(
-            obj2obj_transforms=[att_transform, psf_transform],
-            proj2proj_transforms=[],
-            object_meta=object_meta,
-            proj_meta=proj_meta,
-        )
-
-        self.system_matrix = system_matrix
 
     def A(self, x: torch.Tensor, **kwargs) -> torch.Tensor:
         """Forward operator.
--- a/deepinv/tests/test_external_libraries.py
+++ b/deepinv/tests/test_external_libraries.py
@@ -1,5 +1,6 @@
 import deepinv as dinv
 import torch
+import numpy as np
 import pytest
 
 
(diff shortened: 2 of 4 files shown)


deepinv/physics/EmissionTomography.py
```diff
@@ -0,0 +1,83 @@
+import deepinv.deepinv as dinv
+import torch
+
+
+class Emission_Tomography(dinv.physics.LinearPhysics):
```
tachella:
```suggestion
class SPECT(dinv.physics.LinearPhysics):
```

deepinv/physics/EmissionTomography.py
```diff
@@ -0,0 +1,83 @@
+import deepinv.deepinv as dinv
```
Andrewwango:
```suggestion
import deepinv as dinv
```

deepinv/physics/EmissionTomography.py
```diff
@@ -0,0 +1,83 @@
...
+    :math:`y` is of shape (B, C, N_proj, H', W') where B is the batch size, C is the number of channels (C=1 for emission tomography),
+    N_proj is the number of projections, H' is the height and W' is the width of the projections.
+
+    :param object_meta: Metadata of the object (see `pytomography` documentation for more informations).
+    :param proj_meta: Metadata of the projections (see `pytomography` documentation for more informations).
+    :param att_transform: Attenuation transform (see `pytomography` documentation for more informations).
+    :param psf_transform: Point Spread Function (PSF) transform (see `pytomography` documentation for more informations).
+    """
+
+    def __init__(self, object_meta, proj_meta, att_transform, psf_transform, **kwargs):
```
Andrewwango: I would use the more DeepInverse native interface where an attenuation tensor can be passed in and stored as a physics param:


```suggestion
class SPECT(dinv.physics.LinearPhysics):
    def __init__(self, attenuation, object_meta, proj_meta, psf_meta, device="cpu", verbose: bool = True, **kwargs):

        super().__init__(**kwargs)
        
        pytomography.set_device(device)
        pytomography.set_verbose(verbose)

        self.update_parameters(attenuation=attenuation, psf_meta=psf_meta)

        self.system = SPECTSystemMatrix(
            obj2obj_transforms=[self.att_transform, self.psf_transform],
            proj2proj_transforms=[],
            object_meta=object_meta,
            proj_meta=proj_meta,                
        )

    def update_parameters(self, attenuation=None, psf_meta=None, **kwargs):
        """Update physics parameters.

        :param torch.Tensor attenuation: attenuation map in image domain
        :param SPECTPSFMeta psf_meta: PSF meta object. In an ideal implementation this would be the PSF as a tensor but it would require modifying pytomography
        """
        self.register_buffer("attenuation", attenuation)

        self.att_transform = SPECTAttenuationTransform(attenuation_map=attenuation.squeeze())
        self.psf_transform = SPECTPSFTransform(psf_meta)
        
        super().update_parameters(**kwargs)

    def A...
```

examples/external-libraries/demo_Pytomography.py
```diff
@@ -0,0 +1,151 @@
...
+#       x^{k+1} = \frac{x^k}{A^T 1} A^T \left( \frac{y}{Ax^k + s} \right)
+#
+
+
+def mlem(x0, y, it, physics, scatter=0):
+    x = x0
+    At1 = physics.A_adjoint(torch.ones_like(y))
+    for i in range(it):
+        Ax = physics.A(x)
+        ratio = y / (Ax + scatter + 1e-6)
```
Andrewwango, line 105: It would be nice to have a `ScatterNoise()` or something where it simply does `x+scatter` but so that you could do:
```python
>>> physics=SPECT(..., noise_model=ScatterNoise())
>>> physics.update(scatter=scatter)
>>> ratio = y / physics(x)
etc.
```

deepinv/physics/EmissionTomography.py
Andrewwango: AFAIK, pytomography doesn't test their adjointness of their backward pass, nor the unit norm of their operator.

To integrate the SPECT operator with deepinv, it would be great to make it pass these tests. We can refer to Fessler's group's code whose operator is very similar to pytomography's, but they have an adjointness test: [python](https://github.com/guanhuaw/MIRTorch/blob/dev-docs/tests/spect_tests.py), [julia](https://github.com/JuliaImageRecon/SPECTrecon.jl/blob/main/test/adjoint-project.jl), [paper](https://ieeexplore.ieee.org/document/10032177).

@CoConstan what do you think?

CoConstan: Hi @Andrewwango,

I discussed this with @Tmodrzyk this morning. The method `physics.adjointness_test()` *should* work without any modification. I'll add a pytest class for it in the `test_external_libraries.py` file.

Regarding the normalization: EM-type algorithms in emission tomography kind of implicitly normalize by computing $(A^T \mathbf{1})$. It would be possible to add an option to enforce explicit normalization using `physics.compute_norm()` and dividing the forward result by this value. However, since this operates in 3D, computing the norm at every iteration would be very expensive, so I’m not sure it’s worth the additional cost.

examples/external-libraries/demo_Pytomography.py
CoConstan: Technically, the scatter is estimated only once before the reconstruction, and afterward there is no further update. Because of that, I’m not sure it really deserves its own noise model.

deepinv/physics/EmissionTomography.py
Andrewwango: Thanks @CoConstan - looking forward to seeing the test. From my experiments, the default pytomography SPECT backward is not exactly adjoint, but that could be my problem...

Regarding the normalisation: surely the operator only needs to be normalised once before being used for recon?

examples/external-libraries/demo_Pytomography.py
Andrewwango: sure, it makes sense. I'm just thinking how to provide a minimal interface, but it's true that maybe scatter should be treated differently!

deepinv/physics/EmissionTomography.py
CoConstan: I just tested `physics.adjointness_test()` and it indeed doesn’t work. After some investigation, I found that the issue comes from the method drawing (x) and (y) from a normal distribution $\mathcal{N}(0,1)$. Since our operator is defined only for (x, y > 0), the test fails. It works as expected once the sampling distribution is changed. I’ll push the corresponding test as soon as possible.

Andrewwango:
Note that pytomography SPECT operator forces no grad in the transforms, so we would have to do some overwriting or dirty work to allow gradients for using the operator in deep learning...

Attenuation: https://github.com/PyTomography/PyTomography/blob/main/src/pytomography/transforms/SPECT/attenuation.py#L72

PSF: https://github.com/PyTomography/PyTomography/blob/main/src/pytomography/transforms/SPECT/psf.py#L199

Rotate: https://github.com/PyTomography/PyTomography/blob/main/src/pytomography/transforms/shared/spatial.py#L21

deepinv/physics/EmissionTomography.py
Andrewwango: Thanks, looking forward to seeing it!

Would the `physics.compute_norm` therefore also need a better initialisation then ?

CoConstan: I just pushed the adjointness test in the `test_external_libraries.py` file. It now works both with and without attenuation/PSF.

Regarding normalization, I also tried using `physics.compute_norm` in a more realistic scenario (e.g., a 128×128×128 image with attenuation and a depth-dependent PSF). It took *at least 45 minutes* to compute, likely due to the PSF calculation. Honestly, I’m not sure if this will be practical for real-world use. I used the default parameters of `physics.compute_norm`, so that might be contributing to the long computation time.

CoConstan:
> Note that pytomography SPECT operator forces no grad in the transforms, so we would have to do some overwriting or dirty work to allow gradients for using the operator in deep learning...
> 
> Attenuation: https://github.com/PyTomography/PyTomography/blob/main/src/pytomography/transforms/SPECT/attenuation.py#L72
> 
> PSF: https://github.com/PyTomography/PyTomography/blob/main/src/pytomography/transforms/SPECT/psf.py#L199
> 
> Rotate: https://github.com/PyTomography/PyTomography/blob/main/src/pytomography/transforms/shared/spatial.py#L21

From my previous experiments, the main reason for disabling gradients is to avoid overloading VRAM during the computation graph. For the classical operator and the attenuation transform, this is mainly for safety and to ensure consistency within the library.

Unfortunately, for the PSF transform (which is both depth-dependent and projection-dependent) keeping the computation graph would make VRAM usage explode.

Fortunately, we can wrap `SPECTSystemMatrix.forward` and `SPECTSystemMatrix.backward` in a `torch.autograd.Function`. The backward pass would then simply call `SPECTSystemMatrix.backward` for the first and `SPECTSystemMatrix.forward` for the second, allowing gradient propagation without excessive memory usage.

deepinv/tests/test_external_libraries.py
```diff
@@ -177,3 +178,39 @@ def test_tomography_with_astra_logic(
...
+                          dr= (1, 1),
+                          angles= np.linspace(0, 360, 120, endpoint=False).tolist(), 
+                          radii=[380] * 120)
+        
+        # --- Optional transforms ---
+        attmap = torch.rand([32,32,32]) if use_att else None
+        psf_meta = SPECTPSFMeta(sigma_fit_params= np.random.rand(2)) if use_psf else None
+        physics = dinv.physics.SPECT(object_meta, proj_meta, psf_meta = psf_meta, attmap = attmap, device=device)
+
+        ## --- Test adjointness ---
```
Andrewwango, line 203: This is very cool thanks!

The code here is essentially following that of `physics.adjointness_test`. Maybe we could modify the code there and use that? We could modify it to have:

- a "normalize" parameter which divides by xTATy, which you do here;
- option to use `deepinv.utils.tensorlist.rand_like` rather than `randn_like`, which you do here, for data>0
- Then you could directly integrate this test into `test_physics`. Can we achieve 1e-6 adjointness error?

deepinv/tests/test_external_libraries.py
```diff
@@ -177,3 +178,39 @@ def test_tomography_with_astra_logic(
...
             assert physics.xray_transform.magnification_factor == pytest.approx(1.0)
+
+
+class TestSPECT:
+    @pytest.mark.parametrize("use_att", [False, True])
+    @pytest.mark.parametrize("use_psf", [False, True])
+    def test_adjoint(self, use_att, use_psf):
+        from pytomography.metadata.SPECT import SPECTObjectMeta, SPECTProjMeta, SPECTPSFMeta
+
+        device = "cuda" if torch.cuda.is_available() else "cpu"
```
Andrewwango, line 189: how fast is this test on CPU? We could optionally skip on CPU if it's too slow

deepinv/tests/test_external_libraries.py
```diff
@@ -177,3 +178,39 @@ def test_tomography_with_astra_logic(
             assert physics.xray_transform.magnification_factor == pytest.approx(1.25)
         else:
             assert physics.xray_transform.magnification_factor == pytest.approx(1.0)
+
+
+class TestSPECT:
+    @pytest.mark.parametrize("use_att", [False, True])
+    @pytest.mark.parametrize("use_psf", [False, True])
+    def test_adjoint(self, use_att, use_psf):
+        from pytomography.metadata.SPECT import SPECTObjectMeta, SPECTProjMeta, SPECTPSFMeta
```
Andrewwango, line 187: skip if pytomography not installed

deepinv/physics/emission_tomography.py
```diff
@@ -0,0 +1,109 @@
...
+    D is the depth, H is the height and W is the width of the 3D image.
+    :math:`y` is of shape (B, C, N_proj, H', W') where B is the batch size, C is the number of channels (C=1 for emission tomography),
+    N_proj is the number of projections, H' is the height and W' is the width of the projections.
+
+    :param object_meta: Metadata of the object (see `pytomography` documentation for more informations).
+    :param proj_meta: Metadata of the projections (see `pytomography` documentation for more informations).
+    :param att_transform: Attenuation transform (see `pytomography` documentation for more informations).
+    :param psf_transform: Point Spread Function (PSF) transform (see `pytomography` documentation for more informations).
+    """
+    def __init__(self, object_meta, proj_meta, psf_meta = None, attenuation = None, CT_file = None, device="cpu", verbose: bool = True, **kwargs):
```
Andrewwango, line 31: regarding normalization, I think there's still a use case for GPU users to ask the physics to normalize itself, mirroring behaviour of `deepinv.physics.Tomography`.

We could include this as a GPU test only, and test that the resultant physics has norm 1. For speed, you could precompute the norm for a given setup and hardcode it into the test

PR #869, opened 2025-10-19
Add 3D variants of CNN networks

nucli-vicky:
- [x] DIP (tests: also not done for 2D)
- [x] DRUNet (tests :white_check_mark: ; but might want to add parametrized tests for the different act/downsample/upsample modes)
- [x] UNet (tests :white_check_mark:)
- [x] GAN networks (tests :white_check_mark:)
- [x] DnCNN (tests :white_check_mark:)
- [x] ICNN (tests :white_check_mark:)
- [x] DScCP (tests :white_check_mark:)
- [x] PDNet (tests :white_check_mark:)
- [x] Add or parametrize tests so we have coverage for 3D variants

Will add the 3D variants for diffusion and nets using attention at a different time, as I want to do proper testing with some training runs for it.

This PR will not make `test_onesplit` support 3D. This will be fixed when sliding window inference is added together with a refactor of `PatchDataset`.

This PR should fix #763 and is related to #804 and #743 

**Notes:**
- Also fixed some docs, typing, and added flexibility (e.g. `ESRGANDiscriminator` now has a `filters` argument)
- Most notable: Tests now check all the possible downsampling, upsampling, and activation functions for `DRUNet`. After I wrote the test, it actually failed, and main failed too, when we specify maxpooling. This bug originated from the fact that the default `padding` in `downsample_maxpool` was 0 and is now fixed.

### Checks to be done before submitting your PR

- [x] `python3 -m pytest deepinv/tests` runs successfully.
- [x] `black .` runs successfully.
- [x] `make html` runs successfully (in the `docs/` directory).
- [x] Updated docstrings related to the changes (as applicable).
- [x] Added an entry to the [CHANGELOG.rst](../CHANGELOG.rst).

PR #958, opened 2025-11-25
Option to initialize 3D models with pretrained 2D weights

romainvo:
This PR completes what is already implemented in #869. For the moment, deepinv allows instantiating 3D versions of the denoisers in the library, but no pretrained weights are available. In practice, we can initialize the 3D weights from their 2D counterparts, providing a good starting point for further fine-tuning.

I initialize the weights as follows, by setting `pretrained='download_2d'`:
- for odd kernel size in the depth dimension, the central slice is initialized with the 2D weights and the rest with zeros.
- for even kernel size (in DRUNet up/down conv typically), I make the choice to replicate the 2D weights on both slices, and average by the depth dimension to keep the same norm. 

This new option targets the set of denoisers that have available weights and 3D versions, i.e. DRUNet, DnCNN and DScCP. Please tell me if I missed one.

Here is a toy example of what it produces with a DRUNet
init_3d_from_2d_illustration
here I have replicated the butterfly in a (3,32,256,256) tensor, and I show the central slice.

For the moment, I'm only testing the logic. I could add a performance test with a dummy volume. But it should be close to the training distribution for stable improvements (ideally a natural image replicated along the depth dimension).


nucli-vicky:
This is a super cool idea, thanks!

tachella:
super cool, would it be possible to convolve the kernels the 2D kernels with themselves to get a (separable) 3D kernel? In the same way we can get an N-dimensional Gaussian kernel by convolving a 1-dimensional one N times.

romainvo:
> super cool, would it be possible to convolve the kernels the 2D kernels with themselves to get a (separable) 3D kernel? In the same way we can get an N-dimensional Gaussian kernel by convolving a 1-dimensional one N times.

You mean parameterize 3 orthogonal 2d kernels with the same pretrained 2d weights ? If we do that I think the activations produced at each layer will significantly shift from the 2d model? Or am I misunderstanding something ?

nucli-vicky:
Linking to #804

romainvo:
> super cool, would it be possible to convolve the kernels the 2D kernels with themselves to get a (separable) 3D kernel? In the same way we can get an N-dimensional Gaussian kernel by convolving a 1-dimensional one N times.

@tachella I have added an option to initialize the 3D weights isotropically. Rather than considering the central slice in the depth dimension, you can also initialize the central slice along the 3 directions and average the kernel.

This is when only the central slice in the depth dimension is initialized:
init_3d_from_2d_illustration

And when the 2D weights are added to the central slice along the three directions, `pretrained_2d_isotropic=True`:
init_3d_from_2d_illustration_separable
results are slightly less good, but it is too be interpreted with care as the volume is only a concatenation of replicated slices.

tachella:
hey @romainvo, can you merge main? this will probably fix your docs test issue

docs/source/changelog.rst
```diff
@@ -10,9 +10,11 @@ New Features
 ^^^^^^^^^^^^
 - Add :class:`deepinv.physics.LaplaceNoise` model (:gh:`921` by `Brayan Monroy`_)
 - New way to create optimization models. Standard optimization algorithms (and their unfolded versions) can be created using their class name directly instead of using the `optim_builder` (or `unfolded_builder`) function. (:gh:`592` by `Samuel Hurault`_)
+- Add support for Poisson2Sparse (:gh:`677` by `Jérémy Scanvic`_)
```
Andrewwango: remove the poisson2sparse contrib in "changed"

deepinv/tests/test_models.py
```diff
@@ -1480,3 +1480,46 @@ def test_complex_wrapper(mode, device):
...
+
+    model = model.eval().to(device)
+
+    image_size = (n_channels, 32, 32, 32)
+
+    x = torch.rand(image_size, device=device)[None]
+
+    y = model(x, 0.01)
+    # Check the output tensor shape
+    assert (
```
Andrewwango, line 1575: I guess this test does nothing more than the previous denioser shape tests, except that the pretrained weights don't error. Maybe we could test that images reconstructed with pretrained 2d weights are x number of dB better than no init?

deepinv/models/utils.py
```diff
@@ -444,3 +444,65 @@ def forward(self, x: Tensor):
...
+
+    :param nn.Module model_3d: 3D model to be initialized.
+    :param dict[str, torch.Tensor] ckpt_2d: state_dict of the 2D counterpart of model_3d.
+    :param bool isotropic: If True, for odd kernel sizes, the weights are copied to the center slice
+        of all three dimensions and averaged. For even kernel sizes, the 2D weights are
+        divided by the kernel size and copied to all slices in all three dimensions. Otherwise,
+        only the depth dimension is considered.
+    """
+    for name, module in model_3d.named_modules():
+        if isinstance(module, nn.Conv3d) or isinstance(module, nn.ConvTranspose3d):
```
Andrewwango:
```suggestion
        if isinstance(module, (nn.Conv3d, nn.ConvTranspose3d)):
```

deepinv/models/dncnn.py
```diff
@@ -27,6 +27,7 @@ class DnCNN(Denoiser):
         It is possible to download weights trained via the regularization method in :footcite:t:`pesquet2021learning`, using ``pretrained='download_lipschitz'``.
         Finally, ``pretrained`` can also be set as a path to the user's own pretrained weights.
         See :ref:`pretrained-weights <pretrained-weights>` for more details.
+    :param bool pretrained_2d_isotropic: when loading 2D pretrained weights into a 3D network, whether to initialize the 3D kernels isotropically. By default the weights are loaded axially, i.e., by initializing the central slice of the 3D kernels with the 2D weights.
```
Andrewwango, line 31: need to modify the `pretrained` argument description too to allow `download_2d`

deepinv/models/dncnn.py
```diff
@@ -67,13 +69,15 @@ def __init__(
 
         if pretrained is not None:
             if pretrained.startswith("download"):
-                if dim == 3:  # pragma: no cover
-                    raise RuntimeError(
-                        "No pretrained weights are available for download for 3D DnCNN."
+                if dim == 3 and (
+                    pretrained == "download" or pretrained == "download_lipschitz"
```
Andrewwango:
```suggestion
                    pretrained in ("download", "download_lipschitz")
```

deepinv/tests/test_models.py
romainvo: Indeed you're right. I also thought of testing minimum perf on the 3D denoiser initialized with 2D. On which images/volume would you test it ? 

If it seems sufficient I can do it on a stack of 2D natural images, like the butterfly in my example, but I don't know how relevant it is

Andrewwango: Hmm, the lightweight brainweb in the deepinv HF?

romainvo:
Thanks for the feedback @Andrewwango, I have added the performance check on the brainweb data, the thresholds are set as tight as possible. I am not checking the perf when `n_channels=3`, as I don't think we have a relevant "RGB" volume to test on (if that exists).

nucli-vicky:
@romainvo Technically, it would be cool to apply this to short timeframe video, which in an artificial sense can be thought of as a 3D RGB "volume" ;).

Not sure how performance would be if treating the time-axis with your good old 3x3x3 conv, but it is true that video pretraining --> volume finetuning has seen some initial results ["Video Pretraining Advances 3D Deep Learning
on Chest CT Tasks"](https://proceedings.mlr.press/v227/ke24a/ke24a.pdf)

Also might be worth checking out in the spirit of this PR is ["Reinventing 2D Convolutions for 3D Images"](https://arxiv.org/pdf/1911.10477)
image

romainvo:
> @romainvo Technically, it would be cool to apply this to short timeframe video, which in an artificial sense can be thought of as a 3D RGB "volume" ;).
> 
> Not sure how performance would be if treating the time-axis with your good old 3x3x3 conv, but it is true that video pretraining --> volume finetuning has seen some initial results ["Video Pretraining Advances 3D Deep Learning on Chest CT Tasks"](https://proceedings.mlr.press/v227/ke24a/ke24a.pdf)
> 
> Also might be worth checking out in the spirit of this PR is ["Reinventing 2D Convolutions for 3D Images"](https://arxiv.org/pdf/1911.10477) image

Oh indeed, would be nice ! but I don't think there is that kind of data in the deepinv hugging face ? 

The `"isotropic"` init is kind of similar to the design on the right of your figure, although it's just for initialization, it's not enforced during computation.

Andrewwango:
Happy to add adaptation of 2D networks for dynamic data to a new issue!

Note we already have 2 trivial wrappers that allow 2D networks to be suboptimally used for dynamic data. https://deepinv.github.io/deepinv/user_guide/reconstruction/denoisers.html#dynamic-networks

But for now happy to merge once conflicts addressed (you can merge from main for latest CI fixes)

Andrewwango (approved):

PR #959, opened 2025-11-25
Fix formatting in citation for DeepInverse article

bruAristimunha:
Small typo that I noticed when I was citing the library.

no change log necessary


bruAristimunha:
@mathurinm hard PR for you revision ;p

PR #960, opened 2025-11-26
increase tomography robustness to varying channels

tachella:
The Tomography operator silently fails when evaluated with a tensor of more than 1 channel. This PR fixes it.

### DIFF
--- a/examples/optimization/demo_patch_priors_CT.py
+++ b/examples/optimization/demo_patch_priors_CT.py
@@ -15,7 +15,7 @@ For the reconstruction, we minimize the variational problem
     \underset{x}{\arg\min} \quad \datafid{x}{y} + \lambda g(x).
     \end{equation*}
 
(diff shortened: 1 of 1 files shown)


tachella:
> Thanks for fixing this issue! Looks good to me.
> 
> Small question, maybe it would be nice to add the same test for the Astra wrapper ? It does support multi-channel input, but deals with it a runtime

good idea, I've added it :)

deepinv/tests/test_external_libraries.py
romainvo, line 40:
```suggestion
        :param bool normalize: Initializes the operator with ``normalize=normalize``.
        :param bool fbp: Whether or not to approximate the pseudo-inverse with filtered back-projection.
        :param int channels: Number of input channels. The tomography operator is applied per channel.
```

mh-nguyen712 (approved):

PR #962, opened 2025-11-26
Prevent unnecessary downloads of the FMD dataset

### LINKED ISSUE #499: FMD dataset redownloads
The [FMD](https://deepinv.github.io/deepinv/api/stubs/deepinv.datasets.FMD.html) dataset redownloads itself even if the data has been already downloaded in the indicated `root `folder. I would expect to only download the data when needed, even if `download=True`.

jscanvic:
Fixes #499 

I add a test to check if the directory doesn't exist or is empty. This is a partial fix because the heuristic is very simple but I don't want to spend the time to implement something more robust for now and it's already been more than 6 months this issue is lying around so it should be better than nothing hopefully!

### DIFF
--- a/.github/scripts/diff_sphinx_gallery.py
+++ b/.github/scripts/diff_sphinx_gallery.py
@@ -1,0 +1,246 @@
+import re
+import glob
+from pathlib import Path
+import subprocess
+import sys
+import ast
+
+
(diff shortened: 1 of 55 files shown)


tachella (approved):
LGTM, thanks Jeremy!

Issue #927, opened 2025-11-12
Plot function cannot fix absolute color scale across multiple images


Tmodrzyk:
### Describe the feature

The current ``deepinv.utils.plot`` function only supports two normalization modes before plotting:
- per-image min–max rescaling
- clipping between [0, 1]

However, in many real world applications (for instance tomography), images must be plotted using a shared absolute scale. The plot utility should support a shared color scale.

The current implementation only allows for individual rescales, which makes it unusable for most real-world evaluation or publication purposes. Here's a simple example showing how different the plots can look:
```python
import deepinv as dinv
import matplotlib.pyplot as plt

x = dinv.utils.load_example("butterfly.png", img_size=(128, 128), grayscale=True)
physics1 = dinv.physics.PoissonNoise(gain=1 / 5)
physics2 = dinv.physics.PoissonNoise(gain=1 / 100)

y1 = physics1(x)
y2 = physics2(x)

print(y1.min(), y1.max())
print(y2.min(), y2.max())
dinv.utils.plot(
    [x, y1, y2],
    titles=["Ground-truth image", "Noisy measurement 1", "Noisy measurement 2"],
    figsize=(10, 4),
)
fig, axes = plt.subplots(1, 3, figsize=(10, 4))
vmin = min(x.min().item(), y1.min().item(), y2.min().item())
vmax = max(x.max().item(), y1.max().item(), y2.max().item())

im0 = axes[0].imshow(x.squeeze(), cmap="gray", vmin=vmin, vmax=vmax)
axes[0].set_title("Ground-truth image")
axes[0].axis("off")

im1 = axes[1].imshow(y1.squeeze(), cmap="gray", vmin=vmin, vmax=vmax)
axes[1].set_title("Noisy measurement 1")
axes[1].axis("off")

im2 = axes[2].imshow(y2.squeeze(), cmap="gray", vmin=vmin, vmax=vmax)
axes[2].set_title("Noisy measurement 2")
axes[2].axis("off")

plt.tight_layout()

plt.show()
```
Output:

<img width="1077" height="860" alt="Image" src="https://github.com/user-attachments/assets/ae9f480c-693b-4fc6-ae72-269ca6efdf25" />

### Motivation

Many of the deepinv users work on quantitative applications where units matter, and they can't use our plotting function for their plots.

nucli-vicky:
Good point @Tmodrzyk , I noticed this too and plan to take this up when I add the 3D plotting utilities into the trainer for #804 !

[...]

PR #967, opened 2025-12-02
Add vmin vmax args to ``plot``

Tmodrzyk:
This PR implements the "absolute scale" feature of #927. When selecting the ``clip`` rescale mode, you can pass ``vmin`` and ``vmax`` in ``**kwargs_imshow`` and it will be used to clip the signals and plot the images.
It can be used as follows:
```python
import deepinv as dinv

x1 = dinv.utils.load_example("butterfly.png", img_size=(128, 128), grayscale=True)
physics1 = dinv.physics.PoissonNoise(gain=1 / 5)
physics2 = dinv.physics.PoissonNoise(gain=1 / 100)

y1 = physics1(x1)
y2 = physics2(x1)
vmin = min(x1.min().item(), y1.min().item(), y2.min().item())
vmax = max(x1.max().item(), y1.max().item(), y2.max().item())
dinv.utils.plot(
    [x1, y1, y2],
    titles=["Ground-truth image", "Noisy measurement 1", "Noisy measurement 2"],
    figsize=(10, 6),
    rescale_mode="clip",
    vmin=vmin,
    vmax=vmax,
)
```

### DIFF
--- a/deepinv/utils/plotting.py
+++ b/deepinv/utils/plotting.py
@@ -218,17 +218,7 @@ def preprocess_img(
     # Cast image values to float32 numbers
     # NOTE: Why is it needed?
     im = im.type(torch.float32)
-    if rescale_mode != "clip":
-        if vmin is not None or vmax is not None:
-            warn(
-                "The vmin and vmax arguments are used only when using 'clip' rescaling.",
-                UserWarning,
-                stacklevel=2,
-            )
-    if vmin is not None and vmax is not None and vmin >= vmax:
-        raise ValueError(
-            f"vmin should be strictly less than vmax, got vmin={vmin} and vmax={vmax}."
-        )
+
     # Normalize signal between 0 and 1
     im = normalize_signal(im, mode=rescale_mode, vmin=vmin, vmax=vmax)
 
--- a/deepinv/utils/signals.py
+++ b/deepinv/utils/signals.py
@@ -1,6 +1,7 @@
 """Signal processing utilities"""
 
 from __future__ import annotations
+from warnings import warn
 
 import torch
 
@@ -24,6 +25,17 @@ def normalize_signal(
     :return: the normalized batch of signals.
 
     """
+    if mode != "clip":
+        if vmin is not None or vmax is not None:
+            warn(
+                "The vmin and vmax arguments are used only when using 'clip' rescaling.",
+                UserWarning,
+                stacklevel=2,
+            )
+    if vmin is not None and vmax is not None and vmin >= vmax:
+        raise ValueError(
+            f"vmin should be strictly less than vmax, got vmin={vmin} and vmax={vmax}."
+        )
     if mode == "min_max":
         # Compute the minimum and maximum intensity of the batched signals
         non_batched_dims = list(range(1, inp.ndim))


deepinv/utils/plotting.py
```diff
@@ -189,7 +189,7 @@ def prepare_images(x=None, y=None, x_net=None, x_nl=None, rescale_mode="min_max"
 
 
 @torch.no_grad()
-def preprocess_img(im, rescale_mode="min_max"):
+def preprocess_img(im, rescale_mode="min_max", vmin=None, vmax=None):
```
jscanvic:
```suggestion
def preprocess_img(im, rescale_mode="min_max", *, vmin: float | None = None, vmax: float | None = None):
```

deepinv/utils/signals.py
```diff
@@ -44,7 +55,11 @@ def normalize_signal(inp, *, mode):
         inp[indices] = inp[indices].clamp(min=0.0, max=1.0)
     elif mode == "clip":
         # Clamp every batched signal between zero and one
-        inp = inp.clamp(min=0.0, max=1.0)
+        if vmin is None:
+            vmin = 0.0
+        if vmax is None:
+            vmax = 1.0
+        inp = inp.clamp(min=vmin, max=vmax)
```
jscanvic: You also need to rescale `inp` so the [vmin, vmax] range maps to [0, 1] don't you?

jscanvic (changes requested):
Thanks Thibaut! Some comments

* Can you check if matplotlib supports the use of vmin/vmax in imshow for RGB images? (I think not, we might need to preprocess the image ourselves here)
* It's a bit awkward that `normalize_signal` returns a signal that's not normalized between zero and one despite being the purpose/description of the function
* Update the docstrings
* Can you add a warning for the use of incompatible parameter values? (e.g., vmin and vmax being set while rescale_mode is not set to "clip")
* Update the changelog

Tmodrzyk:
Hi @jscanvic, I took into account your comments. However for:
> _it's a bit awkward that normalize_signal returns a signal that's not normalized between zero and one despite being the purpose/description of the function_

I'm not sure if we should rename the function ``normalize_signal`` to something else, or remove the clipping option and put it in ``dinv.utils.preprocess_img``. The function ``normalize_signal`` (with ``clip`` rescale mode) is already used elsewhere in the repo, but I can fix it manually.

deepinv/utils/plotting.py
```diff
@@ -214,9 +218,19 @@ def preprocess_img(im, rescale_mode="min_max", *, vmin: float | None = None, vma
...
+                "The vmin and vmax arguments are ignored when using 'min_max' rescaling.",
+                UserWarning,
+                stacklevel=2,
+            )
+        if vmin is not None and vmax is not None and vmin >= vmax:
+            raise ValueError(
+                f"vmin should be strictly less than vmax, got vmin={vmin} and vmax={vmax}."
+            )
+        # Normalize values between zero and one
+        im = normalize_signal(im, mode=rescale_mode, vmin=vmin, vmax=vmax)
```
jscanvic: It should be outside the if right?

Tmodrzyk: Oops, thank you !

jscanvic:
Ty for making the change!

RGB images don't seem to work afaict. If I'm not mistaken matplotlib internally overwrites vmin and vmax to 0 and 1 for them. A possible solution is to do the rescaling+clipping ourselves before passing the image to imshow (without passing vmin and vmax to imshow)

**Reproducing**

```python
import deepinv as dinv

x1 = dinv.utils.load_example("butterfly.png", img_size=(128, 128), grayscale=False)
x2 = x1 + 100
x3 = x1 + 100.5
x4 = x1 + 200

dinv.utils.plot(
    [x1, x2, x3, x4],
    rescale_mode="clip",
    vmin=100,
    vmax=101,
    cbar=True,
)
```

**Expected behavior**

The 1st image should be black (clipped out), the 2nd image should show the butterfly normally, the 3rd image should show a out-of-tone butterfly with clipped pixels, the 4th image should be white (clipped out)

**Observed behavior**

Every image is white

Tmodrzyk:
Hi @jscanvic thanks for the test, the one I conducted was not very telling for RGB. Now it should have the expected behavior. 
Basically now in ``preprocess_img`` after calling ``normlize_signal`` to clip the images, I rescale to [0, 1]. ``vmin`` and ``vmax`` are reset to 0 and 1 before being passed to ``imshow``. 
I hope this won't cause some unexpected behaviors down the line...

jscanvic:
Ty! Will have a closer look shortly

deepinv/utils/plotting.py
```diff
@@ -214,10 +218,23 @@ def preprocess_img(im, rescale_mode="min_max"):
     # Cast image values to float32 numbers
     # NOTE: Why is it needed?
     im = im.type(torch.float32)
+    if rescale_mode == "min_max":
```
jscanvic: * I'd move that to `normalize_signal`
* I think it makes more sense to test if `rescale_mode != "clip"` instead of testing that it's `min_max`, it's more that vmin/vmax is only compatible with clip than it's not compatible with min_max

deepinv/utils/plotting.py
```diff
@@ -214,10 +218,23 @@ def preprocess_img(im, rescale_mode="min_max"):
...
+        raise ValueError(
+            f"vmin should be strictly less than vmax, got vmin={vmin} and vmax={vmax}."
+        )
+    # Normalize signal between 0 and 1
+    im = normalize_signal(im, mode=rescale_mode, vmin=vmin, vmax=vmax)
 
-    # Normalize values between zero and one
-    im = normalize_signal(im, mode=rescale_mode)
-
+    if rescale_mode == "clip":
```
jscanvic: I'd move that to `normalize_signal` too

That way normalize_signal always gives out images between 0 and 1

deepinv/utils/plotting.py
```diff
@@ -205,6 +207,8 @@ def preprocess_img(im, rescale_mode="min_max"):
 
     :param torch.Tensor im: the batch of images to preprocess, it is expected to be of shape (B, C, *).
     :param str rescale_mode: the normalization mode, either 'min_max' or 'clip'.
+    ;param float vmin: minimum value for clipping when using 'clip' rescaling.
```
jscanvic:
```suggestion
    :param float, None vmin: minimum value for clipping when using 'clip' rescaling.
```

deepinv/utils/plotting.py
```diff
@@ -205,6 +207,8 @@ def preprocess_img(im, rescale_mode="min_max"):
...
+    ;param float vmax: maximum value for clipping when using 'clip' rescaling.
```
jscanvic:
```suggestion
    :param float, None vmax: maximum value for clipping when using 'clip' rescaling.
```

deepinv/utils/plotting.py
```diff
@@ -214,10 +218,23 @@ def preprocess_img(im, rescale_mode="min_max"):
     # Cast image values to float32 numbers
     # NOTE: Why is it needed?
     im = im.type(torch.float32)
+    if rescale_mode == "min_max":
+        if vmin is not None or vmax is not None:
+            warn(
+                "The vmin and vmax arguments are ignored when using 'min_max' rescaling.",
```
jscanvic: Update the message too if you invert the condition

deepinv/utils/plotting.py
```diff
@@ -353,14 +370,20 @@ def plot(
...
-            col_imgs.append(
+            row_imgs.append(
                 im[i, ...].detach().permute(1, 2, 0).squeeze().cpu().numpy()
             )
-        imgs.append(col_imgs)
+        imgs.append(row_imgs)
+
+    imshow_kwargs = dict(imshow_kwargs)
+    imshow_kwargs.setdefault("vmin", 0)
+    imshow_kwargs.setdefault("vmax", 1)
```
jscanvic: No need for that if you put vmin and vmax as regular parameters to the function instead of having them in the imshow_kwargs in the first place

Tmodrzyk:
I implemented your recommendations, I'll let you know once the CI passes. :)

Tmodrzyk:
I think this is good to go @jscanvic :smile:

deepinv/utils/signals.py
```diff
@@ -25,6 +33,11 @@ def normalize_signal(inp, *, mode):
         # Clone the signal to avoid input mutations
         inp = inp.clone()
 
+        if vmin is not None:
+            minimum_intensity = torch.full_like(minimum_intensity, vmin)
+        if vmax is not None:
+            maximum_intensity = torch.full_like(maximum_intensity, vmax)
+
         # The indices corresponding to the non-constant batched signals
```
jscanvic:
```suggestion
        # The indices corresponding to the non-constant batched signals
```

I think we should leave the minmax rescaling mode as is

deepinv/utils/plotting.py
```diff
@@ -214,9 +218,19 @@ def preprocess_img(im, rescale_mode="min_max"):
...
+        if vmin is not None or vmax is not None:
+            warn(
+                "The vmin and vmax arguments are used only when using 'clip' rescaling.",
+                UserWarning,
+                stacklevel=2,
+            )
+    if vmin is not None and vmax is not None and vmin >= vmax:
+        raise ValueError(
+            f"vmin should be strictly less than vmax, got vmin={vmin} and vmax={vmax}."
+        )
```
jscanvic: Please put this whole block at the beginning of `normalize_signal`

Tmodrzyk: Done in [afc3635](https://github.com/deepinv/deepinv/pull/967/commits/afc3635465bb408003306b69f0ecdcf34b383ef2) !

jscanvic (approved):
LGTM thanks Thibaut!

PR #974, opened 2025-12-04
Add a SNR metric

### LINKED ISSUE #970: Signal-to-noise Ratio
### Environment (optional)

_No response_

### Describe the feature

The library features a PSNR metric but not a SNR metric. It would be useful to have it for unnormalized signals (e.g., k-spaces in MRI)

### Motivation

Comparing measurements

### Alternatives considered

_No response_

### Additional context

_No response_

jscanvic:
Fixes #970

### DIFF
--- a/deepinv/loss/metric/functional.py
+++ b/deepinv/loss/metric/functional.py
@@ -46,8 +46,8 @@ def signal_noise_ratio(x_hat: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
     noise = x_hat - x
     # For a more efficient implementation, we compute the SNR from the signal
     # and noise L2 norms instead of their powers.
-    signal_norm = torch.linalg.vector_norm(x.flatten(1, -1), p=2, dim=1)
-    noise_norm = torch.linalg.vector_norm(noise.flatten(1, -1), p=2, dim=1)
+    signal_norm = torch.linalg.vector_norm(x.flatten(1, -1), ord=2, dim=1)
+    noise_norm = torch.linalg.vector_norm(noise.flatten(1, -1), ord=2, dim=1)
     sqrt_snr = signal_norm / noise_norm
     # The factor 20 instead of 10 comes from the square root.
     return 20 * torch.log10(sqrt_snr)


deepinv/loss/metric/distortion.py
```diff
@@ -295,6 +300,39 @@ def metric(self, x_net, x, *args, **kwargs):
...
 
+class SNR(Metric):
+    r"""
+    Compute the signal-to-noise ratio (SNR)
+
+    For a reference signal :math:`x` corrupted by noise :math:`\varepsilon`
+
+    .. math::
+
+        y = x + \varepsilon,
```
mh-nguyen712: I would use an estimate and a reference here instead of denoising

deepinv/loss/metric/distortion.py
```diff
@@ -295,6 +300,39 @@ def metric(self, x_net, x, *args, **kwargs):
...
+
+    the signal-to-noise ratio expressed in dB is defiend as
+
+    .. math::
+
+        \mathrm{SNR} = 10 \log_{10} \left( \frac{\sum_{i=1}^n |x_i|^2}{\sum_{i=1}^n |x_i - y_i|^2} \right).
```
mh-nguyen712: I would use the norm here
```suggestion
        \mathrm{SNR} = 10 \log_{10} \left( \frac{\|x \|^2}{\|x - y \|^2} \right).
```

deepinv/loss/metric/distortion.py
```diff
@@ -295,6 +300,39 @@ def metric(self, x_net, x, *args, **kwargs):
...
+
+    .. note::
+
+        The input is assumed to be batched and the SNR is computed for each element independently.
+
+    :param torch.Tensor preds: The noisy signal.
```
mh-nguyen712: wrong param name?

deepinv/loss/metric/distortion.py
```diff
@@ -295,6 +300,39 @@ def metric(self, x_net, x, *args, **kwargs):
         return cal_psnr(x_net, x, max_pixel=max_pixel, min_pixel=min_pixel)
 
 
+class SNR(Metric):
+    r"""
```
mh-nguyen712, line 304: add the function signature here (as recently done by Julian)

deepinv/loss/metric/functional.py
```diff
@@ -25,6 +25,37 @@ def cal_psnr(
...
+
+    :param torch.Tensor preds: The noisy signal.
+    :param torch.Tensor target: The reference signal.
+    :return: (torch.Tensor) The SNR value in decibels (dB).
+    """
+    noise = preds - target
+    signal_power = target.abs().pow(2).flatten(1, -1).mean(dim=1)
+    noise_power = noise.abs().pow(2).flatten(1, -1).mean(dim=1)
+    snr = signal_power / noise_power
+    return 10 * torch.log10(snr)
```
mh-nguyen712:
```suggestion
    signal_power = torch.linalg.vector_norm(target.flatten(1, -1), dim=1)
    noise_power = torch.linalg.vector_norm(noise.flatten(1, -1), dim=1)
    snr = noise_power / signal_power    # avoid dividing by zero
    return -20 * torch.log10(snr)
```

deepinv/tests/test_metric.py
```diff
@@ -21,7 +23,7 @@
     "SAM",
     "HaarPSI",
 ]
-FUNCTIONALS = ["cal_mse", "cal_mae", "cal_psnr"]
+FUNCTIONALS = ["cal_mse", "cal_mae", "cal_psnr", "peak_signal_noise_ratio"]
```
mh-nguyen712:
```suggestion
FUNCTIONALS = ["cal_mse", "cal_mae", "cal_psnr", "signal_noise_ratio"]
```

deepinv/tests/test_metric.py
```diff
@@ -186,6 +190,24 @@ def test_functional(functional_name, imsize_2_channel, device, rng):
         torch_psnr = PeakSignalNoiseRatio(data_range=1.0).to(device)
         assert torch.allclose(metric.cal_psnr(x_net, x), torch_psnr(x_net, x))
 
+    elif functional_name == "peak_signal_noise_ratio":
```
mh-nguyen712:
```suggestion
    elif functional_name == "signal_noise_ratio":
```

docs/source/user_guide/training/metric.rst
```diff
@@ -83,6 +83,9 @@ Full reference metrics are used to measure the difference between the original `
    * - :class:`deepinv.loss.metric.PSNR`
      - :math:`\text{PSNR}(\hat{x},x) = 10 \cdot \log_{10} \left( \frac{\text{MAX}^2}{\text{MSE}(\hat{x},x)} \right)`, where :math:`\text{MAX}` is the maximum possible pixel value of the image
 
+   * - :class:`deepinv.loss.metric.SNR`
+     - :math:`\mathrm{SNR} = 10 \log_{10} \left( \frac{\sum_{i=1}^n |x_i|^2}{\sum_{i=1}^n |x_i - \hat{x}_i|^2} \right)`
```
mh-nguyen712: use the norm?

mh-nguyen712 (commented):
comments

deepinv/loss/metric/distortion.py
jscanvic: What do you mean?

deepinv/loss/metric/functional.py
jscanvic: I'd prefer if we stuck with the 1st version

* I think it's confusing that in the intermediate steps you compute the inverse of the square root of the SNR while I simply compute the actual SNR before putting it in dB
* Using `vector_norm` isn't more efficient and it might be slightly less efficient due to the square root
* The numerical error should be smaller with a ratio of means than the ratio of sums from `vector_norm`

What do you think?

jscanvic:
Thanks Hai! I made most of the changes

deepinv/loss/metric/distortion.py
mh-nguyen712: For example: https://github.com/deepinv/deepinv/blob/main/deepinv/loss/metric/distortion.py#L55 or https://github.com/deepinv/deepinv/blob/main/deepinv/loss/metric/distortion.py#L17

deepinv/loss/metric/functional.py
mh-nguyen712: - Yes I understand that it's a bit confusing to compute the inverse ratio in the intermediate step, just to avoid dividing by when x_hat and x are the same but it's rare. Both are ok for me.
- The cuda implementation is extremely fast so I think the square root is not a problem here, it's rather the python overhead when talking about efficiency here: the `vector_norm` only launches one kernel, but the 1st version runs 3 kernels (abs + pow + mean). The difference is not significant here I think (order of ms but we can have 4x difference). 
- I think the numerical error are the same for ratio with mean or sum.

Sorry for bothering you.

jscanvic: Thanks for pushing it through you're not bothering me! Will have a closer look later today

jscanvic: You're right it's more efficient my bad! I'm glad to learn that!

That being said, what do you think of using `torch.compile` here? I did a quick benchmark and it seems that it makes the function faster overall and it makes the `abs().square().mean()` slightly faster than `vector_norm()` too.

**Output**

```python
Running consistency tests on: cuda
✅ All SNR variants match successfully.

Running benchmark on: cuda | Input: 64x4x32000 | Iters: 10000
--------------------------------------------------
SNR1 (Eager, Pow)        : 1.9510 ms
SNR2 (Eager, Norm)       : 0.7352 ms
SNR3 (Compile, Pow)      : 0.3451 ms
SNR4 (Compile, Norm)     : 0.3526 ms
```

**Code**

```python
import torch
import pytest
import time

# --- The Functions ---
def snr1(x_hat: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
    noise = x_hat - x
    signal_power = x.abs().square().flatten(1).mean(dim=1)
    noise_power = noise.abs().square().flatten(1).mean(dim=1)
    return 10 * torch.log10(signal_power / noise_power)

def snr2(x_hat: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
    noise = x_hat - x
    signal_norm = torch.linalg.vector_norm(x.flatten(1), ord=2, dim=1)
    noise_norm = torch.linalg.vector_norm(noise.flatten(1), ord=2, dim=1)
    return 20 * torch.log10(signal_norm / noise_norm)

@torch.compile
def snr3(x_hat: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
    noise = x_hat - x
    signal_power = x.abs().square().flatten(1).mean(dim=1)
    noise_power = noise.abs().square().flatten(1).mean(dim=1)
    return 10 * torch.log10(signal_power / noise_power)

@torch.compile
def snr4(x_hat: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
    noise = x_hat - x
    signal_norm = torch.linalg.vector_norm(x.flatten(1), ord=2, dim=1)
    noise_norm = torch.linalg.vector_norm(noise.flatten(1), ord=2, dim=1)
    return 20 * torch.log10(signal_norm / noise_norm)

# --- The Test ---
def test_snr_consistency():
    device = "cuda"
    torch.manual_seed(42)
    B, C, T = 8, 2, 1024
    x = torch.randn(B, C, T, device=device)
    x_hat = x + torch.randn_like(x) * 0.1

    print(f"\nRunning consistency tests on: {device}")

    # Warmup / Compile
    _ = snr3(x_hat, x)
    _ = snr4(x_hat, x)

    res1, res2 = snr1(x_hat, x), snr2(x_hat, x)
    res3, res4 = snr3(x_hat, x), snr4(x_hat, x)

    torch.testing.assert_close(res1, res2, msg="Mismatch: Mean Power vs Vector Norm")
    torch.testing.assert_close(res1, res3, msg="Mismatch: Eager vs Compile (Method 1)")
    torch.testing.assert_close(res2, res4, msg="Mismatch: Eager vs Compile (Method 2)")
    torch.testing.assert_close(res3, res4, msg="Mismatch: Compiled Power vs Compiled Norm")
    print("✅ All SNR variants match successfully.")

# --- The Benchmark ---
def benchmark_snr():
    device = "cuda" if torch.cuda.is_available() else "cpu"
    # Use larger data for benchmark to saturate GPU/cache
    B, C, T = 64, 4, 32000
    x = torch.randn(B, C, T, device=device)
    x_hat = x + torch.randn_like(x) * 0.1

    iterations = 10000
    funcs = [("SNR1 (Eager, Pow)", snr1),
             ("SNR2 (Eager, Norm)", snr2),
             ("SNR3 (Compile, Pow)", snr3),
             ("SNR4 (Compile, Norm)", snr4)]

    print(f"\nRunning benchmark on: {device} | Input: {B}x{C}x{T} | Iters: {iterations}")
    print("-" * 50)

    results = []

    for name, func in funcs:
        # Warmup (critical for torch.compile to skip compilation time)
        for _ in range(10):
            _ = func(x_hat, x)

        # Sync before timing
        if device == "cuda": torch.cuda.synchronize()

        start = time.perf_counter()
        for _ in range(iterations):
            _ = func(x_hat, x)

        # Sync after timing
        if device == "cuda": torch.cuda.synchronize()
        end = time.perf_counter()

        avg_time = (end - start) / iterations * 1000 # in ms
        results.append((name, avg_time))

    for name, t in results:
        print(f"{name:<25}: {t:.4f} ms")

if __name__ == "__main__":
    test_snr_consistency()
    benchmark_snr()
```

jscanvic: For the divisions by zero (and log of zero) I'm happy with what it currently gives out:

```
Signal ≠ 0, Noise = 0, SNR = inf
Signal = 0, Noise ≠ 0, SNR = -inf
Signal = 0, Noise = 0, SNR = nan
```

Maybe we should put some form of clipping and default value for nans in the base metric class but I wouldn't put it in the functional API

Do you think I should open an issue for that?

mh-nguyen712: Yes indeed torch compile (or torch script) will remove python overhead, but I think we should leave the users do that. 
For the clipping, I think we can leave as it since the values look right to me. 
Thanks for the benchmark anyway!

jscanvic: Okay, I made the change!

PR #623, opened 2025-07-21
Miscelaneous code coverage improvements

jscanvic:
Towards #487 

**Coverage**

* Global: 82.8 ➡️ 83.1 (+0.3)
* `deepinv.models.unet`: 50.6 ➡️ 77.0 (+26.4)
* `deepinv.optim.phase_retrieval`: 10.3 ➡️ 38.0 (+27.7)

**Improvements**

* Test BFBatchNorm2d in `deepinv.models.unet`
* Refactor the UNet forward methods (some of them being untested) into one 
* Improve coverage of `deepinv.optim.phase_retrieval`

PR #976, opened 2025-12-06
Refactor UNet

nucli-vicky:
Right now, the unet network file has different forward methods depending on the `self.compact` passed. This is not necessary and decreases maintainability / readability, since each forward essentially does the same thing. 

In this PR:

- [x] `UNet`'s forward method gets unified without making breaking changes, so model weight files from previous versions can still be used
- [x] `UNet` now gets an additional argument to control the # of feature maps at each stage
- [x] Remove the creation  convblocks for `self.Up_conv{i}` if `self.cat is False`, since they will not be used and therefore just take up memory. 
- [x] Add `self.scales` (or `self.depth`?), same value as `self.compact`, but more informative naming

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -23,7 +23,7 @@ Changed
 - (Breaking) Change :class:`deepinv.physics.TomographyWithAstra` physics interface to better match the interface of the PyTorch-based `Tomography` physics (:gh:`747` by `Alexander Skorikov`_)
 - Add support for Poisson2Sparse (:gh:`677` by `Jérémy Scanvic`_)
 - (Breaking) `Tomography` physics uses the true adjoint by default. `Tomography` and `TomographyWithAstra` implement the pseudo-inverse as the solution of a least-squares problem, with the option to use `fbp`. (:gh:`930` by `Romain Vo`_)
-- :class:`deepinv.models.UNet` now accepts a new (opional) argument `channels_per_scale` to control the number of feature maps at each stage. It now also supports arbitrary number of scales and bias-free batchnorm is supported for 3D variant (:gh:`976` by `Vicky De Ridder`_)
+- :class:`deepinv.models.UNet` now accepts a new (optional) argument `channels_per_scale` to control the number of feature maps at each stage. It now also supports arbitrary number of scales and bias-free batchnorm is supported for 3D variant; also clean-up code (:gh:`976` by `Vicky De Ridder`_)
 - Add a check in `deepinv.datasets.FMD` to avoid unnecessary downloads (:gh:`962` by `Jérémy Scanvic`_)
 - Trainer checkpoint loading verbose (:gh:`982` by `Andrew Wang`_)
 


jscanvic:
Nice! I was working on that but wasn't satisfied with what I came up with and it hasn't been merged in the end #623

nucli-vicky:
I think the logic of how `scales` and `channels_per_scale` interact should be explained better in the docstring, but also don't want to make the param explanations too verbose..

nucli-vicky:
I think the builder improves readability because it has more of a 'detail hierarchy'; people can get a good sense of the overall network architecture code without having to scroll through convblock (handling) details, so reduced cognitive load?

At some point things like this should be fused with more general block construction like the unet tools in `models/utils.py`, and introducing it now in `UNet` is an initial preparatory step.

The only thing I don't like is how now we make the builder object have variable name `b` (single letter) to avoid the formatter introducing too many very small lines. 
I guess one way to bypass this is to have separate builders for `conv_block` and `up_conv` which both inherit init and the shared norm method to avoid duplicate code, but I think this overcomplicates things..

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
 
-    def forward_compact2(self, x: torch.Tensor) -> torch.Tensor:
-        # encoding path
-        cat_dim = 1
-        input = x
-        x1 = self.Conv1(input)
 
-        x2 = self.Maxpool(x1)
-        x2 = self.Conv2(x2)
+class BFBatchNorm2d(nn.BatchNorm2d):
```
jscanvic, line 312: Did you change any of this or has it just been moved to the bottom of the file?

deepinv/models/unet.py
```diff
@@ -1,64 +1,28 @@
...
-
-
 class UNet(Denoiser):
     r"""
     U-Net convolutional denoiser.
 
-    This network is a fully convolutional denoiser based on the U-Net architecture. The number of downsample steps
-    can be controlled with the ``scales`` parameter. The number of trainable parameters increases with the number of
-    scales.
+    This network is a fully convolutional denoiser based on the U-Net architecture. The depth of the network is
```
jscanvic: I don't think we should phrase it as a depth parameter. Depth would more typically be controlled with the number of successive convolutional blocks at each step more than the number of scales as far as I can tell

I think the number of scales makes sense for a UNet already, and maybe clarifying that to have N scales you need to have N - 1 pooling / upsampling layer pairs

deepinv/models/unet.py
```diff
@@ -1,64 +1,28 @@
...
+    controlled by ``scales``, which sets the number of encoder downsampling stages (and corresponding decoder
+    upsampling stages). The width of each stage is controlled by ``channels_per_scale``, which gives the number of
+    feature maps from shallow to deep.
```
jscanvic: It's not entirely clear to me what you mean by each stage. (Haven't looked at the implementation yet). Is it one per scale? One per convolutional block?

Also: I prefer speaking of dimension of the feature space more than the number of feature maps (can be confusing in some cases). It's pretty subjective so I leave it up to you

deepinv/models/unet.py
```diff
@@ -1,64 +1,28 @@
...
+
+    If ``scales`` is not given, it is inferred from ``channels_per_scale`` (its length). If both are omitted, defaults to
+    configuration with ``scales=4`` and ``channels_per_scale=[64, 128, 256, 512, 1024]``. When
+    ``scales`` is specified explicitly together with ``channels_per_scale``, only the first ``scales`` entries
```
jscanvic: I find it weird to dismiss some of the input

deepinv/models/unet.py
```diff
@@ -1,64 +1,28 @@
...
+    of ``channels_per_scale`` are used; its length must be at least ``scales``. The number of trainable parameters
+    increases with both ``scales`` and the values in ``channels_per_scale``.
```
jscanvic: What if scales is specified but not channels_per_scale?

deepinv/models/unet.py
```diff
@@ -70,10 +34,11 @@ class UNet(Denoiser):
     :param bool circular_padding: circular padding for the convolutional layers.
     :param bool cat: use skip-connections between intermediate levels.
     :param bool bias: use learnable biases.
-    :param bool, str batch_norm: if False, no batchnorm applied, if ``True``, use :class:`torch.nn.BatchNorm2d`,
-        if ``batch_norm="biasfree"``, use ``BFBatchNorm2d`` from :footcite:t:`mohan2020robust`.
-    :param int scales: Number of downsampling steps used in the U-Net. The options are 2,3,4 and 5.
+    :param bool, str batch_norm: if False, no batchnorm applied, if ``True``, use :class:`torch.nn.BatchNorm2d` (or 3d variant, when building a 3D network),
```
jscanvic: Does it support 3d now?

deepinv/models/unet.py
```diff
@@ -70,10 +34,11 @@ class UNet(Denoiser):
...
+        if ``batch_norm="biasfree"``, use ``BFBatchNorm2d`` (not yet implemented for 3D) from :footcite:t:`mohan2020robust`.
```
jscanvic: This is getting too dense please explain that the UNet supports 2d and 3d inputs at the top of the docstring plus a note before the parameter list explaining what's not supported yet for 3d inputs. Also, get rid of specific references to dimensionality and instead just say batch normalization / bias free batch normalization

deepinv/models/unet.py
```diff
@@ -70,10 +34,11 @@ class UNet(Denoiser):
...
+    :param int scales: Number of downsampling steps stages (network depth). Must be one of ``{2, 3, 4, 5}``.
```
jscanvic: Same here, I wouldn't use the term depth here

deepinv/models/unet.py
```diff
@@ -70,10 +34,11 @@ class UNet(Denoiser):
...
         The number of trainable parameters increases with the scale.
+    :param Sequence[int] channels_per_scale: Number of feature maps at each encoder stage (from shallow to deep). If None, defaults to ``[64, 128, 256, 512, 1024]``.
```
jscanvic: I find this formulation confusing

Why encoder and not decoder? It's symmetric right?

Perhaps we should add a paragraph explaining that UNets are encoder-decoders that produce feature pyramids (which is a list of feature maps at different scales taking values in possibly different feature spaces). The number of scales is the number of feature maps in the pyramid. I assume `channels_per_scale` maps a scale to the dimension of the corresponding feature space

deepinv/models/unet.py
```diff
@@ -87,7 +52,8 @@ def __init__(
         cat: bool = True,
         bias: bool = True,
         batch_norm: bool | str = True,
-        scales: int = 4,
+        scales: int = None,
```
jscanvic:
```suggestion
        scales: int | None = None,
```

deepinv/models/unet.py
```diff
@@ -87,7 +52,8 @@ def __init__(
...
+        channels_per_scale: Sequence[int] = None,
```
jscanvic:
```suggestion
        channels_per_scale: Sequence[int] | None = None,
```

deepinv/models/unet.py
```diff
@@ -99,161 +65,93 @@ def __init__(
...
+            if channels_per_scale is not None:
+                scales = len(channels_per_scale)
+            else:
+                scales = 4  # legacy default
+
+        if scales not in (2, 3, 4, 5):  # pragma: no cover
+            raise ValueError("`scales` must be one of {2, 3, 4, 5}.")
+
+        if channels_per_scale is None:
+            channels_per_scale = [64, 128, 256, 512, 1024]
```
jscanvic: Okay I get why you say we only take the 1st values in this list, it's because you rely on the number of scales to be finite (in the current implementation) to hard code all of the possible dimensions

I find this confusing and error prone (dismissing input). I think it makes more sense to have this interface:

Either you
* Specify scales and let the dimensions default to `channels_per_scale = [ 64 * 2 ** k for k in range(scales) ]`
* You specify `channels_per_scale` and the number of scales defaults to `scales = len(channels_per_scale)`
* And if you specify both it fails. I don't think there's any reason to allow redundant parametrization here but if we were fine with it we could also just put a warning if the two values match and raising an exception if they don't

deepinv/models/unet.py
```diff
@@ -99,161 +65,93 @@ def __init__(
...
+
+        if len(channels_per_scale) < scales:  # pragma: no cover
+            raise ValueError(
+                f"`channels_per_scale` must have length at least `scales` "
+                f"(got len={len(channels_per_scale)}, scales={scales})."
+            )
+
+        cps = channels_per_scale
```
jscanvic: Put the shorthand right before where it's used to make it clearer

deepinv/models/unet.py
```diff
@@ -99,161 +65,93 @@ def __init__(
...
+
         dim = fix_dim(dim)
 
         conv = conv_nd(dim)
-        batchnorm = batchnorm_nd(dim)
```
jscanvic, line 105: Okay so we already support 3d inputs forget my previous comments about that

deepinv/models/unet.py
```diff
@@ -99,161 +65,93 @@ def __init__(
...
+        self.Maxpool = maxpool_nd(dim)(kernel_size=2, stride=2)
 
         self.in_channels = in_channels
         self.out_channels = out_channels
 
         self.residual = residual
         self.cat = cat
-        self.compact = scales
-        self.Maxpool = maxpool_nd(dim)(kernel_size=2, stride=2)
+        self.depth = scales
```
jscanvic: Does it need to be an attribute? Do we use it outside the constructor?

deepinv/models/unet.py
```diff
@@ -99,161 +65,93 @@ def __init__(
...
+        self.compact = scales  # backward compatibility, old attribute name
```
jscanvic: Nice, if we add a new attribute I'd put it as a property instead to avoid possible inconsistent states

deepinv/models/unet.py
```diff
@@ -99,161 +65,93 @@ def __init__(
...
-                        ch_out,
-                        kernel_size=3,
-                        stride=1,
-                        padding=1,
-                        bias=bias,
-                        padding_mode="circular" if circular_padding else "zeros",
-                    ),
-                    nn.ReLU(inplace=True),
-                )
+        b = UNetConvBlockBuilder(dim, circular_padding, biasfree, bias, batch_norm)
```
jscanvic: * Use keyword arguments instead
* Make the builder class private if you don't intend it to be used outside this module (with an underscore prefix)
* If you make it private, make the name shorter too, even just _Builder would suffice
* It's subjective but I find the word builder a bit too abstract here, maybe just call it _Blocks (or _Layers?), happy to not change that if you prefer

deepinv/models/unet.py
```diff
@@ -99,161 +65,93 @@ def __init__(
...
-                    nn.ReLU(inplace=True),
-                )
-
-        self.Conv1 = conv_block(ch_in=in_channels, ch_out=64)
-        self.Conv2 = conv_block(ch_in=64, ch_out=128)
+        self.Conv1 = b.conv_block(ch_in=in_channels, ch_out=cps[0])
+        self.Conv2 = b.conv_block(ch_in=cps[0], ch_out=cps[1])
         self.Conv3 = (
-            conv_block(ch_in=128, ch_out=256) if self.compact in [3, 4, 5] else None
+            b.conv_block(ch_in=cps[1], ch_out=cps[2]) if self.depth > 2 else None
```
jscanvic: Nice!

Maybe >= 3 instead of > 2 to match Conv3

deepinv/models/unet.py
```diff
@@ -99,161 +65,93 @@ def __init__(
...
 
-        if self.compact == 5:
-            self._forward = self.forward_standard
-        if self.compact == 4:
-            self._forward = self.forward_compact4
-        if self.compact == 3:
-            self._forward = self.forward_compact3
-        if self.compact == 2:
-            self._forward = self.forward_compact2
+        self._forward = self._forward_unet  # avoid breaking changes
```
jscanvic: It's private so I'd argue against leaving it in

deepinv/models/unet.py
```diff
@@ -99,161 +65,93 @@ def __init__(
...
+            b.conv_block(ch_in=cps[1] * 2, ch_out=cps[1])
+            if (self.depth > 2 and self.cat)
+            else None
         )
 
-        self.Up2 = up_conv(ch_in=128, ch_out=64)
-        self.Up_conv2 = conv_block(ch_in=128, ch_out=64)
+        self.Up2 = b.up_conv(ch_in=cps[1], ch_out=cps[0])
+        self.Up_conv2 = (
+            b.conv_block(ch_in=cps[0] * 2, ch_out=cps[0]) if self.cat else None
```
jscanvic: I'm happy with it as is but we could probably get rid of the redundancy between the value of self.cat and the presence of up conv attributes

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
         :param float sigma: noise level (not used).
         """
 
-        factor = 2 ** (self.compact - 1)
+        factor = 2 ** (self.depth - 1)
         if x.size(2) % factor == 0 and x.size(3) % factor == 0:
-            return self._forward(x)
+            return self._forward_unet(x)
```
jscanvic: If we don't keep backwards compatiblity with `self._forward` we can simply name this function `self._forward`

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
 
-    def forward_compact4(self, x: torch.Tensor) -> torch.Tensor:
-        # def forward_compact4(self, x):
-        # encoding path
-        cat_dim = 1
-        input = x
+            enc_feats.append(x)
 
-        x1 = self.Conv1(input)  # 1->64
+        n_levels = len(enc_feats)
```
jscanvic: This is the number of scales / what you call depth, I'd call it `scales`. It's confusing to give the same thing different names

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
+        enc_blocks = [self.Conv1, self.Conv2, self.Conv3, self.Conv4, self.Conv5]
+        enc_feats = []
 
-        d4 = self.Up4(d5)
-        if self.cat:
-            d4 = torch.cat((x3, d4), dim=cat_dim)
-            d4 = self.Up_conv4(d4)
+        for i, block in enumerate(enc_blocks):
+            if block is None:
+                break
```
jscanvic: I find it confusing, can you make enc_blocks only contain the blocks that are actually present?

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
-        input = x
+            enc_feats.append(x)
 
-        x1 = self.Conv1(input)  # 1->64
+        n_levels = len(enc_feats)
 
-        x2 = self.Maxpool(x1)
-        x2 = self.Conv2(x2)  # 64->128
+        up_blocks = [self.Up2, self.Up3, self.Up4, self.Up5]
+        dec_blocks = [self.Up_conv2, self.Up_conv3, self.Up_conv4, self.Up_conv5]
```
jscanvic: Same for those as for enc_blocks

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
 
-        x3 = self.Maxpool(x2)
-        x3 = self.Conv3(x3)  # 128->256
+        for level in range(n_levels - 1):
+            id_decoder = (n_levels - 2) - level
 
-        x4 = self.Maxpool(x3)
-        x4 = self.Conv4(x4)  # 256->512
+            x = up_blocks[id_decoder](x)
```
jscanvic: If we're going the specific indexation route I'd rather use `getattr(self, f"Up{idx}")` and not introduce two indexing of the same thing (the up and up conv blocks)

For the encoder blocks there's no issue because we're simply processing them in order without relying on the indexation

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
-            d3 = self.Up_conv3(d3)
+        x = self.Conv_1x1(x)
 
-        d2 = self.Up2(d3)  # 128->64
-        if self.cat:
-            d2 = torch.cat((x1, d2), dim=cat_dim)
-            d2 = self.Up_conv2(d2)
+        return (
+            x + network_input
+            if self.residual and self.in_channels == self.out_channels
```
jscanvic, line 176: Unrelated but we should probably do something about this conditional (in another PR)

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
+            else x
+        )
 
-        d1 = self.Conv_1x1(d2)
+    def forward_standard(self, x: torch.Tensor) -> torch.Tensor:
+        # These are kept to avoid breaking changes
+        return self._forward_unet(x)
```
jscanvic: Deprecate

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
 
-        out = d1 + x if self.residual and self.in_channels == self.out_channels else d1
-        return out
+    def forward_compact4(self, x: torch.Tensor) -> torch.Tensor:
+        return self._forward_unet(x)
```
jscanvic: Deprecate

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
 
     def forward_compact3(self, x: torch.Tensor) -> torch.Tensor:
-        # encoding path
-        cat_dim = 1
-        input = x
-        x1 = self.Conv1(input)
+        return self._forward_unet(x)
```
jscanvic: Deprecate

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
 
-        x2 = self.Maxpool(x1)
-        x2 = self.Conv2(x2)
-
-        x3 = self.Maxpool(x2)
-        x3 = self.Conv3(x3)
+    def forward_compact2(self, x: torch.Tensor) -> torch.Tensor:
+        return self._forward_unet(x)
```
jscanvic: Deprecate

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
+
+    def load_state_dict(
+        self, state_dict: Mapping[str, Any], strict: bool = True, assign: bool = False
+    ) -> _IncompatibleKeys:
+        if not self.cat:
```
jscanvic: Nice

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
-            d3 = self.Up_conv3(d3)
+        return super().load_state_dict(state_dict, strict=strict, assign=assign)
 
-        d2 = self.Up2(d3)
-        if self.cat:
-            d2 = torch.cat((x1, d2), dim=cat_dim)
-            d2 = self.Up_conv2(d2)
 
-        d1 = self.Conv_1x1(d2)
+class UNetConvBlockBuilder:
```
jscanvic: Make it a dataclass to emphasize it's just a collection of functions that depend on some shared immutable state

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
+    def __init__(
+        self,
+        dim: int,
+        circular_padding: bool,
+        biasfree: bool,
+        bias: bool,
+        norm: bool,
+    ):
+        self.padding_mode = "circular" if circular_padding else "zeros"
```
jscanvic: I'd move that out of the class

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
 
-        d2 = self.Up2(d3)
-        if self.cat:
-            d2 = torch.cat((x1, d2), dim=cat_dim)
-            d2 = self.Up_conv2(d2)
 
-        d1 = self.Conv_1x1(d2)
+class UNetConvBlockBuilder:
+    def __init__(
+        self,
```
jscanvic: Make the rest of the parameters keyword only using the `self, *, other, params` syntax

(Probably not needed if we go the dataclass route)

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
+        norm: bool,
+    ):
+        self.padding_mode = "circular" if circular_padding else "zeros"
+        self.biasfree_norm = biasfree
+        self.use_bias = bias
+        self.norm = norm
+
+        self.dim = dim
+
+    def make_norm(self, ch_out):
```
jscanvic: Add types

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
+            return None
+        return (
+            BFBatchNorm2d(ch_out, use_bias=self.use_bias)
+            if self.biasfree_norm
+            else batchnorm_nd(self.dim)(ch_out)
+        )
 
-        out = d1 + x if self.residual and self.in_channels == self.out_channels else d1
-        return out
+    def conv_block(self, ch_in, ch_out):
```
jscanvic: Add types

deepinv/models/unet.py
```diff
@@ -266,135 +164,207 @@ def forward(self, x: torch.Tensor, sigma=None, **kwargs) -> torch.Tensor:
...
+                padding_mode=self.padding_mode,
+            )
+        )
+        norm = self.make_norm(ch_out)
+        if norm is not None:
+            layers.append(norm)
+        layers.append(nn.ReLU(inplace=True))
+        return nn.Sequential(*layers)
+
+    def up_conv(self, ch_in, ch_out):
```
jscanvic: Did you change those functions compared to the original implementation?

deepinv/models/unet.py
nucli-vicky: No changes, just a move!

deepinv/models/unet.py
nucli-vicky: These should match exactly in behavior

deepinv/models/unet.py
nucli-vicky: removed `self.depth`

deepinv/models/unet.py
nucli-vicky: Do you mean here to throw a warning that it is deprecated? Or already remove altogether?

deepinv/models/unet.py
nucli-vicky: Thanks!

deepinv/models/unet.py
nucli-vicky: Do we keep `self.compact` as is?

deepinv/models/unet.py
nucli-vicky: Hmm, to me the dimension of the feature space is actually more ambiguous (it can mean either resolution, or number of features, or both). I guess number of features makes the most sense?

On "stages": this is what I tend to use as it commonly used in the design of general CNNs (see e.g. the convnext, stage compute ration, ...), there a stage refers to the collection of blocks acting on some feature map spatial resolution (aka, two stages are always separated by a downsampling operation)

deepinv/models/unet.py
nucli-vicky: Agree, depth terminology is not appropriate here!

deepinv/models/unet.py
jscanvic: Sorry I meant throwing a deprecation warning when it's called!

deepinv/models/unet.py
jscanvic: Thanks for clarifying let's keep it your way!

nucli-vicky:
Remaining to do:
- [ ] Fix idx for decoder part --> make some mapper based on the `UNet`'s length
- [ ] Improve docstring
- [ ] Write extensive test for backward compatibility

deepinv/models/unet.py
nucli-vicky: Fixed with requiring that if both are passed, they must match

deepinv/models/unet.py
nucli-vicky: I'm having trouble clearly writing the fact that we do `[64 * (2**k) for k in range(scales)]` in that case; I guess we can literally put that part of the code in the docstring?

deepinv/models/unet.py
nucli-vicky: We could put some explainer on unets. For now I just switched it to "Number of feature maps at each stage"

nucli-vicky:
Thanks for the extensive review, @jscanvic , I tried to address all of your comments.
(i) I made it such that the number of scales can be arbitrary. While a custom `_load_state_dict` guarantees old weights can be loaded, there is no robust fix to support optimizers saved from the `UNet` version on main. Therefore I did end up using `setattr` & `getattr`.
(ii) Similarly, I removed the idea of not building `Up_conv{i}` when `cat=False`, this would break optimizers loaded from the original..
(iii) `BFBatchNorm` is now supported for 3D.
(iv) I feel like we can still improve the docstring a bit, but not sure how.

*Note on the script for backward compat:* I tested locally, but the doctest also loads a pretrained `UNet` in e.g. `self-supervised/demo_multioperator_imaging.py`, and these still work fine

deepinv/models/unet.py
nucli-vicky: indeed inserted it for now

jscanvic:
Thanks for the hard work Vicky! I'll do a second pass shortly

deepinv/models/unet.py
jscanvic: I'm happy with what you put in

deepinv/models/unet.py
```diff
@@ -1,79 +1,43 @@
...
+    :param bool residual: use a skip-connection between input and output.
     :param bool circular_padding: circular padding for the convolutional layers.
     :param bool cat: use skip-connections between intermediate levels.
-    :param bool bias: use learnable biases.
-    :param bool, str batch_norm: if False, no batchnorm applied, if ``True``, use :class:`torch.nn.BatchNorm2d`,
-        if ``batch_norm="biasfree"``, use ``BFBatchNorm2d`` from :footcite:t:`mohan2020robust`.
-    :param int scales: Number of downsampling steps used in the U-Net. The options are 2,3,4 and 5.
-        The number of trainable parameters increases with the scale.
+    :param bool bias: use learnable biases in conv and norm layers.
+    :param bool, str batch_norm: if False, no batchnorm applied, if ``True``, use batchnormalization,
```
jscanvic:
```suggestion
    :param bool, str batch_norm: if False, disable normalization entirely, if ``True``, use batch normalization,
```

deepinv/models/unet.py
```diff
@@ -1,79 +1,43 @@
...
+        if ``batch_norm="biasfree"``, use the bias-free batchnorm from :footcite:t:`mohan2020robust`.
```
jscanvic:
```suggestion
        if ``batch_norm="biasfree"``, use the bias-free batch norm from :footcite:t:`mohan2020robust`.
```

deepinv/models/unet.py
jscanvic: I'd put it as a deprecated read-only property instead of leaving it as a non-deprecated writable attribute

Setting `UNet.compact` doesn't do anything meaningful at the moment so no need to support a setter here

Feel free to add `self._compact = scales` if you need to have access to it when defining the property

deepinv/models/unet.py
```diff
@@ -87,314 +51,305 @@ def __init__(
...
+        return y.view(return_shape).transpose(0, 1)
 
-        x3 = self.Maxpool(x2)
-        x3 = self.Conv3(x3)
 
-        d3 = self.Up3(x3)
-        if self.cat:
-            d3 = torch.cat((x2, d3), dim=cat_dim)
-            d3 = self.Up_conv3(d3)
+class BFBatchNorm2d(_BFBNCore, nn.BatchNorm2d):
```
jscanvic: Please use a function instead of inheriting from a new _BFBNCore class to avoid changing the public interface of BFBatchNorm2d (which we don't really intend)

(Another side benefit of doing that is we avoid a meaningless class that can't be used on its own and which is not type-safe due to relying on methods and attributes that don't exist in the class itself but only in the subclasses. The whole Liskov substitution principle yada yada)

Something like:

```python
# Definition
def _bfbn_forward(m: nn.BatchNorm2d | nn.BatchNorm3d, x: torch.Tensor, *, use_bias: bool) -> torch.Tensor:
      pass

# Usage
return _bf_forward(self, x, use_bias=self.use_bias)
```

(The extra param is for type safety, m doesn't have m.use_bias if it's just a regular BatchNorm)

deepinv/models/unet.py
```diff
@@ -87,314 +51,305 @@ def __init__(
...
+        self,
+        num_features: int,
+        eps: float = 1e-5,
+        momentum: float = 0.1,
+        use_bias: bool = False,
+        affine: bool = True,
+    ):
+        super().__init__(num_features, eps=eps, momentum=momentum, affine=affine)
+        self.use_bias = use_bias
+        self.affine = affine
```
jscanvic: This is mostly a no-op - the affine attribute is set in the parent constructor already

I'd get rid of it

deepinv/models/unet.py
```diff
@@ -87,314 +51,305 @@ def __init__(
...
 
-        d1 = self.Conv_1x1(d2)
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        self._check_input_dim(x)
```
jscanvic: We're nasty to be doing that :( but anyway I'd move it in self._bf_forward / _bfbn_forward

deepinv/models/unet.py
```diff
@@ -87,314 +51,305 @@ def __init__(
...
 
-        if self.compact == 5:
-            self._forward = self.forward_standard
-        if self.compact == 4:
-            self._forward = self.forward_compact4
-        if self.compact == 3:
-            self._forward = self.forward_compact3
-        if self.compact == 2:
-            self._forward = self.forward_compact2
+        self._enc_names = tuple(f"Conv{i+1}" for i in range(scales))
```
jscanvic, line 130: I assume you moved them back to the constructor because scales is not available as an attribute in the forward method?

Did you consider adding self._scales instead?

(Happy with the current option, just a suggestion)

deepinv/models/unet.py
```diff
@@ -65,17 +66,21 @@ def __init__(
...
+            )
+
         if scales is None:
             if channels_per_scale is not None:
                 scales = len(channels_per_scale)
             else:
                 scales = 4  # legacy default
 
-        if scales not in (2, 3, 4, 5):  # pragma: no cover
-            raise ValueError("`scales` must be one of {2, 3, 4, 5}.")
```
jscanvic: Please:

* Raise an explicit ValueError if scales <= 0
* I'm not sure we should support scales=1 maybe put a warning, I'm happy with not supporting it and raising a ValueError if you prefer
* Put it in the changelog that the UNet now accepts an arbitrary number of scales

nucli-vicky: Hmm, normally I'm all for flexibility, and things should work fine for scales = 1. However, in that case it really is not a UNet architecture anymore and just some conv blocks acting purely in the input resolution space. Therefore I think that indeed, raising a `ValueError` when is fine when scales = 1.

deepinv/models/unet.py
nucli-vicky: I think setting these names makes for a cleaner forward code :)

nucli-vicky:
Thanks for the comments @jscanvic , thanks for pointing out `_bf_forward` is better as a function, I was overcomplicating a bit there :)

jscanvic (approved):
Hi Vicky, thanks for making the changes and sorry for the delay! LGTM

docs/source/changelog.rst
```diff
@@ -23,6 +23,7 @@ Changed
 - (Breaking) Change :class:`deepinv.physics.TomographyWithAstra` physics interface to better match the interface of the PyTorch-based `Tomography` physics (:gh:`747` by `Alexander Skorikov`_)
 - Add support for Poisson2Sparse (:gh:`677` by `Jérémy Scanvic`_)
 - (Breaking) `Tomography` physics uses the true adjoint by default. `Tomography` and `TomographyWithAstra` implement the pseudo-inverse as the solution of a least-squares problem, with the option to use `fbp`. (:gh:`930` by `Romain Vo`_)
+- :class:`deepinv.models.UNet` now accepts a new (opional) argument `channels_per_scale` to control the number of feature maps at each stage. It now also supports arbitrary number of scales and bias-free batchnorm is supported for 3D variant (:gh:`976` by `Vicky De Ridder`_)
```
Andrewwango:
```suggestion
- :class:`deepinv.models.UNet` now accepts a new (optional) argument `channels_per_scale` to control the number of feature maps at each stage. It now also supports arbitrary number of scales and bias-free batchnorm is supported for 3D variant; also clean-up code (:gh:`976` by `Vicky De Ridder`_)
```

PR #981, opened 2025-12-09
Typing & small fixes in models

nucli-vicky:
Small effort towards #536 , with some misc. other fixes along the way.

- [x] Add / complete typing in wavdict, tv, tgv, RAM
- [x] Remove duplicate "install ptwt" message in `WaveletDenoiser` docstring.
- [x] Misc. docstring fixes
- [x] More informative error messages.
- [x] Correct the spelling error in method name `thresold_func` & `thresold_2D` + aliases for backward compatibility

*Fix in TV / TGV*
TV / TGV forward method had a default `None` value in forward for `ths`, for which the function would break with non-informative error (`lambd is not defined`). 
Now, we optionally allow the user to specify a `ths` at init which is put as `self.ths`. If no `ths` is provided on forward, we fall back to `self.ths`. If neither is present, we throw an informative error message.

### DIFF
--- a/deepinv/models/ram.py
+++ b/deepinv/models/ram.py
@@ -38,6 +38,7 @@ class RAM(Reconstructor, Denoiser):
     :param Sequence in_channels: Number of input channels. If a list is provided, the model will have separate heads for each channel.
     :param str device: Device to which the model should be moved. If None, the model will be created on the default device.
     :param bool, str pretrained: If `True`, the model will be initialized with pretrained weights. If `str`, load from file.
+
     |sep|
 
       >>> import deepinv as dinv


nucli-vicky:
Thanks @Andrewwango , I think everything passes now, but I've been a bit confused about the unrelated HF error that's often showing up recently in CI/CD docs test..

Andrewwango:
@nucli-vicky I think if you merge from main the CI should work now...

nucli-vicky:
Hmm, even after #987 , the doctest fails here. I understand somewhat why (although `HF_HUB_DISABLE_PROGRESS_BARS=1` would make me think the problem is solved?), but I do not understand why the fail is stochastic, as it did run without error in #987 ..
@Andrewwango (cc @mh-nguyen712)

Andrewwango:
The error is below. It should be sufficient to add the doctest IGNORE OUTPUT thing added in #987 but to the doctest in this file too.
```
File "user_guide/reconstruction/pretrained-models.rst", line 21, in default
Failed example:
    model = dinv.models.RAM(pretrained=True) # or any of the models listed below
Expected nothing
Got:
    Downloading: "https://huggingface.co/mterris/ram/resolve/main/ram.pth.tar" to /home/runner/.cache/torch/hub/checkpoints/ram.pth.tar
```

nucli-vicky:
Thanks @Andrewwango , should be good now!

Minor note: For RAM I always typed as `Tensor`, even though technically it could be a `TensorList`, I think this is fine? Otherwise (and more generally), we could make a `TensorLike` type which covers both?

Andrewwango:
Thanks @nucli-vicky !

And yes it's ok for now but you're right, it would be nice to have a TensorLike type...

Issue #975, opened 2025-12-06
Tomography baseline iterative algorithms


Tmodrzyk:
### Describe the feature

Many of the common iterative algorithms found in Tomography are missing from the library. They usually don't really fit in any first order algorithm canva (forward-backward, primal-dual etc). Now that we have predefined algorithms we could add these to the list of baselines.
Some of these widely used algorithms:
- [ ] ART
- [ ] SART
- [x] SIRT
- [x] MLEM

ART, SART and SIRT are most common in CT, while MLEM is the main algorithm for PET and SPECT reconstruction.
Some others could still fit in the usual first order canva, such as Penalized Weighted Least Square (PWLS), which just adds a preconditioning to the data-fidelity.

Finally most of these algorithms are intended for large-scale 3D volumes, so practitioners are used to variants of these algorithms that run on subsets of the data (so-called ordered-subsets algorithms), which can be thought of as simply taking mini-batches of the measurement data. There are many different ways of selecting and defining these mini-batches. This also means only parts of the forward operator are used at each iteration.

This subset feature is not urgent, as most of these algorithms would run very fast in 2D, ~~and deepinv does not fully support 3D physics yet.~~

### Motivation

This would give users working on tomography ready-to-use baselines to compare to, without relying on outside implementations.

Tmodrzyk:
Updates on this:
- the subsets are now the priority, the best way is probably to use `StackedPhysics`
- MLEM can be adapted to handle Log-Poisson noise which is more accurate for CT 
- MLEM should be adapted to handle the new PET operator of parallelproj #1099 (which I believe will just be adding a constant related to the background activity)

sacha-bouchez:
Hello @Tmodrzyk. It looks like your MLEM implementation does not consider the systematic background error (like we have in PET for example) eventhough it is taken into account in the cost computation (https://github.com/deepinv/deepinv/blob/main/deepinv/optim/optim_iterators/mlem.py#L36).

PS : I'm very new to the library and I might miss something.

Tmodrzyk:
Hi @sacha-bouchez thanks for your comment. This is a known issue, background is taken into account in the OSEM PR #1255 :)

Tmodrzyk:
For emission tomography, I think we're almost done with the addition of BSREM mentioned in #1297 and implemented in #1322.
[...]

PR #985, opened 2025-12-12
SIRT baseline algorithm for Tomography 

Tmodrzyk:
Draft PR towards #975, specifically for the SIRT algorithm.

### DIFF
--- a/docs/source/api/deepinv.optim.rst
+++ b/docs/source/api/deepinv.optim.rst
@@ -34,11 +34,8 @@ Base Class
    deepinv.optim.MD
    deepinv.optim.PMD
    deepinv.optim.PDCP
-<<<<<<< HEAD
    deepinv.optim.SIRT
-=======
    deepinv.optim.MLEM
->>>>>>> 362330ed6af04fb4e8f7ba323de2207e57d8a3d3
 
 
 Potentials


deepinv/optim/utils.py
```diff
@@ -38,6 +38,7 @@ def objective_function(x, data_fidelity, prior, cur_params, y, physics):
     :param deepinv.physics physics: Instance of the physics modeling the observation.
     """
     if prior is not None and prior.explicit_prior:
+        print(data_fidelity(x, y, physics))
```
Andrewwango: remove print

deepinv/optim/optim_iterators/sirt.py
```diff
@@ -0,0 +1,80 @@
...
+        row_sum = row_sum.clamp(min=eps)
+        col_sum = col_sum.clamp(min=eps)
+
+        self._cached_row_sum = row_sum
+        self._cached_col_sum = col_sum
+        self.sinogram_shape = sinogram_shape
+
+        return row_sum, col_sum
+
+    def forward(
```
Andrewwango, line 78: docstring + typing for these two methods, especially for the dict passing

deepinv/optim/optimizers.py
```diff
@@ -2225,3 +2226,63 @@ def custom_init(y, physics):
             trainable_params=trainable_params,
             **kwargs,
         )
+
+
+class SIRT(BaseOptim):
+    r"""Simultaneous Iterative Reconstruction Technique (SIRT) optimization module.
+
+    Implementation of the `Simultaneous Iterative Reconstruction Technique (SIRT) <https://www.sciencedirect.com/science/article/pii/0022519372901804>`_
```
Andrewwango: cite

Andrewwango (changes requested):
Looks good, I think add a small test to check it works? + add to API and user guide

deepinv/optim/optimizers.py
```diff
@@ -2225,3 +2226,63 @@ def custom_init(y, physics):
...
+    algorithm for tomographic reconstruction. This algorithm is especially used in transmission tomography, i.e for X-ray computed tomography.
+    The algorithm minimizes the least-squares data-fidelity term :math:`\|Ax-y\|_2^2` and does not support any regularization.
```
romainvo: I would specify that the algorithm solves a "weighted" least-squares problem, i.e. $\|Ax - y \|_{W}^2$, for reference, you could cite this work: _Computational Analysis and Improvement of SIRT_, 2008, Gregor and Benson.

deepinv/optim/optimizers.py
```diff
@@ -2225,3 +2226,63 @@ def custom_init(y, physics):
...
+        \begin{equation*}
+        x_{k+1} = x_k + \tau V A^{\top} W (y - A x_k)
+        \end{equation*}
+
+    where
+    - :math:`\tau` is a stepsize parameter,
+    - :math:`W = \mathrm{diag}\left(\frac{1}{\sum_{i}a_{ij}}\right)`, a diagonal matrix where each element is the inverse of the row sums of :math:`A`,
+    - :math:`V = \mathrm{diag}\left(\frac{1}{\sum_{j}a_{ij}}\right)`, a diagonal matrix where each element is the inverse of the column sums of :math:`A`.
+
+    The stepsize parameter :math:`\tau`, sometimes called relaxation parameter, should satisfy :math:`0 < \tau < 2` for convergence.
```
romainvo, line 2375: I would also give intuition on what the matrix W and V are doing, i.e.:
- The entries of W are inversely proportional to the length traveled by each ray. The algorithm tolerates larger errors for measurements induced by rays that intersect a larger portion of the object.
- The entries of V are inversely proportional to the number of rays intersecting each voxel. It balances the update based on the voxels' sensitivity to the measurements

```suggestion
    where
    - :math:`\tau` is a stepsize parameter,
    - :math:`W = \mathrm{diag}\left(\frac{1}{\sum_{i}a_{ij}}\right)`, a diagonal matrix where each element is the inverse of the row sums of :math:`A`,
    - :math:`V = \mathrm{diag}\left(\frac{1}{\sum_{j}a_{ij}}\right)`, a diagonal matrix where each element is the inverse of the column sums of :math:`A`.

    The stepsize parameter :math:`\tau`, sometimes called relaxation parameter, should satisfy :math:`0 < \tau < 2` for convergence.
    
    The entries of :math:`W` are inversely proportional to the length traveled by each ray. The algorithm tolerates larger errors for measurements induced by rays that intersect a larger portion of the object. The entries of :math:`V` are inversely proportional to the number of rays intersecting each voxel. It balances the update based on the voxels' sensitivity to the measurements.
```

deepinv/optim/optim_iterators/sirt.py
```diff
@@ -0,0 +1,92 @@
...
+        self._cached_row_sum = None
+        self._cached_col_sum = None
+        self.sinogram_shape = None
+
+    def _get_normalizers(
+        self, x: Tensor, y: Tensor, physics: LinearPhysics, eps: float
+    ):
+        sinogram_shape = tuple(y.shape)
+        # We cache the normalizers to avoid recomputing them at every iteration
+        # If the physics has changed, the shape of y should change, which will trigger a recomputation of the normalizers
```
romainvo, line 56: Underlyingly, it is a bit trickier than checking for the shape of the sinogram. You could have two sinograms with the same shape, even though the geometry of your acquisition has changed, and ultimately that would also change your row/col sums.
It might not be the cleaniest solutions but in addition to what your're doing I would also:

- cache and check the pointer of `y`, i.e. if the measurements are fixed throughout iterations, this means my geometry did not change: `pointer = y.data_ptr()`
- or check the identifier of physics, i.e. `physics_id = id(physics)`, I would not choose this as there is always the possibility the user changes the underlying geometry without instantiating a new object.

deepinv/optim/optim_iterators/sirt.py
```diff
@@ -0,0 +1,92 @@
...
+    - :math:`\tau` is a stepsize parameter which should satisfy :math:`0 < \tau < 2`
+    - :math:`W = \mathrm{diag}\left(\frac{1}{\sum_{i}a_{ij}}\right)`, a diagonal matrix where each element is the inverse of the row sums of :math:`A`,
+    - :math:`V = \mathrm{diag}\left(\frac{1}{\sum_{j}a_{ij}}\right)`, a diagonal matrix where each element is the inverse of the column sums of :math:`A`.
+
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._cached_row_sum = None
+        self._cached_col_sum = None
```
romainvo: Register these two as buffers for proper interaction with the `to()` method. It also makes it clear that they are tensors who live on a specific device.

```suggestion
        self.register_buffer("_cached_row_sum", None)
        self.register_buffer("_cached_col_sum", None)
```

Tmodrzyk:
Thanks a lot for the reviews @romainvo and @Andrewwango I'll take a look shortly. :smile:

deepinv/optim/optim_iterators/sirt.py
Tmodrzyk: Agreed, I implemented the first solution

Tmodrzyk:
Hi @romainvo thank you again for the feedback, I think this is ready for another review !

deepinv/optim/optim_iterators/sirt.py
romainvo: Nice! just the small code comment to fix "the shape of y should change" -> "the measurements data y should change" or something like that

deepinv/tests/test_optim.py
```diff
@@ -1209,3 +1209,32 @@ def test_least_squares_implicit_backward(device, solver, physics_name, batch_siz
...
+
+    y = physics(test_sample)
+
+    # SIRT algorithm
+    sirt = dinv.optim.SIRT(
+        data_fidelity=L2(),
+        max_iter=500,
+        crit_conv="residual",
+        thres_conv=1e-5,
+        verbose=False,
```
romainvo, line 1265: I think it would be nice to also test the SIRT algorithm with the astra wrapper, to see that everything goes smoothly, at least in 3D, what do you think ? we could add the test here or directly into the TestTomographyWithAstra class in test_external_libraries.py .

Could you also add the test for the y pointer, check that everything is working accordingly when y changes ?

romainvo (commented):
Hi @Tmodrzyk, thanks for the changes, looks good to me! just a last few comments with respect to the tests, to get more coverage

deepinv/tests/test_optim.py
Tmodrzyk: I actually hesitated to add another test with the Astra operator ! I think you're right that we should test both.

Tmodrzyk:
So I added these two tests but kept them in ``test_optim`` even though it uses Astra

deepinv/tests/test_optim.py
```diff
@@ -1209,3 +1209,72 @@ def test_least_squares_implicit_backward(device, solver, physics_name, batch_siz
...
+            crit_conv="residual",
+            thres_conv=1e-5,
+            verbose=False,
+            early_stop=True,
+        )
+
+        x_sirt = sirt(y, physics)
+
+        assert sirt.has_converged
+        assert x_sirt is not None
```
romainvo, line 1317: I suggest uing `pytest.importorskip` to skip the second part of test, in case astra is not loaded, like it's done for other optional dependencies in the test suite. Otherwise everything looks good to me 👍

```suggestion
    pytest.importorskip("astra", reason="This test requires the Astra toolbox. It should be installed with `pip install astra-toolbox`")

    # 3D test with Astra
    if device.type != "cpu":
        test_sample = torch.ones((1, 1, 16, 16, 16)).to(device)
        physics = dinv.physics.TomographyWithAstra(
            img_size=(16, 16, 16),
            angles=180,
            angular_range=(0, 360),
            n_detector_pixels=16,
            normalize=True,
        )
        y = physics(test_sample)
        sirt = dinv.optim.SIRT(
            data_fidelity=L2(),
            max_iter=500,
            crit_conv="residual",
            thres_conv=1e-5,
            verbose=False,
            early_stop=True,
        )

        x_sirt = sirt(y, physics)

        assert sirt.has_converged
        assert x_sirt is not None
```

Tmodrzyk:
Hi @romainvo, thanks again for reviewing this PR. Do you want to give this another look or should I go ahead and merge it ?

Tmodrzyk:
@Andrewwango do you want to have another look or should I proceed with the merge?

Andrewwango (approved):
LGTM

PR #988, opened 2025-12-12
Blind Richardson-Lucy algorithm and a demo

Tmodrzyk:
PR for blind deconvolution using the Richardson-Lucy algorithm.
The non-blind version has been implemented in #1051 under the class `deepinv.optim.MLEM`.

[See the original paper](https://opg.optica.org/josaa/abstract.cfm?uri=josaa-12-1-58) for blind deconvolution with Richardson-Lucy
Richardson-Lucy is the same algorithm as MLEM in tomography, so down the line this will also be useful for #975.


jscanvic:
I really really love this!

Liu-Jia padding #934 should make it work even better for real-world (non-circular) deconvolution!

tachella:
a new section on blind inverse problems was merged to main, so we can probably add to the list of blind solvers there :)

Tmodrzyk:
I'll take a look at the new section, thanks @tachella ! Note that this is absolutely not ready for review, I first need to integrate RL correctly into the library :)

Tmodrzyk:
I think this is ready for review even though the Test CPU seems broken right now, I think this is unrelated to this PR ?

deepinv/physics/functional/blur.py
```diff
@@ -1,11 +1,46 @@
 from __future__ import annotations
 import torch
+import torch.nn.functional as F
 
 from math import sqrt, pi
 
 from .dst import dst1
 
 
+def conv2d_filter_adjoint(
```
jscanvic: Isn't it already covered by `deepinv.physics.functional.conv_transpose2d`?

examples/blind-inverse-problems/demo_blind_richardsonlucy.py
```diff
@@ -0,0 +1,419 @@
...
+# The operation :math:`\Pi_{\Delta}` keeps the kernel nonnegative and
+# normalized to unit sum. The :class:`deepinv.optim.BlindRL` class implements
+# these alternating updates.
+#
+# This algorithm is implemented under the :class:`deepinv.optim.BlindRL` class.
+# The number of iterations of each of the two updates can be controlled by the
+# parameters :code:`x_steps` and :code:`k_steps`.
+#
+# For simplicity we first test this algorithm on a clean measurement, i.e. without
+# noise. In this case, the algorithm performs fairly well, with only some oscillation artefacts near the edges known as [Gibbs phenomenon](https://en.wikipedia.org/wiki/Gibbs_phenomenon).
```
jscanvic: Mix up between markdown and RST syntax for the link

jscanvic (changes requested):
Thanks for this great contribution Thibaut!

Just minor comments for now before I dig deeper in the implementation

deepinv/physics/functional/blur.py
Tmodrzyk: It basically does the same but here we use a trick to avoid computing a huge convolution and then cropping the result.
Here's a small comparison with a 13 by 13 kernel, the operation using `torch.nn.functional.conv_transpose2d` quickly becomes too expensive. I'll add comments to make the code and the trick more explicit.
image
Code to reproduce:
```python
import matplotlib.pyplot as plt
import pandas as pd
import torch
from torch.utils.benchmark import Timer
from deepinv.physics import functional as dF

torch.manual_seed(0)
torch.set_num_threads(1)

def filter_adjoint_via_transpose(x, y, filter_size):
    hf, wf = filter_size
    full = dF.conv_transpose2d(y, x, padding="circular")
    height, width = x.shape[-2:]
    rows = (torch.arange(hf) + height // 2 - hf // 2) % height
    columns = (torch.arange(wf) + width // 2 - wf // 2) % width
    return full.index_select(-2, rows).index_select(-1, columns)

def runtime_ms(fn):
    measurement = Timer("fn()", globals={"fn": fn}).blocked_autorange(min_run_time=0.1)
    return 1e3 * measurement.median

filter_size = (13, 13)
rows = []
for size in (16, 32, 64, 128, 256):
    x = torch.randn(1, 1, size, size)
    y = torch.randn_like(x)
    specialized = lambda: dF.conv2d_filter_adjoint(x, y, filter_size)
    via_transpose = lambda: filter_adjoint_via_transpose(x, y, filter_size)
    reference, candidate = specialized(), via_transpose()
    relative_error = torch.linalg.vector_norm(
        reference - candidate
    ) / torch.linalg.vector_norm(reference)
    assert relative_error < 1e-5, f"Relative error: {relative_error:.2e}"
    specialized_ms = runtime_ms(specialized)
    transpose_ms = runtime_ms(via_transpose)
    rows.append(
        (f"{size}x{size}", specialized_ms, transpose_ms, transpose_ms / specialized_ms)
    )
results = pd.DataFrame(
    rows,
    columns=[
        "image size",
        "conv2d_filter_adjoint (ms)",
        "conv_transpose2d + crop (ms)",
        "slowdown",
    ],
)
print(results.to_string(index=False, float_format=lambda value: f"{value:.3f}"))
results.plot(x="image size", y=results.columns[1:3], marker="o", logy=True)
plt.ylabel("median runtime (ms, log scale)")
plt.show()
```

Tmodrzyk: I also added a FFT version of this trick, maybe its more convenient through a gist: https://gist.github.com/Tmodrzyk/7d4ef9d9acf58316d0cc441337b699a4

Tmodrzyk:
I added an FFT version of that modified convolution used to update the kernel, and it speeds up the algorithm quite a lot !

deepinv/optim/optim_iterators/blind_rl.py
```diff
@@ -0,0 +1,152 @@
...
+
+    .. math::
+        y \sim \operatorname{Poisson}(k * x).
+
+    The current iterate is stored as ``X["est"] = (x, k)``. The kernel update
+    assumes 2D circular convolution and a spatially invariant kernel shared by
+    all image channels.
+    """
+
+    def __init__(
```
mh-nguyen712, line 33: Can you add docstring for arguments please, k_prior, normalize_kernel, etc

deepinv/physics/functional/blur.py
mh-nguyen712: Is this different than `deepinv.physics.functional.conv_transpose2d_fft`?

examples/optimization/demo_poisson_mlem.py
```diff
@@ -75,7 +75,7 @@
 
 # Define the blur kernel
 n_channels = 3
-filter_torch = dinv.physics.functional.gaussian_blur(sigma=(2, 2))
+filter_torch = dinv.physics.functional.gaussian_blur(psf_size=(17, 17), sigma=2.0)
```
jscanvic: Why?

deepinv/physics/functional/blur.py
jscanvic: Thanks Thibaut! I'm probably slow but it took me a while to understand the difference is in which variable you take the adjoint with respect to. Maybe it's worth clarifying in the docs with a specific mention to conv_transpose2d? I mean, with a short paragraph instead of just the first line to really emphasize this

Can you also add it to the docs (in the rst file)

deepinv/physics/functional/blur.py
```diff
@@ -1,11 +1,124 @@
...
+from .convolution import (
+    conv2d_fft,
+    _flip_filter_if_needed,
+    _prepare_filter_for_grouped,
+    _raise_value_error_padding_messages,
+)
 from .dst import dst1
 
 
+def conv2d_filter_adjoint(
```
jscanvic: The current name isn't very consistent with conv_transpose2d, I'd prefer transpose to adjoint just for the sake of consistency, I'd also put the 2d at the end, something like: conv_filter_transpose2d

What do you think?

deepinv/optim/optim_iterators/blind_rl.py
```diff
@@ -0,0 +1,133 @@
...
+
+    :param deepinv.optim.Prior k_prior: optional kernel prior. Default: ``None``.
+    :param bool normalize_kernel: whether to normalize the kernel to unit sum. Default: ``True``.
+    :param bool use_fft: whether to use the FFT implementations for convolutions. Default: ``False``.
+    :param float eps: numerical stability constant used for divisions. Default: ``1e-8``.
+    """
+
+    def __init__(
+        self,
+        k_prior=None,
```
jscanvic: Type?

deepinv/optim/optim_iterators/blind_rl.py
```diff
@@ -0,0 +1,133 @@
...
+    Iterator for Blind Richardson-Lucy deconvolution.
+
+    This iterator performs one step to estimate the next kernel, and one step to
+    estimate the next image.
+
+    The current iterate is stored as ``X["est"] = (x, k)``. The kernel update
+    assumes 2D circular convolution and a spatially invariant kernel shared by
+    all image channels.
+
+    :param deepinv.optim.Prior k_prior: optional kernel prior. Default: ``None``.
```
jscanvic:
```suggestion
    :param deepinv.optim.Prior, None k_prior: optional kernel prior. Default: ``None``.
```

deepinv/optim/optim_iterators/blind_rl.py
```diff
@@ -0,0 +1,133 @@
...
+    :param bool normalize_kernel: whether to normalize the kernel to unit sum. Default: ``True``.
+    :param bool use_fft: whether to use the FFT implementations for convolutions. Default: ``False``.
```
jscanvic, line 29: It might be worth to explain the tradeoff between the two in a note, what do you think?

deepinv/optim/optim_iterators/blind_rl.py
```diff
@@ -0,0 +1,133 @@
...
+        :param deepinv.physics.Physics physics: Blur physics updated in-place with the
+            current kernel for the image update.
+        :return: Dictionary ``{"est": (x, k), "cost": F, "it": it}`` containing
+            the updated image, kernel, cost, and iteration number.
+        """
+        x_prev, k_prev = X["est"][:2]
+        x = x_prev.clamp_min(self.eps)
+        k = k_prev
+        if self.normalize_kernel:
+            k = F.normalize(
```
jscanvic: I didn't know about F.normalize, neat!

deepinv/optim/optim_iterators/blind_rl.py
```diff
@@ -0,0 +1,133 @@
...
+                k.clamp_min(0.0).flatten(1), p=1, dim=1, eps=self.eps
```
jscanvic: Is the clamp because the PSF is assumed to take non-negative values? Can we do without it?

deepinv/optim/optim_iterators/blind_rl.py
```diff
@@ -0,0 +1,133 @@
...
+            ).view_as(k)
+        else:
+            k = k.clamp_min(0.0)
+
+        hk, wk = k.shape[-2:]
+        x_steps = int(cur_params.get("x_steps", 1))
```
jscanvic: Why do we need the cast? To deal with floats?

deepinv/optim/optim_iterators/blind_rl.py
```diff
@@ -0,0 +1,133 @@
...
+        g_param = cur_params.get("g_param", None)
+        k_g_param = cur_params.get("g_param_kernel", None)
+
+        ones_y = torch.ones_like(y)
+
+        # Kernel update
+        filter_adjoint = (
+            dF.conv2d_filter_adjoint_fft if self.use_fft else dF.conv2d_filter_adjoint
+        )
+        sensitivity_k = filter_adjoint(x, ones_y, (hk, wk)).sum(dim=1, keepdim=True)
```
jscanvic: I'd put `padding="circular"` even if it's the default. It makes it clear we're consistent with the conv2d below it

deepinv/optim/optim_iterators/blind_rl.py
```diff
@@ -0,0 +1,133 @@
...
+
+        for _ in range(k_steps):
+            y_hat = dF.conv2d(x, k, padding="circular")
+            ratio = y / y_hat.clamp_min(self.eps)
+            numerator_k = filter_adjoint(x, ratio, (hk, wk)).sum(dim=1, keepdim=True)
```
jscanvic: Here too for the padding

deepinv/optim/optim_iterators/blind_rl.py
```diff
@@ -0,0 +1,133 @@
...
+            y_hat = physics.A(x)
+            numerator_x = physics.A_adjoint(y / y_hat.clamp_min(self.eps))
+            denom_x = sensitivity_x + lambda_x * cur_prior.grad(x, g_param)
+            x = x * numerator_x / denom_x.clamp_min(self.eps)
+            x = x.clamp_min(self.eps)
+
+        k_it = 0 if "it" not in X else X["it"]
+        cost = (
+            self.cost_fn(x, cur_data_fidelity, cur_prior, cur_params, y, physics)
+            if self.cost_fn is not None
```
jscanvic, line 131: Is this a recurring pattern in deepinv.optim? Maybe it's worth factoring it in `cost_fn` somehow (in another PR)

(I mean the large if condition)

deepinv/optim/optimizers.py
```diff
@@ -18,10 +18,11 @@
     MDIteration,
     MLEMIteration,
     SIRTIteration,
+    BlindRLIteration,
 )
 from deepinv.optim.fixed_point import FixedPoint
 from deepinv.optim.prior import ZeroPrior, Prior
-from deepinv.optim.data_fidelity import DataFidelity, ZeroFidelity
+from deepinv.optim.data_fidelity import DataFidelity, PoissonLikelihood, ZeroFidelity
```
jscanvic: Nice

deepinv/optim/optimizers.py
```diff
@@ -2360,6 +2361,253 @@ def __init__(
...
+        compute_metrics: bool = False,
+        **kwargs,
+    ):
+        r"""
+        Runs Blind Richardson-Lucy deconvolution.
+
+        :return: ``(x, k)`` if ``compute_metrics`` is ``False``. Otherwise,
+            returns ``((x, k), metrics)``.
+        """
+        from deepinv.physics.blur import Blur
```
jscanvic: Why is it delayed?

jscanvic (changes requested):
It looks good so far!

Minor comments

deepinv/optim/optim_iterators/blind_rl.py
Tmodrzyk: Nope, removing !

deepinv/optim/optim_iterators/blind_rl.py
Tmodrzyk: It is indeed a recurring pattern, pretty much all iterators share it.

deepinv/optim/optimizers.py
Tmodrzyk: No valid reasons, putting it back to the top imports

deepinv/physics/functional/blur.py
Tmodrzyk: Done ! You're not slow, it also took me a while to understand the trick, let me know if the explanation paragraph is still too opaque

deepinv/optim/optim_iterators/blind_rl.py
Tmodrzyk: Yes but removing it seems fine. I think as long as positive inputs are passed it's ok.

deepinv/optim/optim_iterators/blind_rl.py
Tmodrzyk: Done, in the parent class `BlindRL`

examples/optimization/demo_poisson_mlem.py
Tmodrzyk: leftovers from an old revamp of this demo, removing !

examples/blind-inverse-problems/demo_blind_richardsonlucy.py
```diff
@@ -0,0 +1,422 @@
...
+setting, where both the underlying clean image and the blur kernel are unknown.
+
+We first consider the non-blind problem
+
+.. math::
+
+    y = A_h x,
+
+where :math:`A_h` is the convolution operator with known kernel :math:`h`.
+Richardson-Lucy :footcite:t:`richardsonBayesianBasedIterativeMethod1972,lucyIterativeTechniqueRectification1974`
```
jscanvic, line 15:
```suggestion
Richardson-Lucy :footcite:p:`richardsonBayesianBasedIterativeMethod1972,lucyIterativeTechniqueRectification1974`
```

Currently it renders as:

```
Richardson-Lucy Richardson[1], Lucy[2] is the Maximum-Likelihood Expectation-Maximization (MLEM) algorithm for Poisson deconvolution. 
```

examples/blind-inverse-problems/demo_blind_richardsonlucy.py
```diff
@@ -0,0 +1,422 @@
...
+    vmax=1,
+)
+dinv.utils.plot_curves(metrics)
+
+# %%
+# One-Step-Late regularization
+# ----------------------------
+#
+# A standard heuristic to extend EM methods to the regularized setting is called
+# One-Step-Late (OSL) regularization :footcite:t:`greenUseEmAlgorithm1990`.
```
jscanvic, line 280:
```suggestion
# One-Step-Late (OSL) regularization :footcite:p:`greenUseEmAlgorithm1990`.
```

deepinv/optim/optimizers.py
```diff
@@ -2614,6 +2616,261 @@ def forward(
...
+        \frac{x^{(k)}}{A_{h^{(k+1)}}^\top \mathbf{1}}
+        \odot
+        A_{h^{(k+1)}}^\top\left(\frac{y}{A_{h^{(k+1)}} x^{(k)}}\right).
+
+    where the kernel is constrained to be nonnegative and, by default, normalized to
+    unit sum by the \Pi_{\Delta} operation after each kernel update.
+
+    Image and kernel priors can be used.
+    The regularized algorithm is implemented using the the One-Step-Late (OSL) heuristic
+    of Green :footcite:t:`greenUseEmAlgorithm1990`.
```
jscanvic, line 2656:
```suggestion
    of Green :footcite:p:`greenUseEmAlgorithm1990`.
```

jscanvic (commented):
Sorry for the delay!

It looks great! Just minor changes and I'll approve once main is merged & CI passes

deepinv/optim/optimizers.py
```diff
@@ -2614,6 +2616,261 @@ def forward(
...
+
+    def forward(
+        self,
+        y: torch.Tensor,
+        x_gt: torch.Tensor = None,
+        compute_metrics: bool = False,
+        **kwargs,
+    ):
+        r"""
+        Runs Blind Richardson-Lucy deconvolution.
```
tachella: Add input param descriptions

deepinv/optim/optimizers.py
```diff
@@ -2614,6 +2616,261 @@ def forward(
...
+
+        super(BlindRL, self).__init__(
+            BlindRLIteration(
+                k_prior=k_prior,
+                normalize_kernel=normalize_kernel,
+                use_fft=use_fft,
+                eps=eps,
+                cost_fn=cost_fn,
+            ),
+            data_fidelity=PoissonLikelihood(),
```
tachella, line 2777: This is always fixed?

I wonder if this is the same choice that fluorescence microscopists use for blind deconvolution?

examples/blind-inverse-problems/demo_blind_richardsonlucy.py
```diff
@@ -0,0 +1,422 @@
...
+===================================
+
+This example introduces the Richardson-Lucy algorithm for deconvolution in the blind
+setting, where both the underlying clean image and the blur kernel are unknown.
+
+We first consider the non-blind problem
+
+.. math::
+
+    y = A_h x,
```
tachella: Why not simply h * x (see the blur math definition)

Tmodrzyk:
Thanks for the reviews @tachella @jscanvic I really appreciate it ! 
I'll try to get back to this asap :smile:

tachella:
@Tmodrzyk my main question is whether we want to frame the algorithm as a general RL blind deconv (eg in the noiseless demo), or something specific to Poisson (eg in the method docstring)

Tmodrzyk:
@tachella from my understanding the workflow in microscopy would be non blind RL with an experimentally estimated PSF (so just our MLEM implementation). 
They have some extra pre and post processing steps, but that's basically it. See this paper for reference: https://www.nature.com/articles/s41592-024-02294-7

The blind version is more on the research side. Both blind and non-blind RL are intended for Poisson noise, but some additional steps can make it usable in other settings (Poissong-Gaussian etc). Usually, just assuming the data to be Poisson is enough. 
Again this is info I gathered here and there, I don't personally work on biological imaging !

(edit) Conclusion: I would present it as a Poisson deconvolution algorithm, with possibility to extend it to other settings later on.

deepinv/optim/optimizers.py
Tmodrzyk: So as I mentioned in the other comment, for now it is indeed fixed but with some modifications later on we might be able to handle other noise models !

examples/blind-inverse-problems/demo_blind_richardsonlucy.py
Tmodrzyk: It was to be consistent with my other notation of A_x for the convolution of the image through the kernel. Switching to "*" notations !

jscanvic (approved):
LGTM ty!

Tmodrzyk:
Hi @tachella should we merge or do you want to take another look ?

tachella (approved):
LGTM, thanks a lot Thibaut !

PR #790, opened 2025-09-29
Distributed framework for physics, priors, and data fidelity

bmalezieux:
# Draft implementation of distributed Plug and Play

Large reconstructions easily exceed single-GPU memory and/or underutilize multi-GPU setups. This PR introduces a distributed framework that shards work across processes/devices while keeping user-facing algorithms close to their single-GPU counterparts.

Quick start example: see `examples/distrib`

---

### What’s included

- **DistributedContext**
 Auto PG init (NCCL/Gloo), device pick via `LOCAL_RANK`, sharding (`round_robin` / `block`), simple `all_reduce_/broadcast_/barrier`, seeds & deterministic mode.

- **DistributedMeasurements**
Holds **only local** `y` shards via `factory` or list; aligned with physics sharding.

- **DistributedSignal**
Replicated `x` with **auto sync** on init/updates.

- **DistributedPhysics**
Local-only operator list.

- **DistributedLinearPhysics**
Linear adjoint/VJP with **single all-reduce**; `reduction={"sum","mean"}`.

- **DistributedDataFidelity**
**Loss**: local `df.d(A(x),y)` → scalar all-reduce.  
**Grad**: local VJP → x-shaped all-reduce. Supports `sum/mean`.

- **DistributedPrior + 2D strategies**
Patch → batch → prior (`grad`/`prox`) → reduce → all-reduce.  
Strategies: `basic` (naive HW splits) , `smart_tiling` (uniform tiles, RF padding, batching).

- **Utils**
`tiling_splitting_strategy`, `extract_and_pad_patch`, `tiling2d_reduce_fn` for slicing/padding/reduction metadata.

### Checklist

- [x] New modules: `distrib_framework.py`, `strategies.py`, `utils.py`
- [x] 2D tiling strategies (`basic`, `smart_tiling`) with padding/batching/reduce
- [x] Distributed loss & gradient with single-tensor/ scalar all-reduces
- [x] Example at **`examples/distrib`**
- [ ] Benchmarks + docs polish
- [ ] Tests
- [ ] 3D / N-D tiling support

### Checks to be done before submitting your PR

- [ ] `python3 -m pytest deepinv/tests` runs successfully.
- [ ] `black .` runs successfully.
- [ ] `make html` runs successfully (in the `docs/` directory).
- [ ] Updated docstrings related to the changes (as applicable).
- [ ] Added an entry to the [CHANGELOG.rst](../CHANGELOG.rst).

PR #989, opened 2025-12-12
Fix #968 and remove self.device from internal LinearPhysics logic

### LINKED ISSUE #968: Physics holding reference to self.device causes bug when updating buffers
### Environment

All

### Describe the bug

When instantiating a `MultiCoilMRI` on `"cpu"`, calling `to("cuda")` on the physics does not update the attribute `self.device`. Subsequently, when updating buffers like `mask` with a Tensor that is on GPU, the `update_parameters`  and `check_mask` method use `self.device` to automatically move the buffer to the device that was used at initialization, causing inconsistency when calling the physics with cuda inputs.

### Steps to reproduce

```python
import deepinv as dinv
import torch

if __name__ == '__main__':

    size = 64
    n_coil = 2

    # init on CPU by default
    physics = dinv.physics.MultiCoilMRI(img_size=(size, size))

    # move physics (buffers and parameters) to GPU if available
    physics = physics.to("cuda")
    # .to("cuda") does not update the device attribute
    print(physics.device)

    # both mask and coil_maps are on "cuda"
    mask = physics.mask.clone()
    coil_maps = physics.coil_maps.clone()
    physics.update_parameters(mask=mask, coil_maps=coil_maps)
    
    print(physics.coil_maps.device, mask.device)
    
    # raises an error because input is on "cuda" but buffers were moved to "cpu"
    y = physics.A(torch.randn((1,2,size,size), device="cuda"))
```

### Expected behavior

When calling `to(device)` every buffers and module parameters should live on `device`, even when updated, until another call to `to(other_device)` changes it. I've checked that at least `MRI` suffers from the same bug, and maybe other physics that have a `self.device` attribute.

If we want to keep the `self.device` and change it accordingly when calling `to()` we would have to override the method in order to propagate the change to `self.device`.
What I suggest instead is to get rid of the `self.device` attributes, which are not really useful, and instead apply the torch convention, i.e., not the nn.Module holding reference to the device but the Tensors, so buffers or parameters. 

The pattern that often appears is that we want the device as input during initialization to instantiate Tensors on the GPU. Then it is saved in a `self.device` attribute, which is used by `update_parameters` methods to push buffers on `self.device`, which is potentially not up to date because of `to()` not propagating the new device.

Not exhaustively, it appears in:
- MultiCoilMRI,
- MRI,
- All the physics in `blur.py`
- CmpressedSensing

Instead of using `self.device` to push Tensors on a specific device. I would pass `device` during initialization, and instantiate buffers on said `device`. Then, when updating buffers, the new values are assigned to the buffers without changing its device, like what happens in torch:

```python
import torch

a = torch.randn(64, device='cuda')

# This does not raise an error, and `a` still is on the GPU after new assignment
a[:] = torch.randn_like(a, device='cpu')
```

What do you think ? Please let me know if I missed any use cases where `self.device` is really handy.

```
cpu
cpu cuda:0
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/projects/UDIP/rvo/deepinv/experimental/mcmri_device_bug.py", line 25, in <module>
    y = physics.A(torch.randn((1,2,size,size), device="cuda"))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/projects/UDIP/rvo/deepinv/deepinv/physics/mri.py", line 269, in A
    Sx = self.coil_maps * self.to_torch_complex(x)[:, None]  # [B,N,...,H,W]
         ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

### Additional context

related are #383 #460

romainvo:
#As explained in #968, currently, almost all classes inheriting `LinearPhysics` hold a `self.device` attribute. This attribute is not updating correctly when calling `to(device)` on the physics, which can cause inconsistency when updating device then buffers in a certain order.

This PR fixes issue #968 and also aims to improve the internal logic of buffer updates to avoid future bugs.

An easy fix, is to define `self.device` as a `@property` (thanks @jscanvic). Each physics instance instantiates a dummy buffer, whose device is updated correctly when calling the `to()` method. If the user manually tries to update the device property by doing something like `physics.device = device`, it calls a setter which moves every buffers/parameters on the specified device.

```python
    @property
    def device(self) -> torch.device | str:
        r"""
        Returns the device where the physics parameters/buffers are stored.

        :return: device of the physics parameters.
        """
        return self._device_holder.device
    
    @device.setter
    def device(self, device: torch.device | str):
        r"""
        Sets the device where the physics parameters are stored.

        :param torch.device | str device: device to move the physics parameters to.
        """
        if not hasattr(self, '_device_holder'):
            _device_holder = torch.tensor(0., device=device)
            self.register_buffer('_device_holder', _device_holder)    
        
        # updates the device property by moving a dummy buffer
        # and also moves all parameters/buffers to the target device, for consistency
        self.to(device)
```

The other point is more a matter of design. In pytorch, `nn.Module` don't have a `self.device` attribute. We can either enforce this design choice, by removing all the `self.device` from the internal logic or just push the fix above, and maintain current design. The `self.device` is used a lot by the `update_parameters` function, but in practice, we could just maintain the device of the buffer when updated. And only use the `to(device)` logic when switching devices.

The bug needs to be fixed, but the rest is more of a discussion. I can do a separate issue/PR if necessary ?

### Removing `self.device` from every physics internal logic
- [x] `blur.py`
- [x] `forward.py`
- [x] `mri.py`
- [x] `compressed_sensing.py`
- [x] `phase_retrieval.py`
- [x] `radio.py`
- [x] `tomography.py`
- [x] `functional/radon.py`
- [x] `generator/base.py`


jscanvic:
It's really clever to use a dummy buffer! Will review asap

romainvo:
> I know it's gonna be ugly but don't we need to filter out the dummy tensor from the state dict?

Oh forgot to do it, torch provides the option already, by setting `persistent=False`. So quite clean 👌🏼

mh-nguyen712:
Thanks for the initiatives @romainvo!
I would move the behavior toward the logic of Pytorch as follows:
- we still allow a `device` parameter in `__init__` so the module will be initialized in that device. 
- we should remove all dependencies on `self.device`, so it raises error if input and buffer (if any) are living on different devices (as in native pytorch) 
- I think we don't need the `self.device` as a property and overwrite the setter. Users must call `physics.to(device)`, not (physics.device = device). We also need to make sure that the `noise_model` (if any, and its buffers) is moved to the device as well  

There are 2 reasons for this: 
- Some Physics do not have buffers (eg Denoising) 
- We're integrating distributed framework in #790 and `self.device` could create issues.

I'm happy to discuss more and correct me if I'm wrong!

deepinv/physics/forward.py
```diff
@@ -259,6 +259,11 @@ def update_parameters(self, **kwargs):
                     and hasattr(self, key)
                     and isinstance(value, torch.Tensor)
                 ):
+                    if hasattr(value, "device"):
```
jscanvic: Drop the hasattr("device") which is redundant with isinstance(torch.Tensor)

deepinv/physics/forward.py
```diff
@@ -259,6 +259,11 @@ def update_parameters(self, **kwargs):
...
+                        assert getattr(self, key).device == value.device, (
```
jscanvic: Why this check?

deepinv/physics/forward.py
```diff
@@ -440,6 +446,32 @@ def __init__(
             warnings.warn(
                 "Using implicit_backward_solver with a low number of iterations may produce inaccurate gradients during the backward pass. If you are not doing backpropagation through `A_dagger` or `prox_l2`, ignore this message. If you are training unfolded models, consider increasing max_iter."
             )
+            
+        self.device = device
```
jscanvic:
```suggestion
        device_holder = torch.empty(0, device=device)
        self.register_buffer("_device_holder", device_holder)
```

I'd avoid relying on the setter internally

deepinv/physics/forward.py
```diff
@@ -440,6 +446,32 @@ def __init__(
...
+        return self._device_holder.device
+    
+    @device.setter
+    def device(self, device: torch.device | str):
+        r"""
+        Sets the device where the physics parameters are stored.
+
+        :param torch.device | str device: device to move the physics parameters to.
+        """
+        if not hasattr(self, '_device_holder'):
```
jscanvic: Is this initialization logic needed if we initialize in the constructor?

deepinv/physics/forward.py
```diff
@@ -440,6 +446,32 @@ def __init__(
...
+            _device_holder = torch.tensor(0., device=device)
+            self.register_buffer('_device_holder', _device_holder)    
+        
+        # updates the device property by moving a dummy buffer
+        # and also moves all parameters/buffers to the target device, for consistency
+        self.to(device)
```
jscanvic: Nice

deepinv/physics/mri.py
```diff
@@ -82,19 +82,18 @@ def __init__(
...
-        self.device = device
         self.three_d = three_d
         self.img_size = img_size
 
         if mask is None:
             mask = torch.ones(*img_size, device=device)
 
         # Check and update mask
-        self.register_buffer("mask", self.check_mask(mask).to(self.device))
+        self.register_buffer("mask", self.check_mask(mask))
```
jscanvic, line 96: Nice

deepinv/physics/mri.py
```diff
@@ -155,16 +154,12 @@ def update_parameters(self, mask: Tensor = None, check_mask: bool = True, **kwar
...
                 )
                 if check_mask
                 else mask
             )
 
-            self.register_buffer("mask", mask)
-        if kwargs:
-            super().update_parameters(**kwargs)
-
+        super().update_parameters(mask=mask, **kwargs)
```
jscanvic, line 162: Nice

deepinv/physics/mri.py
```diff
@@ -229,26 +224,27 @@ def __init__(
...
                 (self.img_size[-2:] if not self.three_d else self.img_size[-3:]),
                 dtype=torch.complex64,
+                device=device
             )
         elif isinstance(coil_maps, int):
-            coil_maps = self.simulate_birdcage_csm(n_coils=coil_maps)
+            coil_maps = self.simulate_birdcage_csm(n_coils=coil_maps).to(device)
 
-        self.update_parameters(mask=mask, coil_maps=coil_maps)
+        self.register_buffer("mask", self.check_mask(mask, three_d=self.three_d, device=device))
```
jscanvic: Is the device parameter meant to be used here in check_mask?

deepinv/physics/mri.py
```diff
@@ -346,35 +342,48 @@ def update_parameters(
...
-            self.register_buffer("mask", mask)
 
         if coil_maps is not None:
-            while len(coil_maps.shape) < (
-                4 if not self.three_d else 5
-            ):  # to B,N,H,W or B,N,D,H,W
-                coil_maps = coil_maps.unsqueeze(0)
+            coil_maps = (
+                self.check_coil_maps(coil_maps, three_d=self.three_d)
+                if check_coil_maps
```
jscanvic, line 366: Why adding an option to skip this pre-processing?

deepinv/physics/mri.py
```diff
@@ -346,35 +342,48 @@ def update_parameters(
...
+                else coil_maps
+            )
 
-            if not coil_maps.is_complex():
-                raise ValueError("coil_maps should be of torch complex dtype.")
+        super().update_parameters(mask=mask, coil_maps=coil_maps, **kwargs)
+
+    @staticmethod
+    def check_coil_maps(coil_maps: Tensor, three_d: bool) -> Tensor:
```
jscanvic, line 373: It can be clearer to call it something like prepare coil maps. To me checking means just verifying everything is in order without trying to fix things

I'd mark it as a private method using an underscore prefix

deepinv/physics/mri.py
```diff
@@ -584,7 +593,7 @@ def to_static(self, mask: torch.Tensor | None = None) -> MRI:
         return MRI(
             mask=torch.clip(self.mask.sum(2), 0.0, 1.0) if mask is None else mask,
             img_size=self.img_size,
-            device=self.device,
+            device=device,
```
jscanvic, line 599: Nice

deepinv/physics/mri.py
```diff
@@ -654,6 +663,6 @@ def A_adjoint(
             return super().A_adjoint(y, mask, **kwargs)
         else:
             mask = mask if mask is not None else self.mask
-            return self.to_static().A_adjoint(
+            return self.to_static(device=y.device).A_adjoint(
```
jscanvic, line 669: Did you add the new device parameter to every call to `to_static`? (It's probably easier for you to check)

deepinv/utils/mixins.py
```diff
@@ -111,31 +112,30 @@ class MRIMixin:
 
     Base class that provides helper functions for FFT and mask checking.
     """
-
+    
+    @staticmethod
```
jscanvic, line 116: Nice

deepinv/utils/mixins.py
```diff
@@ -111,31 +112,30 @@ class MRIMixin:
...
+    @_deprecated_argument("device")
```
jscanvic, line 117: Nice!

deepinv/utils/decorators.py
```diff
@@ -2,6 +2,30 @@
 import functools
 
 
+def _deprecated_argument(*arg_names):
```
jscanvic, line 5: How redundant it is with the other warning functions in this module? (I haven't checked in details)

I think it should be good as is but it might make future refactoring easier if we have a good idea of what's in common

romainvo:
> Thanks for the initiatives @romainvo! I would move the behavior toward the logic of Pytorch as follows:
> 
>     - we still allow a `device` parameter in `__init__` so the module will be initialized in that device.
> 
>     - we should remove all dependencies on `self.device`, so it raises error if input and buffer (if any) are living on different devices (as in native pytorch)
> 
>     - I think we don't need the `self.device` as a property and overwrite the setter. Users must call `physics.to(device)`, not (physics.device = device). We also need to make sure that the `noise_model` (if any, and its buffers) is moved to the device as well
> 
> 
> There are 2 reasons for this:
> 
>     - Some Physics do not have buffers (eg Denoising)
> 
>     - We're integrating distributed framework in [Distributed framework for physics, priors, and data fidelity #790](https://github.com/deepinv/deepinv/pull/790) and `self.device` could create issues.
> 
> 
> I'm happy to discuss more and correct me if I'm wrong!

I do agree with  your points. My point with switching `self.device` as a property is mainly to avoid breaking changes, so that current users can switch more softly to not use physics.device if they do.

In practice there shouldn't be any mismatch between the device property and the buffers' device (unless a user manually pushes a specific buffer to another device, but torch allows it so it's outside our scope) as its directly taken from the dummy buffer I've instantiated.

Regarding the  distributed framework that is soon to be merged, there shouldn't be any conflict either as the device property is directly infered from the dummy buffer, whose rank is managed by the distributed logic.

I don't have strong opinion on keeping or removing self.device, so both are good with me, just wanted to point out some details. Let me know what you prefer and I'll push forward with it.

mh-nguyen712:
Yes you're right about `self.device` as a property. In anycase, our physics only supports single-device computation so it should not create issues.
However, I would not add `self.device`  as a setter (I still prefer `self.to(...)`, but make the `self.dtype` a property as well.

romainvo:
Okay, I'll remove the setter on `self.device`, so if someone tries `physics.device = device`, it'll throw an `AttributeError`.

Good point for `self.dtype` indeed, I did not see that it was also an attribute.

mh-nguyen712:
Thanks! If you have any opinions, please let us know too @jscanvic!

jscanvic:
Thanks for your input! @mh-nguyen712 I believe we're all 3 on the same page when it comes to the points you raised

I'm not convinced removing the setter is worth it but it's better we get Romain's work merged quicker so not gonna argue against it

deepinv/physics/mri.py
romainvo: No forgot to remove it

deepinv/physics/mri.py
romainvo: I agree with you. Initially, it was to be consistent with the naming of `check_mask` as both functions do the same thing basically. If we change for one, I would change for the other, but this would be breaking in case of `check_mask` ? I can still change `check_coil_maps` if you prefer in the end.

deepinv/physics/mri.py
romainvo: Just did it, propagated it to `models/dynamic.py`

deepinv/utils/decorators.py
romainvo: They almost do the same thing, in `deprecated_alias` we replace an old argument with a new one. In `depracated_argument`, we plan to remove an argument entirely from the function signature, without replacement, so here `device` in the case of `check_mask`. 

I tried explaining it in the docstring but not sure it is that clear ?

romainvo:
I'm facing issues with some of the tests in `test_physics` (`test_operators_differentiability` and `test_physics_state_dict`), some of them could be buggy, I'll try to document it as soon as possible. 

- Related to #999, `test_operators_differentiability` fails on the `"composition"` operators, for the moment I'm just skipping the test for this config.
- `test_physics_state_dict` also fails. Both tests rely on `physics._buffers` which is not well behaved and does not return buffers in nested modules. Similary the `get_all_tensor_attrs` does not yield the proper set of keys when a `nn.ModuleList` is part of the physics (as in Composed or Stacked ones). I am fixing both to account for this.

@mh-nguyen712 in the end I did not touch the `self.dtype` logic. It was becoming messy to deal with trying to fix `self.device` logic at the same time. Also some classes like `CompressedSensing` seem to explicitly rely on type casting with `self.dtype` at runtime, I am not sure if we want to deprecate that the same way we're doing it with `self.device` ?

If you're good with it, I can open another issue afterwards and work on it more properly.

mh-nguyen712:
> I'm facing issues with some of the tests in `test_physics` (`test_operators_differentiability` and `test_physics_state_dict`), some of them could be buggy, I'll try to document it as soon as possible.
> 
> * Related to [Updating logic in Composed and Stacked Physics is not well behaved #999](https://github.com/deepinv/deepinv/issues/999), `test_operators_differentiability` fails on the `"composition"` operators, for the moment I'm just skipping the test for this config.
> * `test_physics_state_dict` also fails. Both tests rely on `physics._buffers` which is not well behaved and does not return buffers in nested modules. Similary the `get_all_tensor_attrs` does not yield the proper set of keys when a `nn.ModuleList` is part of the physics (as in Composed or Stacked ones). I am fixing both to account for this.
> 
> @mh-nguyen712 in the end I did not touch the `self.dtype` logic. It was becoming messy to deal with trying to fix `self.device` logic at the same time. Also some classes like `CompressedSensing` seem to explicitly rely on type casting with `self.dtype` at runtime, I am not sure if we want to deprecate that the same way we're doing it with `self.device` ?
> 
> If you're good with it, I can open another issue afterwards and work on it more properly.

Feel free to leave the `dtype` for now. It's cool if you can open an issue for this to keep track of everything. Thanks!

romainvo:
> Okay, I'll remove the setter on `self.device`, so if someone tries `physics.device = device`, it'll throw an `AttributeError`.
> 
> Good point for `self.dtype` indeed, I did not see that it was also an attribute.

Thinking back on this point, and discussing with other users, I think having a setter for the device, at least until the next release, shouldn't cost us that much. It's not going to influence the current refactoring and may allow users to switch more softly, e.g. some people surely have resolved the current bug by doing `physics.device = device` themselves, noticing that `self.to(device)` was not working.

I fully agree that `physics.device = device` is not the proper way, but with the warnings we should be able to remove it soon, without breaking too much. What do you think @mh-nguyen712 ?

mh-nguyen712:
> > Okay, I'll remove the setter on `self.device`, so if someone tries `physics.device = device`, it'll throw an `AttributeError`.
> > Good point for `self.dtype` indeed, I did not see that it was also an attribute.
> 
> Thinking back on this point, and discussing with other users, I think having a setter for the device, at least until the next release, shouldn't cost us that much. It's not going to influence the current refactoring and may allow users to switch more softly, e.g. some people surely have resolved the current bug by doing `physics.device = device` themselves, noticing that `self.to(device)` was not working.
> 
> I fully agree that `physics.device = device` is not the proper way, but with the warnings we should be able to remove it soon, without breaking too much. What do you think @mh-nguyen712 ?

Thanks for the feedback that you have. Indeed, we can leave a warning for now and then consider removing it in the future.

deepinv/physics/blur.py
```diff
@@ -75,16 +75,99 @@ def __init__(
...
+            img_size=self.imsize, filter=filter, factor=factor, device=device
+        )
+        self.factor = filter_parameters["factor"]
+        if filter is not None:
+            self.register_buffer("filter", filter_parameters["filter"])
+            # Used for prox_l2 computation when `use_fft=True` and `padding='circular'`
+            self.register_buffer("Fh", filter_parameters["Fh"])
+            self.register_buffer("Fhc", filter_parameters["Fhc"])
+            self.register_buffer("Fh2", filter_parameters["Fh2"])
+        else:
```
mh-nguyen712: I think we can register None as a buffer too: 
https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_buffer

deepinv/physics/blur.py
```diff
@@ -106,7 +189,8 @@ def A(
 
         """
         self.imsize_dynamic = x.shape[-3:]
-        self.update_parameters(filter=filter, factor=factor, **kwargs)
+
+        self.update_parameters(filter=filter, factor=factor, device=x.device, **kwargs)
```
mh-nguyen712, line 273: I'm not sure if `update_parameters` of all physics  support `device`.

deepinv/physics/blur.py
```diff
@@ -135,17 +219,14 @@ def A_adjoint(
             If `factor` is passed, `filter` must also be passed as a `str` or `Tensor`, in order to update the filter to the new factor.
 
         """
-        if factor is not None:
-            self.factor = self.check_factor(factor)
+        self.update_parameters(filter=filter, factor=factor, device=y.device, **kwargs)
```
mh-nguyen712: here too

deepinv/physics/blur.py
```diff
@@ -389,14 +442,22 @@ class Blur(LinearPhysics):
...
+        device: str | torch.device = "cpu",
+        **kwargs,
+    ):
+        super().__init__(device=device, **kwargs)
         self.padding = padding
         assert (
             isinstance(filter, Tensor) or filter is None
         ), f"The filter must be a torch.Tensor or None, got filter of type {type(filter)}."
-        self.register_buffer("filter", filter)
+        if filter is not None:
```
mh-nguyen712: same we can use `register_buffer` I think

deepinv/physics/blur.py
```diff
@@ -523,26 +595,40 @@ def U_adjoint(self, x: Tensor) -> Tensor:
...
+        parameters = {}
+
         if filter is not None and isinstance(filter, Tensor):
-            filter = filter.to(self.device)
-            if self.img_size[0] > filter.shape[1]:
-                filter = filter.repeat(1, self.img_size[0], 1, 1)
-            mask = filter_fft_2d(filter, self.img_size)
+            filter = filter.to(device)
+            if img_size[0] > filter.shape[1]:
+                filter = filter.repeat(1, img_size[0], 1, 1)
```
mh-nguyen712, line 672: I would use `expand` (that does not allocate new memory) instead of `repeat` (allocate more memory)

deepinv/physics/blur.py
```diff
@@ -598,9 +684,21 @@ def __init__(
         device: torch.device | str = "cpu",
         **kwargs,
     ):
-        super().__init__(**kwargs)
-        self.device = device
-        self.update_parameters(filters, multipliers, padding, **kwargs)
+        super().__init__(device=device, **kwargs)
+
+        if filters is not None and isinstance(filters, Tensor):
```
mh-nguyen712: We can use a single call to `register_buffer` and remove `to(device)` (there is a call `self.to(device)` at the end.

deepinv/physics/forward.py
```diff
@@ -259,7 +259,11 @@ def update_parameters(self, **kwargs):
                     and hasattr(self, key)
                     and isinstance(value, torch.Tensor)
                 ):
-                    self.register_buffer(key, value)
+
+                    # Move `value` to the buffer's device before updating
+                    # regardless of where the `value` tensor is located.
+                    # Also performs type casting if necessary.
+                    setattr(self, key, value.to(getattr(self, key)))
```
mh-nguyen712, line 276: it seems to me that `value.to(getattr(self, key)) = value`, maybe I'm wrong. 
And why not use `register_buffer`? If a tensor is set as attribute, it does not work with `self.to(...)`

deepinv/physics/functional/radon.py
```diff
@@ -216,7 +215,7 @@ def __init__(
     ):
         super().__init__()
         self.circle = circle
-        theta = theta if theta is not None else torch.arange(180).to(self.device)
+        theta = theta if theta is not None else torch.arange(180).to(device)
```
mh-nguyen712:
```suggestion
        theta = theta if theta is not None else torch.arange(180, device=device)
```

deepinv/physics/functional/radon.py
```diff
@@ -365,19 +368,18 @@ def __init__(
     ):
         super().__init__()
         self.circle = circle
-        self.device = device
-        theta = theta if theta is not None else torch.arange(180).to(self.device)
+        theta = theta if theta is not None else torch.arange(180).to(device)
```
mh-nguyen712:
```suggestion
        theta = theta if theta is not None else torch.arange(180, device=device)
```

deepinv/physics/mri.py
```diff
@@ -229,26 +225,29 @@ def __init__(
...
 
         if coil_maps is None:
             coil_maps = torch.ones(
                 (self.img_size[-2:] if not self.three_d else self.img_size[-3:]),
                 dtype=torch.complex64,
+                device=device,
             )
         elif isinstance(coil_maps, int):
-            coil_maps = self.simulate_birdcage_csm(n_coils=coil_maps)
+            coil_maps = self.simulate_birdcage_csm(n_coils=coil_maps).to(device)
```
mh-nguyen712, line 245: minor: there is already `self.to(device)`

deepinv/physics/radio.py
```diff
@@ -90,8 +90,10 @@ def __init__(
         else:
             self.adj_projection = lambda x: x
 
+        self.to(device)
+
     def setWeight(self, w: torch.Tensor):
-        self.dataWeight = w.to(self.device)
+        self.dataWeight = w.to(self.dataWeight)
```
mh-nguyen712, line 96: maybe set as buffer?

examples/external-libraries/demo_ri_basic.py
```diff
@@ -118,8 +120,10 @@ def __init__(
         else:
             self.adj_projection = lambda x: x
 
+        self.to(device)
+
     def setWeight(self, w):
-        self.dataWeight = w.to(self.device)
+        self.dataWeight = w.to(self.dataWeight)
```
mh-nguyen712, line 126: set as buffer?

deepinv/physics/radio.py
romainvo: It's already set as a buffer in the `__init__` function ?

deepinv/physics/blur.py
romainvo: Yep, old artifacts to make some of the state_dict test pass, but should be good with the fix on the test

deepinv/physics/blur.py
romainvo: No you're right. I have removed the custom `update_parameters` function from `Downsampling` as it was redundant with the super() function. So now the correct device is pass when calling `self.get_filter_parameters`.

To make it work properly though, I've slightly altered `update_parameters`  in the `LinearPhysics` class. Now, it also updates non torch.Tensor parameters, like `self.factor`, which should be the expected behaviour as there is no need for some parameters to be Tensor.

deepinv/physics/forward.py
romainvo: Not necessarily, `value` could be on a different device than `getattr(self, key)`. I could use `register_buffer` in practice, though it's already registered as a buffer, so there is no real use. So no problem with `self.to`

deepinv/physics/blur.py
romainvo: So I did not take into account a use-case where the user might call `physics.update_parameters(filter="gaussian")` for example, outside `A` or `A_adjoint`. If we do not allow `update_parameters` to take a `device` as a kwarg, related kernels and tensors will be initialized on the "cpu" and then pushed to the correct device in `super().update_parameters()`, which is less efficient than initializing directly on the correct device.

I don't think there are easy workaround, without authorizing `device` as input to `update_parameters` @mh-nguyen712 . Unfortunately, the previous implementation relied on `self.device` for efficient initialization :')

mh-nguyen712: Ah ok sorry I didn't think about that either. So feel free to leave `device` as an argument of the custom `update_parameters()`

mh-nguyen712: Another issue of allowing non-tensor in `update_parameters()` is that it can update the class method.
Eg: the MRI example is broken, and the CI is not passing (there, `rss` is an agument of `A` and also a class method)

deepinv/physics/blur.py
romainvo: In the end I reverted that change. It made the `test_physics_state_dict` failed at line 1865

```
E           RuntimeError: Error(s) in loading state_dict for ComposedLinearPhysics:
E               While copying the parameter named "physics_list.0.filter", whose dimensions in the model are torch.Size([1, 3, 9, 9]) and whose dimensions in the checkpoint are torch.Size([1, 3, 9, 9]), an exception occurred : ('unsupported operation: more than one element of the written-to tensor refers to a single memory location. Please clone() the tensor before performing the operation.',).
```

I dug into it quickly: https://discuss.pytorch.org/t/about-error-more-than-one-element-of-the-written-to-tensor-refers-to-a-single-memory-location/85526/4 but found no easy alternative except reverting to the previous state. It's unrelated to this PR so I think it's okay to keep it that way for the moment.

deepinv/physics/blur.py
romainvo: You're right, I did not take that into account.. I "unallowed" non-tensor update in `Physics.update_parameters`.

I've also added back the custom `update_parameters` function for `Downsampling` and `BlurFFT`. They all rely to a final call to `super().udpate_parameters` to forward the update to the super class, but there is some class-specific processing done before. 

Originally, I wanted to fix a few caveats in the update logic of `Downsampling`/`BlurFFT`, but we can open an issue to track it. Dealing with it in this PR was a bit too much.

mh-nguyen712:
Can you fix the GPU test and docs please @romainvo?

romainvo:
`test_trainer_speed` is failing by a small margin (1.22 > 1.1), but I don't see anything that would slow down `physics.Inpainting`. Would it be possible to run it again to confirm the test is failing consistently @mh-nguyen712 ?

mh-nguyen712:
> `test_trainer_speed` is failing by a small margin (1.22 > 1.1), but I don't see anything that would slow down `physics.Inpainting`. Would it be possible to run it again to confirm the test is failing consistently @mh-nguyen712 ?

I think I already see this test failed randomly. I just re-run the test on my local and it's passing. So I think it's good to go.
Thanks for the fixes!

romainvo:
With respect to new physics being implemented (e.g. #1020 ), removing `self.device` induces some changes in how `Physics` and related classes should handle parameter updates. A big change is how we handle the device/dtype of the pushed Tensors.

What was done before is that every tensor is pushed/initialized onto `self.device` (with `self.dtype` if it exists). Following the same logic, if we remove `self.device` internally, this means we can rely on current buffers/parameters, which also hold the information of what the implicit device is. This is what I've implemented in this PR, i.e. every new tensor is pushed/casted to the already existing parameter/buffer it updates: see `Physics.update_parameters`. This also means that, for correct operation, every update must be performed via `super().update_parameters` (except in the __init__ function).

For the moment, the currently expected design is mostly implicit. Would it make sense to add some section detailing it to["Bring your own physics"](https://deepinv.github.io/deepinv/auto_examples/basics/demo_custom_physics.html) ? As the codebase grows larger, I feel it would help maintaining/contributing to have clear guidelines @mh-nguyen712 @jscanvic

mh-nguyen712:
> With respect to new physics being implemented (e.g. #1020 ), removing `self.device` induces some changes in how `Physics` and related classes should handle parameter updates. A big change is how we handle the device/dtype of the pushed Tensors.
> 
> What was done before is that every tensor is pushed/initialized onto `self.device` (with `self.dtype` if it exists). Following the same logic, if we remove `self.device` internally, this means we can rely on current buffers/parameters, which also hold the information of what the implicit device is. This is what I've implemented in this PR, i.e. every new tensor is pushed/casted to the already existing parameter/buffer it updates: see `Physics.update_parameters`. This also means that, for correct operation, every update must be performed via `super().update_parameters` (except in the **init** function).
> 
> For the moment, the currently expected design is mostly implicit. Would it make sense to add some section detailing it to["Bring your own physics"](https://deepinv.github.io/deepinv/auto_examples/basics/demo_custom_physics.html) ? As the codebase grows larger, I feel it would help maintaining/contributing to have clear guidelines @mh-nguyen712 @jscanvic

Yes It would be really cool if you can add a small note to the physics user guide or `Bring your own physics` example. Great idea! It helps the maintainers / developers too!

docs/source/contributing.rst
```diff
@@ -80,8 +80,8 @@ How to write and run tests:
 
   If you've contributed a new class (e.g. dataset, physics etc.), you should add it to any existing tests, e.g. those that check physics adjointness, dataset return format, etc.
 
-3. Check your tests pass locally by running ``python -m pytest deepinv/tests`` in the root directory after making the desired changes. Learn more `here <https://realpython.com/pytest-python-testing/>`__.
-4. Push your code to your PR. A maintainer will run the tests on CPU and GPU in the CI, and you will see the results in the `Test PR...` GitHub action.
+1. Check your tests pass locally by running ``python -m pytest deepinv/tests`` in the root directory after making the desired changes. Learn more `here <https://realpython.com/pytest-python-testing/>`__. You can also run specific tests by providing the path to the test file, e.g. ``python -m pytest deepinv/tests/test_physics.py``, or even to a specific test function, e.g. ``python -m pytest -k "test_operators_adjointness" deepinv/tests/test_physics.py``.
```
jscanvic: Why change the item numbering?

docs/source/contributing.rst
```diff
@@ -220,6 +220,28 @@ Below is a minimal working example of a typical docstring that includes all thes
         def __init__(self, in_channels: int, out_channels: int, pretrained: bool = None):
             pass
 
+Contributing new physics
+-------------------------
+
+Providing new physics is one of the most impactful contributions you can make to DeepInverse, as it will allow users to solve new inverse problems. The process for contributing a new physics is similar to what is described above with some specificities:
```
jscanvic: Thanks for making the effort to document that!

I think the section can be made a bit terser an clearer. Can you polish it?

For instance, some comments:
- It's a bit repetitive ("new" repeated 6 times, "new physics" 4 times)
- it will allow => it allows?
- The justification in the 1st sentence might not be very convincing and might weaken the premise
- "The process for contributing a new physics" might be verbose
- "what is described above" it's not clear that it refers to contributing datasets and people might not have read that section first neither
- same for "in the previous section"
- The verb forms are inconsistent in "registering your physics ... and check that ..."
- Some of it sounds conversational to me (you/your repeated 8 times)

It's already really good as is so happy to go with whatever changes you end up making!

examples/basics/demo_custom_physics.py
```diff
@@ -48,6 +49,12 @@
 # .. note::
 #     To make the new physics compatible with other torch functionalities, all physics parameters (i.e. attributes of type :class:`torch.Tensor`) should be registered as `module buffers <https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_buffer>`_ by using `self.register_buffer(param_name, param_tensor)`. This ensures methods like `.to(), .cuda()` work properly, allowing one to train a model using Distributed Data Parallel.
 #
+# .. important::
+#     As per :class:`torch.nn.Module` functionalities, it is expected that all learnable parameters and registered buffers of the physics are on the same device and have the same data type. To make the initialization of the physics more flexible and efficient, you may want to pass a ``device`` and ``dtype`` argument to the physics constructor, and use them to initialize the physics parameters/buffers. It is important that these arguments are not set as instance variables, e.g. ``self.device = device``, since ``self.device`` is not automatically updated when calling ``physics.to(device)``.
```
jscanvic, line 53: Nice!

examples/basics/demo_custom_physics.py
```diff
@@ -48,6 +49,12 @@
...
+#
+# .. note::
+#     Below, when calling ``super().update_parameters(coefficients=coefficients, **kwargs)`` in ``Decolorize.A``, the ``super().update_parameters`` method takes care of pushing the incoming parameters to the correct device and data type. For efficiency, it is nonetheless recommended to initialize the incoming ``coefficients`` parameters on the correct device, to avoid unnecessary data transfers.
```
jscanvic, line 56: Nice!

docs/source/contributing.rst
romainvo: Thanks for the comments! I've pushed a new version, feel free to modify if need be :) :

jscanvic:
Ty! LGTM I'm waiting for CI to finish and I'll approve

PR #943, opened 2025-11-19
Workflows

mh-nguyen712:
This PRs change `pip` to `uv` in all workflows. Hopefully this will reduce the cache size.

It's also faster for installing dependencies (eg 55s comparing to 2m45s in the docs workflow)

### Checks to be done before submitting your PR

- [ ] `python3 -m pytest deepinv/tests` runs successfully.
- [ ] `black .` runs successfully.
- [ ] `make html` runs successfully (in the `docs/` directory).
- [ ] Updated docstrings related to the changes (as applicable).
- [ ] Added an entry to the [changelog.rst](https://github.com/deepinv/deepinv/blob/main/docs/source/changelog.rst).

PR #991, opened 2025-12-12
fix auto-release

tachella:
I'm reverting the changes applied to the auto pypi release workflow (updated to `uv`) since they broke the workflows and the workflows were already very fast (so no need for `uv` here).

The files should match those present before #943 


jscanvic (approved):
LGTM

Issue #533, opened 2025-06-13
Composed physics docs


Andrewwango:
We should have some docs explaining the benefits and limitations of `ComposedPhysics` proposed in #462 :

- Advertising that physics (physics, linear physics) are composable (although not decomposable physics)
- Composition is not only useful for composing with other physics, but the general case where you want to add a function to A and A_adjoint (i.e. `physics2 = physics * LinearPhysics(A=..., A_adjoint=...)`)
- Potential common pitfalls for users when composing (see below)

Furthermore, we should consider a way of "hijacking" physics (i.e. modifying `A` and `A_adjoint`) rather than using composition, leading to two benefits:

- Currently, using `ComposedPhysics`, for example, let's say a user wants to add a premultiplier to their inpainting physics so they do `physics_new = physics * LinearPhysics(A=lambda x: x*3)`. I think it would be reasonable that the user would expect `torch.all(physics_new.mask == physics.mask) == True` . However, in the current implementation this will not be true as whenever you compose, you lose the physics attributes, which is expected for general composition but not for this simple "hijacking" case. 
- Currently, if composing a `DecomposablePhysics` with something else, the resultant physics will lose decomposability (i.e. lose closed-form prox and dagger). This is expected in the general case, but maybe not desirable for everyone (i.e. #489).

Final optional edge case: composing a `DecomposablePhysics` and a `DecomposablePhysics` where it is valid e.g. `Inpainting `+ circular `Shift`.

Add subissue #489

Andrewwango:
In #524, we are using a different method of explicitly defining a wrapper class for the `Pad` and the `MultiScalePhysics`:

```python
class Wrapper(LinearPhysics):
    def __init__(self, base: Physics):
    def A(self, x):
        return self.base.A(hijack(x))
    def A_adjoint(self, y):
        return hijack_adjoint(self.base.A_adjoint(y))
```

It would be good to unify a) this method, b) composed physics provided in #462, and c) the physics method hijacking proposed in #489 . I am generally scared that unifying them now will subtly break things, as observed in #489. They should:
- Solve issues mentioned in this issue
- Be a separate feature to composed physics?
- Thoroughly tested: namely prox and A_dagger remain closed form, A_A_adjoint remains correct...

Andrewwango:
Interesting and possibly a solution: https://python-patterns.guide/gang-of-four/decorator-pattern/

Andrewwango:
very partially fixed by #1000 (just adds a small example to the user guide)

PR #1000, opened 2025-12-17
ComposedPhysics class and related are not in the published doc

romainvo:
`ComposedPhysics` and related methods are not present in the online doc:

composed_physics_doc

I've updated the necessary .rst files. I've also exposed `ComposedPhysics`, `ComposedLinearPhysics` and `compose` to `deepinv.physics`.

After the update:
composed_physics_doc_after


deepinv/physics/forward.py
```diff
@@ -817,7 +817,7 @@ class ComposedPhysics(Physics):
 
     where :math:`A_i(\cdot)` is the ith physics operator and :math:`N_k(\cdot)` is the noise of the last operator.
 
-    :param list[deepinv.physics.Physics] *physics: list of physics to compose.
+    :param list[deepinv.physics.Physics] physics: list of physics to compose.
```
Andrewwango: it's not a list - it's arbitrary number of physics objects.

deepinv/physics/forward.py
```diff
@@ -889,7 +889,7 @@ class ComposedLinearPhysics(ComposedPhysics, LinearPhysics):
 
     where :math:`A_i(\cdot)` is the i-th physics operator and :math:`N_k(\cdot)` is the noise of the last operator.
 
-    :param list[deepinv.physics.Physics] *physics: list of physics operators to compose.
+    :param list[deepinv.physics.Physics] physics: list of physics operators to compose.
```
Andrewwango: likewise

Andrewwango:
note this is mentioned in #533

deepinv/physics/forward.py
romainvo: I removed it because sphinx was raising an error while building the doc, like "the beginning of emphasis not ended" or something like that. I copy-pasted the format from the `compose` function below 

https://github.com/deepinv/deepinv/blob/114d1efc640ba8e2cda554b5d2f74b8e28843c22/deepinv/physics/forward.py#L914-L931

I don't know what's the intended format in the docstring for *args.. I'll look into it quick

romainvo: Not super clear to me how we should document *args and **kwargs (https://stackoverflow.com/questions/1137161/what-is-the-correct-way-to-document-a-kwargs-parameter)

If we align on `update_parameters` functions
https://github.com/deepinv/deepinv/blob/114d1efc640ba8e2cda554b5d2f74b8e28843c22/deepinv/physics/forward.py#L248-L254

what do you think of this ? (with `from collections.abc import Iterable`)
```python
    r"""
    :param Iterable[deepinv.physics.Physics] physics: variable number of physics to compose.
    """

    def __init__(self, *physics: Iterable[Physics], device: str | torch.device = "cpu", **kwargs):
```

PR #1001, opened 2025-12-18
Fix Zero prior

vsechaud:
A small fix for the `Zero` class in prior, where the method `fn` was returning a tensor with the wrong dimension.
I took the opportunity to rename it `ZeroPrior` so that it is consistent with the other priors (e.g., `L1Prior`, `WaveletPrior`, `TVPrior`, etc.). I am also adding it to the documentation and test files.

### DIFF
--- a/deepinv/optim/prior.py
+++ b/deepinv/optim/prior.py
@@ -60,16 +60,16 @@ class ZeroPrior(Prior):
         Computes the gradient of the zero prior :math:`\reg{x} = 0` at :math:`x`.
 
         :param torch.Tensor x: Variable :math:`x` at which the prior is computed.
-        :return: (:class:`torch.Tensor`) zero tensor.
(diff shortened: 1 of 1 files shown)


vsechaud:
I had to adjust the maximum number of iterations for the test with FISTA + `ZeroPrior`, which takes longer to converge.

deepinv/optim/prior.py
```diff
@@ -37,41 +37,39 @@ def __init__(self, g=None, *args, **kwargs):
...
 
 
-class Zero(Prior):
+class ZeroPrior(Prior):
     r"""
     Zero prior :math:`\reg{x} = 0`.
     """
 
     def __init__(self):
         super().__init__()
```
samuro95, line 46: Thanks ! it was indeed to much these lines

Issue #601, opened 2025-07-11
Trainer profiling


Andrewwango:
It would be good to have trainer profiling for its individual methods, like [lightning](https://lightning.ai/docs/pytorch/stable/tuning/profiler_basic.html) does, for example using `torch.profiler`.

E,g,:
1. setup the profiler:
```python
            prof = profile(activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA],
                           with_stack=True)
                           # schedule=schedule(wait=1, warmup=1, active=8, repeat=1),
                           # on_trace_ready=tensorboard_trace_handler(f"./profile/{self.wandb_setup['name']}"),
                           # profile_memory=True)
```
3. `prof.start()`
4. `prof.step()`
5. `prof.stop()`
6. `prof.export_chrome_trace("trace.json")`
7. then you can read the trace with https://ui.perfetto.dev

otherwise use torch cuda event

Andrewwango:
Should be closed by https://github.com/deepinv/deepinv/discussions/700#discussioncomment-14218650

PR #1002, opened 2025-12-18
Add benchmarks

tachella:
Draft towards adding benchmarks

This PR automatically adds benchmarks to the docs, modifying sphinx-build to gather the benchmark info and create the rst pages and tables automatically.

Additionally, the PR adds timing to `deepinv.test`, which now adds the `runtime` and `runtime_std` fields to the output dictionary. Solves a sub-issue of https://github.com/deepinv/deepinv/issues/601

Solvers and datasets should be placed on repo https://github.com/deepinv/benchmarks generated using, which should be organized as

```
.
├── benchmark1/
│   ├── solver1.py
│   ├── solver2.py
│   ├── ...
├── benchmark2/
│   ├── solver1.py
│   ├── solver2.py
│   ├── ...
├── ...
└── benchmarkN/
    ├── solver1.py
    ├── solver2.py
    ├── ...
```

Benchmarks are read from `result.parquet` files on the [huggingface dataset](https://huggingface.co/datasets/deepinv/benchmarks/tree/main), and a leaderboard table is added automatically to sphinx

@tomMoral 

### Benchmarks list
image

### Benchmark info
image


deepinv/training/trainer.py
```diff
@@ -1606,6 +1622,10 @@ def test(
             if self.verbose:
                 print(f"{name}: {l.avg:.3f} +- {l.std:.3f}")
 
+        # add runtime info
+        out["runtime"] = self.timing_meter.avg
+        out["runtime_std"] = self.timing_meter.std
+
```
Andrewwango, line 1620: Could we very simply add a CUDA peak memory check too when device is not cpu?

deepinv/training/trainer.py
```diff
@@ -813,6 +819,7 @@ def model_inference(self, y, physics, x=None, train=True, **kwargs):
         if self._model_accepts_update_parameters:
             kwargs["update_parameters"] = True
 
+        start_time = time.time()
```
Andrewwango: time.time is meaningless for cuda, and not the most accurate even on CPU. If here we are to support CUDA then we would have to use cuda Events + cuda synchronise after the warmup + at the end, so we measure the whole execution time rather than just kernel launch time. On CPU it would be better to use time perf counter.

docs/source/_static/custom.css
```diff
@@ -79,4 +79,26 @@ img.sphx-glr-single-img {
     max-width: 100%;
     height: auto;
     padding: 5px;
-}
\ No newline at end of file
+}
+
+/* Ensure the header is clickable and shows pointer */
```
Andrewwango: remove comments

docs/source/_static/main.js
```diff
@@ -0,0 +1,12 @@
+$(document).ready(function () {
+    // Check if the table exists and isn't already initialized
```
Andrewwango: same

docs/source/_templates/myclass_template.rst
```diff
@@ -12,3 +12,10 @@
 
 .. minigallery:: {{module}}.{{objname}}
     :add-heading: Examples using ``{{objname}}``:
+
+{% if benchmark_mapping and objname in benchmark_mapping %}
```
Andrewwango: super cool! does this work ? can you send me a screenshot?

docs/source/conf.py
```diff
@@ -167,9 +170,22 @@ def _noindex_viewcode(app, pagename, templatename, context, doctree):
         )
 
 
+class ShortClassRole(PyXRefRole):
```
Andrewwango, line 185: What is going on here? The PyXRefRole docs on sphinx don't seem to exist...

docs/source/conf.py
```diff
@@ -399,7 +415,19 @@ def __call__(self, filename):
 html_theme = "pydata_sphinx_theme"
 html_favicon = "figures/logo.ico"
 html_static_path = ["_static"]
-html_css_files = ["custom.css"]
+html_js_files = [
+    # Load jQuery first
```
Andrewwango: remove comments

docs/source/conf.py
```diff
@@ -399,7 +415,19 @@ def __call__(self, filename):
...
+    "https://code.jquery.com/jquery-3.7.1.min.js",
+    # Load DataTables second
+    "https://cdn.datatables.net/1.13.7/js/jquery.dataTables.min.js",
+    # Load your script last
```
Andrewwango: remove bloat

docs/source/generate_benchmarks.py
```diff
@@ -0,0 +1,322 @@
...
+    for folder, parquet_file in parquet_files:
+        dataset, physics, noise, benchmark_name = generate_rst_from_parquet(
+            parquet_file, output_dir
+        )
+        benchmark_info.append((benchmark_name, dataset, physics, noise))
+
+        # We save mappings for classes used in dataset, physics, and noise
+        # to later be included in the docstrings of these classes (see source/_templates/myclass_template.rst)
+        if dataset not in benchmark_mapping:
+            benchmark_mapping[dataset] = [benchmark_name]
```
Andrewwango, line 60: hmmmm surely you need to set the var as global in the func def? Why does this need to be global? It's not very pythonic...

docs/source/generate_benchmarks.py
```diff
@@ -0,0 +1,322 @@
...
+
+
+
+.. dropdown:: Benchmark information 
+    :chevron: down-up
+
+    """,
+    ]
+
+    lines.append(f"    - *Dataset*: :sclass:`deepinv.datasets.{dataset}` \n")
```
Andrewwango: Wait - so we assume that every benchmark added to the benchmarks is a deepinv dataset?

If so, this clearly needs to be added to the benchmarks instructions for new contributions: if someone wants to contribute a benchmark, both the dataset and the physics need to be approved and merged into deepinv main first.

This makes sense: maintainers should have the opportunity to screen benchmark quality by reviewing dataset and physics implementation.

docs/source/generate_benchmarks.py
```diff
@@ -0,0 +1,322 @@
...
+    benchmarks_content = """.. _benchmarks:
+
+Benchmarks
+=================
+
+This section provides benchmark results for various datasets and physics models.
+
+.. note::
+
+    Benchmarks are defined on the https://github.com/deepinv/benchmarks repository.
```
Andrewwango:
```suggestion
    Benchmarks are defined in the https://github.com/deepinv/benchmarks repository.
```

docs/source/generate_benchmarks.py
```diff
@@ -0,0 +1,322 @@
...
+    To contribute a new benchmark or add your solver to an existing benchmark, please refer to this repository.
+
+You can try your solver on one of the benchmarks, make sure that it receives `(y, physics)` as input and outputs the reconstructed image as `x`,
+then install (or upgrade) the `deepinv_bench` package and run the benchmark as follows:
```
Andrewwango:
```suggestion
then install (or upgrade) the `deepinv_bench` package as follows:
```

docs/source/generate_benchmarks.py
```diff
@@ -0,0 +1,322 @@
...
+=================
+
+This section provides benchmark results for various datasets and physics models.
+
+.. note::
+
+    Benchmarks are defined on the https://github.com/deepinv/benchmarks repository.
+    To contribute a new benchmark or add your solver to an existing benchmark, please refer to this repository.
+
+You can try your solver on one of the benchmarks, make sure that it receives `(y, physics)` as input and outputs the reconstructed image as `x`,
```
Andrewwango:
```suggestion
You can try your own solver on any benchmark of your choice. Make sure that it receives `(y: torch.Tensor, physics: dinv.physics.Physics)` as input and outputs the reconstructed image as `x: torch.Tensor`,
```
link to Reconstructor user guide

Andrewwango (changes requested):
exciting so far. One big concern for me is the text on the deepinv benchmark repo README, which is important to present the rules of the benchmarks + how people can contribute. How can I review that?

docs/source/_templates/myclass_template.rst
tachella: it does work :)

image

deepinv/training/trainer.py
tachella: good idea - I've added it

deepinv/training/trainer.py
tachella: I've simplified the timing and kept a global timer over the whole test with perf_counter only. I think timing at the single model inference level requires synchronizing and could slow down testing

docs/source/_static/custom.css
tachella: done

docs/source/_static/main.js
tachella: done

docs/source/conf.py
tachella: It does exist, why do you ask?

docs/source/conf.py
tachella: done

docs/source/conf.py
tachella: done

docs/source/generate_benchmarks.py
tachella: I didn't find any other way to make it work with sphinx..

docs/source/generate_benchmarks.py
tachella: I will add this to the instructions in the benchmarks repo

tachella:
> exciting so far. One big concern for me is the text on the deepinv benchmark repo README, which is important to present the rules of the benchmarks + how people can contribute. How can I review that?

Good point. I can open a PR on that repo?

jscanvic (commented):
Thanks Julián! Just a minor comment, I'm not fond of letting jQuery creep in and I'm pretty confident I can vibe code 20 lines of vanilla JavaScript to do the same thing the current jQuery+DataTables solution does. Happy to have a go at it if you're happy with that :)

tachella:
> Thanks Julián! Just a minor comment, I'm not fond of letting jQuery creep in and I'm pretty confident I can vibe code 20 lines of vanilla JavaScript to do the same thing the current jQuery+DataTables solution does. Happy to have a go at it if you're happy with that :)

If you can find an alternative solution that works it would be great 😃

deepinv/training/trainer.py
```diff
@@ -1543,6 +1543,7 @@ def test(
 
         batches = min([len(loader) - loader.drop_last for loader in test_dataloader])
 
+        perf_counter_start = time.perf_counter()
```
Tmodrzyk, line 1556: We should use `torch.cuda.synchronize()` here before setting each of the start and end time.

deepinv/training/trainer.py
```diff
@@ -1591,4 +1596,18 @@ def test(
             if self.verbose:
                 print(f"{name}: {l.avg:.3f} +- {l.std:.3f}")
 
+        perf_counter_end = time.perf_counter()
+        elapsed_time = perf_counter_end - perf_counter_start
+        avg_time_per_sample = elapsed_time / (batches * test_dataloader[0].batch_size)
```
Tmodrzyk: Are we really measuring the correct thing here ? 
This gives us the average duration of a step.
But I believe what we want to measure here is avg inference time for one sample. 
We should definitely remove from the time counter all the things that are not related to the model: loss / metric computations, generating measurements, prints, etc...

deepinv/training/testing.py
```diff
@@ -53,7 +53,7 @@ def test(
         ``'A_adjoint'``, ``'prox_l2'``, or ``'y'``. Default is ``'A_adjoint'``. The user can modify the no-learning method
         by overwriting the :func:`no_learning_inference <deepinv.Trainer.no_learning_inference>` method
     :returns: A dictionary with the metrics computed on the test set, where the keys are the metric names, and include
-        the average and standard deviation of the metric.
+        the average and standard deviation of the metric, timing and peak GPU memory usage information. Timings correspond to average inference time per batch in seconds.
```
Tmodrzyk: This is not correctly describing what is currently being measured. We can either keep the current timings and change this description, or measure exactly that.

docs/source/conf.py
```diff
@@ -391,13 +407,19 @@ def __call__(self, filename):
...
-
 html_theme = "pydata_sphinx_theme"
 html_favicon = "figures/logo.ico"
 html_static_path = ["_static"]
-html_css_files = ["custom.css"]
+html_js_files = [
+    "https://code.jquery.com/jquery-3.7.1.min.js",
+    "https://cdn.datatables.net/1.13.7/js/jquery.dataTables.min.js",
+    "main.js",
+]
```
Tmodrzyk: We should avoid relying on external js libraries for such a small feature (displaying the benchmark tables with an arrow to order them ?)
It seems dataTables is the most straightforward solution but it forces us to us jQuery...
We should at least delay loading these libs on the benchmark pages, and not for the whole doc.

docs/source/generate_benchmarks.py
```diff
@@ -0,0 +1,385 @@
...
+
+        if noise not in benchmark_mapping:
+            benchmark_mapping[noise] = [benchmark_name]
+        else:
+            benchmark_mapping[noise].append(benchmark_name)
+
+    generate_main_rst(benchmark_info, source_dir)
+
+
+def process_csv_file(csv_path):
```
Tmodrzyk: We should have more specific naming, ~~for instance `rst_from_csv`~~ `strings_from_csv` ?

deepinv/training/trainer.py
```diff
@@ -1570,6 +1571,10 @@ def test(
         if self.verbose:
             print("Test results:")
 
+        # reset peak GPU memory usage counter
+        if torch.cuda.is_available():
+            torch.cuda.reset_peak_memory_stats()
```
Tmodrzyk: I believe that doing this will wipe out the memory usage reported later on with `max_memory_allocated`, so we don't measure the allocated memory in the steps above, but in the whole section below this code.

docs/source/_templates/myclass_template.rst
```diff
@@ -13,3 +13,9 @@
 .. minigallery:: {{module}}.{{objname}}
     :add-heading: Examples using ``{{objname}}``:
 
+{% if benchmark_mapping and objname in benchmark_mapping %}
+Find in benchmarks
+------------------
+
+- :ref:`{{ benchmark_mapping[objname] }}
```
Tmodrzyk:
```suggestion
- :ref:`{{ benchmark_mapping[objname] }}`
```

Tmodrzyk (changes requested):
Thanks for the PR @tachella. The main mechanisms to create rst files to display from the benchmark cvs files look ok.
However, the way we currently measure both the time and the memory usage looks incorrect, and will systematically wildly over-estimate time and under-estimate memory usage.
We need to fix that before merging.

Otherwise I have the same concerns as @jscanvic for the javascript parts. I get that sphinx does not support sortable tables by default, but at the same time we should stay away from anything related to js. 
We also load those libraries globally while I believe we could delay them to load them only for the benchmark pages.

deepinv/training/trainer.py
tachella: I agree that this is not very exact. Before I had done something more exact inside model_inference, but then we would add annoying overhead to every test call. I think for the sake of benchmarks, we just need the total test time, so I'd remove the 'per sample' and just report the total test time which includes everything in the loop, but it will include it for all compared methods.

deepinv/training/trainer.py
tachella: you are right, it was in the wrong place!

deepinv/training/testing.py
tachella: you're right, I fixed (see my other comment), and now it just report total test time

docs/source/_templates/myclass_template.rst
tachella: thanks for pointing out this. Actually this wasn't really being used (the if was never activated), asthe printing of benchmarks into the class docstrings is done by `add_benchmark_section` in `generate_benchmarks.py`, so I ended up removing these whole block of code

deepinv/training/trainer.py
tachella: I'm not sure if the synchronize will do much to the final test time, especially in test datasets with many images. I'm afraid that adding it might add unnecessary overhead to all testing in deepinv

docs/source/conf.py
tachella: I removed the .js dependency by adding a standalone much lighter table sorter (with the help of LLMs). Let me know if you think it's okay now

deepinv/training/trainer.py
Tmodrzyk: I agree that this will add overhead. These measure sections should be guarded by a condition so that measure and sync take place only when we are benchmarking for speed.
If we don't use `torch.cuda.synchronize()` we do not measure the actual runtime, and the times output by the benchmark will be useless.

deepinv/training/trainer.py
Tmodrzyk: Thanks for the detailed answer @tachella. Again, measuring stuff clearly adds overhead, but we should not be measuring it all the time, only for benchmarks. 
The issue here is that if a random disk access takes longer than usual, which does happen, the model will be penalized for it. I'm not confident with measuring performances with this formula, this is too far from what we actually want to measure.

tachella:
@Tmodrzyk @Andrewwango 

I implemented all the suggested changes - can you please have a look again at the PR?

Many thanks!!

deepinv/training/trainer.py
```diff
@@ -1507,7 +1508,7 @@ def test(
         :param bool log_raw_metrics: if `True`, also return non-aggregated metrics as a list.
         :param Metric, list[Metric], None metrics: Metric or list of metrics used for evaluation. If
             ``None``, uses the metrics provided during Trainer initialization.
-        :returns: dict of metrics results with means and stds.
+        :returns: dict of metrics, timings (in sec) and peak GPU memory usage (in GBy) results with means and stds. Timings correspond to total test time.
```
Andrewwango: GBy?

deepinv/training/trainer.py
Andrewwango: I agree with @Tmodrzyk that we need at least two cuda synchronizes, before the testing loop and after it, but not during. This will not slow down testing significantly, it only happens twice, to measure the average inference time. The cuda times are meaningless otherwise. Typically, users don't use `.test` in production scenarios, so they don't mind a small gap from the previous command to test and from test to the next command. Note it shouldn't affect the timing of test itself.

deepinv/training/trainer.py
```suggestion
        :returns: dict of metrics, timings (in sec) and peak GPU memory usage (in GBy) results with means and stds. Note: timings correspond to total time for `test` to run, which also includes time to load data and compute metrics. Therefore, the reported runtime will be greater than just model inference timings.
```

deepinv/training/trainer.py
Andrewwango: I agree with @Tmodrzyk, but I also am happy with @tachella 's latest code which calls it "runtime" which is evidently not just model inference time anymore. Whether runtime is useful is a different story. Actually measuring model inference speed is also extremely important but is an orthogonal issue. I'm reopening #601; we can continue discussion there

Andrewwango (changes requested):
thanks Julian, almost there, just some comments on timing

Tmodrzyk:
Hi @tachella thanks for taking the feedback into account ! I'll give another round of review after @Andrewwango's suggestions are processed. :smile:

deepinv/training/trainer.py
tachella: moved to GB

tachella:
Thanks to both! I just addressed @Andrewwango's comments.

I've added the synchronize to limit any potential biases

Andrewwango (approved):
LGTM @Tmodrzyk

Tmodrzyk (approved):
LGTM, rendered doc looks incredible, congrats !

tomMoral:
Great that this was merged!
Is it live somewhere in the deepinverse doc? (Could not find it)

> Otherwise I have the same concerns as @jscanvic for the javascript parts. I get that sphinx does not support sortable tables by default, but at the same time we should stay away from anything related to js. 

Benchopt reports now generate sortable tables using a Js framework, might be interesting to see if this could be reused directly so we don't maintain two versions based on the same code. (Later on)

jscanvic:
@tomMoral It's now live at https://deepinv.org/auto_benchmarks/benchmarks.html

PR #1007, opened 2025-12-21
Make CI fail on deprecation warnings in tests

jscanvic:
Our own code should not trigger deprecation warnings yet it does - I suggest to remedy that


Andrewwango:
Looks good, but are we fixing the CI in this too then?

PR #1014, opened 2026-01-04
Enable ruff F821 F822 F823

nucli-vicky:
This PR adds rules F821-823:

- F821: undefined name not allowed (*very useful for catching bugs, see for instance the missing warnings import*)
- F822: everything in `__all__`  should exist. Critical, but also seems a bit too evident, might not be worth it to have it as a rule? Fine with removing it.
- F823: don't allow "variable used before assignment" --> also good for catching bugs.

I think adding these rules makes sense and does not significantly raise the bar for contributions, while still being good for preventing bugs, which is the philosophy agreed upon in #539.  In recent months, a few bugs in main were already identified manually and fixed, which would have been caught by these rules too, e.g. part of the small fixes in #981 , #955, so these rules do correspond with real bugs which occur during development, not some hypothetical scenarios (note also the change in `optim_iterator` fixes a live bug (albeit small).

### DIFF
--- a/deepinv/datasets/bsds500.py
+++ b/deepinv/datasets/bsds500.py
@@ -69,7 +69,7 @@ class BSDS500(ImageDataset):
             )
             download_sum = calculate_md5(fpath=zip_path)
             if not download_sum == checksum:
-                return ValueError(
+                raise ValueError(
                     "Verification of the dataset failed (unexpected md5 checksum of the downloaded zip-file)"
                 )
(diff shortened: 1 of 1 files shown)


nucli-vicky:
Just realized enabling all the `F` rules is already worked on in #705. However, as already mentioned by @jscanvic , it does make sense to split these up into smaller PRs, and I believe these rules have high signal; anything caught should be an actual bug that can raise errors.

PR #1011, opened 2025-12-29
refactor least squares

tachella:
- add a standalone section to the user guide for pseudoinverse reconstruction
- refactor least squares solvers from a single `utils.py` file to one file per solver, with a subfolder inside optim. The refactoring should be backwards compatible, since I kept 
```
.
└── deepinv/
    └── optim/
        └── linear/
            ├── lsqr.py
            ├── minres.py
            └── ...
```
Solvers can now be imported as `deepinv.optim.linear.solverxx`
- following #992, included a test checking the behaviour of linear least squares solvers where the input is exactly zero
 

- I tried removing the `eps` parameter in `conjugate_gradient`, following the suggestions in #992, only optimizing dimensions that haven't converged. However, the stability was lower than the current version, and many tests were not passing, so I kept the current version with some minor improvements (handle b=0 case + recompute residual every 50 iterations for stability).

### Checks to be done before submitting your PR

- [x] `python3 -m pytest deepinv/tests` runs successfully.
- [x] `black .` runs successfully.
- [x] `make html` runs successfully (in the `docs/` directory).
- [x] Updated docstrings related to the changes (as applicable).
- [x] Added an entry to the [changelog.rst](https://github.com/deepinv/deepinv/blob/main/docs/source/changelog.rst).

PR #1016, opened 2026-01-07
HOTFIX func ref

mh-nguyen712:
The recent PR #1011 refactor the least_squares function but forgot to update the reference ```:func:`...least_squares` ```, which breaks the CI in other PRs, eg #790.

This PR updates these func refs.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -104,11 +104,6 @@ jobs:
           echo "Running Sphinx with filename_pattern: $SG_PATTERN"
           uv run sphinx-build -W docs/source _build -D "sphinx_gallery_conf.filename_pattern=$SG_PATTERN"
 
-      - name: Build html documentation without examples
-        run: |
-          echo "Building full Sphinx documentation without examples"
-          uv run sphinx-build -W docs/source _build -D plot_gallery=0 -b html
-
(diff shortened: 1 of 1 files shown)


tachella:
thanks Hai for fixing this! I don't get why the tests passed on my PR but then are not passing on others.. It is weird that the wrong links did not raise warnings?

mh-nguyen712:
Hi, I think in the PR, the workflow only runs the modified examples. Since in your PR, you did not chang any examples so they are not checked.

tachella:
> Hi, I think in the PR, the workflow only runs the modified examples. Since in your PR, you did not chang any examples so they are not checked.

but most of the errors are outside the examples - normally we should be building the whole docs except the examples...

mh-nguyen712:
> > Hi, I think in the PR, the workflow only runs the modified examples. Since in your PR, you did not chang any examples so they are not checked.
> 
> but most of the errors are outside the examples - normally we should be building the whole docs except the examples...

Hmm you're right. As can be seen here https://github.com/deepinv/deepinv/actions/runs/20655821162/job/59308461197 it was passing, but I dont know why.

Anyway, I think these fixes are necessary, maybe I should remove the make html-fast to avoid redundancy

tachella (approved):

PR #1017, opened 2026-01-09
Integrate MAPz/VBLE

MaudBqrd:
## Summary
This PR adds **latent optimization inverse problem solvers** (CSGM, MAP-z, VBLE) to DeepInverse. It implements two neural networks architectures (VAE, CAE), a latent optimizer class, and a posterior sampling class for VBLE sampling.

## Motivation

As discussed with @tachella, I propose adding "latent optimization" methods into deepinv, namely:
- CSGM [1]
- MAP-z [1]
- VBLE [2,3]

These methods solve inverse problems by optimizing/sampling in the latent space of a latent variable model.

### Neural networks architectures

Two neural network architectures have been added:

- A simple VAE (in `models/vae.py`)
- A compressive autoencoder (CAE) (in `models/cae.py`)

### Latent optimization reconstructor

In `models/latent_optimizer.py`, the `LatentOptimizer` class (based on `Reconstructor` class) handles the latent optimization, with the forward function returning one solution to the inverse problem. The implemented algorithm is compatible with VAEs, CAEs and also `DCGANGenerator` from `models/gan.py`.

### Posterior sampling with VBLE

For the posterior sampling method VBLE, the `VBLESampling` class (in `sampling/vble.py`) enables to (batch) sample several posterior samples and compute posterior statistics. 

## Demos

Two demos are provided:

- `examples/sampling/demo_sampling_vble.py`: Deblurring with VBLE
- `examples/models/demo_training_cae.py`: Training / finetuning a CAE model  (as this is a bit specific, not sure about the relevance of its integration).

The CAE training demo is mainly included for reproducibility and experimentation. I am not yet sure whether it should be part of the core library.

## Remarks

- In the current version, the pretrained VAE and CAE models are downloaded from my Hugging Face repository. This should probably change in the final version?
- I added docstrings to all functions and classes, but I have not yet:
  - Integrated content into `docs/`
  - Added unit tests

I can add both if the general structure and relevance of the contribution are validated.

## References 

[1] [CSGM/MAP-z paper](https://arxiv.org/pdf/1703.03208): Bora et al., Compressed Sensing using Generative Models, ICML, 2017
[2] [VBLE-z paper](https://arxiv.org/pdf/2311.17744): Biquard et al., Variational Bayes image restoration with compressive autoencoders, IEEE TIP, 2025
[3] [VBLE-xz paper](https://arxiv.org/pdf/2412.04130?): Biquard et al. Deep priors for satellite image restoration with accurate uncertainties, IEEE TGRS, 2025


deepinv/models/vae.py
```diff
@@ -0,0 +1,395 @@
...
+        img_size: int | None = None,
+        activation: callable = nn.LeakyReLU,
+    ):
+        super().__init__()
+
+        self.fully_conv_bottleneck = fully_conv_bottleneck
+
+        dim = 2
+        conv = conv_nd(dim)
+        batchnorm = batchnorm_nd(dim)
```
nucli-vicky: Since this structure is already in place, any reason why we don't allow specifying dim starting from the VAE and decoder architecture? Apart from architectural aspects (which would be solved this way), it doesn't seem there is code which relies on data being 2D right now?

deepinv/models/vae.py
```diff
@@ -0,0 +1,395 @@
...
+
+    def __init__(
+        self,
+        in_channels: int,
+        latent_dim: int,
+        filters: tuple[int] = (64, 64, 64, 64),
+        fully_conv_bottleneck: bool = False,
+        use_batchnorm: bool = False,
+        img_size: int | None = None,
+        activation: callable = nn.LeakyReLU,
```
nucli-vicky: lowercase `callable` is a function, not a type. 
```suggestion
        activation: Callable[[torch.Tensor], torch.Tensor] = nn.LeakyReLU,
```
or

```suggestion
        activation: nn.Module = nn.LeakyReLU,
```

deepinv/models/vae.py
MaudBqrd: There is no real reason :) I have added dim as a VAE input argument (and thus as Encoder and Decoder input argument). This only requires to change `n_pixels` (line 202) computation in `compute_kl_loss` to a generic version.

deepinv/models/vae.py
MaudBqrd: Done!

PR #1020, opened 2026-01-19
Add inverse scattering problem

tachella:
This PR adds the `deepinv.physics.Scattering` operator, which allows simulating measurements of a 2D non-linear scattering problem, with applications in microwave tomography and optical diffraction tomography among others. 

**Summary of contributions**

- A new `deepinv.physics.Scattering` operator
- A demo example showcasing the main features of the operator
- A generalization of the `compute_norm` method to non-linears physics
- A series of tests checking the correctness of the operator using closed form formulas for scattered waves (Mie Theory)

I've updated `black` which generated a bunch of new small style changes

**New operator description**

For each of the $i=1,\dots,T$ transmitters, the 2D forward model is given by the inhomogeneous Helmholtz equation (see e.g., Soubies et al. 2017):

$$
\nabla^2 u_i(\mathbf{r}) + k^2(\mathbf{r}) u_i(\mathbf{r}) = - (k^2(\mathbf{r}) - k_b^2) v_i(\mathbf{r}) \quad \mathbf{r} \in \mathbb{R}^2
$$

where
- $u_i$ is the (unknown) scattered field,
- $k_b$ is the (known scalar) wavenumber of the incident wave in the background medium,
- $k(\mathbf{r})$ is the (unknown) spatially-varying wavenumber of the object to be recovered,
- $v_i$ is the incident field generated by the ith transmitter in the absence of the object.

The total field (scattered + incident) is measured at \(R\) different receiver locations surrounding the object.

Parametrizing the unknown spatially-varying wavenumber as $k^2(\mathbf{r}) = k_b^2 (x(\mathbf{r})+1)$, where
- $k_b$ is the background wavenumber,
- $x = k^2/k_b^2 - 1$ is the scattering potential of the object to be recovered,

the problem can be reformulated in the **Lippmann-Schwinger** integral equation form:

$$
\begin{align*}
u_i &= g * \left(x \circ (u_i+v_i)\right) \\
y_i &= G_s \left(x \circ (u_i+v_i)\right)
\end{align*}
$$

where
- $g(\mathbf{r}) = k_b^2 \frac{i}{4} H_0^1(k_b\|\mathbf{r}\|)$ is Green's function in 2D (normalized by $k_b^2$),
- $y \in \mathbb{C}^{R}$ are the measurements at the receivers for the ith transmitter,
- $G_s$ denotes the convolution with Green's operator plus sampling at the \(R\) different receiver locations.

### DIFF
--- a/deepinv/models/kernel_network.py
+++ b/deepinv/models/kernel_network.py
@@ -40,6 +40,7 @@ class KernelIdentificationNetwork(nn.Module):
     >>> import torch
     >>> device = dinv.utils.get_freer_gpu() if torch.cuda.is_available() else "cpu"
     >>> kernel_estimator = dinv.models.KernelIdentificationNetwork(device=device)
+    ...
     >>> physics = dinv.physics.SpaceVaryingBlur(device=device, padding="constant")
     >>> y = torch.randn(1, 3, 128, 128).to(device)  # random blurry image for demonstration
     >>> with torch.no_grad():
--- a/deepinv/models/wrapper.py
+++ b/deepinv/models/wrapper.py
@@ -233,6 +233,7 @@ class ComplexDenoiserWrapper(Denoiser):
         >>> import torch
         >>> from deepinv.models import ComplexDenoiserWrapper, DRUNet
         >>> denoiser = DRUNet()
+        ...
         >>> complex_denoiser = ComplexDenoiserWrapper(denoiser, mode="real_imag")
         >>> y = torch.randn(2, 3, 32, 32, dtype=torch.complex64)  # complex input
         >>> sigma = 0.1


deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
+from .forward import Physics, LinearPhysics
```
Andrewwango: Full import path + sort imports

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+
+    where :math:`u_i` is the (unknown) scattered field, :math:`k_b` is the (known scalar) wavenumber of the incident wave in the background medium,
+    :math:`k(\mathbf{r})` is the (unknown) spatially-varying wavenumber of the object to be recovered,
+    and :math:`v_i` is the incident field generated by the ith transmitter in the absence of the object.
+    The total field (scattered + incident) is measured at :math:`R` different receiver locations surrounding the object.
+
+
+    Parametrizing the unknown spatially-varying wavenumber as :math:`k^2(\mathbf{r}) = k_b^2 (x(\mathbf{r})+1)`, where
+    :math:`k_b` is the background wavenumber, and :math:`x = k^2/k_b^2 - 1` is the scattering potential of the object to be recovered,
+    the problem can be reformulated in the **Lippmann-Schwinger** integral equation form:
```
Andrewwango, line 27: The forward problem

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+    .. math::
+
+        u_i &= g * \left( x \circ (u_i+v_i) \right) \\
+        y_i &= G_s \left( x \circ (u_i+v_i) \right)
+
+    where :math:`g(\mathbf{r}) = k_b^2 \frac{i}{4} H_0^1(k_b\|\mathbf{r}\|)` is Green's function in 2D (normalized by :math:`k_b^2`),
+    :math:`y \in \mathbb{C}^{R}` are the measurements at the receivers for the ith transmitter,
+    and :math:`G_s` denotes the convolution with Green's operator plus sampling at the :math:`R` different receiver locations.
+
+
```
Andrewwango, line 38: Small note in the context of inverse problems: the goal is to find the source given the scattering measurements

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+        If the objects' contrast and/or the wavenumber is very large, the linear solver may not converge, leading to inaccurate results.
+        Setting `verbose=True` in the constructor will print solver convergence information, which can help diagnose convergence issues.
+
+
+    .. note::
+
+        The forward operator uses the adjoint state method to compute gradients, e.g., :meth:`deepinv.physics.Scattering.A_jvp` and :meth:`deepinv.physics.Scattering.A_vjp` by default,
+        which is more memory efficient than using autograd, especially for large-scale problems. The adjoint state method requires
+        one additional solve of the Lippmann-Schwinger equation per gradient computation, and avoids storing the entire computational graph (done by standard autograd).
+        This can be disabled by setting `adjoint_state=False` in the :class:`deepinv.physics.Scattering.SolverConfig` object passed to the constructor,
```
Andrewwango:
```suggestion
        This can be disabled by setting `adjoint_state=False` in the :class:`deepinv.physics.Scattering.SolverConfig` object passed to the constructor
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+        .. math::
+
+                g_L(\mathbf{r}) = g(\mathbf{r}) * \text{rect}\left(\frac{\|\mathbf{r}\|}{L}\right),
+
+
+    :param int img_width: Number of pixels per image side (`H=W`). The minimum required number of pixels depends on the background wavenumber to avoid spatial aliasing.
+        We require at least :math:`\text{img_width} \geq 2 k_b L/(2 \pi)`, where :math:`k_b` is the background wavenumber and :math:`L` is the box length.
+    :param torch.Tensor receivers: Tensor of shape `(2, R)` (shared receivers) or `(2, T, R)` (per-transmitter receivers) with receiver x/y positions.
+    :param torch.Tensor transmitters: Tensor of shape `(2, T)` with transmitter x/y positions.
+    :param float, torch.Tensor background_wavenumber: background wavenumber :math:`k_b` (real or complex). It controls the frequency of the incident wave, and thus the maximum achievable resolution of the contrast image.
```
Andrewwango: How are complex numbers stored - in complex64 or two channel?

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+    :param deepinv.physics.scattering.Scattering.SolverConfig solver_config: Configuration for the Lippmann-Schwinger solver used to
+        compute the total field (first equation in the forward model). If `None`, default configuration is used.
```
Andrewwango: What is the default config

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+    :param float box_length: Physical side length of the square domain :math:`L`, which can be specified relative to the wavenumber.
+    :param str wave_type: incident wave type. Options are `'circular_wave'` or `'plane_wave'`.
+        The circular wave corresponds to a point source transmitter, while the plane wave corresponds to a far-field transmitter.
+    :param str device: Torch device string, e.g. `'cpu'` or `'cuda'`.
+    :param torch.dtype dtype: Torch `dtype` for tensors (e.g. `torch.complex128`).
+    :param bool verbose: Enable verbose/debug printing.
+    """
+
+    def __init__(
```
Andrewwango, line 94: Static typing for these please :)

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+    :param torch.Tensor transmitters: Tensor of shape `(2, T)` with transmitter x/y positions.
+    :param float, torch.Tensor background_wavenumber: background wavenumber :math:`k_b` (real or complex). It controls the frequency of the incident wave, and thus the maximum achievable resolution of the contrast image.
+    :param deepinv.physics.scattering.Scattering.SolverConfig solver_config: Configuration for the Lippmann-Schwinger solver used to
+        compute the total field (first equation in the forward model). If `None`, default configuration is used.
+    :param float box_length: Physical side length of the square domain :math:`L`, which can be specified relative to the wavenumber.
+    :param str wave_type: incident wave type. Options are `'circular_wave'` or `'plane_wave'`.
+        The circular wave corresponds to a point source transmitter, while the plane wave corresponds to a far-field transmitter.
+    :param str device: Torch device string, e.g. `'cpu'` or `'cuda'`.
+    :param torch.dtype dtype: Torch `dtype` for tensors (e.g. `torch.complex128`).
```
Andrewwango: Must be complex?

deepinv/physics/functional/special.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # 3. Fallback to SciPy (requires CPU transfer)
+    # Transfer to CPU, convert to numpy, compute, then back to Torch
+    try:
+        import scipy.special
+    except ImportError:
+        raise ImportError(
+            "SciPy or PyTorch version >= 2.6 is required for hankel1 computation."
+        )
+    out = scipy.special.hankel1(n, x.to("cpu"))
```
mh-nguyen712, line 62: Missing `.numpy()` then send it back to a `tensor` ?

deepinv/physics/functional/special.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # 2. Fallback to SciPy
+    try:
+        import scipy.special
+    except ImportError:
+        raise ImportError(
+            "SciPy or PyTorch version >= 2.6 is required for jv computation."
+        )
+    # We detach and move to CPU to avoid breaking the graph or moving the whole model
+    out = scipy.special.jv(n, x.to("cpu"))
```
mh-nguyen712, line 101: same here

deepinv/physics/functional/special.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    # Use torch.special functions for J and Y
+    jn = torch.special.bessel_j(n, x)
+    yn = torch.special.bessel_y(n, x)
+
+    # Combine into a complex tensor: J + iY
+    return torch.complex(jn, yn)
+
+
+def hankel1(n, x, device=None):
```
mh-nguyen712: I would remove the `device`, it should be the same as `x`

deepinv/physics/functional/special.py
```diff
@@ -0,0 +1,73 @@
...
+    except ImportError:
+        raise ImportError(
+            "SciPy or PyTorch version >= 2.6 is required for hankel1 computation."
+        )
+    out = scipy.special.hankel1(n, x.to("cpu"))
+
+    return out.to(device=device or x.device)
+
+
+def jv(n, x, device=None):
```
mh-nguyen712: same here for `device`, and I would prefer call it `bessel_j` as in torch rather than `jv`. but both is ok.

deepinv/physics/functional/special.py
```diff
@@ -0,0 +1,73 @@
+import torch
```
Andrewwango, line 1: Not a very descriptive file name

deepinv/physics/functional/special.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def native_hankel1(n, x):
```
Andrewwango: Docstring and types for input and output

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+        """
+        if transmitters is not None:
+            self.transmitters = transmitters.to(self.device).to(self.dtype)
+            # recompute the incident field
+            self.generate_incident_field()
+
+        if receivers is not None:
+            receivers = receivers.to(self.device).to(self.dtype)
+
+            # Handle receivers: convert (2, R) to (2, T, R) if needed
```
Andrewwango, line 242: Remove this logic from init

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+            # multiply distances by scalar wavenumber
+            dist_transmitter_circles = (
+                dist_transmitter_circles * self.wavenumber
+            )  # (T, H*W)
+            incident_field = green_function(dist_transmitter_circles)
+            incident_field = incident_field.reshape(
+                1, incident_field.size(0), self.img_width, self.img_width
+            )  # (1, T, H, W)
+
+        self.incident_field = incident_field.to(self.device).to(self.dtype)
```
Andrewwango: It would be more pythonic to return this and set the attribute in the update parameters

deepinv/physics/functional/matrix.py
```diff
@@ -0,0 +1,44 @@
+import torch
+import warnings
+
+
+def power_method(
```
mh-nguyen712, line 5: Cool. NOTE: we're starting to have many hiding functions everywhere. Eg this function in `physics.functional` and a similar one `least_squares` in `optim`. maybe we can reorganize a bit...

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+                img_width=self.img_width,
+                verbose=self.verbose,
+            )
+
+    def generate_incident_field(self):
+        """
+        Generate incident fields on the image grid and at receiver positions.
+
+        Produces:
+            - self.incident_field of shape (1, T, H, W)
```
Andrewwango: Whats T

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+
+        :param torch.dtype dtype: torch.dtype used for the generated fields
+        """
+        x_domain = self.x_domain.flatten()
+        y_domain = self.y_domain.flatten()
+        x_transmitters, y_transmitters = (
+            self.transmitters[0, :],
```
Andrewwango: Shouldn't need indexing here: should be able to just unpack directly from: x, y = self.transmitters

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+            self.transmitters[1, :],
+        )
+
+        if (
+            self.wave_type == "plane_wave"
+        ):  # u = exp(i*k*x) = exp(i*2*pi*(cos(theta)*x + sin(theta)*y))
```
Andrewwango: Comment on next like or in docstring

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+    ):
+        super(Scattering, self).__init__()
+        if wave_type not in wave_type in ["circular_wave", "plane_wave"]:
+            raise ValueError(
+                'Wave type not recognized, options are "circular_wave" or "plane_wave"'
+            )
+
+        # store a single scalar wavenumber (no wavenumber dimension)
+        if not isinstance(background_wavenumber, torch.Tensor):
+            self.wavenumber = torch.tensor(
```
mh-nguyen712: save it as a `buffer`

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+    def compute_total_field(self, x, **kwargs):
+        """
+        Solve for the total field given scattering potential x using the provided solver.
+
+        :param torch.Tensor x: Scattering potential tensor of shape (B,1,H,W).
+        :returns: (:class:`torch.Tensor`) Total field tensor of shape (B,T,H,W).
+        """
+        # This solves the following linear problem: (I - G_d*diag(x))u = v
+
+        if x.abs().sum() <= 1e-5:
```
Andrewwango: What if x is scaled low intensity?

deepinv/physics/scattering.py
mh-nguyen712, line 94: Can you add typing please?

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+            warn("The input x is zero")
+
+        k02 = self.wavenumber**2
+        k2 = k02 * (x + 1)
+        s = (k2 - k02) * self.incident_field
+        return self.solver(k2, s) + self.incident_field
+
+    def set_verbose(self, verbose):
```
Andrewwango: Why is this needed? Can we not just set them in the init?

If needed, it would be nicer to use property setter rather than a dedicated method

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+        self.update_parameters(receivers=self.receivers, transmitters=self.transmitters)
+
+    @dataclass
+    class SolverConfig:
+        r"""
+        Configuration class for the Lippmann-Schwinger solver.
+
+        Defines the parameters for the iterative solver used to solve the Lippmann-Schwinger equation.
+
+        :param int min_iter: Minimum number of iterations. By default, 1.
```
mh-nguyen712, line 160:
```suggestion
        :param int min_iter: Minimum number of iterations. By default, `1`.
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+
+    def set_verbose(self, verbose):
+        """
+        Toggle verbosity.
+
+        :param bool verbose: Enable verbose/debug printing.
+        """
+        self.verbose = verbose
+        self.born_operator.verbose = verbose
+        self.solver.set_verbose(verbose)
```
Andrewwango, line 377: Same with this method

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+    @dataclass
+    class SolverConfig:
+        r"""
+        Configuration class for the Lippmann-Schwinger solver.
+
+        Defines the parameters for the iterative solver used to solve the Lippmann-Schwinger equation.
+
+        :param int min_iter: Minimum number of iterations. By default, 1.
+        :param int max_iter: Maximum number of iterations. By default, 500.
+        :param str solver: Linear solver to use ('lsqr', 'BiCGStab' or 'CG'). By default, 'lsqr'.
```
mh-nguyen712:
```suggestion
        :param str solver: Linear solver to use (`'lsqr'`, `'BiCGStab'` or `'CG'`). By default, `'lsqr'`.
```

mh-nguyen712: and other values

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+        Compute sensor outputs y = G_s * diag(x) u.
+
+        :param torch.Tensor x: Scattering potential (B,1,H,W).
+        :param torch.Tensor total_field: Total field u (B,T,H,W).
+        """
+        # This computes y = G_s*diag(x)*Et
+        self.born_operator.update_total_field(total_field)
+        return self.born_operator.A(x)
+
+    def A(self, x, receivers=None, transmitters=None, **kwargs):
```
Andrewwango: Static typing for external methods please, and add the params to the docstring

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+        """
+        Forward operator wrapper: updates parameters, solves for total field and returns measurements.
+
+        :param torch.Tensor x: Scattering potential (B,1,H,W).
+        :param torch.dtype dtype: torch.dtype used for inputs/outputs
```
Andrewwango: Remove

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+
+    def update_parameters(self, receivers=None, transmitters=None, **kwargs):
+        """
+        Update transmitter and receiver parameters and recompute dependent fields/operators.
+
+        :param torch.Tensor receivers: New receiver positions of shape `(2, R)` or `(2, T, R)`.
+        :param torch.Tensor transmitters: New transmitter positions of shape `(2, T)`, where T is the number of transmitters.
+        """
+        if transmitters is not None:
+            self.transmitters = transmitters.to(self.device).to(self.dtype)
```
mh-nguyen712: use `register_buffer` so it works with `.to()`

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+        self.incident_field = incident_field.to(self.device).to(self.dtype)
+
+    def A_jvp(self, x, u, **kwargs):
+        r"""
+        Jacobian-vector product for the forward operator.
+
+        :param torch.Tensor x: Scattering potential (B,1,H,W).
+        :param torch.Tensor u: Vector to multiply with the Jacobian (B,1,H,W).
+        """
+        return torch.autograd.functional.jvp(self.A, x, v=u)[1]
```
mh-nguyen712, line 339: I'm not sure if it's supported for this operator `A`, but forward-mode auto diff can be faster here (`torch.func.jvp`).

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+    """
+    out = 1j / 4 * hankel1(0, r)
+    if remove_nans:
+        out[torch.isnan(out)] = out.abs().max()  # singularity at 0
+    return out
+
+
+def green_fourier(img_width, box_length, wavenumber):
+    r"""
+    Green function discretizaton for :cite:t:`vico2016fast` in 2D.
```
mh-nguyen712:
```suggestion
    Green function discretization for :cite:t:`vico2016fast` in 2D.
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+                1, incident_field.size(0), self.img_width, self.img_width
+            )  # (1, T, H, W)
+
+        self.incident_field = incident_field.to(self.device).to(self.dtype)
+
+    def A_jvp(self, x, u, **kwargs):
+        r"""
+        Jacobian-vector product for the forward operator.
+
+        :param torch.Tensor x: Scattering potential (B,1,H,W).
```
mh-nguyen712:
```suggestion
        :param torch.Tensor x: Scattering potential `(B,1,H,W)`.
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+        background_wavenumber=10,
+        solver_config=None,
+        box_length=1.0,
+        wave_type="circular_wave",
+        device="cpu",
+        dtype=torch.complex128,
+        verbose=False,
+    ):
+        super(Scattering, self).__init__()
+        if wave_type not in wave_type in ["circular_wave", "plane_wave"]:
```
mh-nguyen712:
```suggestion
        if wave_type not in ["circular_wave", "plane_wave"]:
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+            -box_length / 2, box_length / 2, img_width, device=device, dtype=dtype
+        )
+        self.y_domain, self.x_domain = torch.meshgrid(
+            -image_domain, image_domain, indexing="ij"
+        )
+        self.img_width = img_width
+
+        # incident field
+        self.total_field = None
+        self.img_width = img_width
```
mh-nguyen712: duplicate

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+    total_field = torch.zeros_like(incident_field)
+    jv_prime = lambda n, x: 0.5 * (jv(n - 1, x) - jv(n + 1, x))
+    hankel1_prime = lambda n, x: 0.5 * (hankel1(n - 1, x) - hankel1(n + 1, x))
+
+    list_n = [0]
+    for i in range(1, n_coeffs):
+        list_n.append(i)
+        list_n.append(-i)
+
+    for p in range(angles.shape[0]):
```
mh-nguyen712, line 1044: some variable is recomputed in the loop but they are constant eg `extra_contrast`

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+    :param torch.Tensor receivers: Receiver positions of shape (2, T, R).
+    :param torch.Tensor x_domain: x coordinates of image grid points (H*W,).
+    :param torch.Tensor y_domain: y coordinates of image grid points (H*W,).
+    :param float, torch.Tensor wavenumber: Scalar wavenumber.
+    :param float pixel_area: Area of each pixel in the discretized grid.
+    :param int img_width: Number of img_width per image side (H=W).
+    :param bool verbose: Enable verbose/debug printing.
+    """
+
+    def __init__(
```
Andrewwango, line 509: Static typing

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+            y_domain,
+            wavenumber,
+            pixel_area,
+            img_width,
+            dtype=total_field.dtype,
+            device=total_field.device,
+        )
+        self.verbose = verbose
+
+    def update_total_field(self, total_field):
```
Andrewwango: I prefer implementing setter methods as properties

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+    """
+    Generate equispaced sensors on a circle.
+
+    :param int number: Number of sensors.
+    :param float radius: Radius of the circle.
+    :param float max_angle: Maximum angle in degrees covered by sensors.
+    :param float offset_angle: Offset angle in degrees.
+    :param str device: Torch device for tensors.
+    :param torch.dtype dtype: torch.dtype used for output tensors (inferred from device)
+    :return: dictionary with:
```
Andrewwango: This is wrong

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+def circular_sensors(number, radius, max_angle=360, offset_angle=0, device="cpu"):
+    """
+    Generate equispaced sensors on a circle.
+
+    :param int number: Number of sensors.
+    :param float radius: Radius of the circle.
+    :param float max_angle: Maximum angle in degrees covered by sensors.
+    :param float offset_angle: Offset angle in degrees.
+    :param str device: Torch device for tensors.
+    :param torch.dtype dtype: torch.dtype used for output tensors (inferred from device)
```
Andrewwango: There seems to be many methods with incorrectly dtype in the docstring

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+    :param int img_width: Number of img_width per image side (`H=W`).
+    :param float box_length: Physical side length of the square domain.
+    :param torch.Tensor wavenumber: Wavenumber (real or complex).
+    :param SolverConfig config: Solver configuration parameters.
+    :param bool verbose: Enable verbose/debug printing.
+    :param str device: Torch device string, e.g. 'cpu' or 'cuda'.
+    :param torch.dtype dtype: Torch `dtype` for tensors (e.g. `torch.complex128`).
+    """
+
+    def __init__(
```
Andrewwango, line 762: Static typing

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+        self.e = 0.01
+        _, self.g_fourier = green_fourier(
+            img_width, box_length, (wavenumber.pow(2) + 1j * self.e).sqrt()
+        )
+        self.k02 = wavenumber.pow(2).to(device)
+        self.g_fourier = self.g_fourier.to(device).to(dtype)
+        self.adjoint_state = config.adjoint_state
+
+        # Store solver parameters in a dictionary for easy passing
+        self.solver_params = {
```
Andrewwango: Why the need for a dict when you already have a config class? I think either have a dict for everything, or a config for everything, or just have everything as keyword args - I prefer that even though it's more verbose

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+            "tol": config.tol,
+        }
+
+    def set_verbose(self, verbose):
+        """
+        Set the verbosity of the solver.
+        """
+        self.solver_params["verbose"] = verbose
+
+    def forward(self, k2, source, init=None, **kwargs):
```
Andrewwango: need descriptions of inputs and outputs here especially with nonstandard variable name k2

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+        A = lambda u: u - apply_filter(m * u, g_fourier)
+
+        # Define its adjoint AT(u) = u - m.conj() * G.conj()(u)
+        AT = lambda u: u - m.conj() * apply_filter(u, g_fourier.conj())
+
+        # The right-hand side is the incident field u_i = G * source
+        b_incident_field = apply_filter(source, g_fourier)
+
+        # Solve A(u_s) = b for the scattered field u_s
+        scattered_field = least_squares(
```
Andrewwango: this function is largely the same as the solver forward and is repeated code - can we abstract this and have it similar to how we have it in Tomography?

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+
+def green_fourier(img_width, box_length, wavenumber):
+    r"""
+    Green function discretizaton for :cite:t:`vico2016fast` in 2D.
+
+    :param int img_width: image width H=W
+    :param float box_length: physical box length
+    :param torch.Tensor or scalar wavenumber: scalar wavenumber expected
+    :param torch.dtype dtype: torch.dtype used for intermediate and returned tensors
+    """
```
Andrewwango, line 929: return type

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+    H, W = field.shape[-2], field.shape[-1]
+    H2 = H * (factor - 1) // 2
+    W2 = W * (factor - 1) // 2
+    # pad input
+    field = torch.nn.functional.pad(field, (W2, W2, H2, H2), mode=padding_mode, value=0)
+
+    # filter in fourier space
+    xf = torch.fft.fft2(field, norm="ortho")
+    yf = xf * filterf.unsqueeze(0)  # .conj()
```
Andrewwango:
```suggestion
    yf = xf * filterf.unsqueeze(0)
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1007 @@
...
+
+    for p in range(angles.shape[0]):
+        for n in list_n:
+            # calculate incident and total fields
+            ind = r < cylinder_radius
+            extra_contrast = torch.sqrt(
+                torch.tensor(1 + cylinder_contrast, dtype=dtype)
+            )
+            w = wavenumber
+            jvn = jv(n, w * extra_contrast * cylinder_radius).to(dtype)
```
Andrewwango: seems to be a lot of casting going on inside the loop

examples/physics/demo_scattering.py
```diff
@@ -0,0 +1,298 @@
...
+    This parametrization ensures that the scattering potential :math:`x` is dimensionless, and can be used for different physical
+    modalities:
+    In **microwave tomography** applications, the scattering potential is related to the object's relative permittivity :math:`\epsilon_r` as :math:`x(\mathbf{r}) = \epsilon_r(\mathbf{r}) - 1`.
+    In **optical diffraction tomography**, the scattering potential is related to the refractive index :math:`n` as :math:`x(\mathbf{r}) = n^2(\mathbf{r}) - 1`.
+
+    Moreover, the wavenumber can be also provided in a dimensionless form by normalizing it with respect to the box length :math:`L` as :math:`k_b = 2 \pi L / \lambda`,
+    where :math:`\lambda` is the wavelength of the incident wave.
+
+This example shows how to define the scattering forward model, generate measurements,
+and perform reconstructions using both a linear (Born approximation) solver and a non-linear
```
Andrewwango: describe what image reconstruction is in this case: recovering the sources from the scatter measurements

examples/physics/demo_scattering.py
```diff
@@ -0,0 +1,298 @@
...
+#
+# .. math::
+#
+#         \left(I - G_s \text{diag}(x)\right) u_i = b_i
+#
+# where :math:`G_s` is the convolution operator with Green's function :math:`g`, and :math:`b_i = G_s (x \circ v_i)`.
+#
+# This linear system becomes highly ill-conditioned for high contrast objects (i.e., large :math:`\|x\|_{\infty}`) and/or high wavenumber
+# (which induces a high spectral norm of the Green operator :math:`\|G_s\|_2`), and the solver
+# may fail to converge. In that case, one can try to increase the number of iterations, or change the solver.
```
Andrewwango: change the solver: link to other solvers?

examples/physics/demo_scattering.py
```diff
@@ -0,0 +1,298 @@
...
+    device=device,
+    background_wavenumber=wavenumber,
+    solver_config=config,
+    transmitters=transmitters,
+    receivers=receivers,
+)
+physics.normalize(x)
+
+
+physics.set_verbose(True)
```
Andrewwango, line 124: this is a bit cumbersome - better to set verbose as an argument in the init

examples/physics/demo_scattering.py
```diff
@@ -0,0 +1,298 @@
...
+
+x_gd = gd_solver(y, physics)
+
+print("PSNR of gradient descent reconstruction:", psnr(x, x_gd).item())
+
+dinv.utils.plot(
+    [x, x_lin, x_gd],
+    titles=[
+        "ground truth",
+        "Born approximation\nPSNR={:.2f}dB".format(psnr(x, x_lin).item()),
```
Andrewwango: use fstring or the new subtitle arg of plot

examples/physics/demo_scattering.py
```diff
@@ -0,0 +1,298 @@
...
+# using its linear pseudo-inverse, computed with a linear solver.
+
+physics.set_verbose(False)
+
+x_lin = physics.A_dagger(y, linear=True)
+
+print(f"PSNR of Born approximation:", psnr(x, x_lin).item())
+
+# %%
+# Reconstruction with gradient descent
```
Andrewwango, line 252: highlight that because we are pytorch native, we can autodiff to compute the grads of the nonlinear op!

deepinv/physics/scattering.py
tachella: it is via torch.complex, I've clarified it in the doc

deepinv/physics/scattering.py
tachella: added the description of default config

deepinv/physics/scattering.py
tachella: yes, added a clarification

deepinv/physics/scattering.py
tachella: done

tachella: done

deepinv/physics/scattering.py
tachella: thanks for spotting this!

deepinv/physics/scattering.py
tachella: done

deepinv/physics/scattering.py
tachella: thanks, fixed

deepinv/physics/scattering.py
tachella: not sure we need code mode here

deepinv/physics/scattering.py
tachella: I've added it to strings and bools, but not floats

deepinv/physics/scattering.py
tachella: number of transmitters (aka angles of tomography) - added a clarification

deepinv/physics/scattering.py
tachella: done

deepinv/physics/scattering.py
tachella: I moved the registering from the init to here.

Is the device handling okay now? I'm not sure given the latest updates

deepinv/physics/scattering.py
tachella: good idea, thanks!

deepinv/physics/scattering.py
tachella: I've added a math explanation in the docstring which should be clearer

deepinv/physics/scattering.py
tachella: agreed

deepinv/physics/scattering.py
tachella: thanks done

deepinv/physics/scattering.py
tachella: I tried this at the beginning, but I didn't manage to make it work with double differentiation, whereas this solution works

deepinv/physics/scattering.py
tachella: this was actually a debugging code that should be removed (now gone)

deepinv/physics/scattering.py
tachella: the issue is that I don't know how to set the verbosity of the solver automatically when setting the verbosity of the parent class

do you know a better solution?

deepinv/physics/scattering.py
tachella: see my comment above

deepinv/physics/scattering.py
tachella: done

deepinv/physics/scattering.py
tachella: done

deepinv/physics/scattering.py
tachella: ok, I removed the method

deepinv/physics/scattering.py
tachella: thanks fixed to "tuple"

deepinv/physics/scattering.py
tachella: true sorry for this, some artifact of the past, should be fixed now

deepinv/physics/scattering.py
tachella: thanks for spotting this

deepinv/physics/scattering.py
tachella: done

deepinv/physics/scattering.py
tachella: done

deepinv/physics/scattering.py
tachella: removed them

deepinv/physics/scattering.py
tachella: done, I've moved extra_contrast, w, and ind outside the loop

deepinv/physics/scattering.py
tachella: done

deepinv/physics/scattering.py
tachella: done

deepinv/physics/scattering.py
tachella: done

examples/physics/demo_scattering.py
tachella: it is not recovering the sources (known), but rather the contrast of the object. I've clarified this

examples/physics/demo_scattering.py
tachella: I've added a ref to the solver config class

examples/physics/demo_scattering.py
tachella: it is because I want to avoid  verbosity when normalizing, but I want verbosity when doing physics(x)

examples/physics/demo_scattering.py
tachella: I've added a new section dedicated to gradients. 

we compute gradients by hand doing the 'state-adjoint' method which reduces memory by a lot (instead of default autodiff grads). this is also explained in the text.

examples/physics/demo_scattering.py
tachella: done moved to fstring

deepinv/physics/functional/special.py
tachella: it is consistent with the special modules of torch and scipy

deepinv/physics/functional/special.py
tachella: agreed

deepinv/physics/functional/special.py
tachella: this is handled by scipy

deepinv/physics/functional/special.py
tachella: done

deepinv/physics/functional/special.py
tachella: done

deepinv/physics/functional/matrix.py
tachella: true, my reasoning was that least_squares is almost an inverse solver which uses optimization, whereas the power method is only related to some properties of the forward operator

tachella:
@mh-nguyen712 @Andrewwango 
Many thanks for your very nice and exhaustive reviews :) I've applied all the changes, except for some very few ones where I answered to the comments. let me know if the changes are fine!

deepinv/physics/functional/special.py
```diff
@@ -0,0 +1,87 @@
...
+    # Transfer to CPU, convert to numpy, compute, then back to Torch
+    try:
+        import scipy.special
+    except ImportError:
+        raise ImportError(
+            "SciPy or PyTorch version >= 2.6 is required for hankel1 computation."
+        )
+    out = scipy.special.hankel1(n, x.to("cpu"))
+
+    return out.to(device=device or x.device)
```
mh-nguyen712:
```suggestion
    return out.to(device=device)
```

deepinv/physics/functional/special.py
```diff
@@ -0,0 +1,87 @@
...
+        import scipy.special
+    except ImportError:
+        raise ImportError(
+            "SciPy or PyTorch version >= 2.6 is required for jv computation."
+        )
+    # We detach and move to CPU to avoid breaking the graph or moving the whole model
+    out = scipy.special.jv(n, x.to("cpu"))
+
+    # Return as a torch tensor on the original or requested device
+    return out.to(device=device or x.device)
```
mh-nguyen712:
```suggestion
    return out.to(device=device)
```

deepinv/physics/scattering.py
mh-nguyen712: From the latest PR #989, we should not rely on `self.device` internally in the future, since for example `physics.to(device)` with move buffers to `device` but not `self.device`. So either we add a dummy tensor `_device_holder` to dynamically update the `device` when calling `update_parameters()` or only register the buffer on the same device of the tensor and users are responsible for giving the tensor `transmitter` on the right device. Can you confirm that @romainvo?

deepinv/physics/functional/special.py
```diff
@@ -0,0 +1,87 @@
+import torch
+
```
mh-nguyen712, line 2: Should we expose these functions to the API? It could be useful

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1070 @@
...
+        )
+        self.img_width = img_width
+
+        # incident field
+        self.total_field = None
+
+        self.update_parameters(receivers=receivers, transmitters=transmitters)
+
+    @dataclass
+    class SolverConfig:
```
mh-nguyen712, line 154: Can we put this solver config dataclass outside of `Scattering` class?

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1070 @@
...
+        :param torch.Tensor x: Scattering potential `(B,1,H,W)`.
+        :param torch.Tensor u: Vector to multiply with the Jacobian `(B,1,H,W)`.
+        """
+        return torch.autograd.functional.jvp(self.A, x, v=u)[1]
+
+    def A_vjp(self, x: torch.Tensor, v: torch.Tensor, **kwargs):
+        """
+        Vector-Jacobian product for the forward operator.
+
+        :param torch.Tensor x: Scattering potential (B,1,H,W).
```
mh-nguyen712:
```suggestion
        :param torch.Tensor x: Scattering potential of shape `(B,1,H,W)`.
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1070 @@
...
+        :param torch.Tensor v: Vector to multiply with the Jacobian (B,T,R).
```
mh-nguyen712:
```suggestion
        :param torch.Tensor v: Vector to multiply with the Jacobian, a tensor of shape `(B,T,R)`.
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1070 @@
...
+            x.requires_grad_(True)
+            y = self.A(x)
+            return torch.autograd.grad(y, x, grad_outputs=v, retain_graph=False)[0]
+
+    def compute_total_field(self, x: torch.Tensor, **kwargs) -> torch.Tensor:
+        """
+        Solve for the total field given scattering potential x using the provided solver.
+
+        :param torch.Tensor x: Scattering potential tensor of shape (B,1,H,W).
+        :returns: (:class:`torch.Tensor`) Total field tensor of shape (B,T,H,W).
```
mh-nguyen712:
```suggestion
        Solve for the total field given scattering potential `x` using the provided solver.

        :param torch.Tensor x: Scattering potential tensor of shape `(B,1,H,W)`.
        :returns: (:class:`torch.Tensor`) Total field tensor of shape `(B,T,H,W)`.
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1070 @@
...
+            img_width=self.img_width,
+            box_length=self.box_length,
+            wavenumber=self.wavenumber,
+            verbose=self.verbose,
+            device=self.device,
+            dtype=self.dtype,
+            config=solver_config,
+        )
+
+    def update_parameters(self, receivers=None, transmitters=None, **kwargs):
```
mh-nguyen712:
```suggestion
    def update_parameters(self, receivers: torch.Tensor=None, transmitters: torch.Tensor=None, **kwargs):
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1070 @@
...
+        """
+        Update transmitter and receiver parameters and recompute dependent fields/operators.
+
+        :param torch.Tensor receivers: New receiver positions of shape `(2, R)` or `(2, T, R)`.
+        :param torch.Tensor transmitters: New transmitter positions of shape `(2, T)`, where T is the number of transmitters.
```
mh-nguyen712:
```suggestion
        :param torch.Tensor transmitters: New transmitter positions of shape `(2, T)`, where `T` is the number of transmitters.
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1070 @@
...
+                y_domain=self.y_domain,
+                wavenumber=self.wavenumber,
+                pixel_area=self.pixel_area,
+                img_width=self.img_width,
+                verbose=self.verbose,
+            )
+
+        super().update_parameters(**kwargs)
+
+    def generate_incident_field(self):
```
mh-nguyen712: Can this function run on GPU? I think yes, then we can provide `device, dtype` argument to prevent `.to()` in line 220

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1070 @@
...
+        if flag and not linear and max_iter != 2:
+            if self.verbose:
+                print("Alternated optimization pseudo inverse did not converge")
+
+        return x
+
+
+class BornOperator(LinearPhysics):
+    r"""
+    Linear operator implementing the Born approximation for the scattering problem.
```
mh-nguyen712, line 489: I would add a bit more details about this operator here, it will be helpful

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1070 @@
...
+
+    :param torch.Tensor total_field: Total field tensor (1,T,H,W) or (B,T,H,W).
+    :param torch.Tensor receivers: Receiver positions of shape (2, T, R).
+    :param torch.Tensor x_domain: x coordinates of image grid points (H*W,).
+    :param torch.Tensor y_domain: y coordinates of image grid points (H*W,).
+    :param float, torch.Tensor wavenumber: Scalar wavenumber.
+    :param float pixel_area: Area of each pixel in the discretized grid.
+    :param int img_width: Number of img_width per image side (H=W).
+    :param bool verbose: Enable verbose/debug printing.
```
mh-nguyen712:
```suggestion
    :param torch.Tensor total_field: Total field tensor `(1,T,H,W)` or `(B,T,H,W)`.
    :param torch.Tensor receivers: Receiver positions of shape `(2, T, R)`.
    :param torch.Tensor x_domain: x coordinates of image grid points `(H*W,)`.
    :param torch.Tensor y_domain: y coordinates of image grid points `(H*W,)`.
    :param float, torch.Tensor wavenumber: Scalar wavenumber.
    :param float pixel_area: Area of each pixel in the discretized grid.
    :param int img_width: Number of img_width per image side (H=W).
    :param bool verbose: Enable verbose/debug printing.
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1070 @@
...
+
+    def __init__(
+        self,
+        total_field: torch.Tensor,
+        receivers: torch.Tensor,
+        x_domain: torch.Tensor,
+        y_domain: torch.Tensor,
+        wavenumber: float | torch.Tensor,
+        pixel_area: float,
+        img_width: int,
```
mh-nguyen712, line 517: a bit confusing that `img_width` is not the width of the image but the number of `img_width`, a bit unclear to me

deepinv/physics/scattering.py
romainvo: Yes, exactly, the same is also true for `self.dtype`, it is not going to be updated appropriately when calling `to()`. Ideally, the `_device_holder` should just be a temporary measure to avoid breaking current workflows; I think newly implemented physics should not rely on this.

The deprecation of `self.device` also induces some changes in how parameters/buffers should be updated:
- in #989, when updating parameters inside `A` or `A_adjoint`, it is implied that the inputs are on the same device as the current buffers, and if new buffers are created, they are initialized on the device of the input (so `update_parameters`, if needed, takes a `device` kwarg as input.
- if some parameters should only be of a certain dtype, I suggest we hardcode it instead of setting `self.dtype = torch.complex128`. If multiple dtype precisions are supported (like float16 or float32), then it should be up to the user to initialize correctly first, and then to manage the dtype of the parameters (by allowing them to pass `dtype` as a kwarg in the update logic e.g.)
- a rule I've implemented in #989, is that when updating a parameters, the dtype and device of the current parameter takes precedence over those of the pushed tensor, in practice this is handled by the super class in `Physics.update_parameters`. This means that you need to ensure all parameters updates are done via `super().update_parameters`. Because we don't have access to self.dtype or self.device, this prevents discrepency between the set of buffers, i.e. all buffers have the same dtype/device at all time, and we only rely on the `to()` method to change it. (https://github.com/deepinv/deepinv/pull/989/changes#diff-69993452ba3f714d704e427f51b84732addeb9b5fe672965827d28725955fe93R270-R275

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1070 @@
...
+
+        .. math::
+
+                g_L(\mathbf{r}) = g(\mathbf{r}) * \text{rect}\left(\frac{\|\mathbf{r}\|}{L}\right),
+
+
+    :param int img_width: Number of pixels per image side (`H=W`). The minimum required number of pixels depends on the background wavenumber to avoid spatial aliasing.
+        We require at least :math:`\text{image width} \geq 2 k_b L/(2 \pi)`, where :math:`k_b` is the background wavenumber and :math:`L` is the box length.
+    :param torch.Tensor receivers: Tensor of shape `(2, R)` (shared receivers) or `(2, T, R)` (per-transmitter receivers) with receiver x/y positions.
+    :param torch.Tensor transmitters: Tensor of shape `(2, T)` with transmitter x/y positions.
```
romainvo: I would indicate the expected `dtype` of both Tensors for clarity

tachella: these should be simple floats, I added for clarity

deepinv/physics/scattering.py
tachella: since it is only for scattering, I would keep it here

deepinv/physics/scattering.py
tachella: it should run on GPU indeed
I've removed the .to() in line 220

deepinv/physics/scattering.py
tachella: thanks for the clarifications. 

I removed all the `to.(self.device)` and `to(self.dtype)` from `update_parameters()`. I chose not to include the dtype and device as optional arguments, as the user can  do `physics.to(device)` instead of changing devices in `update_parameters`

deepinv/physics/scattering.py
tachella: agreed, I've added a math description below

deepinv/physics/scattering.py
tachella: I've fixed the docstring, it is the number of pixels on one side of the image (H=W), the same naming as tomography

deepinv/physics/scattering.py
tachella: done

deepinv/physics/scattering.py
tachella: Scattering is the only physics that needs to solve a least squares problem to compute the forward

deepinv/physics/functional/special.py
tachella: done

deepinv/physics/functional/special.py
tachella: done

deepinv/physics/functional/special.py
tachella: you are right, I've added them to the functional API

deepinv/physics/scattering.py
tachella: I agree - I have removed the dictionary and kept the config class

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1102 @@
...
+        """
+        self.verbose = verbose
+        self.born_operator.verbose = verbose
+        self.solver.set_verbose(verbose)
+
+    def compute_field_out(
+        self, x: torch.Tensor, total_field: torch.Tensor
+    ) -> torch.Tensor:
+        """
+        Compute sensor outputs y = G_s * diag(x) u.
```
mh-nguyen712:
```suggestion
        Compute sensor outputs :math:`y = G_s diag(x) u`.
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1102 @@
...
+            if linear:
+                total_field = self.incident_field
+            else:
+                total_field = self.compute_total_field(
+                    x, init=total_field if use_init else None
+                )
+
+            self.born_operator.total_field = total_field
+            x = self.born_operator.A_dagger(y, init=x if use_init else None)
+            rel_err = (x - prev_x).abs().pow(2).mean() / prev_x.abs().pow(2).mean()
```
mh-nguyen712, line 470: remove `.abs` would save 2 kernel calls 🤭
```suggestion
            rel_err = (x - prev_x).pow(2).mean() / prev_x.pow(2).mean()
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1102 @@
...
+            prev_x = x.clone()
+
+            if linear:
+                total_field = self.incident_field
+            else:
+                total_field = self.compute_total_field(
+                    x, init=total_field if use_init else None
+                )
+
+            self.born_operator.total_field = total_field
```
mh-nguyen712, line 468: use `register_buffer` instead since in Born init `total_field` is a buffer

deepinv/physics/scattering.py
mh-nguyen712: and add this to the API or user guide

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1102 @@
...
+        y: torch.Tensor,
+        init: torch.Tensor = None,
+        solver: str = "lsqr",
+        gamma: float = 1e3,
+        **kwargs,
+    ) -> torch.Tensor:
+        """
+        Solve least-squares for x given y using the operator A and its adjoint.
+
+        :param torch.Tensor y: Measurements (B,T,R).
```
mh-nguyen712:
```suggestion
        :param torch.Tensor y: Measurements of shape `(B,T,R)`.
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1102 @@
...
+    Generate equispaced sensors on a circle.
+
+    :param int number: Number of sensors.
+    :param float radius: Radius of the circle.
+    :param float max_angle: Maximum angle in degrees covered by sensors.
+    :param float offset_angle: Offset angle in degrees.
+    :param str device: Torch device for tensors.
+    :return:: Tuple of tensors:
+         - `transmitters`: Tensor of shape `(2, number)` with (x,y) positions.
+         - `receivers`: Tensor of shape `(2, number, number-1)` with (x,y) positions.
```
mh-nguyen712, line 689: add a newline (I'm not sure at the end or the beginning) to make the itemize be rendered nicely

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1102 @@
...
+    """
+    angles = (
+        torch.linspace(0, max_angle / 360 * 2 * torch.pi, number + 1, device=device)[
+            :-1
+        ]
+        + offset_angle / 360 * 2 * torch.pi
+    )
+    receiver_radii = radius * torch.ones_like(angles)
```
mh-nguyen712: keep it a as a scalar should be ok here (will be broadcasted when * torch.sin(...))

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1102 @@
...
+    x_pos = receiver_radii * torch.cos(angles)
+    y_pos = receiver_radii * torch.sin(angles)
+    transmitters = torch.stack([x_pos, y_pos])  # 2, number
+    receivers = torch.zeros((2, number, number - 1), device=device)
+    for i in range(number):
+        receivers[:, i, :] = transmitters[:, torch.arange(number) != i]
```
mh-nguyen712: We can remove the loop by indexing (can be advantageous if `number` is large)

```suggestion
    mask = ~torch.eye(number, dtype=torch.bool, device=device)
    all_indices = torch.arange(number, device=device).expand(number, number)
    valid_indices = all_indices[mask].view(number, number - 1)
    receivers = transmitters[:, valid_indices]
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1102 @@
...
+        box_length: float,
+        wavenumber: torch.Tensor,
+        config: "Scattering.SolverConfig" = None,
+        device: torch.device | str = "cpu",
+        dtype: torch.dtype = torch.complex128,
+    ):
+        super().__init__()
+
+        # Pre-compute Green's function in Fourier space and move to device
+        self.e = 0.01
```
mh-nguyen712: is this constant important?

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1102 @@
...
+    def forward(
+        self,
+        k2: torch.Tensor,
+        source: torch.Tensor,
+        init: torch.Tensor = None,
+        **kwargs,
+    ) -> torch.Tensor:
+        """
+        Calls the custom autograd function to solve the equation.
+        PyTorch will automatically use the defined .backward() method during backpropagation.
```
mh-nguyen712:
```suggestion
        PyTorch will automatically use the defined `.backward()` method during backpropagation.
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1102 @@
...
+    :param torch.dtype dtype: torch.dtype used for intermediate and returned tensors.
+    :param str device: Torch device string, e.g. 'cpu' or 'cuda'.
+    :param int n_coeffs: Number of Fourier coefficients to use in the series expansion.
+    :param float transmitter_radius: Radius of the circular wave transmitter (only for 'circular_wave' type).
+    :return: total_field, incident_field tensors of shape (1, P, img_width, img_width).
+    """
+    if not isinstance(wavenumber, torch.Tensor):
+        wavenumber = torch.tensor(wavenumber, dtype=dtype, device=device)
+    else:
+        wavenumber = wavenumber.to(dtype).to(device)
```
mh-nguyen712:
```suggestion
        wavenumber = wavenumber.to(dtype=dtype, device=device)
```

deepinv/physics/scattering.py
```diff
@@ -0,0 +1,1102 @@
...
+    # filter in fourier space
+    xf = torch.fft.fft2(field, norm="ortho")
+    yf = xf * filterf.unsqueeze(0)
+    y = torch.fft.ifft2(yf, norm="ortho")
+    # crop output
+    y = y[..., H2 : H2 + H, W2 : W2 + W]
+    return y
+
+
+def mie_theory(
```
mh-nguyen712, line 988: Is this function used often? If yes I think we could try to optimize it a bit

mh-nguyen712: ah sorry just saw that this function is for testing only

deepinv/physics/functional/special.py
```diff
@@ -0,0 +1,90 @@
...
+
+def native_hankel1(n: int, x: torch.Tensor) -> torch.Tensor:
```
mh-nguyen712, line 4: I think we can remove this function. As of torch2.10 there is only `bessel_j0` or `bessel_j1` and no `bessel_j` or `bessel_y`

deepinv/physics/functional/special.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    # Use torch.special functions for J and Y
+    jn = torch.special.bessel_j(n, x)
+    yn = torch.special.bessel_y(n, x)
+
+    # Combine into a complex tensor: J + iY
+    return torch.complex(jn, yn)
+
+
+def hankel1(n: int, x: torch.Tensor) -> torch.Tensor:
```
mh-nguyen712, line 29: So since pytorch don't have hankel or bessel function implemented for higher order, I'm not sure if we should keep these functions or just use the function from scipy directly.

deepinv/physics/scattering.py
tachella: done

deepinv/physics/scattering.py
tachella: they are complex numbers, so we need the abs

deepinv/physics/scattering.py
tachella: done

deepinv/physics/scattering.py
tachella: done

deepinv/physics/scattering.py
tachella: done, thanks for spotting!

deepinv/physics/scattering.py
tachella: true! thanks, done

deepinv/physics/scattering.py
tachella: nice suggestion, done!

deepinv/physics/scattering.py
tachella: it can add stability in very non linear settings. I've moved it to the Config class so the user can choose it instead of having it hard-coded. Thanks for seeing this!

deepinv/physics/scattering.py
tachella: done

deepinv/physics/scattering.py
tachella: done

deepinv/physics/scattering.py
tachella: yep, only testing

deepinv/physics/functional/special.py
tachella: Many thanks for spotting this!

I prefer to keep it, since the most important calls are for n=0 and n=1 (except for the testing function). I've fixed the implementation and now check if n in (0,1)

deepinv/physics/functional/special.py
tachella: same as above, I kept and added a check in (0,1)

deepinv/physics/functional/special.py
```diff
@@ -0,0 +1,104 @@
...
+
+    :param int n: Order of the Hankel function.
+    :param torch.Tensor x: Input tensor.
+    :return: Complex tensor representing :math:`H_1(n, x)`.
+    """
+    # 1. Check if we can use native Torch (requires version >= 1.9 for torch.special)
+    # and specifically bessel_j/y which were stabilized in later 2.x versions.
+    has_native_bessel = hasattr(torch.special, "bessel_j") and hasattr(
+        torch.special, "bessel_y"
+    )
```
mh-nguyen712:
```suggestion
    has_native_bessel = hasattr(torch.special, "bessel_j0") and hasattr(
        torch.special, "bessel_y0"
    )
```

deepinv/physics/functional/special.py
```diff
@@ -0,0 +1,104 @@
...
+    :param int n: Order of the Bessel function.
+    :param torch.Tensor x: Input tensor.
+    :return: Tensor representing :math:`J_v(n, x)`.
+    """
+    # 1. Attempt Native PyTorch (Available in torch >= 1.9)
+    device = x.device
+    if hasattr(torch.special, "bessel_j0") and n == 0:
+        try:
+            # Note: bessel_j supports float/double and supports autograd
+            return torch.special.bessel_j0(n, x).to(device=device)
```
mh-nguyen712:
```suggestion
            return torch.special.bessel_j0(x)
```

deepinv/physics/functional/special.py
```diff
@@ -0,0 +1,104 @@
...
+        except (TypeError, RuntimeError):
+            # Fallback if the specific n (order) or x (dtype) is unsupported natively
+            pass
+
+    if hasattr(torch.special, "bessel_j1") and n == 1:
+        try:
+            return torch.special.bessel_j1(n, x).to(device=device)
```
mh-nguyen712:
```suggestion
            return torch.special.bessel_j1(x)
```

examples/physics/demo_scattering.py
Andrewwango: ok

deepinv/physics/scattering.py
Andrewwango: no problem

Andrewwango (approved):
LGTM

PR #1022, opened 2026-01-20
Fix backward anderson acceleration for DEQ 

dylansechet:
`DEQConfig` allows you to configure specific Anderson acceleration parameters to solve the backward equilibrium. These parameters are being ignored, and the extra pass currently uses the default `BaseOptim.anderson_acceleration_config` instead.

I suspect this is just a typo, as the correct object is defined but never used.


dylansechet:
Rebased on main to fix the black-related lint failure, and fixed a dead reference I'd added to the changelog that was preventing the docs from building.

dylansechet:
Looking into it more closely, I think Anderson acceleration is broken in more scenarios than this. I will be closing this PR and reopening a larger one.

PR #1023, opened 2026-01-21
Fix Anderson acceleration

dylansechet:
This is a successor to #1022. I'd initially found a bug in anderson acceleration with DEQs, but realized the bug was more pervasive and decided to open a new PR.

### Issue
In its current state, Anderson acceleration is completely broken (the code for `FixedPoint.init_anderson_acceleration` references attributes that don't exist and crashes as soon as it is run).
However, this was never caught by the test suite because of a second bug, which causes anderson acceleration to be silently disabled (setting `PGD(anderson_acceleration=True)` doesn't actually enable anderson acceleration).

### Fix

This PR:
- Fixes syntax in `FixedPoint.init_anderson_acceleration` and `FixedPoint.anderson_acceleration_step` so the code can run
- Ensures anderson acceleration is actually used when enabled in PGD, HQS and GD
- Fixes `AndersonAccelerationConfig` default values (history_size and eps were swapped)
- Ensures `DEQConfig` settings are used for anderson acceleration in the backward equilibrium

### Follow-up fixes
Doing the fixes above and running the test suite reveals more bugs, which are also fixed:
- Make anderson acceleration autograd-compatible for unrolled networks (comes at the cost of extra memory. The current solution is just to clone the tensors that were being modified in-place, but it could maybe be done more efficiently)
- Allow anderson acceleration to support arbitrary shapes instead of just b,d,w,h

I've also added a very basic new test to hopefully prevent this from reoccurring. 

- [X] `python3 -m pytest deepinv/tests` runs successfully. (I've only run test_optim and test_unfolded)

### DIFF
--- a/deepinv/optim/fixed_point.py
+++ b/deepinv/optim/fixed_point.py
@@ -198,7 +198,6 @@ class FixedPoint(nn.Module):
         m = min(
             it + 1, self.anderson_acceleration_config.history_size
         )  # effective history length
-        idx_m = idx if idx < m else (m - 1)
         if self.anderson_acceleration_config.full_backprop:
             # Full gradient through history
             # avoid in-place modification, which messes with autograd
@@ -206,8 +205,8 @@ class FixedPoint(nn.Module):
             x_hist = self.x_hist.clone()
             H = self.H.clone()
             # update history and buffers
-            x_hist[:, idx_m] = x_prev.reshape(b, -1)
-            T_hist[:, idx_m] = Tx_prev.reshape(b, -1)
+            x_hist[:, idx] = x_prev.reshape(b, -1)
+            T_hist[:, idx] = Tx_prev.reshape(b, -1)
             self.x_hist = x_hist
             self.T_hist = T_hist
             # use the m first elements of the history
@@ -217,8 +216,8 @@ class FixedPoint(nn.Module):
             # No gradient through history, only through current iterate
             # update history and buffers
             H = self.H.clone().detach()
-            self.x_hist[:, idx_m] = x_prev.reshape(b, -1).detach()
-            self.T_hist[:, idx_m] = Tx_prev.reshape(b, -1).detach()
+            self.x_hist[:, idx] = x_prev.reshape(b, -1).detach()
+            self.T_hist[:, idx] = Tx_prev.reshape(b, -1).detach()
             # old values : don't use grad
             X_old = self.x_hist[:, :m].detach()
             T_old = self.T_hist[:, :m].detach()
@@ -227,9 +226,9 @@ class FixedPoint(nn.Module):
             t_cur = Tx_prev.reshape(b, 1, -1)
             # use masking to only have the current value updating the gradient
             mask = torch.zeros((1, m, 1), device=x_prev.device, dtype=x_prev.dtype)
-            mask[:, idx_m, :] = 1
-            X = X_old + mask * (x_cur - X_old[:, idx_m : idx_m + 1, :])
-            T = T_old + mask * (t_cur - T_old[:, idx_m : idx_m + 1, :])
+            mask[:, idx, :] = 1
+            X = X_old + mask * (x_cur - X_old[:, idx : idx + 1, :])
+            T = T_old + mask * (t_cur - T_old[:, idx : idx + 1, :])
 
         G = T - X  # (b,m,d)
         H[:, 1 : m + 1, 1 : m + 1] = (


dylansechet:
The CI failure appears unrelated to this PR's changes, and is caused by an outdated dependency. I've opened #1024 to fix it.

deepinv/optim/fixed_point.py
```diff
@@ -178,13 +191,19 @@ def anderson_acceleration_step(
         x_prev = X_prev["est"][0]  # current iterate Tx
         Tx_prev = TX_prev["est"][0]  # current iterate x
         b = x_prev.shape[0]  # batchsize
-        x_hist[:, it % self.history_size] = x_prev.reshape(
+
+        # Create a copy to avoid in-place modification, which messes with autograd
```
samuro95: The clone() operation was fixing autograd in-place triggered errors ?  However, the clone operation keeps the same autograd history so it does not stop gradient computation. It could be nice to avoid full backpropagation within history for fast backprop. We could imagine having a flag and then detach `x_hist` and `H_hist` (and probably `p`) when it is true (by default ?). If you want, I can update the code accordingly.

deepinv/optim/fixed_point.py
```diff
@@ -178,13 +191,19 @@ def anderson_acceleration_step(
         x_prev = X_prev["est"][0]  # current iterate Tx
         Tx_prev = TX_prev["est"][0]  # current iterate x
```
samuro95:
```suggestion
        Tx_prev = TX_prev["est"][0]  # current iterate Tx
```

deepinv/optim/fixed_point.py
```diff
@@ -178,13 +191,19 @@ def anderson_acceleration_step(
         x_prev = X_prev["est"][0]  # current iterate Tx
```
samuro95:
```suggestion
        x_prev = X_prev["est"][0]  # current iterate x
```

dylansechet: Fixed in [a32cee9](https://github.com/deepinv/deepinv/pull/1023/commits/a32cee9b30ba5a3505d259d4d60206ef593272c2)

deepinv/optim/fixed_point.py
dylansechet: Fixed in [a32cee9](https://github.com/deepinv/deepinv/pull/1023/commits/a32cee9b30ba5a3505d259d4d60206ef593272c2)

deepinv/optim/fixed_point.py
dylansechet: Yes. The error shows up when running the backward pass : `RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [torch.FloatTensor [5, 19, 3441]], which is output 0 of AsStridedBackward0, is at version 20; expected version 19 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).`

In practice, it only happens for unrolled networks, which seem to be the only scenario where you try to run a backward pass through anderson acceleration steps. 

It would be great if you could update the code, I'm not sure at all what the best interface would look like here.

samuro95:
@dylansechet I added the possibility to do backprop only through the last iterate using the boolean parameter `full_backprop`which is added in the AndersonAccelerationConfig. I also fixed the Anderson Acceleration because it was actually not working properly, adding buffers for the update histories, H and q. Finally, I added tests in optim to check convergence. Can you double check that it works for you please.

deepinv/optim/fixed_point.py
```diff
@@ -167,39 +184,73 @@ def anderson_acceleration_step(
         :param dict X_prev: previous iterate.
         :param dict TX_prev: output of the fixed-point operator evaluated at X_prev
         :param torch.Tensor x_hist: history of last ``history-size`` iterates.
-        :param torch.Tensor T_hist: history of T evlauaton at the last ``history-size``, where T is the fixed-point operator.
+        :param torch.Tensor T_hist: history of T evaluations at the last ``history-size``, where T is the fixed-point operator.
         :param torch.Tensor H: H in the Anderson acceleration linear system Hp = q .
         :param torch.Tensor q: q in the Anderson acceleration linear system Hp = q .
```
dylansechet: Some of the removed parameters are still in the docstring, we can get rid of `x_hist`, `T_hist`, `H` and `q`

deepinv/optim/fixed_point.py
```diff
@@ -167,39 +184,73 @@ def anderson_acceleration_step(
...
-            self.anderson_acceleration_config.beta * (p[:, None] @ T_hist[:, :m])[:, 0]
-            + (1 - self.anderson_acceleration_config.beta)
-            * (p[:, None] @ x_hist[:, :m])[:, 0]
-        )  # Anderson acceleration step.
+
+        # Anderson update
+        beta = self.anderson_acceleration_config.beta
+        x = beta * (p[:, None] @ T)[:, 0] + (1 - beta) * (p[:, None] @ X)[:, 0]
+        x = x.view_as(x_prev)
         x = x.view(x_prev.shape)
```
dylansechet:
```suggestion
```
This is redundant with the `view_as`, we can get rid of it

deepinv/optim/optimizers.py
```diff
@@ -50,9 +50,10 @@ class AndersonAccelerationConfig:
     :param  float eps: Regularization parameter for Anderson acceleration.
     """
```
dylansechet, line 53:
```suggestion
    :param  float eps: Regularization parameter for Anderson acceleration.
    :param  bool full_backprop: Compute backpropagation through all iterates of Anderson acceleration instead of the last iterate only . Default: ``False``.    
    """

```

deepinv/optim/fixed_point.py
```diff
@@ -167,39 +184,73 @@ def anderson_acceleration_step(
...
-        m = min(it + 1, self.history_size)
-        G = T_hist[:, :m] - x_hist[:, :m]
+
+        idx = (
+            it % self.anderson_acceleration_config.history_size
+        )  # "current" position in the ring buffer
+        m = min(
+            it + 1, self.anderson_acceleration_config.history_size
+        )  # effective history length
+        idx_m = idx if idx < m else (m - 1)
```
dylansechet: `idx_m` is always equal to `idx`

samuro95:
Thanks for your suggestions, I let you do the update.

dylansechet:
Done!

PR #1029, opened 2026-01-26
fix cassi sd

Andrewwango:
Fixes CASSI in single-disperser mode (mode="sd") - previously the adjointness test failed (but it wasn't present) - so now the test is added. The fix is to pad with zeros instead of ones.


matthieutrs:
Thanks for fixing!

PR #1033, opened 2026-01-28
Add space-varying blur operator with tiles processing.

mh-nguyen712:
This PR adds a new physics: `TiledSpaceVaryingBlur` and a new generator `TiledBlurGenerator`.

The forward operator is essentially:
- Unfold input image into (overlapping) patches
- Perform local convolution between a local patch and a PSF 
- Fold these patches back to the image, with **smooth blending** in the overlapping zone between adjacent patches

The physics is tested for various `img_size`, `patch_size` and `stride`.

cc @paul-escande 


examples/physics/demo_blur_tour.py
```diff
@@ -324,3 +324,63 @@
     rescale_mode="clip",
     figsize=(5, 5),
 )
+
+
+# %%
+# Space varying blur with tiles
```
Andrewwango, line 330: Maybe explain the reasoning behind why you might want to use this - e.g. for local PSF estimation

deepinv/physics/functional/tiled_product_convolution.py
```diff
@@ -0,0 +1,337 @@
...
+
+from torch import Tensor
+
+import torch
+import torch.nn.functional as F
+from deepinv.physics.functional.utils import _as_pair, _add_tuple
+
+
+@dataclass
+class TiledPConv2dConfig:
```
Andrewwango: while this supports only 2D now, the config itself isn't 2D-specific so maybe we can generalise this in the:
- name of the config & handler
- the typing

docs/source/api/deepinv.physics.rst
```diff
@@ -87,6 +88,7 @@ Generators
    deepinv.physics.generator.DiffractionBlurGenerator
    deepinv.physics.generator.DiffractionBlurGenerator3D
    deepinv.physics.generator.ProductConvolutionBlurGenerator
+   deepinv.physics.generator.TiledBlurGenerator
```
Andrewwango, line 94: this needs to go in user guide too

deepinv/tests/test_physics.py
```diff
@@ -2327,3 +2347,41 @@ def test_squared_or_non_squared_norms(seed, device):
...
+
+
+@pytest.mark.parametrize("batch_size", [1, 2])
+@pytest.mark.parametrize("n_channels", [1, 3])
+@pytest.mark.parametrize("img_size", [(32, 32), (33, 33), (32, 33)])
+@pytest.mark.parametrize("patch_size", [(8, 8), (9, 9), (8, 9), 8])
+@pytest.mark.parametrize("stride", [(4, 4), (5, 5), (4, 5), 3])
+@pytest.mark.parametrize("psf_size", [(5, 5), (6, 6), (5, 6)])
+@pytest.mark.parametrize("use_fft", [False, True])
+def test_tiled_product_physics_adjointness(
```
Andrewwango, line 2490: what's the purpose of this adjointness test when you already include the physics above in the main physics tests?

deepinv/physics/functional/utils.py
```diff
@@ -0,0 +1,23 @@
+# Some utility functions
+from __future__ import annotations
+
+
+def _add_tuple(a: tuple, b: tuple, constant: float = 1) -> tuple:
+    """Add two tuples element-wise with optional scaling of second tuple.
+    It computes: output[i] = a[i] + b[i] * constant
+    """
+    assert len(a) == len(b), "Input tuples must have the same length"
```
Andrewwango: use if ... raise ValueError or RuntimeError rather than asserts

deepinv/physics/functional/utils.py
```diff
@@ -0,0 +1,23 @@
...
+    return tuple(a[i] + constant * b[i] for i in range(len(a)))
+
+
+def _as_pair(value: int | tuple | list) -> tuple[int, int]:
```
Andrewwango: should we allow floats for generality?

deepinv/physics/blur.py
```diff
@@ -682,6 +693,312 @@ def update_parameters(
...
+            patch_size=patch_size,
+            stride=self.stride,
+            psf_size_init=filters.shape[-2:] if filters is not None else None,
+        )
+        self.tiled_handler = TiledPConv2dHandler(self.config)
+
+        self.adjoint_config = self.config.adjoint_config()
+        self.adjoint_tiled_handler = TiledPConv2dHandler(self.adjoint_config)
+
+        self.register_buffer("filters", filters)
```
Andrewwango, line 976: call update parameters instead

deepinv/physics/blur.py
```diff
@@ -682,6 +693,312 @@ def update_parameters(
...
+    >>> psf_generator = MotionBlurGenerator(psf_size=(31, 31))
+    >>> generator = TiledBlurGenerator(
+    ...     psf_generator=psf_generator,
+    ...     patch_size=patch_size,
+    ...     stride=stride,
+    ... )
+    >>> filters = generator.step(batch_size=1, img_size=img_size)["filters"]
+    >>> physics = TiledSpaceVaryingBlur(patch_size=patch_size, stride=stride)
+    >>> y = physics(x, filters=filters)
+    >>> dinv.utils.plot([x, y], titles=["Original", "Blurred"])   # DOCTEST: +SKIP
```
Andrewwango: print shapes of x, y

deepinv/physics/blur.py
```diff
@@ -682,6 +693,312 @@ def update_parameters(
...
+        This class supports both FFT-based convolutions and direct convolutions, see :func:`deepinv.physics.functional.conv2d_fft` and :func:`deepinv.physics.functional.conv2d`.
+        FFT-based convolutions are typically faster for large filters. This can be selected via the `use_fft` parameter (default: `True`).
+
+    .. note::
+
+        This class supports broadcast between of batch and channel dimensions of the filters and the input image.
+        See :func:`deepinv.physics.functional.conv2d` for more details.
+
+
+    :param torch.Tensor filters: Filters :math:`h_k`. Tensor of size `(B, C, K, h, w)` where
```
Andrewwango, line 900: say what happens when filters is None

deepinv/physics/blur.py
```diff
@@ -682,6 +693,312 @@ def update_parameters(
...
+        `B` is the batch size, `C` the number of channels, `K` the number of filters, `h` and `w` the filter height and width which should be smaller or equal than the image :math:`x` height and width respectively.
+    :param int, tuple[int, int] patch_size: Size of the patches (tiles). If `int`, the same size is used for both spatial dimensions.
+
```
Andrewwango, line 905: you're missing a docstring param for blending

deepinv/physics/blur.py
```diff
@@ -682,6 +693,312 @@ def update_parameters(
...
+            )
+
+        self.rearrange = rearrange
+        self.patch_size = _as_pair(patch_size)
+        self.stride = (
+            _as_pair(stride)
+            if stride is not None
+            else tuple(p // 2 for p in self.patch_size)
+        )
+        assert (
```
Andrewwango: 1. use raise ValueError instead of asserts, everywhere
2. maybe add this check into the config setter instead?

deepinv/physics/blur.py
```diff
@@ -682,6 +693,312 @@ def update_parameters(
...
+        try:
+            from einops import rearrange
+        except ImportError:
+            raise ImportError(
+                "einops is required for TiledSpaceVaryingBlur. Please install it via 'pip install einops'."
+            )
+
+        self.rearrange = rearrange
+        self.patch_size = _as_pair(patch_size)
+        self.stride = (
```
Andrewwango: again surely this input validation stuff should be in the config?

If not, then what's the purpose of the config? there seems to be two options:
- Remove the config, and just perform all logic and param configuration here. Pros: cleaner, avoids bloating with config classes.
- Keep the config and put all patching logic and validation in there. Pros: separates the patching from the physics - we could then use this for a deepinv-wide patcher (#1039)

deepinv/physics/functional/tiled_product_convolution.py
Andrewwango: reflecting on this, maybe the name should be changed just to be PatchConfig: since this is not specific to the spacevaryingblur physics at all

deepinv/physics/blur.py
```diff
@@ -682,6 +693,312 @@ def update_parameters(
...
+            else tuple(p // 2 for p in self.patch_size)
+        )
+        assert (
+            self.stride[0] <= self.patch_size[0]
+            and self.stride[1] <= self.patch_size[1]
+        ), f"Stride {self.stride} must be smaller or equal than patch_size {self.patch_size}."
+        self._dynamic_img_size = None  # To track image size changes
+
+        self.use_fft = use_fft
+        self.conv2d_fn = conv2d if not use_fft else conv2d_fft
```
Andrewwango: A lot of these are similar to the normal Blur physics. Should we inheirt from Blur?

deepinv/physics/functional/tiled_product_convolution.py
```diff
@@ -0,0 +1,337 @@
...
+        # Compute number of patches and required padding
+        n_h = (img_size[0] - self.patch_size[0]) // self.stride[0] + 1
+        n_w = (img_size[1] - self.patch_size[1]) // self.stride[1] + 1
+
+        pad_h = self.patch_size[0] + n_h * self.stride[0] - img_size[0]
+        pad_w = self.patch_size[1] + n_w * self.stride[1] - img_size[1]
+
+        return (img_size[0] + pad_h, img_size[1] + pad_w), (pad_h, pad_w)
+
+    def adjoint_config(self) -> "TiledPConv2dConfig":
```
Andrewwango: What's the purpose of this (rather than, say, `def clone(self)`?)

deepinv/physics/functional/tiled_product_convolution.py
```diff
@@ -0,0 +1,337 @@
...
+        self, img_size: tuple[int, int]
+    ) -> tuple[tuple[int, int], tuple[int, int]]:
+        """
+        Get compatible image size and required padding.
+
+        :param img_size: Original image size (height, width).
+        :return: Tuple of (compatible_size, padding).
+        """
+        # Compute number of patches and required padding
+        n_h = (img_size[0] - self.patch_size[0]) // self.stride[0] + 1
```
Andrewwango: use get_num_patches?

deepinv/physics/functional/tiled_product_convolution.py
```diff
@@ -0,0 +1,337 @@
...
+    @property
+    def margin(self) -> tuple[int, int]:
+        """
+        Compute margin based on PSF size: psf_size - 1.
+        """
+        if self.psf_size is None:
+            return (0, 0)
+        return (self.psf_size[0] - 1, self.psf_size[1] - 1)
+
+    def get_num_patches(self, img_size: tuple[int, int]) -> tuple[int, int]:
```
Andrewwango: make this property too?

deepinv/physics/functional/tiled_product_convolution.py
```diff
@@ -0,0 +1,337 @@
...
+        expansion = self.margin
+        return TiledPConv2dConfig(
+            patch_size=_add_tuple(self.patch_size, expansion),
+            stride=self.stride,
+            psf_size_init=self.psf_size,
+        )
+
+
+# =============================================================================
+# PATCH HANDLER CLASS
```
Andrewwango: remove

deepinv/physics/functional/tiled_product_convolution.py
```diff
@@ -0,0 +1,337 @@
...
+# =============================================================================
+class TiledPConv2dHandler:
```
Andrewwango: This is nice, but it probably would be cleaner as a mixin (see https://deepinv.github.io/deepinv/user_guide/other/utils.html#mixins)

deepinv/physics/functional/tiled_product_convolution.py
```diff
@@ -0,0 +1,337 @@
...
+        return patches.contiguous()
+
+    def patches_to_image(
+        self, patches: Tensor, img_size: tuple[int, int] | None = None
+    ) -> Tensor:
+        r"""
+        Reconstruct an image from overlapping patches.
+
+        This is the inverse operation of `image_to_patches`. Note that overlapping
+        regions are summed, so proper normalization may be needed for correct reconstruction.
```
Andrewwango: "may" be needed: does the user need to do that themselves or is it handled automatically?

deepinv/physics/functional/tiled_product_convolution.py
```diff
@@ -0,0 +1,337 @@
...
+        # Crop to target size if specified
+        if img_size is not None:
+            img_size = _as_pair(img_size)
+            output = output[:, :, : img_size[0], : img_size[1]]
+
+        return output.contiguous()
+
+
+# =============================================================================
+# UNITY PARTITION FUNCTIONS
```
Andrewwango: remove

deepinv/physics/blur.py
mh-nguyen712: thanks!

deepinv/physics/blur.py
mh-nguyen712: I can't because we need `img_size` to initialize the mask

deepinv/physics/blur.py
mh-nguyen712: I think inherit from LinearPhysics is ok, since `Blur` supports 3D, but this is not.

deepinv/physics/functional/tiled_product_convolution.py
mh-nguyen712: It depends on `img_size` so can not be a property

deepinv/physics/functional/tiled_product_convolution.py
mh-nguyen712: No this is a different function. the num patches here is the maximum num patches that we can split from an img_size (the last part are ignore if it's smaller than `patch_size`). 
The num patches given by `get_num_patches` is the num patch after padding image, to cover all pixel.

deepinv/physics/functional/tiled_product_convolution.py
mh-nguyen712: It's not a clone. It gives the corresponding config used for `A_adjoint`. Since we used valid conv in `A`, we need large patches in `A_adjoint`

deepinv/physics/functional/tiled_product_convolution.py
mh-nguyen712: This one is specific to the physics indeed. It handles also the padding linked to valid convolution, that why it has `psf_size` argument

mh-nguyen712: Agee. But currently the physics is only 2D so I would leave as it for now.

deepinv/physics/functional/tiled_product_convolution.py
mh-nguyen712: I can move this to `deepinv.utils.TiledMixin`, what do you think? In fact it only depend on `patch_size` and `stride` so we can remove the `config`

deepinv/physics/functional/utils.py
mh-nguyen712: yes, done!

docs/source/api/deepinv.physics.rst
mh-nguyen712: done

deepinv/tests/test_physics.py
mh-nguyen712: as this physics works on patches, I want to make sure it works for all img_size / patch_size / padding. In the physics tests only one value is tested

mh-nguyen712: This is quite fast as tests

deepinv/physics/functional/tiled_product_convolution.py
mh-nguyen712: I meant this is not mathematically the inverse of `image_to_patches`, since the overlapping zone are sum (by unfold).
That's why we do smooth blending in `TiledSpaceVaryingBlur`. I'll clarify this

mh-nguyen712:
Thanks for the review @Andrewwango!
Following your suggestion, I removed the `TiledConfig` and create a mixin instead. It's much cleaner now. Thanks!

deepinv/tests/test_physics.py
Andrewwango: cool

deepinv/physics/blur.py
```diff
@@ -682,6 +692,318 @@ def update_parameters(
...
+        )
+
+        w = self.multipliers  # (B, C, K, H, W)
+        h = self.filters  # (B, C, K, h, w)
+
+        # Extract patches: (B, C, K1, K2, P1, P2)
+        patches = self.image_to_patches(x)
+
+        n_rows, n_cols = patches.size(2), patches.size(3)
+        assert n_rows * n_cols == h.size(2), (
```
Andrewwango: some asserts still remain -> valueerror or runtimeerror

deepinv/physics/blur.py
```diff
@@ -682,6 +692,318 @@ def update_parameters(
...
+        original_patch_size = self.patch_size
+
+        # Update patch size to account for margin in adjoint
+        self.patch_size = _add_tuple(original_patch_size, margin)
+        patches = self.image_to_patches(y)
+        # Restore original patch size
+        self.patch_size = original_patch_size
+
+        n_rows, n_cols = patches.size(2), patches.size(3)
+        assert n_rows * n_cols == h.size(
```
Andrewwango: here too

deepinv/physics/blur.py
Andrewwango: looks good

deepinv/utils/mixins.py
```diff
@@ -278,3 +282,151 @@ def rss(
...
+        return output.contiguous()
+
+    def get_needed_pad(self, img_size: tuple[int, int]) -> tuple[int, int]:
+        """
+        Get required padding.
+
+        :param img_size: Original image size (height, width).
+        :return: Tuple of (compatible_size, padding).
+        """
+        # Compute number of patches and required padding
```
Andrewwango: Clarify what "number of patches" means here then compared to `get_num_patches` meaning

Andrewwango (changes requested):
It looks great Hai! Just a couple of small minor comments, then good to go!

PR #1038, opened 2026-01-31
Add versioned documentations

mh-nguyen712:
Adds versioned documentation:
- old release
- stable (latest release)
- dev: main branch

### DIFF
--- a/examples/physics/demo_mri_tour.py
+++ b/examples/physics/demo_mri_tour.py
@@ -464,13 +464,11 @@ dataset = dinv.datasets.CMRxReconSliceDataset(
 
 x, y, params = next(iter(DataLoader(dataset)))
 
-print(
-    f"""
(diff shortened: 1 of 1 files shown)


tachella:
nice! I'll review asap

tachella:
I have a few questions:

- is it complicated to store all versions? (or at least versions from 0.3.7 onwards
- can we add the version switcher to the bottom left corner, as most libraries do?
- there seems to be a small issue in the switcher next to the logo, since it gives a wider nav bar

image

mh-nguyen712:
I have this in my browser:
image
Yes we can move the version switcher to anywhere, but torch has this in the top-left and numpy in the top-right.

For the storage, it will push to our githug/gh_pages so I'm don't how much storage we have.  We can for example keep the top 5 latest version

PR #1046, opened 2026-02-06
fix: deprecate verbose_individual_losses

### LINKED ISSUE #928: Deprecate `verbose_individual_losses` in trainer
In `deepinv.training.Trainer`, if `verbose_individual_losses=False` but `train=False` (i.e. being used in tester mode) and `compute_eval_losses=False` (as usual), then trainer will fail as the progress bar will try to print the total loss but this hasn't been calculated.

https://github.com/deepinv/deepinv/blob/2bc07427272ac56d20f501bb8e266e0d3aeff9b2/deepinv/training/trainer.py#L1026

I would suggest to deprecate the option completely, the argument doesn't add much, we could just always print the total loss to the end of the progress bar either way, if the user doesn't want it they can ignore it. We should conditionally add it if and only if it exists.

Furthermore, there's no point conditioning this below, adding a value to the logs is no overhead anyway.

https://github.com/deepinv/deepinv/blob/2bc07427272ac56d20f501bb8e266e0d3aeff9b2/deepinv/training/trainer.py#L827

tibisabau:

Closes #928.

### DIFF
--- a/deepinv/__init__.py
+++ b/deepinv/__init__.py
@@ -27,10 +27,6 @@ from deepinv import utils
 
 __all__ += ["utils"]
 
-from deepinv import models
-
-__all__ += ["iterative"]
-
 from deepinv import physics
 
 __all__ += ["physics"]
--- a/deepinv/distributed/__init__.py
+++ b/deepinv/distributed/__init__.py
@@ -1,0 +1,17 @@
+from deepinv.distributed.distrib_framework import (
+    DistributedContext,
+    DistributedStackedPhysics,
+    DistributedStackedLinearPhysics,
+    DistributedProcessing,
+    DistributedDataFidelity,
+)
+from deepinv.distributed.distribute import distribute
+
+__all__ = [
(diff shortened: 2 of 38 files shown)


deepinv/training/trainer.py
```diff
@@ -302,7 +300,6 @@ class Trainer:
     compute_eval_losses: bool = False
     log_train_batch: bool = False
     verbose: bool = True
-    verbose_individual_losses: bool = True
```
Andrewwango, line 305: Removing this parameter adds a breaking change, since users' code who use the parameter, will fail. Instead, we should keep the parameter, default it to None, and check if it's not None in `__post_init__`, and if it's not None, raise a deprecation warning. See for example the `display_losses_eval` argument, achieving the same thing.

Andrewwango, line 305: Also please add an entry to our changelog (in docs/source/changelog.rst) to document that this argument has been deprecated, + to recognise your contribution!

PR #1050, opened 2026-02-10
Fix least-square with zero-valued gamma.

paquiteau:
A simple fix to make `gamma=0` possible as an input for the least-squares solvers (this is for instance the value passed down from the `least_squares_implicit_backward`). 

### DIFF
--- a/deepinv/optim/linear/least_squares.py
+++ b/deepinv/optim/linear/least_squares.py
@@ -82,20 +82,17 @@ def least_squares(
         parallel_dim = [parallel_dim]
 
     if gamma is None:
-        gamma = torch.tensor(0.0, device=y.device)
-        gamma_provided = False
-    else:
-        gamma_provided = True
+        # set gamma to infinity for unregularized least-square.
+        gamma = torch.tensor(torch.inf, device=y.device)
+    elif not isinstance(gamma, torch.Tensor):
+        gamma = torch.tensor(gamma, device=y.device)
 
-        if not isinstance(gamma, torch.Tensor):
-            gamma = torch.tensor(gamma, device=y.device)
-
-        if torch.any(gamma <= 0):
-            warnings.warn(
-                "Regularization parameter of least squares problem (gamma) should be positive."
-                "Otherwise, the problem can become non-convex and the solvers are not designed for that."
-                "Continuing anyway..."
-            )
+    if torch.any(gamma <= 0):
+        warnings.warn(
+            "Regularization parameter of least squares problem (gamma) should be positive."
+            "Otherwise, the problem can become non-convex and the solvers are not designed for that."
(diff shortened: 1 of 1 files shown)


romainvo:
Hi @paquiteau, thank you for pushing this and sorry for the delay! 

I was going through the `least_squares` function, and I'm not sure I understand your fix. If I'm correct, in the current state we have the following two edge cases:
- passing 0 to `gamma` will lead to NaN errors when setting `eta = 1 / gamma` below.
- passing `None` to gamma` actually behaves as setting gama to infinity as we ignore the quadratic term and only solve $\min_x \|Ax-y\|^2$

In `least_squares_implicit_backward`, the default value for gamma is None, which applies my second case. We currently set a dummy value to 0. when gamma is None which is kind of ambiguous
```python
if gamma is None:
    gamma = torch.tensor(0.0, device=y.device)
    gamma_provided = False
```
maybe we could set it to infinity to reflect the real behavior.

Does what I'm describing feel correct to you? Could you detail what use case you had in mind, if I missed something, please?

paquiteau:
we could set it to infinity (or NaN) and use this value for the checks instead of relying on the extra variable `gamma_provided`. You tell me what looks best to you, I can update the fix accordingly.

romainvo:
> we could set it to infinity (or NaN) and use this value for the checks instead of relying on the extra variable `gamma_provided`. You tell me what looks best to you, I can update the fix accordingly.

I think setting it to infinity is fine, that way the value for `1 / gamma` is resolved automatically.

paquiteau:
Alright, I pushed a new proposal, there remain a bit of a fuzzy area for batched gamma, it seems that the behavior is not clearly defined in the case when a batched case hit differents paths: what if we have `gamma=[0, inf, 1]` provided for batch-size of 3 ?

romainvo:
Normally, each batch element should be treated independently, at least the algorithm runs on the entire batch as long as a sub-element has not yet converged (max_iter or some threshold). 

That said, though, for the "lsqr" solver for example, when passing `gamma=0`, the condition number explodes, and the algorithm breaks the loop for all batch elements.

deepinv/optim/linear/least_squares.py
```diff
@@ -192,7 +188,7 @@ def least_squares(
                 f"Solver {solver} not recognized. Choose between 'CG', 'lsqr' and 'BiCGStab'."
             )
 
-        if not gamma_provided and not overcomplete and not complete:
+        if not overcomplete and not complete:
```
romainvo, line 191: I think that is what's making the tests fail

```suggestion
        if torch.isinf(gamma).all() and not overcomplete and not complete:
```

romainvo (changes requested):

PR #1051, opened 2026-02-11
MLEM baseline algorithm for Poisson inverse problems

Tmodrzyk:
This PR aims at implementing a classic baseline for emission tomography called [Maximum-Likelihood Expectation-Maximization]((https://pubmed.ncbi.nlm.nih.gov/18238264/)). The algorithm is also widely used for Poisson deconvolution, where it is known as [Richardson-Lucy](https://en.wikipedia.org/wiki/Richardson%E2%80%93Lucy_deconvolution). It is mainly used in astronomy and microscopy. 
The algorithm is implemented in a new optimization class inheriting from BaseOptim. A notebook to showcase the algorithm was also added. 

The algorithm also supports extension to smooth and non-smooth regularizations through an heuristic called One-Step-Late which is implemented in the new MLEM class.  

Towards #975 

### DIFF
--- a/deepinv/optim/optimizers.py
+++ b/deepinv/optim/optimizers.py
@@ -2268,7 +2268,7 @@ class MLEM(BaseOptim):
     .. math::
         x_{k+1} = \frac{x_k}{A^T \mathbf{1} + \lambda \partial g(x_k)} \odot A^T \left(\frac{y}{A x_k}\right)
 
-        where :math:`\partial g(x_k)` is a subgradient of :math:`g` at point :math:`x_k`.
+    where :math:`\partial g(x_k)` is a subgradient of :math:`g` at point :math:`x_k`.
 
     :param deepinv.optim.DataFidelity, list[DataFidelity] data_fidelity: data fidelity term.
         If ``None``, the data fidelity term is not used. Default: ``None``.


Tmodrzyk:
@samuro95 I think this is ready for review, the doc is failing on the same issue as the other PRs (ram / kernel id network download)

deepinv/optim/optim_iterators/mlem.py
```diff
@@ -0,0 +1,72 @@
...
+    def __init__(self, **kwargs):
+        super(MLEMIteration, self).__init__(**kwargs)
+        self.g_step = gStepMLEM(**kwargs)
+        self.f_step = fStepMLEM(**kwargs)
+
+    def forward(
+        self, X, cur_data_fidelity, cur_prior, cur_params, y, physics, *args, **kwargs
+    ):
+        r"""
+        Single Maximum-Likelihood Expectation-Maximization (MLEM) iteration.
```
samuro95, line 21: Could you add a a refence to the description you did in the MLEM class ? and maybe say that it is used there.

deepinv/optim/optimizers.py
```diff
@@ -2225,3 +2226,96 @@ def custom_init(y, physics):
             trainable_params=trainable_params,
             **kwargs,
         )
+
+
+class MLEM(BaseOptim):
+    r"""
+    Maximum Likelihood Expectation Maximization (MLEM) algorithm for Poisson inverse problems.
```
samuro95, line 2233: Can you add a reference for the algorithm ?

deepinv/optim/optimizers.py
```diff
@@ -2225,3 +2226,96 @@ def custom_init(y, physics):
...
+
+    Implements the MLEM iterative algorithm, which is a classic baseline reconstruction method
+    for inverse problems with Poisson noise statistics. At each iteration, the algorithm
+    performs a multiplicative update of the form:
+
+    .. math::
+        x_{k+1} = \frac{x_k}{A^T \mathbf{1}} \odot A^T \left(\frac{y}{A x_k}\right)
```
samuro95, line 2244: Is it related to an oprimization step over the Poisson neg likelihood ?

deepinv/optim/optimizers.py
```diff
@@ -2225,3 +2226,96 @@ def custom_init(y, physics):
...
+
+    where :math:`A` is the forward operator, :math:`y` is the observed data,
+    :math:`\mathbf{1}` is a tensor of ones, and :math:`\odot` denotes element-wise multiplication.
+
+    The algorithm can be used with a prior term (e.g., for MAP-EM variants) or without
+    (standard MLEM). See :class:`deepinv.optim.optim_iterators.MLEMIteration` for the details of the iteration.
+
+    Any data-fidelity can be used but the algorithm only minimizes the Poisson negative log-likelihood data-fidelity.
```
samuro95: So is it an insance of a more general algorithm, here specified for the Poisson negative log-likelihood data-fidelity ?

deepinv/optim/optimizers.py
```diff
@@ -2225,3 +2226,96 @@ def custom_init(y, physics):
...
+                torch.Tensor,
+                Physics,
+            ],
+            torch.Tensor,
+        ] = None,
+        **kwargs,
+    ):
+        if params_algo is None:
+            params_algo = {
+                "lambda": lambda_reg,
```
samuro95, line 2333: There is not stepsize in the algorithm ?

deepinv/optim/optim_iterators/mlem.py
```diff
@@ -0,0 +1,72 @@
...
+
+class fStepMLEM(fStep):
+    def __init__(self, **kwargs):
+        super(fStepMLEM, self).__init__(**kwargs)
+
+    def forward(self, x, cur_data_fidelity, cur_params, y, physics):
+        return x * physics.A_adjoint(y / (physics.A(x).clamp(min=1e-15)))
+
+
+class gStepMLEM(gStep):
```
samuro95: Maybe specify in the description of the algorithm what is the optimization step that you do on g. I seems to be grad by default, and prox if grad not available ? Could the choice between prox and grad be decided with a parameter ?

examples/optimization/demo_poisson_mlem.py
```diff
@@ -0,0 +1,302 @@
+r"""
+Poisson Inverse Problems with Maximum-Likelihood Expectation-Maximization (MLEM)
+====================================================================================================
+
+This example demonstrates how to solve Poisson inverse problems using the
+**Maximum-Likelihood Expectation-Maximization (MLEM)** algorithm :footcite:`sheppMaximumLikelihoodReconstruction1982b`,
+also known as the Richardson-Lucy algorithm in the deconvolution setting :footcite:`richardsonBayesianBasedIterativeMethod1972a,lucyIterativeTechniqueRectification1974a`.
```
samuro95: Ok maybe it would make sense to actually specify that it is Richardson-Lucy in the above. And one question I have is, would it make sense to implement a general Richardson-Lucy that could be then applied to any data fidelity term ?

deepinv/optim/optimizers.py
Tmodrzyk: I added some context in the description about the main interpretations. 
It is a form of majorization-minimization method, but it is not related to a Euclidean gradient or proximal step.
Some also see it as a preconditioned gradient descent with an unusual preconditioner (1 / x)
Recent research tend to fit it in the Bregman mirror descent framework but it is not the consensus yet.

deepinv/optim/optimizers.py
Tmodrzyk: I added the main reference for tomography + some others that are well-known in their respective fields (deconvolution, NMF).

deepinv/optim/optimizers.py
Tmodrzyk: No it is quite specific, it works only for the Poisson negative log-likelihood.
I reformulated this part to make it clearer that the data-fidelity will only be measured, but only the Poisson data-fidelity is minimized by these iterations.

deepinv/optim/optimizers.py
Tmodrzyk: Indeed there is no step-size ! 
Some people use a form of over-relaxation where they put the iterates to a power between 1 and 2 but it is not common at all.

deepinv/optim/optim_iterators/mlem.py
Tmodrzyk: I added a small description + ref to the MLEM algorithm description in the class docstring.

deepinv/optim/optim_iterators/mlem.py
Tmodrzyk: Thanks, I added a param to choose between prox and grad if both are available + the description in the main algorithm.

examples/optimization/demo_poisson_mlem.py
Tmodrzyk: The algorithm is known as Richardson-Lucy only in the deconvolution / signal processing community, in tomography everyone calls it MLEM. 
It is actually the same algorithm though. 
Unfortunately it can only handle Poisson noise, so only this data-fidelity !

Tmodrzyk:
Hi @samuro95 thank you very much for the review ! I tried to answer your questions. I also added context / description in many parts of the code. 

To sum it up, MLEM / Richardson-Lucy / MU is an optimization algorithm but it can't be written as grad / prox step. Rather it relates more to majorization-minimization with a specific surrogate function designed for the Poisson negative log-likelihood. 
Originally it was derived from the EM framework (from statistics) hence the name. It is not very well-known in the optimization community but it is a very common baseline for many application communities (especially in tomography since the 80s and Non-Negative Matrix Factorization since 2000).

deepinv/optim/optim_iterators/mlem.py
```diff
@@ -0,0 +1,99 @@
...
+        self.g_step = gStepMLEM(compute_prox=kwargs.get("compute_prox", False))
+        self.f_step = fStepMLEM(**kwargs)
+
+    def forward(
+        self, X, cur_data_fidelity, cur_prior, cur_params, y, physics, *args, **kwargs
+    ):
+        r"""
+        Single Maximum-Likelihood Expectation-Maximization (MLEM) iteration.
+
+        :param torch.Tensor x: Current iterate :math:`x_k`.
```
samuro95, line 25: I am not sure that it makes sense to keep the f and g steps for this algorithm, as it does not write in the format f_step \circ g_step. Maybe we could just keep the forward here.

deepinv/optim/optim_iterators/mlem.py
```diff
@@ -0,0 +1,99 @@
...
+        x_prev = X["est"][0]
+        k = 0 if "it" not in X else X["it"]
+        sensitivity = physics.A_adjoint(ones_like(y))
+        x = self.f_step(x_prev, cur_data_fidelity, cur_params, y, physics)
+        if cur_prior is not None:
+            denom = sensitivity + self.g_step(x, cur_prior, cur_params)
+        else:
+            denom = sensitivity
+        x = x / denom.clamp(min=1e-15)
+        F = (
```
samuro95, line 46: Maybe it could worth precising again in the docstring here, that this correspond to an update on f + g where f is the Poisson Log-Likelihood.

deepinv/optim/optimizers.py
```diff
@@ -2225,3 +2226,117 @@ def custom_init(y, physics):
...
+    :param bool g_first: if ``True``, the prior step is applied before the data fidelity step.
+        Default: ``False``.
+    :param dict params_algo: dictionary of algorithm parameters. If ``None``, the parameters
+        are set from ``lambda_reg``, ``stepsize``, and ``g_param``. Default: ``None``.
+    :param Callable cost_fn: custom cost function. Default: ``None``.
+    :param bool compute_prox: whether to use the proximal operator of the prior (if ``True``) or its gradient (if ``False``) if both are available. Default: ``False``.
+    :param kwargs: additional keyword arguments passed to :class:`deepinv.optim.BaseOptim`.
+    """
+
+    def __init__(
```
samuro95, line 2299: Try to keep the arguments as ADMM here (missing unfold, trainable_params, sigma_denoiser ...)

examples/optimization/demo_poisson_mlem.py
```diff
@@ -0,0 +1,302 @@
+r"""
```
samuro95, line 1: Super nice example :)

deepinv/optim/optim_iterators/mlem.py
Tmodrzyk: Totally agree !

Tmodrzyk:
Hi @samuro95 thank you for all the feedback, I think this is ready for another review !

deepinv/optim/optimizers.py
```diff
@@ -2225,3 +2226,125 @@ def custom_init(y, physics):
...
+    of the provided ``data_fidelity`` argument.
+
+    A regularization can be included via the ``prior`` argument, which will lead to a MAP-EM variant of the MLEM algorithm.
+    Our implementation is based on the One-Step-Late (OSL) heuristic of Green :footcite:t:`greenUseEmAlgorithm1990`.
+    It leads to the following update rule:
+
+    .. math::
+        x_{k+1} = \frac{x_k}{A^T \mathbf{1} + \lambda \nabla g(x_k)} \odot A^T \left(\frac{y}{A x_k}\right)
+
+        where :math:`g` is the prior function and :math:`\lambda` is the regularization parameter.
```
mh-nguyen712: Render issue here
```suggestion
    where :math:`g` is the prior function and :math:`\lambda` is the regularization parameter.
```

deepinv/optim/optimizers.py
```diff
@@ -2225,3 +2226,125 @@ def custom_init(y, physics):
...
+
+    In the case of a non-differentiable prior, the gradient term :math:`\nabla g(x_k)` is replaced by a subgradient:
+
+    .. math::
+        x_{k+1} = \frac{x_k}{A^T \mathbf{1} + \lambda \partial g(x_k)} \odot A^T \left(\frac{y}{A x_k}\right)
+
+        where :math:`\partial g(x_k)` is a subgradient of :math:`g` at point :math:`x_k`.
```
mh-nguyen712:
```suggestion
        x_{k+1} = \frac{x_k}{A^T \mathbf{1} + \lambda \partial g(x_k)} \odot A^T \left(\frac{y}{A x_k}\right)

    where :math:`\partial g(x_k)` is a subgradient of :math:`g` at point :math:`x_k`.
```

Tmodrzyk:
Hi @mh-nguyen712  thanks a lot for the fixes !

PR #1056, opened 2026-02-12
Add ability to pass noise_level_map to DRUNet

LeRatonLaveurSolitaire:
Hello everyone, 

Today I needed to edit DRUNet to be able to directly pass a noise level map so I thought that it might be useful to add this feature into DeepInv.

Let me know if any change needs to be done !

- [X] `python3 -m pytest deepinv/tests` runs successfully.


Tmodrzyk:
Hi @LeRatonLaveurSolitaire, thanks a lot for your PR. I think the CI was broken for something unrelated to your contribution, could you try to make an empty commit to re-run it ? After that I think it will be ready for reviewing. :smiley:

deepinv/models/drunet.py
```diff
@@ -200,12 +200,17 @@ def forward(self, x: torch.Tensor, sigma: torch.Tensor | float) -> torch.Tensor:
...
-            If ``sigma`` is a tensor, it must be of shape ``(batch_size,)``.
+            If ``sigma`` is a tensor, it can be of shape ``(batch_size,)`` or ``(batch_size, 1, height, width)``.
         """
         if isinstance(sigma, torch.Tensor):
             if sigma.ndim > 0:
-                noise_level_map = sigma.view(x.size(0), 1, 1, 1)
-                noise_level_map = noise_level_map.expand(-1, 1, x.size(2), x.size(3))
+                if sigma.shape == (x.size(0), 1, *x.shape[2:]):
+                    noise_level_map = sigma
+                else:
```
Andrewwango: This should be an elif is a singleton, and else: raise ValueError (incorrect shape)

Andrewwango (changes requested):
Thanks for the contribution!

deepinv/models/drunet.py
LeRatonLaveurSolitaire: Thanks, I took your comment into account in the last commit

tachella:
Thanks for this very useful PR :)

can you please also add this option to [RAM](https://deepinv.github.io/deepinv/api/stubs/deepinv.models.RAM.html#deepinv.models.RAM), which follows the same DRUNET way of conditioning on noise maps?

LeRatonLaveurSolitaire:
I tried something for RAM that seems to work but there is a lot more preprocessing of sigma in RAM than in DRUNet which makes the implementation a bit ugly in my opinion. 
For the computation of the snr in `realign_input`, I took the max of each image but I am not sure if it is the best way to do it. 
Let me know if any changes are necessary.

tachella:
Many thanks, I think your implementation with the max looks good :)

Could you add some test checking that the performance is okay for both DRUNet and RAM?
Something similar to [this test](https://github.com/deepinv/deepinv/blob/3359744218daf4af9eefa015a9d9932a7df7621b/deepinv/tests/test_models.py#L1203), but specific to space varying noise maps (it can be a much smaller and self-contained test)

LeRatonLaveurSolitaire:
I added some tests. I had to fix `ComplexDenoiserWrapper` to make it work with complex numbers. I also found a bug in RAM, if the input is only composed of 0, `max_val` is then 0 and ram returns a table of NaN so I wrote a small fix to prevent this.
I don't know why but the tests work when called with `test_denoiser_perf_noise_map(dinv.utils.get_freer_gpu(),'real_imag')` and `test_denoiser_perf_noise_map(dinv.utils.get_freer_gpu(),'abs_angle')` but fail when called with pytest.

deepinv/tests/test_models.py
```diff
@@ -1320,6 +1320,73 @@ def test_denoiser_perf(device):
...
+        / sigma_filter**2
+    ).view((1, 1, 1, -1))
+
+    rng = torch.Generator(device=device).manual_seed(123)
+    sigma = sigma * gaussian_x
+    y = x + sigma * torch.randn(x.shape, generator=rng, device=device)
+
+    psnr_fn = PSNR(max_pixel=1)
+
+    # Only test the trained denoisers and the correspinding expected performance
```
tachella:
```suggestion
    # Only test the trained denoisers and the corresponding expected performance
```

deepinv/tests/test_models.py
```diff
@@ -1320,6 +1320,73 @@ def test_denoiser_perf(device):
...
+        kwargs = {}
+
+        with torch.no_grad():
+            x_hat = denoiser(y, sigma=sigma, **kwargs)
+        improvement = psnr_fn(x_hat, x) - psnr_fn(y, x)
+        assert torch.all(
+            improvement >= torch.tensor(expected_perf).to(device)
+        ), f"Got improvement {improvement.tolist()}, expected at least {expected_perf} with denoiser={type(denoiser).__name__}"
+
+    for denoiser, expected_perf in learned_denoisers:
```
tachella: why is this needed? not sure what we are testing here

deepinv/tests/test_models.py
```diff
@@ -1320,6 +1320,73 @@ def test_denoiser_perf(device):
...
+    # Only test the trained denoisers and the correspinding expected performance
+    learned_denoisers = [
+        (dinv.models.DRUNet(pretrained="download").to(device), (6.8, 10.4, 11.0)),
+        (dinv.models.RAM(pretrained=True).to(device), (6, 9.5, 10)),
+    ]
+
+    for denoiser, expected_perf in learned_denoisers:
+        kwargs = {}
+
+        with torch.no_grad():
```
tachella: can you compare with the isotropic sigma denoiser instead of the noisy image?

deepinv/models/drunet.py
```diff
@@ -200,12 +200,21 @@ def forward(self, x: torch.Tensor, sigma: torch.Tensor | float) -> torch.Tensor:
 
         :param torch.Tensor x: noisy image
         :param float, torch.Tensor sigma: noise level. If ``sigma`` is a float, it is used for all images in the batch.
-            If ``sigma`` is a tensor, it must be of shape ``(batch_size,)``.
+            If ``sigma`` is a tensor, it can be of shape ``(batch_size,)`` or ``(batch_size, 1, height, width)``.
```
tachella, line 218: can you add a ..tip:: block above in the docstring explaining that the model can be also used with non-uniform noise levels? same for the RAM docstring

deepinv/models/ram.py
```diff
@@ -360,12 +387,38 @@ def forward(
         sigma = torch.maximum(
             sigma, torch.tensor(self.sigma_threshold, device=x_in.device)
         )
-        sigma = self._handle_sigma(sigma)
+
+        if isinstance(sigma, torch.Tensor):
```
tachella: this can be simplified to a single if else
```suggestion
        if isinstance(sigma, torch.Tensor) and sigma.shape == (y.shape[0], 1, *y.shape[2:])
```

deepinv/models/ram.py
```diff
@@ -360,12 +387,38 @@ def forward(
...
+                sigma = self._handle_sigma(sigma)
+        else:
+            sigma = self._handle_sigma(sigma)
 
         gain = torch.maximum(
             gain, torch.tensor(self.gain_threshold, device=x_in.device)
         )
-        gain = self._handle_sigma(gain)
+
+        if isinstance(gain, torch.Tensor):
```
tachella: same here, simplify to a single else if

deepinv/models/wrapper.py
```diff
@@ -267,32 +267,44 @@ def forward(self, x: torch.Tensor, sigma: float | torch.Tensor) -> torch.Tensor:
...
         """
         # Duplicate sigma in the batch dimension for real and imaginary parts
-        sigma = self._handle_sigma(
-            sigma,
-            batch_size=x.size(0) * 2,
-            ndim=x.ndim,
-            device=x.device,
-            dtype=x.real.dtype,
-        )
+        if isinstance(sigma, torch.Tensor):
```
tachella: same here, simplify ifs

tachella (changes requested):
thanks for the contribution! some additional feedback

deepinv/models/wrapper.py
LeRatonLaveurSolitaire: Thanks, I took your comment into account in the last commit

deepinv/models/ram.py
LeRatonLaveurSolitaire: Thanks, I took your comment into account in the last commit

deepinv/models/ram.py
LeRatonLaveurSolitaire: Thanks, I took your comment into account in the last commit

deepinv/models/drunet.py
LeRatonLaveurSolitaire: Thanks, I took your comment into account in the last commit

deepinv/tests/test_models.py
LeRatonLaveurSolitaire: Thanks, I took your comment into account in the last commit

deepinv/tests/test_models.py
LeRatonLaveurSolitaire: Thanks, I took your comment into account in the last commit

deepinv/tests/test_models.py
LeRatonLaveurSolitaire: This helped me debug the test when it wasn't working by displaying the PSNR in the terminal. So I thought it would be useful to leave it that way in case it stopped working again.

deepinv/models/drunet.py
```diff
@@ -196,16 +196,27 @@ def forward_unet(self, x0):
 
     def forward(self, x: torch.Tensor, sigma: torch.Tensor | float) -> torch.Tensor:
         r"""
+        .. tip:: `sigma` can be a non uniform noise map with a size `(batch_size, 1, height, width)`.
```
tachella: sorry I wasn't clear enough in my previous comment. I would move this to the docstring in the init defining the class, so users won't miss this functionality

```suggestion
        .. tip:: This model can handle non-uniform `sigma` maps, which can be of size `(batch_size, 1, height, width)`.
```

deepinv/models/ram.py
```diff
@@ -299,10 +324,12 @@ def forward(
             :class:`PoissonNoise <deepinv.physics.PoissonNoise>` or :class:`PoissonGaussianNoise <deepinv.physics.PoissonGaussianNoise>`
             is specified in the physics. If both are provided, the `sigma` and `gain` values provided to the model will be used.
 
+        .. tip:: `sigma` and `gain` can be a non uniform noise map with a size `(batch_size, 1, height, width)`.
```
tachella: same here, move to class definition

tachella (changes requested):
many thanks for applying the changes! I have some small comments left

deepinv/models/drunet.py
LeRatonLaveurSolitaire: No worries, I modified it in the last commit  !

deepinv/tests/test_models.py
```diff
@@ -1320,6 +1320,76 @@ def test_denoiser_perf(device):
...
+        img_size=64,
+        resize_mode="resize",
+    ).to(device)
+
+    x = torch.cat([x1, x2, x1], dim=0)
+    # three different noise levels
+    sigma_values = torch.ones(x.shape, device=device)[:, 0].unsqueeze(1) * torch.tensor(
+        [0.05, 0.1, 0.2]
+    ).to(device).view((3, 1, 1, 1))
+    # non uniform noise, gaussian amplitude along the width
```
matthieutrs, line 1382: Is it possible to add a test on sigma maps = sigma (constant everywhere) and check that
```python
out_map = denoiser(x, sigma=sigma_map)
out_val = denoiser(x, sigma = sigma)
```
yield exact same results?

matthieutrs (changes requested):
Thanks for the PR, great addition! I would just add one test for safety - that the sigma map and sigma (single value) yield exact same performance

deepinv/tests/test_models.py
LeRatonLaveurSolitaire: I took your comment into account in the last commit

matthieutrs (approved):

tachella:
hey @LeRatonLaveurSolitaire, can you please merge main into your PR (I think you haven't set the option to allow maintainers to push to your branch, so I can't do it myself)?

I think tests should pass once main is merged

many thanks and sorry for the delay in getting this PR through :)

tachella (approved):
thanks for this contribution!

deepinv/models/drunet.py
```diff
@@ -213,12 +215,21 @@ def forward(self, x: torch.Tensor, sigma: torch.Tensor | float) -> torch.Tensor:
 
         :param torch.Tensor x: noisy image
         :param float, torch.Tensor sigma: noise level. If ``sigma`` is a float, it is used for all images in the batch.
-            If ``sigma`` is a tensor, it must be of shape ``(batch_size,)``.
+            If ``sigma`` is a tensor, it can be of shape ``(batch_size,)`` or ``(batch_size, 1, height, width)``.
```
Andrewwango, line 218: Small note: it allows 3D as well?

deepinv/models/ram.py
```diff
@@ -433,7 +495,11 @@ def obtain_sigma_gain(
             if isinstance(sigma, TensorList):
                 sigma = sigma.abs().max()
         else:
-            sigma = sigma / rescale_val
+            if isinstance(sigma, torch.Tensor) and sigma.ndim == 4:
+                rescale_val_map = rescale_val.view(rescale_val.shape[0], 1, 1, 1)
```
Andrewwango, line 499: Rescale map view_as sigma for robustness?

deepinv/tests/test_models.py
```diff
@@ -1346,6 +1346,87 @@ def test_denoiser_perf(device, load_example_image):
...
+    ).view((1, 1, 1, -1))
+
+    rng = torch.Generator(device=device).manual_seed(123)
+    sigma_map = sigma_cst_map * gaussian_x
+    y = x + sigma_map * torch.randn(x.shape, generator=rng, device=device)
+
+    psnr_fn = PSNR(max_pixel=1)
+
+    # Only test the trained denoisers and the corresponding expected performance
+    learned_denoisers = [
```
Andrewwango: Parametrise the pytest instead

Andrewwango (changes requested):
Thanks for this great PR! Some small comments

deepinv/models/drunet.py
LeRatonLaveurSolitaire: I haven't tested this, but since I am only checking that sigma has the same size as the image (except for the color channel, which should have a size of 1), it should theoretically works

deepinv/models/ram.py
LeRatonLaveurSolitaire: I am very sorry but since it's been a few month, I don't remember why I wrote it like this, it might be because I copy/paste the structure from before or I might have encounter an issue with rescale as a float

deepinv/tests/test_models.py
LeRatonLaveurSolitaire: I took your comment into account in the last commit

Andrewwango (approved):

PR #1059, opened 2026-02-16
fix tests

tachella:

### DIFF
--- a/examples/optimization/demo_3D_denoising.py
+++ b/examples/optimization/demo_3D_denoising.py
@@ -187,14 +187,11 @@ assert psnr_tv > 29.5
 # A convenient algorithm in this situation is the Dykstra-like algorithm proposed by :footcite:t:`combettes2009iterative`, writing
 #
 # .. math::
-#     \begin{equation}
-#     \begin{aligned}
+#
 #     &\text{For}\;n=0,1,\ldots \\
 #     &\quad x_{n+1} = \sum_{\ell=1}^L \omega_{\ell} \operatorname{prox}_{g_\ell}z_{\ell,n}, \\
 #     &\quad \text{For}\;\ell=1,\ldots,L \\
 #     &\quad \quad z_{\ell,n+1} = x_{n+1} + z_{\ell,n} - \operatorname{prox}_{g_\ell}z_{\ell,n},
-#     \end{aligned}
-#     \end{equation}
 #
 #
 # where :math:`\omega_{\ell} = 1/L` for all :math:`\ell` and :math:`g_{\ell} = \lambda_{\ell} \|\Psi_{\ell} \cdot\|_1`.
--- a/examples/sampling/demo_diffpir.py
+++ b/examples/sampling/demo_diffpir.py
@@ -64,8 +64,7 @@ plot(
 # The algorithm writes as follows, for :math:`t` decreasing from :math:`T` to :math:`1`:
 #
 # .. math::
-#         \begin{equation*}
-#         \begin{aligned}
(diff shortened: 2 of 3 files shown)


deepinv/models/drunet.py
```diff
@@ -49,6 +49,16 @@ class DRUNet(Denoiser):
     :param torch.device, str device: Device to put the model on.
     :param str, int dim: Whether to build 2D or 3D network (if str, can be "2", "2d", "3D", etc.)
 
+    ::
```
mh-nguyen712, line 52: Is this the new way to do doctests?

deepinv/tests/test_physics.py
```diff
@@ -559,6 +560,10 @@ def find_nonlinear_operator(name, device):
             verbose=False,
         )
         x = torch.rand(1, 1, 32, 32, dtype=dtype, device=device) * 0.1  # low contrast
+        if version.parse(torch.__version__) < version.parse("2.6.0"):
```
mh-nguyen712, line 563: We can still test this, since it will fall back to `scipy`, right?

deepinv/tests/test_physics.py
```diff
@@ -2360,6 +2365,11 @@ def test_scattering_mie(device, wavenumber, contrast, wave_type):
 
     We limit the number of tests, since this is a rather long test
     """
+    if version.parse(torch.__version__) < version.parse("2.6.0"):
```
mh-nguyen712, line 2368: similar here `scipy`

deepinv/tests/test_physics.py
tachella: I didn't want to add scipy as a dependency, so I prefer to keep it this way

deepinv/tests/test_physics.py
tachella: same as above

deepinv/models/drunet.py
tachella: no, it removes the doctest while keeping the python code snippet

I think we tried many solutions and nothing seems to work, so I was thinking of deactivating doctests on these cases (at least for the moment, until we find a solution)

mh-nguyen712: ah ok I see. I thought that the +ELLIPSIS and ... on a newline was working.

mh-nguyen712: and do you know why the gpu tests are passing now? Did the admin update the drivers or it's something else?

.github/workflows/test_gpu.yml
```diff
@@ -46,9 +46,6 @@ jobs:
       - name: Install the project
         run: |
           uv sync --extra dataset --extra denoisers --extra test --extra training
-          uv pip install --python .venv torch==2.6.0 torchvision==0.21.0 torchaudio==2.6.0 --index-url https://download.pytorch.org/whl/cu124
```
Andrewwango, line 49: why is this removed?

deepinv/tests/test_physics.py
```diff
@@ -559,6 +560,10 @@ def find_nonlinear_operator(name, device):
             verbose=False,
         )
         x = torch.rand(1, 1, 32, 32, dtype=dtype, device=device) * 0.1  # low contrast
+        if version.parse(torch.__version__) < version.parse("2.6.0"):
+            pytest.skip(
+                "This test requires PyTorch 2.6.0 or higher due to the use of torch.special."
```
Andrewwango, line 565: Should we mention this in the scattering docstring too?

PR #1069, opened 2026-02-23
fix scattering imports

tachella:
Fix remaining scattering imports - double check by hand that the required torch.special functions are indeed there.

The current bug is due to asking for torch 2.6 or higher, whereas we should ask for 2.8 or higher

### DIFF
--- a/deepinv/tests/test_physics.py
+++ b/deepinv/tests/test_physics.py
@@ -8,7 +8,6 @@ import warnings
 import random
 
 import torch
-from packaging import version
 
 import numpy as np
 from deepinv.physics.forward import adjoint_function


tachella:
> It seems that we skip all the tests related to scattering due to the missing of `scipy`? Should we include `scipy` as dependency, at least in the CI for testing?

if we test with pytorch 2.10 then it run most tests (except mie_scattering)?

mh-nguyen712:
yes I think it's a good option too (we can test for torch2.8 first for a smooth transition?)

tachella:
> yes I think it's a good option too (we can test for torch2.8 first for a smooth transition?)

we are already testing with torch 2.10, see in [Check Installed Packages](https://github.com/deepinv/deepinv/actions/runs/22299974384/job/64505216118?pr=1069) and scipy is there (probably as a dependency of another package), so we are indeed testing things

PR #1071, opened 2026-02-24
Warning when reduction="mean" or "none" in StackedLinearPhysics

### LINKED ISSUE #1066: StackedLinearPhysics "mean" and "none" reduction options break adjointness of the operator
### Environment

All

### Describe the bug

When stacking operators $\\{ A_i \\}_{0\leq i \leq n}$ to compute 

```math
y = \begin{bmatrix} A_1(x) \\ A_2(x) \\ \vdots \\ A_n(x) \end{bmatrix}
```

the adjoint is only obtained by summing $A^{\top}y = \sum_{i=1}^{n} A_i^{\top}y_i$. Computing the mean or returning the list of independent adjoint evaluation (`reduction = 'none'`) breaks adjointness.

### Steps to reproduce

```python
import deepinv as dinv
import torch

pansharpening = dinv.physics.Pansharpen(img_size=(3,128,128))
pansharpening.reduction = lambda x: sum(x) / len(x) # override and set to "mean"

test = pansharpening.adjointness_test(torch.randn(1,3,128,128))
print(test) # something not close to zero at all
```

### Expected behavior

The options `reduction="mean"` or `"none"` are not used internally, though for compatibility, we could just deprecate them in case the other options are used by anyone.

_No response_

### Additional context

_No response_

romainvo:
This PR fixes #1066. It adds a warning when the user stacks operator with `reduction="mean"` or `"none"`, informing that it breaks the adjointness property of the resulting stacked linear operator.

I've indicated "coming deprecation" in the warning though if it's a functionality we want to provide to the user, I can remove it.

It also concerns the `DistributedStackedLinearPhysics` class @bmalezieux, the default `reduce_op` is `"sum"`, but the user can also choose `"mean"`, which will also break adjointness. 

The functionality is quite recent, I think it would be okay to remove the `"mean"` option. If users really want to average, they can still do so manually.  Or similarly just indicate that it breaks adjointness.

What do you think? 

---

I have removed the `reduce_op` argument from `DistributedStackedLinearPhysics`, the user still has the option to put `reduce_op="mean"` or `None` when calling `A_adjoint`, in case they want to use it with custom algorithms and defer reduction. @bmalezieux 

### DIFF
--- a/deepinv/loss/metric/perceptual.py
+++ b/deepinv/loss/metric/perceptual.py
@@ -1,7 +1,8 @@
-from deepinv.loss.metric.metric import import_pyiqa, Metric
+import sys, io
+import math
 import torch
 import torch.nn.functional as F
-import math
+from deepinv.loss.metric.metric import import_pyiqa, Metric
 
 
 class LPIPS(Metric):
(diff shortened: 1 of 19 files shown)


bmalezieux:
I agree with you that the cleanest and simplest option is to just remove the option in the distributed framework

PR #1085, opened 2026-03-02
Sample a unique factor per batch in DownsamplingGenerator

romainvo:
The current implementation of `DownsamplingGenerator` allows the user to sample multiple factors in a single batch, which is not permitted by the `Downsampling` physics, as it will produce measurements with different shapes. 

This PR fixes this, and modifies the `step` function to sample a unique factor when `batch_size > 1` and `len(self.list_factors) > 1`. I have also implemented a test checking that a unique factor is sampled per batch.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -112,6 +112,7 @@ jobs:
           # using plot_gallery=0 to avoid building examples
           export TQDM_DISABLE=1
           export HF_HUB_DISABLE_PROGRESS_BARS=1
+          export DEEPINV_DOWNLOAD_VERBOSE=0
           uv run sphinx-build -M doctest docs/source docs/build -D plot_gallery=0 -v
 
       - name: Notebook generation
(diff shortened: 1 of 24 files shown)


jscanvic:
Thanks for raising this issue @romainvo!

There was some discussion about batches of different downsampling rates a while back but I forgot what it was all about. Anyway, I'll review asap!

Andrewwango (approved):
LGTM, maybe just add a comment on the file to explain what's going on

romainvo:
> LGTM, maybe just add a comment on the file to explain what's going on

I have added a note 👍!

deepinv/physics/generator/downsampling.py
```diff
@@ -107,13 +107,27 @@ def step(
...
+        # have the same shape.
+        if batch_size > 1 and len(self.list_factors) > 1:
+            factor_index = torch.randint(
+                low=0,
+                high=len(self.list_factors),
+                size=(1,),
+                generator=self.rng,
+                **self.factory_kwargs,
+            )
+            factor_indices = factor_index.repeat(batch_size)
```
mh-nguyen712: I remember that we have a `random_choice` function in deepinv. I think that it could be nice to use it, but not important.

romainvo: Done 👍 !

Issue #1039, opened 2026-02-04
Unified patching/tiling functionality


Andrewwango:
Many models, physics, algorithms and of course dataloaders require patching of some sort:

1. patch dataset: https://github.com/deepinv/deepinv/blob/main/deepinv/datasets/patch_dataset.py
2. noise estimation module #1015: https://github.com/deepinv/deepinv/pull/1015/changes#diff-c6c9f7755beea585a39e264f29a6d282bdf417a117fe97a685dac34c16c3731cR67
3. patch sampler https://deepinv.github.io/deepinv/_modules/deepinv/datasets/random_sampler.html#RandomPatchSampler
4. #790 
5. patch likelihoods

Each of the above have their own patching logic and validation, which all perform essentially the same thing.

Moving forward, we should have one patching function/class/mixin in deepinv/utils that all of the above (and future code) uses. Acceptance criteria:

- Easy to use via functional, adapter class or mixin (e.g. see TiledMixin2D proposed in #1033)
- Patch/tile with args patch_size, stride
- Perform padding if needed (like in #1033)
- Validate input params and shape
- Blending would be nice (like in #1033)
- arbitrarily support nD where n in (1,2,3)

PR #1088, opened 2026-03-03
Distributed backward

bmalezieux:
# Adding backward pass for distributed framework

1. **Backward synchronization across distributed components**  
Physics and DataFidelity follow a map-reduce pattern on local operators, and use `DistributedGradientSync` to all-reduce gradients for replicated inputs during backward.  
`DistributedProcessing` uses `DistributedParameterSync` to queue a single end-of-backward all-reduce of processor parameter gradients.  
Algorithm parameters (e.g., unfolded step sizes) are synchronized via `DistributedReplicatedParameters` hooks.

2. **`distribute()` API update**  
`type_object="module"` is now restricted to `BaseOptim` models with `unfold=True`, handled through `_distribute_base_optim`.  

3. **New distributed + unrolling example**  
`examples/distributed/demo_unrolled_distributed.py` demonstrates end-to-end integration: stacked distributed physics, tiled distributed denoising inside unfolded DRS, and training with `deepinv.Trainer`.  

4. **`DistributedContext` communication updates (`dist.nn` + `dist`)**  
`DistributedContext` now centralizes collective calls and automatically selects autograd-aware functional collectives (`torch.distributed.nn.functional`) when tensors require gradients.  
For non-grad/inference paths, it falls back to in-place `torch.distributed` collectives for lower overhead and memory efficiency. 

TODO:
- [x] documentation
- [x] changelog


deepinv/distributed/distrib_framework.py
```diff
@@ -1061,11 +1259,22 @@ def _apply_op(
...
+        ):
+            x = DistributedGradientSync.apply(x, self.ctx)
+
         # Get local measurements
         y_local = [y[i] for i in physics.local_indexes]
 
-        # Compute A(x) locally
-        Ax_local = physics.A(x, gather=False, **kwargs)
+        # Compute A_i(x) locally without re-entering distributed gather/reduce logic.
+        Ax_local = [p.A(x, **kwargs) for p in physics.local_physics]
```
romainvo: Why fall back to manual iteration through local physics instead of doing `physics.A(x, gather=False, **kwargs)` ?

deepinv/distributed/distrib_framework.py
```diff
@@ -1061,11 +1259,22 @@ def _apply_op(
...
+        # accumulated gradient once. For non-leaf x, cross-rank aggregation is
+        # already propagated through preceding distributed collectives.
+        if (
+            isinstance(x, torch.Tensor)
+            and x.requires_grad
+            and x.is_leaf
+            and self.ctx.use_dist
+        ):
+            x = DistributedGradientSync.apply(x, self.ctx)
+
```
romainvo, line 1329: Isn't this redundant with line 403-406 in DistributedStackedPhysics._map_reduce_gather ? If we call physics.A(x, gather=False, **kwargs) below line 1277, isn't the DistributedGradientSync in _map_reduce_gather doing the job or am I missing something ?

deepinv/distributed/distrib_framework.py
```diff
@@ -214,14 +223,87 @@ def local_indices(self, num_items: int) -> list[int]:
...
+    def all_reduce(
+        self,
+        x: torch.Tensor,
+        op: dist.ReduceOp = dist.ReduceOp.SUM,
+        group=None,
+    ) -> torch.Tensor:
+        return self._collective("all_reduce", x, op=op, group=group)
+
+    def broadcast(self, x: torch.Tensor, src: int = 0, group=None) -> torch.Tensor:
+        return self._collective("broadcast", x, src=src, group=group)
```
mh-nguyen712: I think we could simplify this a bit. Moreover, calling the lookup `getattr` in each training step might be an overhead. 
Proposition: 
- Remove `_should_use_functional`: a simple `if x.requires_grad` check is ok
- Refactor `_collective` to accept a `fn` and `fn_functional`, eg: 

```python
def _collective(
    self,
    fn: Callable,
    fn_functional: Callable,
    x: torch.Tensor,
    **kwargs: Any,
) -> torch.Tensor:
    if not self.use_dist:
        return x

    if x.requires_grad:
        return fn_functional(x, **kwargs)

    fn(x, **kwargs)
    return x
```
then 
```python
return self._collective(dist.all_reduce, dist_nn.all_reduce, x, op=op, group=group)
```
and so on

deepinv/distributed/distrib_framework.py
```diff
@@ -780,6 +867,15 @@ class DistributedProcessing:
     :param int | None max_batch_size: maximum number of patches to process in a single batch.
         If ``None``, all local patches are batched together. Set to ``1`` for sequential processing
         (useful for memory-constrained scenarios). Higher values increase throughput but require more memory. Default is `None`.
+    :param str checkpoint_batches: activation checkpointing mode for patch-batches during backward.
+        Supported values are ``'auto'``, ``'always'`` and ``'never'``.
+        - ``'auto'`` (default): enable checkpointing only when gradients are enabled and there are multiple local patch-batches.
+        - ``'always'``: always checkpoint patch-batches when gradients are enabled.
+        - ``'never'``: disable checkpointing.
```
mh-nguyen712, line 926: add a newline here to get a nice documentation rendering
```suggestion
        - ``'never'``: disable checkpointing.
        
```

deepinv/distributed/distrib_framework.py
```diff
@@ -780,6 +867,15 @@ class DistributedProcessing:
     :param int | None max_batch_size: maximum number of patches to process in a single batch.
         If ``None``, all local patches are batched together. Set to ``1`` for sequential processing
         (useful for memory-constrained scenarios). Higher values increase throughput but require more memory. Default is `None`.
+    :param str checkpoint_batches: activation checkpointing mode for patch-batches during backward.
+        Supported values are ``'auto'``, ``'always'`` and ``'never'``.
+        - ``'auto'`` (default): enable checkpointing only when gradients are enabled and there are multiple local patch-batches.
```
mh-nguyen712, line 924:
```suggestion

        - ``'auto'`` (default): enable checkpointing only when gradients are enabled and there are multiple local patch-batches.
```

deepinv/distributed/distrib_framework.py
```diff
@@ -944,15 +1086,66 @@ def _apply_op(
...
+        self.average = average
+        self.parameters = [p for p in parameters if isinstance(p, torch.nn.Parameter)]
+        self._hooks = []
+        self._register_hooks()
+
+    def _sync_grad_value(self, grad: torch.Tensor | None) -> torch.Tensor | None:
+        if grad is None or not self.ctx.use_dist:
+            return grad
+        grad = self.ctx.all_reduce(grad, op=dist.ReduceOp.SUM)
+        if self.average and self.ctx.world_size > 1 and not grad.requires_grad:
```
mh-nguyen712, line 1176: I'm not sure if `not grad.requires_grad` was intentional here?

deepinv/distributed/distrib_framework.py
```diff
@@ -1181,3 +1391,29 @@ def _local_grad_op(idx, data, **kw):
...
+
+        Currently supported when a single shared DataFidelity object is used for all
+        operators (the standard unfolded setup). In that case, the prox is delegated
+        to the wrapped DataFidelity with the distributed physics object.
+        """
+        self._check_is_distributed_physics(physics)
+        if self.single_fidelity is None:
+            raise NotImplementedError(
+                "prox is only supported when DistributedDataFidelity wraps a single "
+                "shared DataFidelity instance."
```
mh-nguyen712:
```suggestion
                "prox is only supported when DistributedDataFidelity wraps a single shared DataFidelity instance."
```

deepinv/distributed/distribute.py
```diff
@@ -211,6 +226,168 @@ def data_fidelity_factory(
...
+    tiling_strategy: str | DistributedSignalStrategy | None = None,
+    patch_size: int = 256,
+    overlap: int = 64,
+    tiling_dims: int | tuple | None = None,
+    max_batch_size: int | None = None,
+    checkpoint_batches: str = "auto",
+    checkpoint_use_reentrant: bool = False,
+    checkpoint_preserve_rng_state: bool = True,
+    dtype: torch.dtype | None = torch.float32,
+    gather_strategy: str = "concatenated",
```
mh-nguyen712: Seem that this parameter is not used

deepinv/distributed/distribute.py
```diff
@@ -211,6 +226,168 @@ def data_fidelity_factory(
...
+
+
+def _distribute_base_optim(
+    model: BaseOptim,
+    ctx: DistributedContext,
+    *,
+    average: bool = True,
+    tiling_strategy: str | DistributedSignalStrategy | None = None,
+    patch_size: int = 256,
+    overlap: int = 64,
```
mh-nguyen712, line 269: Might not related to this PR, but we're starting to have different logic / implementation for patching / tiling, eg see [#1039](https://github.com/deepinv/deepinv/issues/1039). It would be nice to have a unify implementation across the library

deepinv/distributed/distribute.py
```diff
@@ -211,6 +226,168 @@ def data_fidelity_factory(
...
+
+    - ``model.data_fidelity`` (``nn.ModuleList``): each
+      :class:`~deepinv.optim.DataFidelity` entry is replaced with a
+      :class:`DistributedDataFidelity`.
+    - ``model.prior`` (``nn.ModuleList``): for every :class:`~deepinv.optim.Prior`
+      that holds a ``.denoiser`` attribute (e.g. :class:`~deepinv.optim.PnP`,
+      :class:`~deepinv.optim.RED`), the denoiser is replaced with a
+      :class:`DistributedProcessing` (overlap-tiled, gradient-synced).
+    - ``model.params_algo`` (``nn.ParameterDict``): all ``nn.ParameterList`` values
+      (trainable step sizes, regularisation weights, etc.) receive distributed
```
mh-nguyen712:
```suggestion
      (trainable step sizes, regularization weights, etc.) receive distributed
```

deepinv/distributed/distribute.py
```diff
@@ -234,12 +414,18 @@ def distribute(
...
+    checkpoint_batches: str = "auto",
+    checkpoint_use_reentrant: bool = False,
+    checkpoint_preserve_rng_state: bool = True,
     **kwargs,
 ) -> (
     DistributedStackedPhysics
     | DistributedStackedLinearPhysics
     | DistributedProcessing
     | DistributedDataFidelity
+    | torch.nn.parameter.Parameter
```
mh-nguyen712:
```suggestion
    | torch.nn.Parameter
```

deepinv/distributed/distribute.py
```diff
@@ -234,12 +414,18 @@ def distribute(
...
+    | list[torch.nn.parameter.Parameter]
```
mh-nguyen712:
```suggestion
    | list[torch.nn.Parameter]
```

deepinv/distributed/distribute.py
```diff
@@ -317,6 +518,23 @@ def distribute(
...
         ...     ddenoiser = distribute(denoiser, ctx)
+
+        Distribute a full unfolded PGD model in one call:
+
+        >>> from deepinv.models import DnCNN
+        >>> from deepinv.optim import PGD
+        >>> from deepinv.optim.data_fidelity import L2
+        >>> from deepinv.optim.prior import PnP
+        >>> from deepinv.distributed import DistributedContext, distribute
+        >>> with DistributedContext() as ctx: # doctest: +SKIP
```
mh-nguyen712, line 526: Can this code be run on CPU? (I think yes, so we can remove the +SKIP)

deepinv/distributed/distributed_utils.py
```diff
@@ -189,6 +253,83 @@ def reduce_local_results(
...
 
 
+class DistributedGradientSync(torch.autograd.Function):
+    """
+    Autograd function that performs all_reduce on gradients during backward.
+    Used to synchronize gradients for replicated inputs in distributed physics.
+    """
+
+    @staticmethod
+    def forward(autograd_ctx, x, dist_ctx):
```
mh-nguyen712, line 265: I would use the standard naming -- ctx for custom autograd function
```suggestion
    def forward(ctx, x, dist_ctx):
```

deepinv/distributed/distributed_utils.py
```diff
@@ -189,6 +253,83 @@ def reduce_local_results(
...
+
+
+class DistributedParameterSync(torch.autograd.Function):
+    """
+    Autograd function that reduces parameters gradients at the end of backward.
+    Used to synchronize model updates in distributed processing.
+    """
+
+    @staticmethod
+    def forward(autograd_ctx, x, dist_ctx, *parameters):
```
mh-nguyen712, line 295: same here

deepinv/distributed/distributed_utils.py
```diff
@@ -189,6 +253,83 @@ def reduce_local_results(
...
+    @staticmethod
+    def backward(autograd_ctx, grad_output):
+        if autograd_ctx.dist_ctx.use_dist and grad_output is not None:
+            higher_order_path = grad_output.requires_grad
+            # Use returned tensor so functional all_reduce paths (create_graph=True)
+            # are preserved for higher-order differentiation.
+            grad_output = autograd_ctx.dist_ctx.all_reduce(
+                grad_output, op=dist.ReduceOp.SUM
+            )
+            if autograd_ctx.dist_ctx.world_size > 1 and not higher_order_path:
```
mh-nguyen712, line 283: Can you explain why we prevent averaging when for higher_order?

mh-nguyen712 (changes requested):
Many thanks for this very nice feature! Only some minor comments / questions.

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,329 @@
...
+        return x_batch, [torch.stack([yy[i] for yy in ys], dim=0) for i in range(n_ops)]
+    return x_batch, torch.stack(ys, dim=0)
+
+
+def move_measurement_to_device(y, device):
+    """Move TensorList/list/tensor measurements to the target device."""
+    if hasattr(y, "to"):
+        return y.to(device)
+    if isinstance(y, TensorList):
+        return TensorList([m.to(device) for m in y])
```
romainvo: TensorList supports the `to` method

```suggestion
    if hasattr(y, "to"):
        return y.to(device)
```

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,382 @@
+r"""
+Distributed Image Tiling Strategies
+======================================
+
+This example shows two tiling strategies for distributed processing of large images.
+The distributed framework enables processing large images by automatically dividing them into tiles,
+processing tiles independently or in parallel, and reconstructing the final result.
```
romainvo, line 7: Very cool example!!

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,382 @@
...
+    img_size=clean_image.shape,
+    tiling_dims=(-2, -1),
+    patch_size=patch_size,
+    overlap=overlap,
+    stride=stride,
+    pad_mode=pad_mode,
+)
+
+print(f"BasicStrategy: {basic_strategy.get_num_patches()} patches")
+print(f"OverlapTilingStrategy: {overlap_strategy.get_num_patches()} patches")
```
romainvo, line 89: For readability, I would split the code here and put each part in its respective section, i.e below under BasicStrategy and OverlapTilingStrategy respectively

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,382 @@
...
+    return h, w
+
+
+# %%
+# .. raw:: html
+#
+#    </details>
+
+# %%
+# Extract patch 0 from both strategies to compare their handling of padding and overlap.
```
romainvo, line 351: I would add a Comparison section here, .e.g

```suggestion
# %%
# Comparison
# --------------------------------
#


# %%
# Extract patch 0 from both strategies to compare their handling of padding and overlap.
```

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,382 @@
...
+# ---------------
+#
+# Define parameters for image size, patch size, and overlap settings.
+
+device = torch.device("cpu")
+img_size = 512
+patch_size = 256
+stride = None  # Defaults to patch_size when None
+overlap = 64  # Receptive field padding for OverlapTilingStrategy
+pad_mode = "reflect"
```
romainvo: I would put these under the OverlapTilingStrategy section

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,329 @@
...
+            physics=stacked_physics,
+            save_dir=str(data_root),
+            dataset_filename=f"{dataset_name}_val",
+            device=ctx.device,
+            train_datapoints=len(val_base),
+            num_workers=num_workers,
+        )
+
+    if ctx.use_dist:
+        dist.barrier()
```
romainvo: Why dist.barrier() here ?

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,329 @@
...
+-------------------------------------------------------------
+
+This example shows how to combine:
+
+- the distributed framework (image/model parallelism over large images),
+- unfolded optimization with :class:`deepinv.optim.DRS`,
+- standard training with :class:`deepinv.Trainer`.
+
+Each rank processes different parts/operators of the same image. This is not
+standard data-parallel training over different images.
```
romainvo, line 12: I wouldn't spare details here and if possible provide the bits of intuition you built while deploying these unrolled distributed training here and that could be useful to others:
- the part about synchronizing the DataGenerator is crucial and could be missed within the code of the helper functions
- you could sell the checkpointing feature here, adding that even with smaller budgets, you can mitigate memory with time and distribute across huge images
- surely other crucial stuff that I'm missing ?

romainvo (commented):
Thanks @bmalezieux and @triem1998 for the very cool work ! I put some comments, not sure all are relevants.

A global comment, the distributed features are getting increasingly complex, distrib_framework.py has 1420 lines!

Do you think it would make sense to split the different distributed components into separate files, matching the module names we already have in the library (e.g physics, models, optim or data_fidelity) ?

deepinv/distributed/distribute.py
bmalezieux: There's a Sphinx issue when doing that, [here's the doc CI result](https://github.com/deepinv/deepinv/actions/runs/24566365214/job/71827484308)

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,321 @@
...
+    # Single process
+    python examples/distributed/demo_unrolled_distributed.py
+
+.. code-block:: bash
+
+    # Multi-process (2 ranks)
+    python -m torch.distributed.run --nproc_per_node=2 examples/distributed/demo_unrolled_distributed.py
+"""
+
+# %%
```
tachella, line 30: remove this section, we don't typically add a section for imports

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,321 @@
+r"""
+Distributed Training of an Unfolded DRS Algorithm on Urban100
+-------------------------------------------------------------
+
```
tachella, line 4: I would add a bit more text motivating the use of distributed training. Something along the lines of:

In many large-scale imaging problems, the size of the image/volume to reconstruct is very large, making it impossible to train reconstruction networks (in this example, unfolded networks) with a single GPU. The `deepinv.distributed` framework enables training a model on multiple GPUs, by carefully parallelizing the data fidelity and denoising steps inside the network.

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,321 @@
...
+from deepinv.optim.prior import PnP
+from deepinv.physics import Denoising, GaussianNoise, stack
+from deepinv.utils import get_data_home
+from deepinv.utils.plotting import plot, plot_curves
+from deepinv.utils.tensorlist import TensorList
+
+# %%
+# Helper functions
+# -----------------------------------------------------------------------------
+
```
tachella, line 64: explain a bit here what you're trying to do with these helper functions

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,321 @@
...
+        return x_batch, TensorList(
+            [torch.stack([yy[i] for yy in ys], dim=0) for i in range(n_ops)]
+        )
+    if isinstance(ys[0], (list, tuple)):
+        n_ops = len(ys[0])
+        return x_batch, [torch.stack([yy[i] for yy in ys], dim=0) for i in range(n_ops)]
+    return x_batch, torch.stack(ys, dim=0)
+
+
+def move_measurement_to_device(y, device):
```
tachella: is this really necessary?

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,321 @@
...
+    """Move TensorList/list/tensor measurements to the target device."""
+    if hasattr(y, "to"):
+        return y.to(device)
+    if isinstance(y, list):
+        return [m.to(device) for m in y]
+    return y
+
+
+def first_measurement(y):
```
tachella: same here, is this necessary?

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,321 @@
...
+from deepinv.utils import get_data_home
+from deepinv.utils.plotting import plot, plot_curves
+from deepinv.utils.tensorlist import TensorList
+
+# %%
+# Helper functions
+# -----------------------------------------------------------------------------
+
+
+def collate_deepinv_batch(batch):
```
tachella:
```suggestion
def collate_batch(batch):
```
(I would avoid any 'deepinv' inside the code, since it's clear we're using deepinv)

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,321 @@
...
+        collate_fn=collate_deepinv_batch,
+    )
+
+    return stacked_physics, train_loader, val_loader
+
+
+# %%
+# Configuration
+# -----------------------------------------------------------------------------
+
```
tachella, line 196: explain here a bit what we are defining, and why we make these choices

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,321 @@
...
+Distributed Training of an Unfolded DRS Algorithm on Urban100
+-------------------------------------------------------------
+
+This example shows how to combine:
+
+- the distributed framework (image/model parallelism over large images),
+- unfolded optimization with :class:`deepinv.optim.DRS`,
+- standard training with :class:`deepinv.Trainer`.
+
+Each rank processes different parts/operators of the same image. This is not
```
tachella: I think the term 'rank' is a bit hard to process for beginners, maybe something like
```suggestion
Each GPU (rank) processes different parts/operators of the same image. This is not
```

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,321 @@
...
+num_workers = 4 if torch.cuda.is_available() else 0
+patch_size = crop_size // 2
+overlap = max(8, patch_size // 8)
+torch.manual_seed(seed)
+
+
+# %%
+# Build distributed physics/model and train with deepinv.Trainer
+# -----------------------------------------------------------------------------
+
```
tachella, line 231: explain here what are the different steps needed to create the model (and what differs from standard unfolded)

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,396 @@
+r"""
+Distributed Image Tiling Strategies
+======================================
+
+This example shows two tiling strategies for distributed processing of large images.
```
tachella:
```suggestion
This example shows tiling strategies for distributed processing of large images.
```
I guess these are the only tiling strategies that we'll consider?

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,396 @@
...
+processing tiles independently or in parallel, and reconstructing the final result.
+
+This example covers:
+
+- :func:`deepinv.distributed.strategies.BasicStrategy`: Simple non-overlapping tiling
+- :func:`deepinv.distributed.strategies.OverlapTilingStrategy`: Overlapping tiles with padding for artifact-free reconstruction
+
+"""
+
+# %%
```
tachella: not sure this is needed

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,396 @@
...
+import matplotlib.pyplot as plt
+from matplotlib import patheffects as pe
+from matplotlib.patches import Rectangle
+from matplotlib.patches import Patch
+import numpy as np
+import torch
+
+from deepinv.distributed.strategies import BasicStrategy, OverlapTilingStrategy
+from deepinv.utils.demo import load_example
+from deepinv.utils.plotting import plot
```
tachella:
```suggestion
from deepinv.utils import plot
```

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,396 @@
...
+
+import matplotlib.pyplot as plt
+from matplotlib import patheffects as pe
+from matplotlib.patches import Rectangle
+from matplotlib.patches import Patch
+import numpy as np
+import torch
+
+from deepinv.distributed.strategies import BasicStrategy, OverlapTilingStrategy
+from deepinv.utils.demo import load_example
```
tachella:
```suggestion
from deepinv.utils import load_example
```

docs/source/user_guide/training/distributed.rst
```diff
@@ -0,0 +1,276 @@
...
+model processes large images or volumes.
+
+The API is the same as for :ref:`distributed reconstruction <distributed>`:
+
+1. :class:`deepinv.distributed.DistributedContext` manages the processes and devices
+2. :func:`deepinv.distributed.distribute` converts deepinv objects to distributed versions
+
+.. note::
+
+    This is different from data-parallel training. In data parallelism,
```
tachella:
```suggestion
    This is different from `data-parallel training offered by pytorch <https://docs.pytorch.org/tutorials/intermediate/ddp_tutorial.html>`_. In data parallelism,
```

docs/source/user_guide/distributed/training.rst
```diff
@@ -0,0 +1,276 @@
...
+
+    import torch
+    import deepinv as dinv
+    from deepinv.distributed import DistributedContext, distribute
+    from deepinv.optim import DRS
+    from deepinv.optim.data_fidelity import L2
+    from deepinv.optim.prior import PnP
+
+    with DistributedContext(seed=0, seed_offset=False) as ctx:
+        physics = distribute(stacked_physics, ctx)
```
tachella, line 47:
```suggestion
        # distribute the forward operator
        physics = distribute(stacked_physics, ctx)
```

docs/source/user_guide/distributed/training.rst
```diff
@@ -0,0 +1,276 @@
...
+
+        model = DRS(
+            data_fidelity=L2(),
+            prior=PnP(denoiser),
+            max_iter=5,
+            unfold=True,
+            trainable_params=["stepsize", "sigma_denoiser"],
+        )
+        model = distribute(model, ctx, patch_size=256, overlap=64)
```
tachella, line 58:
```suggestion
        # distribute the model
        model = distribute(model, ctx, patch_size=256, overlap=64)
```

docs/source/user_guide/training/distributed.rst
```diff
@@ -0,0 +1,276 @@
...
+    for a complete example training an unfolded DRS model.
+
+
+When to Use It
+--------------
+
+Distributed training is useful when you want to train a model but one process
+cannot hold the full computation or would be too slow. Common scenarios include:
+
+**Many physics operators**: if your measurements come from several operators
```
tachella:
```suggestion
**Many physics operators**: if your measurements can be split across several sub-operators
```

docs/source/user_guide/training/distributed.rst
```diff
@@ -0,0 +1,276 @@
...
+dataloader should usually return the same batch on every rank.
+
+In practice:
+
+- Do not use :class:`torch.utils.data.distributed.DistributedSampler` for this use case.
+- Use the same dataset, batch size, and shuffle seed on all ranks.
+- Move both images and measurements to ``ctx.device``.
+- If measurements are stored as a list or :class:`deepinv.utils.TensorList`, move each tensor to the device.
+
+This is different from data parallelism, where each process receives
```
tachella:
```suggestion
This is different from PyTorch's data parallelism, where each process receives
```

docs/source/user_guide/training/distributed.rst
```diff
@@ -0,0 +1,276 @@
...
+different samples and gradients are synchronized only between model replicas.
+
```
tachella: I'm not sure what this means: do we need custom data loaders? If yes, can we include a small snippet of an example?

tachella (changes requested):
Many thanks for the PR! some comments on the docs/examples

docs/source/user_guide/training/distributed.rst
bmalezieux: It just means that we're not doing data parallelism in the usual sense which means replicating a model on several devices, then computing the gradient with different mini batches on each device and synchronizing. Here we need tiling to compute the gradient on one sample. There's no need for custom dataloaders.

deepinv/distributed/framework/distributed_context.py
```diff
@@ -0,0 +1,296 @@
...
+class DistributedContext:
+    r"""
+    Context manager for distributed computing.
+
+    Handles:
+      - Initialization/destruction of the process group (if `RANK` / `WORLD_SIZE` environment variables exist)
+      - Backend choice: NCCL when one-GPU-per-process per node, else Gloo.
+      - Device selection based on `LOCAL_RANK` and visible GPUs
+      - Sharding helpers and tiny communication helpers
+
```
tachella, line 21: I think it would be nice here to add a 

.. note::

explaining what "rank" and WORLD SIZE is for non-experts

deepinv/distributed/framework/distributed_data_fidelity.py
```diff
@@ -0,0 +1,288 @@
...
+    a single reduction, avoiding redundant communication.
+
+    The key operations are:
+
+        - ``fn(x, y, physics)``: Computes the data fidelity :math:`\sum_i d(A_i(x), y_i)`
+        - ``grad(x, y, physics)``: Computes the gradient :math:`\sum_i A_i^T \nabla d(A_i(x), y_i)`
+
+    Both operations use an efficient pattern:
+
+        1. Compute local forward operations (A_local)
```
tachella:
```suggestion
        1. Compute local forward operations :math:`A_i`
```

deepinv/distributed/framework/distributed_data_fidelity.py
```diff
@@ -0,0 +1,288 @@
...
+    ) -> torch.Tensor:
+        r"""
+        Compute the gradient of the distributed data fidelity term.
+
+        For distributed physics with operators :math:`\{A_i\}` and measurements :math:`\{y_i\}`,
+        computes:
+
+        .. math::
+
+            \nabla_x f(x) = \sum_i A_i^T \nabla d(A_i(x), y_i)
```
tachella:
```suggestion
            \nabla_x f(x) = \sum_i \frac{\partial A_i}{\partial \x} \nabla d(A_i(x), y_i)
```
This works for non-linear A as well?

deepinv/distributed/framework/distributed_data_fidelity.py
```diff
@@ -0,0 +1,288 @@
...
+
+        This is computed efficiently by:
+
+            1. Each rank computes :math:`A_i(x)` for its local operators
+            2. Each rank computes :math:`\nabla d(A_i(x), y_i)` for its local operators
+            3. Each rank computes :math:`\sum_{i \in \text{local}} A_i^T \nabla d(A_i(x), y_i)` using A_vjp_local
```
tachella: same here with A_i^T

deepinv/distributed/framework/distributed_processing.py
```diff
@@ -0,0 +1,273 @@
...
+
+    **Example use cases:**
+
+        - Distributed denoising of large images/volumes
+        - Applying neural network priors across multiple GPUs
+        - Processing signals too large to fit on a single device
+
+    :param DistributedContext ctx: distributed context manager.
+    :param Callable[[torch.Tensor], torch.Tensor] processor: processing function to apply to signal patches.
+        Should accept a batched tensor of shape ``(N, C, ...)`` and return a tensor of the same shape.
```
tachella:
```suggestion
        Should accept a batched tensor of shape ``(B, C, ...)`` and return a tensor of the same shape.
```

deepinv/distributed/framework/distributed_processing.py
```diff
@@ -0,0 +1,273 @@
...
+        Examples: denoiser, neural network, prior gradient function, etc.
+    :param str | DistributedSignalStrategy | None strategy: signal processing strategy for patch extraction
+        and reduction. Either a strategy name (``'basic'``, ``'overlap_tiling'``) or a custom strategy instance.
+        Default is ``'overlap_tiling'`` which handles overlapping patches with smooth blending.
```
tachella: what does "basic" do?

deepinv/distributed/framework/distributed_processing.py
```diff
@@ -0,0 +1,273 @@
...
+    :param dict | None strategy_kwargs: additional keyword arguments passed to the strategy constructor
+        when using string strategy names. Examples: ``patch_size``, ``overlap``, ``tiling_dims``. Default is `None`.
+    :param int | None max_batch_size: maximum number of patches to process in a single batch.
+        If ``None``, all local patches are batched together. Set to ``1`` for sequential processing
+        (useful for memory-constrained scenarios). Higher values increase throughput but require more memory. Default is `None`.
+    :param str checkpoint_batches: activation checkpointing mode for patch-batches during backward.
```
tachella: I'm not sure I fully get from the docstring what "activation checkpointing" means here, could you please detail a bit more?

docs/source/user_guide/training/distributed.rst
```diff
@@ -0,0 +1,281 @@
+.. _distributed-training:
+
+Distributed Training
+====================
+
+The distributed framework can be used during training. This is useful when a
+single inverse problem is too large for one GPU, for example because reconstruction
```
tachella:
```suggestion
forward model is too large for one GPU, for example because the reconstruction
```

docs/source/user_guide/training/distributed.rst
```diff
@@ -0,0 +1,281 @@
...
+model processes large images or volumes.
```
tachella:
```suggestion
model processes large images or volumes, and the operator acts globally over them.
```

docs/source/user_guide/training/distributed.rst
```diff
@@ -0,0 +1,281 @@
...
+Using ``seed_offset=False`` keeps random streams aligned across ranks. This is
+important because ranks should usually consume the same data in the same order.
+
+**Step 2: Distribute the physics**
+
+.. code-block:: python
+
+    physics = distribute(stacked_physics, ctx)
+
+For a stacked physics :math:`A = [A_1, \ldots, A_N]`, each rank owns a subset of
```
tachella:
```suggestion
For a :class:`stacked physics <deepinv.physics.StackedPhysics>` :math:`A = [A_1, \ldots, A_N]`, each rank owns a subset of
```

docs/source/user_guide/training/distributed.rst
```diff
@@ -0,0 +1,281 @@
...
+
+- Do not use :class:`torch.utils.data.distributed.DistributedSampler` for this use case.
+- Use the same dataset, batch size, and shuffle seed on all ranks.
+- Move both images and measurements to ``ctx.device``.
+- If measurements are stored as a list or :class:`deepinv.utils.TensorList`, move each tensor to the device.
+
+This is different from Pytorch's data parallelism, where each process receives
+different samples and gradients are synchronized only between model replicas.
+There's no need for custom data loaders or samplers for the distributed framework,
+as long as all ranks consume the same data in the same order.
```
tachella:
```suggestion
as long as all ranks operate on the same image/volume.
```

docs/source/user_guide/training/distributed.rst
```diff
@@ -0,0 +1,281 @@
...
+
+How Backward Works
+------------------
+
+You normally do not need to write custom backward code. The distributed objects
+are built so that PyTorch autograd can follow the full computation.
+
+**Through distributed physics**: each rank applies its local operators
+:math:`A_i`. During backward, the gradient with respect to the shared input is
+computed from local contributions and synchronized across ranks. For linear
```
tachella: Is it only working for linear physics? If yes, not sure "vector jacobian products" makes sense here

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,396 @@
...
+    BasicStrategy,
+    OverlapTilingStrategy,
+)
+from deepinv.utils import load_example, plot
+
+# %%
+# Configuration
+# ---------------
+#
+# Define parameters for image size, patch size, and overlap settings.
```
tachella: overlapping settings are not set here?

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,396 @@
...
+plot(clean_image, titles="Original Image")
+
+# %%
+# Instantiate strategies
+# ----------------------------
+#
+# Create both patching strategies to compare their behavior.
+# BasicStrategy creates non-overlapping patches.
+
+basic_strategy = BasicStrategy(
```
tachella: A silly question: why do we need a special case of non-overlapping as a dedicated "BasicStrategy" class? Can this just be a special case of overlapping=0?

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,396 @@
...
+print(f"BasicStrategy: {basic_strategy.get_num_patches()} patches")
+
+
+# %%
+# ----------------------------
+#
+# OverlapTilingStrategy creates overlapping patches with padding to provide context at boundaries, reducing artifacts.
+
+stride = None  # Defaults to patch_size when None
+overlap = 64  # Receptive field padding for OverlapTilingStrategy
```
tachella: why "receptive field"? is this just the overlap between tiles?

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,396 @@
...
+
+# %%
+# OverlapTilingStrategy
+# --------------------------------
+#
+
+
+# %%
+# OverlapTilingStrategy addresses the boundary artifact problem by adding overlapping patches with padding
+# (the "receptive field"). Instead of non-overlapping tiles, this strategy:
```
tachella:
```suggestion
# Instead of non-overlapping tiles, this strategy:
```
to me, receptive field denotes the patch size used by a network to make a prediction at a given output pixel - I'm not sure how this relates to the overlap?

tachella (changes requested):
Many thanks for applying my previous comments and this PR which represents a great effort!

Please see my new set of comments - my main question is whether is necessary to have two separate classes for no overlapping "Basic Strategy" and overlapping setups - can't we have a single class where an overlapping parameter can be 0?

examples/distributed/demo_tiling_strategies.py
bmalezieux: Good point, as discussed I removed `BasicStrategy`

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,250 @@
...
+    grayscale=False,
+    device=device,
+    img_size=img_size,
+    resize_mode="resize",
+)
+
+plot(clean_image, titles="Original Image")
+
+# %%
+# OverlapTilingStrategy creates overlapping patches with padding to provide context at boundaries, reducing artifacts.
```
tachella:
```suggestion
# :func:`deepinv.distributed.strategies.distributed_strategies.OverlapTilingStrategy <OverlapTilingStrategy>` creates overlapping patches with padding to provide context at boundaries, reducing artifacts.
```

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,250 @@
...
+"""
+
+import matplotlib.pyplot as plt
+from matplotlib import patheffects as pe
+from matplotlib.patches import Rectangle
+from matplotlib.patches import Patch
+import numpy as np
+import torch
+
+from deepinv.distributed.strategies.distributed_strategies import (
```
tachella: I think this import is a bit too long, can we do 
```suggestion
from deepinv.distributed import (
```
instead?

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,250 @@
...
+plt.show()
+
+
+# %%
+# Summary
+# -------
+#
+# Display summary showing patch counts, stride settings, and overlap parameters.
+
+print("\n" + "=" * 60)
```
tachella:
```suggestion
```
remove

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,250 @@
...
+print(f"SUMMARY\nImage: {img_size}×{img_size}")
+print(
+    f"OverlapTilingStrategy: {overlap_strategy.get_num_patches()} patches, stride={patch_size}, overlap={overlap}, Patch: {overlap_strategy._metadata['window_shape']}"
+)
+print("=" * 60)
```
tachella:
```suggestion
```

examples/distributed/demo_tiling_strategies.py
```diff
@@ -0,0 +1,250 @@
+r"""
```
tachella: I'm not sure this example is super relevant, it plots how overlapping patches are taken from an image, which it's not super informative...

I would remove it, unless we want to add more comparisons of tiling strategies and other features

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,317 @@
+r"""
+Distributed Training of an Unfolded DRS Algorithm on Urban100
```
tachella:
```suggestion
Distributed Training of Unfolded Networks
```

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,317 @@
...
+The `deepinv.distributed` framework enables training a model on multiple GPUs, by carefully parallelizing the data fidelity and denoising steps inside the network.
+
+This example shows how to combine:
+
+- the distributed framework (image/model parallelism over large images),
+- unfolded optimization with :class:`deepinv.optim.DRS`,
+- standard training with :class:`deepinv.Trainer`.
+
+Each GPU (rank) processes different parts/operators of the same image. This is not
+standard data-parallel training over different images.
```
tachella:
```suggestion
standard data-parallel training (e.g., via :class:`torch.nn.parallel.DistributedDataParallel`) over different images.
```

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,317 @@
...
+                "Before training",
+                "After training",
+            ],
+            save_fn="distributed_unrolled_result.png",
+        )
+        if train_history and val_history:
+            plot_curves({"train_psnr": [train_history], "val_psnr": [val_history]})
+
+        print("Saved: distributed_unrolled_result.png")
+        print("=" * 78)
```
tachella:
```suggestion
```

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,317 @@
...
+# - call :func:`deepinv.distributed.distribute` to distribute the physics and model across ranks
+# - Train with :class:`deepinv.Trainer` as usual.
+# The framework takes care of synchronizing the forward/backward passes across ranks, and communicating the necessary information between them.
+
+
+# Keep identical random streams across ranks: this framework splits each image
+# across devices, so all ranks should consume the same minibatches.
+with DistributedContext(seed=seed, seed_offset=False) as ctx:
+    if ctx.rank == 0:
+        print("=" * 78)
```
tachella:
```suggestion
```

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,317 @@
...
+        print("Distributed Unfolded DRS Demo (Urban100)")
```
tachella:
```suggestion
```

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,317 @@
...
+        print("=" * 78)
```
tachella:
```suggestion
```

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,317 @@
...
+        epochs=epochs,
+        device=ctx.device,
+        losses=[dinv.loss.SupLoss(metric=dinv.metric.MSE())],
+        metrics=psnr_metric,
+        optimizer=optimizer,
+        train_dataloader=train_loader,
+        eval_dataloader=val_loader,
+        grad_clip=1.0,
+        compare_no_learning=False,
+        save_path=f"ckpts/distributed_unfolded_drs_rank{ctx.rank}",
```
tachella: how do you load the model afterwards? might be worth saying it in the end?

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,317 @@
...
+            titles=[
+                "Ground truth",
+                "One noisy measurement",
+                "Before training",
+                "After training",
+            ],
+            save_fn="distributed_unrolled_result.png",
+        )
+        if train_history and val_history:
+            plot_curves({"train_psnr": [train_history], "val_psnr": [val_history]})
```
tachella, line 337: I would only plot curves if there's more than 1 epoch (GPU-case), otherwise we get this

Image

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,317 @@
...
+    )
+
+    return stacked_physics, train_loader, val_loader
+
+
+# %%
+# Configuration
+# -----------------------------------------------------------------------------
+# Settings for training and distributed processing.
+# patch_size and overlap control the size of the image patches that each rank processes, and how much they overlap with each other.
```
tachella, line 192: add a .. note here saying that we train for a few epochs for to reduce the demo time, but that we recommend using more epochs to get optimal results

docs/source/user_guide/training/distributed.rst
```diff
@@ -0,0 +1,281 @@
+.. _distributed-training:
+
+Distributed Training
+====================
+
+The distributed framework can be used during training. This is useful when a
```
tachella:
```suggestion
The distributed framework can be used during training when a
```

docs/source/user_guide/distributed/training.rst
```diff
@@ -0,0 +1,281 @@
+.. _distributed-training:
```
tachella, line 1: I think it would be clearer to have a 'Distributed computing' section after 'Training, Testing and Utilities' (and at the same level of 'Reconstruction') rather than having one subsection inside each of them..

This would also allow doing something like

```
Training, Testing and Utilities
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

....


Distributed Computing
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

In large-scale image reconstruction problems, one GPU might not be enough to handle large images/volumes or large forward operators. We provide multi-GPU tools specific to image reconstruction with `deepinv.distributed`:

::
    physics = deepinv.distributed.distribute(physics) # distrubute physics
    denoiser = deepinv.distributed.distribute(denoiser) # distribute denoiser

two blocks here, one for distributed inference, and other for distributed training (the two blocks that you already did :))
```

docs/source/user_guide/distributed/training.rst
```diff
@@ -0,0 +1,281 @@
...
+
+Distributed Training
```
tachella, line 3: This doc doesn't seem to appear in the generated docs, I think it's missing from the toctree

deepinv/distributed/strategies/distributed_strategies.py
```diff
@@ -181,114 +180,6 @@ def unpack_batched_results(
...
-        r"""Simple assignment of patches to output tensor."""
-        for idx, patch in local_pairs:
-            out_tensor[self._patch_slices[idx]] = patch
-
-    def get_num_patches(self) -> int:
-        r"""Return total number of patches."""
-        return len(self._patch_slices)
-
-
 class OverlapTilingStrategy(DistributedSignalStrategy):
```
tachella:
```suggestion
class TilingStrategy(DistributedSignalStrategy):
```
I think we can simplify the name here if we don't make the distinction between strategies

deepinv/distributed/strategies/distributed_strategies.py
```diff
@@ -466,7 +357,7 @@ def create_strategy(
     r"""
     Create a distributed signal strategy by name.
 
-    :param str strategy_name: name of the strategy (`'basic'`, `'overlap_tiling'`).
+    :param str strategy_name: name of the strategy (`'overlap_tiling'`).
```
tachella: Can we get rid of this? If we are having a single strategy?

tachella (changes requested):
Thanks a lot for taking into account my previous comments! I think the PR is converging :)

I have a few more comments with the organization of the user guide and examples

deepinv/distributed/distribute.py
```diff
@@ -211,6 +222,165 @@ def data_fidelity_factory(
...
+    checkpoint_batches: str = "auto",
+    checkpoint_use_reentrant: bool = False,
+    checkpoint_preserve_rng_state: bool = True,
+    dtype: torch.dtype | None = torch.float32,
+) -> BaseOptim:
+    r"""
+    In-place distribute a :class:`deepinv.optim.BaseOptim` unfolded model.
+
+    Uses the structured interface of :class:`~deepinv.optim.BaseOptim` directly —
+    rather than an opaque module walk — to distribute each component precisely:
```
tachella:
```suggestion
    to distribute each component precisely:
```

Remove unnecessary comparison

docs/source/user_guide/training/distributed.rst
```diff
@@ -0,0 +1,281 @@
...
+        trainer.train()
+
+The rest of your training code can stay close to a standard
+:class:`deepinv.Trainer` workflow. The distributed objects handle the
+communication needed by the forward pass and by backpropagation.
+
+.. note::
+
+    See :ref:`sphx_glr_auto_examples_distributed_demo_unrolled_distributed.py`
+    for a complete example training an unfolded DRS model.
```
tachella:
```suggestion
    for a complete example training an unfolded model.
```

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,327 @@
...
+.. code-block:: bash
+
+    # Single process
+    python examples/distributed/demo_unrolled_distributed.py
+
+.. code-block:: bash
+
+    # Multi-process (2 ranks)
+    python -m torch.distributed.run --nproc_per_node=2 examples/distributed/demo_unrolled_distributed.py
+"""
```
tachella, line 28: Add a new section here titled

"Import modules" or similar, to make a break from the intro above

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,327 @@
...
+    # rebuild the same model and Trainer, then call
+    # `trainer.load_model(
+    #     "ckpts/distributed_unfolded_drs/<timestamp>/ckp_best.pth.tar"
+    # )`.
+    # No rank-specific path or distributed checkpoint API is needed.
+
+    with torch.no_grad():
+        demo_rec_after = model(demo_y, distributed_physics)
+
+    # %%
```
tachella: remove this since it doesn't render as a new section in Sphinx

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,327 @@
...
+    # Display training summary and qualitative result (rank 0 only)
+    # -------------------------------------------------------------------------
```
tachella: remove this line
```suggestion
```

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,327 @@
...
+    return stacked_physics, train_loader, val_loader
+
+
+# %%
+# Configuration
+# -----------------------------------------------------------------------------
+# Settings for training and distributed processing.
+# patch_size and overlap control the size of the image patches that each rank processes, and how much they overlap with each other.
+
+# .. note::
```
tachella, line 194: this is not rendering ok, I think it's because of the missing #. Please double check once you applied the fix

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,327 @@
...
+
+    return stacked_physics, train_loader, val_loader
+
+
+# %%
+# Configuration
+# -----------------------------------------------------------------------------
+# Settings for training and distributed processing.
+# patch_size and overlap control the size of the image patches that each rank processes, and how much they overlap with each other.
+
```
tachella:
```suggestion
*
```

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,327 @@
...
+        )
+    if isinstance(ys[0], (list, tuple)):
+        n_ops = len(ys[0])
+        return x_batch, [torch.stack([yy[i] for yy in ys], dim=0) for i in range(n_ops)]
+    return x_batch, torch.stack(ys, dim=0)
+
+
+def prepare_dataset(
+    ctx: DistributedContext,
+    *,
```
tachella: is this * really needed?

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,327 @@
...
+from deepinv.utils.plotting import plot, plot_curves
+from deepinv.utils.tensorlist import TensorList
+
+# %%
+# Helper functions
+# -----------------------------------------------------------------------------
+# These functions assist with data collation, device management, and dataset preparation.
+
+
+def collate_batch(batch):
```
tachella, line 66: all the tensor list code is not really used in this example? I would remove then, it's adding more boilerplate

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,327 @@
...
+from deepinv.optim.prior import PnP
+from deepinv.physics import Denoising, GaussianNoise, stack
+from deepinv.utils import get_data_home
+from deepinv.utils.plotting import plot, plot_curves
+from deepinv.utils.tensorlist import TensorList
+
+# %%
+# Helper functions
+# -----------------------------------------------------------------------------
+# These functions assist with data collation, device management, and dataset preparation.
```
tachella: Please be more explicit of what's the goal of this section - "Helper functions" is too generic.

I would say something like


```suggestion
In this example, we parallelize ... operators.. 
We also introduce a batch preprocessing function that ...
```

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,327 @@
...
+    """Create/load Urban100 measurements for training and validation.
+
+    Important: all ranks iterate over the same batches since distribution is over
+    image content/operators, not over different images.
+    """
+    noise_levels = (0.06, 0.08)
+    physics_list = []
+    for i, sigma in enumerate(noise_levels):
+        torch.manual_seed(seed + i)
+        physics_list.append(Denoising(noise_model=GaussianNoise(sigma=sigma)))
```
tachella: Can we move to a mixed operator here? Eg blur and denoising?

My worry is that if you have 2 observations with Gaussian noise, the right approach is to construct the sufficient statistic:

Image

which defeats the point of using the distributed physics...

docs/source/user_guide/distributed/reconstruction.rst
```diff
@@ -16,8 +16,8 @@ The framework provides an API centered around two key functions:
     The distributed framework is particularly useful when:
     
     - *Multiple physics operators* with individual measurements need to be processed in parallel
-    - *Large images* are too large to fit in a single device's memory
-    - *Denoising priors* need to be applied to large images using spatial tiling
+    - *Large images or volumes* are too large to be processed in a single device's memory
```
tachella, line 19: I have an important question here:

The distributed tools assume that one fully copy of the signal fits into a single GPU? Or is it possible to keep global copies on CPU?

deepinv/tests/test_distributed.py
```diff
@@ -1639,7 +1642,887 @@ def _test_distribute_helper_data_fidelity_worker(rank, world_size, args):
...
+            # But x.grad accumulates, so it's fine.
+            # actually we don't need retain_graph for x, but we need it for intermediate graph if shared?
+            # independent branches only join at x.
+            for i in range(len(y_dist) - 1, -1, -1):
+                (y_dist[i].sum() * grad_scaler).backward()
+        else:
+            loss_dist = sum([yi.sum() for yi in y_dist])
+            loss_dist.backward()
+
+        # Aggregate gradients from all ranks IS NOW AUTOMATIC via DistributedGradientSync!
```
tachella: should we remove this comment, and the commented lines below?

deepinv/tests/test_distributed.py
```diff
@@ -1639,7 +1642,887 @@ def _test_distribute_helper_data_fidelity_worker(rank, world_size, args):
...
+            return "failure: no param gradients"
+
+        if x_grad_dist is None:
+            return "failure: no input gradients"
+
+        # Check against reference simulated LOCALLY
+        local_weight_grad = denoiser.conv.weight.grad
+        local_bias_grad = denoiser.conv.bias.grad
+
+        # Gradient reduction is now AUTOMATIC via DistributedParameterSync!
```
tachella: remove

deepinv/tests/test_distributed.py
```diff
@@ -1639,7 +1642,887 @@ def _test_distribute_helper_data_fidelity_worker(rank, world_size, args):
...
+        # if ctx.use_dist:
+        #    # Reduce gradients to see if they match the global run
+        #    dist.all_reduce(local_weight_grad, op=dist.ReduceOp.SUM)
+        #    dist.all_reduce(local_bias_grad, op=dist.ReduceOp.SUM)
+
+        # Now local_weight_grad contains the SUM of gradients from all ranks.
```
tachella: remove

deepinv/tests/test_distributed.py
```diff
@@ -1639,7 +1642,887 @@ def _test_distribute_helper_data_fidelity_worker(rank, world_size, args):
...
+        # This should match the gradient if we ran the whole thing on one GPU.
+
+        # Let's compute the Single-GPU reference locally
+        from deepinv.distributed.strategies import create_strategy
```
tachella: why the import here?

deepinv/distributed/framework/distributed_processing.py
```diff
@@ -0,0 +1,271 @@
...
+        4. Combining processed patches back into the full signal with proper overlap handling
+
+    The processor can be any callable that operates on tensors (e.g., denoisers, priors,
+    neural networks, etc.). The class handles all distributed coordination automatically.
+
+    |sep|
+
+    **Example use cases:**
+
+        - Distributed denoising of large images/volumes
```
tachella: remove indentation of all the bullet points

Image

deepinv/distributed/framework/distributed_processing.py
```diff
@@ -0,0 +1,271 @@
...
+from deepinv.distributed.framework.distributed_context import DistributedContext
+
+
+class DistributedProcessing(torch.nn.Module):
+    r"""
+    Distributed signal processing using pluggable tiling and reduction strategies.
+
+    This class enables distributed processing of large signals (images, volumes, etc.) by:
+
+        1. Splitting the signal into patches using a chosen strategy
```
tachella: remove indentation in all bullet points

deepinv/distributed/framework/distributed_processing.py
```diff
@@ -0,0 +1,271 @@
...
+        Should accept a batched tensor of shape ``(B, C, ...)`` and return a tensor of the same shape.
+        Examples: denoiser, neural network, prior gradient function, etc.
+    :param DistributedSignalStrategy | None strategy: signal processing strategy for patch extraction
+        and reduction. Either a custom strategy instance or `None`, which corresponds to the default tiling strategy.
+    :param dict | None strategy_kwargs: additional keyword arguments passed to the strategy constructor
+        when using string strategy names. Examples: ``patch_size``, ``overlap``, ``tiling_dims``. Default is `None`.
+    :param int | None max_batch_size: maximum number of patches to process in a single batch.
+        If ``None``, all local patches are batched together. Set to ``1`` for sequential processing
+        (useful for memory-constrained scenarios). Higher values increase throughput but require more memory. Default is `None`.
+    :param str checkpoint_batches: activation checkpointing mode for patch-batches during backward.
```
tachella: This point and the one below doesn't seem to be rendering well. Can you please fix and double check that it renders well?

Image

deepinv/distributed/framework/distributed_processing.py
```diff
@@ -0,0 +1,271 @@
...
+        2. Distributing patches across multiple processes/GPUs
+        3. Processing each patch independently using a provided processor function
+        4. Combining processed patches back into the full signal with proper overlap handling
+
+    The processor can be any callable that operates on tensors (e.g., denoisers, priors,
+    neural networks, etc.). The class handles all distributed coordination automatically.
+
+    |sep|
+
+    **Example use cases:**
```
tachella: I would give a snippet on how to use the class rather than text?

deepinv/distributed/framework/distributed_physics.py
```diff
@@ -0,0 +1,568 @@
...
+from deepinv.distributed.framework.distributed_utils import (
+    DistributedGradientSync,
+    map_reduce_gather,
+)
+from deepinv.distributed.framework.distributed_context import DistributedContext
+
+
+class DistributedStackedPhysics(Physics):
+    r"""
+    Holds only local physics operators. Exposes fast local and compatible global APIs.
```
tachella: This is a bit cryptic to me? Do we need this line?

deepinv/distributed/framework/distributed_physics.py
```diff
@@ -0,0 +1,568 @@
...
+
+    .. note::
+
+        It is intended to parallelize models naturally expressed as a stack/list of operators
+        (e.g., :class:`deepinv.physics.StackedPhysics` or an explicit Python list of
+        :class:`deepinv.physics.Physics` objects) and is **not** meant to split a
+        single monolithic physics operator across ranks.
+
+    If your forward model is a single operator that can be decomposed into multiple
+    sub-operators, it is up to you to perform that decomposition (e.g., build a
```
tachella:
```suggestion
    sub-operators, you can build any custom decomposition (e.g., build a
```

deepinv/distributed/framework/distributed_physics.py
```diff
@@ -0,0 +1,568 @@
...
+
+        Default is `'concatenated'`.
+    """
+
+    def __init__(
+        self,
+        ctx: DistributedContext,
+        num_operators: int,
+        factory: Callable[[int, torch.device, dict | None], Physics],
+        *,
```
tachella, line 52: is this necessary?

deepinv/distributed/framework/distributed_data_fidelity.py
```diff
@@ -0,0 +1,288 @@
...
+    Distributed data fidelity term for use with distributed physics operators.
+
+    This class wraps a standard DataFidelity object and makes it compatible with
+    DistributedStackedLinearPhysics by implementing efficient distributed computation patterns.
+    It computes data fidelity terms and gradients using local operations followed by
+    a single reduction, avoiding redundant communication.
+
+    The key operations are:
+
+        - ``fn(x, y, physics)``: Computes the data fidelity :math:`\sum_i d(A_i(x), y_i)`
```
tachella: remove indentation

deepinv/distributed/framework/distributed_data_fidelity.py
```diff
@@ -0,0 +1,288 @@
...
+        - ``grad(x, y, physics)``: Computes the gradient :math:`\sum_i A_i^T \nabla d(A_i(x), y_i)`
+
+    Both operations use an efficient pattern:
+
+        1. Compute local forward operations :math:`A_i`
```
tachella: remove indentation

deepinv/distributed/framework/distributed_data_fidelity.py
```diff
@@ -0,0 +1,288 @@
...
+    """
+
+    def __init__(
+        self,
+        ctx: DistributedContext,
+        data_fidelity: (
+            DataFidelity | Callable[[int, torch.device, dict | None], DataFidelity]
+        ),
+        num_operators: int | None = None,
+        *,
```
tachella, line 55: do you need this?

tachella (changes requested):
many thanks for incorporating my previous comments!! Please see a new round of comments, I think we are getting closer!

One key question I have (and I guess we need to clarify in the docs): the distributed computing still assumes that a whole copy of the volume/image fits in a single device? Or is there a way to keep it on cpu?

examples/distributed/demo_unrolled_distributed.py
bmalezieux: Actually only the tensor list code is used, i kept only this

docs/source/user_guide/distributed/reconstruction.rst
bmalezieux: For now the full signal should fit the APU memory, since it's necessary for physics computations when they act globally. I'll modify the doc.

deepinv/distributed/framework/distributed_physics.py
bmalezieux: I think it is, it makes optional arguments keyword-only, so it makes the API safer and clearer in my opinion. I would leave it like that.

deepinv/distributed/framework/distributed_data_fidelity.py
bmalezieux: Same as for distributed_physics, i would leave it for API clarity and safety.

deepinv/distributed/framework/distributed_context.py
```diff
@@ -0,0 +1,300 @@
...
+        # Get number of processes and ranks
+        self.local_world_size = int(os.environ.get("LOCAL_WORLD_SIZE", 1))
+        self.local_rank = int(os.environ.get("LOCAL_RANK", 0))
+        self.rank = int(os.environ.get("RANK", 0))
+        self.world_size = int(os.environ.get("WORLD_SIZE", 1))
+
+        if should_init_pg:
+            backend = self.backend
+            if backend is None:
+                # Backend decision considering device_mode:
```
tachella: remove this comment, as it is bloating a bit the code

deepinv/distributed/framework/distributed_context.py
```diff
@@ -0,0 +1,300 @@
...
+                self.device = torch.device(f"cuda:{dev_index}")
+                torch.cuda.set_device(self.device)
+            else:
+                self.device = torch.device("cpu")
+
+        self._post_init_setup()
+        return self
+
+    def __exit__(self, exc_type, exc, tb):
+        # Only destroy process group if:
```
tachella: remove comment

docs/source/user_guide.rst
```diff
@@ -282,4 +273,46 @@ Reconstruction networks ``model`` can be trained on datasets to improve their pe
...
+            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+            Distributed training for large-scale images and volumes.
+
+
+.. toctree::
+   :maxdepth: 2
+   :hidden:
+   :caption: Distributed Computing
+
+   user_guide/reconstruction/distributed
```
tachella: Create a separate folder user_guide/distributed/ and put the two files in there

docs/source/user_guide/training/distributed.rst
```diff
@@ -0,0 +1,281 @@
...
+
+How Backward Works
+------------------
+
+You normally do not need to write custom backward code. The distributed objects
+are built so that PyTorch autograd can follow the full computation.
+
+**Through distributed physics**: each rank applies its local operators
+:math:`A_i`. During backward, the gradient with respect to the shared input is
+computed from local contributions and synchronized across ranks. For linear
```
tachella: I'm not sure what

"For linear physics, adjoints are reduced so that the model sees
the gradient of the full stacked problem, not only the local operators."

means.

Do we need this here? Does the backward also work for non linear physics?

docs/source/user_guide/distributed/training.rst
```diff
@@ -0,0 +1,281 @@
...
+   * - Mode
+     - Description
+   * - ``"auto"``
+     - Checkpoint patch batches only when gradients are enabled and there are multiple local batches.
+   * - ``"always"``
+     - Checkpoint patch batches whenever gradients are enabled.
+   * - ``"never"``
+     - Disable checkpointing.
+
+Set ``max_batch_size=1`` to process local patches sequentially when memory is
```
tachella, line 228: might be worth saying here what is max_batch_size by default

docs/source/user_guide/training/distributed.rst
```diff
@@ -0,0 +1,281 @@
...
+:class:`deepinv.distributed.DistributedContext` detects whether distributed
+environment variables are present and selects the device for each rank.
+
+
+Troubleshooting
+---------------
+
+**Training hangs**
+
+- Make sure every rank enters the same training loop.
```
tachella: not sure this is very helpful for the user, what should they do?

examples/distributed/demo_unrolled_distributed.py
```diff
@@ -0,0 +1,343 @@
...
+        compare_no_learning=False,
+        save_path=checkpoint_root,
+        verbose=(ctx.rank == 0),
+        show_progress_bar=(ctx.rank == 0),
+        freq_update_progress_bar=5,
+        check_grad=True,
+        non_blocking_transfers=False,
+    )
+    trainer.train()
+    # Reload checkpoints with the usual DeepInverse Trainer API. In this run,
```
tachella: move this to the docstring of the subsection

deepinv/distributed/framework/distributed_utils.py
```diff
@@ -190,6 +256,147 @@ def reduce_local_results(
...
+        if autograd_ctx.dist_ctx.use_dist and grad_output is not None:
+            higher_order_path = grad_output.requires_grad
+            # Use returned tensor so functional all_reduce paths (create_graph=True)
+            # are preserved for higher-order differentiation.
+            grad_output = autograd_ctx.dist_ctx.all_reduce(
+                grad_output, op=dist.ReduceOp.SUM
+            )
+            # First-order training expects "mean across ranks" semantics:
+            # all_reduce gives a SUM, so we divide by world_size.
+            # Higher-order path (grad_output.requires_grad=True): this gradient is
```
tachella: remove this comment from "Higher-order ..."

tachella (changes requested):
Many thanks for the changes, should be a last set of comments :)!

tachella (approved):
LGTM, thanks Benoit !

PR #1093, opened 2026-03-05
Fix generator mixture and extend test suite in test_generators.py

### LINKED ISSUE #1090: GeneratorMixture raises error at initialization when provided random generator is on "cuda" device
### Environment

All

### Describe the bug

The current implementation of  `GeneratorMixture` does not take a device as an argument, so its device always defaults to "cpu". When the user provides an already existing `torch.Generator` that lives on a cuda device, it raises an error as the init function of `PhysicsGenerator` checks its device ("cpu" by default) against the generator's device.

On a side note, this also means that if the user does not provide a generator, the rng of a `GeneratorMixture` always lives on the "cpu".

### Steps to reproduce

```python
import torch
import deepinv as dinv

rng = torch.Generator(device="cuda")

mb_generator = dinv.physics.generator.MotionBlurGenerator(
    psf_size=(3, 3), num_channels=1, device="cuda", rng=rng
)
sigma_generator = dinv.physics.generator.SigmaGenerator(
    device="cuda", rng=rng
)   

# cannot specify device
generator = dinv.physics.generator.GeneratorMixture(
    generators=[mb_generator, sigma_generator],
    probs=[0.5, 0.5],
    rng=rng
)
```

### Expected behavior

Add a `device` argument to `GeneratorMixture`.

_No response_

### Additional context

_No response_

romainvo:
This PR fixes #1090: `GeneratorMixture` raised an error when initialized with a `torch.Generator` on a CUDA device.

New feature:
- If the mixed generators are compatible (in terms of keys and params shape), the `GeneratorMixture` automatically samples a different generator for each batch element. Previously, a single generator was sampled for the whole batch.

I have also extended the test suite in `test_generators.py`:
- uses the fixtures defined in `conftest.py` (for device mainly)
- I have added a fixture for the random generator, now every `PhysicsGenerator` is tested with rng=None and rng=torch.Generator(device)
- I have added a test for `GeneratorMixture` which was not tested for the moment

### DIFF
--- a/deepinv/physics/generator/base.py
+++ b/deepinv/physics/generator/base.py
@@ -309,8 +309,10 @@ class GeneratorMixture(PhysicsGenerator):
                 for key, value in params.items():
                     if key not in result:
                         result[key] = torch.empty(
-                            batch_size, *value.shape[1:],
-                            dtype=value.dtype, device=value.device
+                            batch_size,
+                            *value.shape[1:],
+                            dtype=value.dtype,
+                            device=value.device,
                         )
                     result[key][batch_positions] = value
 


deepinv/physics/generator/base.py
```diff
@@ -212,15 +212,61 @@ def __init__(
...
+        for g in generators:
+            params = g.step(batch_size=1)
+            generators_params.append(params)
+            generators_keys.append(set(params.keys()))
+
+        for i in range(len(generators_keys)):
+            for j in range(i + 1, len(generators_keys)):
+                if generators_keys[i] != generators_keys[j]:
+                    if verbose:
+                        print(
```
Andrewwango: use warn instead?

deepinv/physics/generator/base.py
```diff
@@ -212,15 +212,61 @@ def __init__(
...
     ) -> None:
-        super().__init__(rng=rng)
+        super().__init__(device=device, rng=rng)
         probs = torch.tensor(probs)
         assert torch.sum(probs) == 1, "The sum of the probabilities must be 1."
         self.generators = generators
         self.probs = probs
         self.cum_probs = torch.cumsum(probs, dim=0)
 
+        self.batch_compatible = self._compatible_generators(generators, verbose=verbose)
```
Andrewwango: My main worry here is that some generators' `step` are slow (e.g. https://deepinv.github.io/deepinv/api/stubs/deepinv.physics.generator.PolyOrderMaskGenerator.html or we don't have it yet in in deepinv, but in particular poisson disk mask generation). But I don't have a solution, and maybe this is ok for now

deepinv/physics/generator/base.py
```diff
@@ -230,6 +276,39 @@ def step(self, batch_size: int = 1, seed: int = None, **kwargs):
...
+            p = torch.rand(batch_size, generator=self.rng, device=self.device)
+            # Get generator index for each sample
+            generator_indices = torch.searchsorted(self.cum_probs, p)
+
+            # Group batch elements by their assigned generator
+            result = {}
+            for gen_idx, generator in enumerate(self.generators):
+                mask = generator_indices == gen_idx
+                if mask.any():
+                    # Find which batch positions use this generator
```
Andrewwango, line 301: @jscanvic  do you fear we descend into a similar chaos here as we did with the transforms `n_trans > 1, batch_size > 1` situation?

deepinv/physics/generator/base.py
romainvo: You fear the evaluation of `_compatible_generators` will be too slow you mean? I I could still add an option for not sampling different generators per batch, this way we can still keep the efficient approach if need be

deepinv/physics/generator/base.py
jscanvic: It definitely looks like it to me @Andrewwango

romainvo: What would be potentially problematic here? @Andrewwango @jscanvic

deepinv/physics/generator/base.py
romainvo: I've added a `use_batch_sampling` option in case the user mixes slow generators and wants to disable it.

deepinv/physics/generator/base.py
```diff
@@ -189,6 +190,10 @@ class GeneratorMixture(PhysicsGenerator):
 
     :param list[PhysicsGenerator] generators: the generators instantiated from :class:`deepinv.physics.generator.PhysicsGenerator`.
     :param list[float] probs: the probability of each generator to be used at each step
+    :param: bool use_batch_sampling: whether to sample a different generator for each element in the batch. This is only possible if all generators in the mixture produce parameters with the same keys and shapes. If not, a single generator will be sampled per batch. Defaults to True.
```
mh-nguyen712: This is super cool, thanks Romain!
```suggestion
    :param: bool use_batch_sampling: whether to sample a different generator for each element in the batch. This is only possible if all generators in the mixture produce parameters with the same keys and shapes. If not, a single generator will be sampled per batch. Defaults to `True`.
```

deepinv/physics/generator/base.py
```diff
@@ -212,15 +217,66 @@ def __init__(
         self,
         generators: list[PhysicsGenerator],
         probs: list[float],
+        use_batch_sampling: bool = True,
+        device: str | torch.device = "cpu",
         rng: torch.Generator = None,
```
mh-nguyen712, line 222: A note one rng: since each generator in the list `generators` also has their own rng, should we overwrite them by the rng from Mixture?

deepinv/physics/generator/base.py
```diff
@@ -230,6 +286,39 @@ def step(self, batch_size: int = 1, seed: int = None, **kwargs):
...
+            generator_indices = torch.searchsorted(self.cum_probs, p)
+
+            # Group batch elements by their assigned generator
+            result = {}
+            for gen_idx, generator in enumerate(self.generators):
+                mask = generator_indices == gen_idx
+                if mask.any():
+                    # Find which batch positions use this generator
+                    batch_positions = torch.where(mask)[0]
+                    num_samples = mask.sum().item()
```
mh-nguyen712:
```suggestion
                    num_samples = batch_positions.size(0)

File: deepinv/physics/generator/base.py [hunk 7]
```diff
@@ -230,6 +286,39 @@ def step(self, batch_size: int = 1, seed: int = None, **kwargs):
...
+                        for i, pos in enumerate(batch_positions):
+                            result[key][pos.item()] = value[i]
+
+            # Convert lists back to stacked tensors
+            return {k: torch.stack(v, dim=0) for k, v in result.items()}
+
+        else:
+            p = torch.rand(
+                1, generator=self.rng, device=self.device
+            ).item()  # np.random.uniform()
```
mh-nguyen712:
```suggestion
            )
```

deepinv/physics/generator/downsampling.py
```diff
@@ -113,7 +114,7 @@ def step(
         # unique factor for the whole batch to ensure that all produced measurements
         # have the same shape.
         factors = random_choice(
-            torch.as_tensor(self.list_factors),
+            torch.as_tensor(self.list_factors).to(self.device),
```
mh-nguyen712: call this in __init__

deepinv/physics/generator/base.py
romainvo: I like the idea of using a single generator for the whole. Nevertheless, I think enforcing it here could be risky e.g. if the user does

```python

rng = torch.Generator()
generator1 = MotionblurGenerator(psf=(15,15), rng=rng)
generator2 = MotionblurGenerator(psf=(31,31), rng=rng)
generator = GeneratorMixture([generator1, generator2], probs=[0.5,0.5])
```

here, if under the hood GeneratorMixture overwrite the rng of generator1 and generator2, then `rng` does not control the random state of the generator1 and generator2 anymore which is not good.

mh-nguyen712: Yes that why I propose to overwrite the rng of generator_1 and generator_2 by the one from mixture_generator. Or am I missing something?

deepinv/physics/generator/base.py
```diff
@@ -230,6 +286,39 @@ def step(self, batch_size: int = 1, seed: int = None, **kwargs):
...
+                    # Store with position information for later reordering
+                    for key, value in params.items():
+                        if key not in result:
+                            result[key] = [None] * batch_size
+
+                        for i, pos in enumerate(batch_positions):
+                            result[key][pos.item()] = value[i]
+
+            # Convert lists back to stacked tensors
+            return {k: torch.stack(v, dim=0) for k, v in result.items()}
```
mh-nguyen712: I think we could improve a bit this by pre-allocating tensor and do a vectorized assignment to avoid for loop:

```suggestion
                for key, value in params.items():
                    if key not in result:
                        result[key] = torch.empty(
                            batch_size, *value.shape[1:],
                            dtype=value.dtype, device=value.device
                        )
                    result[key][batch_positions] = value

            return result
```

deepinv/physics/generator/base.py
romainvo: What I meant is that after the mixture, the user might, for any reason, want to reset the state of `generator1` and `generator2` through the `rng` they instantiated themselves before the mixture
- with the current GeneratorMixture, it is possible
- but with the overriding part, we would strip this control from the user. If they did not pass a specific rng to GeneratorMixture, they would have to access `GeneratorMixture.rng` instead of the plain `rng they created themselves, which might not be convenient. 

In my opinion, having a single generator for the whole is good only if the user does it willingly, e.g by passing the same `rng` which controls generator1 and generator2, to the GeneratorMixture, in which case we dont need the overriding part.

Sorry if I'm not making sense.. (I've updated my comment above, but not sure it's clearer ^^')

deepinv/physics/generator/base.py
romainvo: Very nice 👍

deepinv/physics/generator/base.py
romainvo: Hello @mh-nguyen712, I feel like this feature raises concerns outside the scope of GeneratorMixture and is more about properly controlling the random state of possibly multiple PhysicsGenerator throughout some code. Since this is outside the scope of the PR, I can move the conversation to a separate issue?

mh-nguyen712: Hi @romainvo  totally agree. That just was a side note, feel free to skip for now.

PR #1099, opened 2026-03-11
Integrate non ToF PET with parallelproj

tachella:
- Integrate `PET` via `parallelproj`. 
- Adds 2D and 3D non-ToF PET examples
- Small fix to colorbar normalization


docs/source/user_guide/physics/physics.rst
```diff
@@ -79,6 +79,11 @@ This is particular useful when dealing with blind inverse problems or parameter
        | :class:`deepinv.physics.TomographyWithAstra`
      -
 
+   * - Positron Emission Tomography
+     -
+       | :class:`deepinv.physics.PET`
+     -
+
```
Tmodrzyk, line 86: This should be under `Tomography` along with the two other tomography classes.

deepinv/physics/pet.py
```diff
@@ -0,0 +1,483 @@
...
+
+if TYPE_CHECKING:
+    import parallelproj
+
+
+class PET(LinearPhysics):
+    r"""
+    Non time-of-flight Positron emission tomography (PET) physics model.
+
+    This operator relies on the `parallelproj` library :footcite:t:`schramm2024parallelproj`.
```
Tmodrzyk, line 17: Paranthesis around this citation otherwise it looks odd in the rendered doc:
```
This operator relies on the parallelproj library Schramm and Thielemans[1].
```

pyproject.toml
```diff
@@ -131,6 +131,9 @@ pytorch = "*"
 torchvision = "*"
 python = ">=3.10"
 
+[tool.pixi.feature.physics.dependencies]
+parallelproj = "*"
```
Tmodrzyk, line 135: Shouldn't we set a minimum version here ? 
E.g. to avoid people reporting bugs when they use very old versions of `parallelproj`.

deepinv/physics/pet.py
```diff
@@ -0,0 +1,483 @@
...
+    :meth:`physics.A <deepinv.physics.Physics.A>` or :meth:`physics.A_adjoint <deepinv.physics.LinearPhysics.A_adjoint>`.
+
+    .. note::
+
+        This operator requires the `parallelproj` package to be installed.
+        You can install it via conda/mamba:
+
+        ::
+
+            conda install -c conda-forge parallelproj
```
Tmodrzyk, line 45: It might be me but the installation was not straightforward from this note.
We should indicate that `deepinv` should be installed either through `pixi` or `conda` in order to handle this conda-only dependency.
If installed with `pixi`, the user should install the full environment with `pixi install -e full`.
Side note, (probably related to the pixi PR #1108 and not this one) I had warnings when trying to install `deepinv` with all additionnal dependencies:
```
❯ pixi add --pypi "deepinv[dataset,denoisers,physics]"
 WARN The package `deepinv==0.4.0` does not have an extra named `physics`
 WARN The package `deepinv==0.4.0` does not have an extra named `physics`
 WARN The package `deepinv==0.4.0` does not have an extra named `physics`
 WARN The package `deepinv==0.4.0` does not have an extra named `physics`
 WARN The package `deepinv==0.4.0` does not have an extra named `physics`
 WARN The package `deepinv==0.4.0` does not have an extra named `physics`
✔ Added deepinv[dataset,denoisers,physics] >=0.4.0, <0.5
Added these as pypi-dependencies.
```

Tmodrzyk: It seems that the pixi configuration is not working for CUDA. 
When running `pixi install -e full` to create a fresh venv and trying to import `parallelproj` (for instance when running one of the examples), I'm getting an error:
``ImportError: Cannot find parallelproj cuda lib. Consider settting the environment variable PARALLELPROJ_CUDA_LIB.``

deepinv/physics/pet.py
```diff
@@ -0,0 +1,483 @@
...
+    :param tuple img_size: shape of the input 2D `(H, W)` or 3D volumes `(D, H, W)`.
+    :param tuple voxel_size: voxel size in mm. Default is 2 x 2 x 2 mm.
+    :param float fwhm_data_mm: full width at half maximum (FWHM) of the Gaussian blur :math:`g`. It has a crucial impact on the maximum achievable resolution,
+        which is typically a fraction of the FWHM.
+    :param None, parallelproj.pet_scanners.ModularizedPETScannerGeometry scanner: Scanner configuration. If None, the default scanner from parallelproj is used.
+    :param int radial_trim: radial trim of rays on the sides of the volume to improve efficiency.
+    :param float gain: gain factor :math:`\gamma` for the Poisson noise model.
+    :param bool normalize: If `True` the forward operator is normalized such that :math:`\|A\|=1`.
+    :param bool normalize_counts: If `False` the :math:`\gamma` term in from of the Poisson noise is removed,
+        that is the measurements are true counts.
```
Tmodrzyk, line 64:
```suggestion
    :param bool normalize_counts: If `False` the :math:`\gamma` term in front of the Poisson noise is removed,
        that is the measurements are true counts.
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,483 @@
...
+    :param bool projected_attenuation: If `False` (default), the attenuation should be provided in image space as :math:`\mu` (ie auxiliary CT scan).
+        If `True`, the attenuation should be provided in the image space as :math:`c`. This should be set as `False` to efficiently compute
+        gradients wrt the attenuation.
```
Tmodrzyk, line 67:
```suggestion
    :param bool projected_attenuation: If `False` (default), the attenuation should be provided in image space as :math:`\mu` (ie auxiliary CT scan).
        If `True`, the attenuation should be provided in the projection space as :math:`c`. This should be set as `False` to efficiently compute
        gradients with respect to the attenuation.
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,483 @@
...
+        **kwargs,
+    ) -> torch.Tensor:
+        r"""
+        Apply the linear operator :math:`Ax=c \circ H(g*x)` to a signal :math:`x`
+
+        :param torch.Tensor x: input image or volume
+        :param torch.Tensor add_background: whether to add background :math:`s`. By default, no background is added.
+        :param torch.Tensor background: If not `None`, update the background :math:`s` of the operator.
+        :param torch.Tensor attenuation: If not `None`, update the attenuation :math:`s` of the operator.
+        """
```
Tmodrzyk, line 212:
```suggestion
        :param torch.Tensor attenuation: If not `None`, update the attenuation (:math:`c` if `projected_attenuation=True`, :math:`\mu` if `projected_attenuation=False`) of the operator.
        """
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,483 @@
...
+            if not self.projected_attenuation:
+                if self.is_2d:
+                    attenuation = attenuation.unsqueeze(-1)
+
+                proj_att = LinearSingleChannelOperator.apply(attenuation, self.proj)
+                if self.is_2d:
+                    proj_att = proj_att.squeeze(-1)
+                self.attenuation = torch.exp(-proj_att)
+            else:
+                self.attenuation = attenuation
```
Tmodrzyk, line 306: We should be able to handle the case where a attenuation sinogram is provided, as we support it in the init. 
A good option would be to issue a warning if there is a mismatch between the `projected_attenuation` value and the shape of the provided `attenuation`, but to still proceed with the conversion to update the param.

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,223 @@
...
+:math:`g \in \mathbb{R}_{+}^{n}` is a Gaussian blur kernel, :math:`x\in\mathbb{R}_{+}^{n}`
+is the emission image, :math:`s \in \mathbb{R}_{+}^{m}` is the (expected) background,
+:math:`\mathcal{P}` denotes Poisson noise with gain :math:`\gamma > 0`,
+:math:`c=\exp(-H\mu)\in \mathbb{R}_{+}^{m}` is an (optional) attenuation term
+with :math:`\mu \in \mathbb{R}_{+}^{n}` an attenuation map (typically obtained through an auxiliary CT scan).
+
+.. note::
+
+    PET physics relies on the `parallelproj` library.
+    Please `download it <https://parallelproj.readthedocs.io/en/stable/installation.html>`_ to run this example.
```
Tmodrzyk, line 24: It would make the example easier for beginner to directly provide the install command, depending on the different ways supported to install deepinv (conda vs pixi)

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,223 @@
...
+# The maximum achievable resolution (in high count settings) is typically proportional to the full-width at half
+# maximum (FWHM) of the Gaussian blur kernel, which here is set to 4 mm.
+#
+# We use a PET scanner with a single ring of detectors, which is a polygon of
+# 32 sides, with each side containing 16 detectors. This gives us a total of 32*16=512 detectors in total.
+#
+# .. tip::
+#
+#       You can play with different geometries and voxel sizes to get a good grasp of
+#       the scanner geometry.
```
Tmodrzyk, line 52:
```suggestion
#       You can play with different geometries and voxel sizes to get a good grasp of
#       the scanner geometry, and visualize it with `physics.plot_geometry()`
```

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,223 @@
...
+#      In this example, we normalize the forward operator :math:`\|A\|=1` (`normalize=True`) and
+#      the Poisson counts to be between 0 and 1 (`normalize_counts=True`), to simplify reconstruction.
+#      If you want to use the operator with real PET measurements, you will need to
+#      carefully handle the normalization. See also the :ref:`normalized 2D PET example <sphx_glr_auto_examples_physics_demo_pet2d.py>`.
+
+device = "cuda" if torch.cuda.is_available() else "cpu"
+img_size = (128, 128)
+voxel_size = (3, 3)
+
+# number of sides of the polygone approximating a circle
```
Tmodrzyk, line 65:
```suggestion
# number of sides of the polygon approximating a circle
```

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,223 @@
...
+num_sides = 32
+
+# number of detectors per polygone side
```
Tmodrzyk, line 68:
```suggestion
# number of detectors per polygon side
```

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,223 @@
...
+scanner = parallelproj.pet_scanners.DemoPETScannerGeometry(
+    torch_compat,
+    dev=device,
+    num_rings=num_rings,
+    num_sides=num_sides,
+    num_lor_endpoints_per_side=num_lor_endpoints_per_side,
+)
+
+# gain of the device (associated with amount of dose/acq. time)
+# larger gain -> more poisson noise -> harder reconstruction
```
Tmodrzyk, line 83:
```suggestion
# gain of the device. 
# higher gains are associated to lower dose and/or shorter acquisition times, while lower gains are associated to higher dose and/or longer acquisition times.
# larger gain -> more poisson noise -> harder reconstruction
```

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,223 @@
...
+#
+# Both attenuation and background are stored as "physics parameters" which are patient dependent
+# and can be updated via :meth:`physics.update(...) <deepinv.physics.Physics.update>` or by passing them as kwargs in
+# :meth:`physics(x, ...) <deepinv.physics.Physics.forward>`, :meth:`physics.A(x, ...) <deepinv.physics.Physics.A>` or
+# :meth:`physics.A_adjoint(y, ...) <deepinv.physics.LinearPhysics.A_adjoint>`.
+#
+# .. note::
+#
+#   The attenuation is stored in the physics in sinogram space as :math:`\exp(-\mu)` to speed up computations,
+#   but it should be provided in image space, i.e., :math:`\mu`, to the physics.
```
Tmodrzyk, line 150: I already pointed this out in the implementation, but since we support both at initialization, we should also support both during inference.
See my specific comments in `pet.py`

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,223 @@
...
+dinv.utils.plot(
+    [physics.attenuation, y, y2], ["sino. atten.", "meas.", "corrected meas."]
+)
+
+# %%
+# Backprojection and sensitivities
+# --------------------------------
+# We backproject the data to visualize the sensitivity map of the scanner.
+# The sensitivity map is defined as the back-projection of a sinogram of ones, which corresponds to the number of rays intersecting each voxel.
+#
```
Tmodrzyk, line 166:
```suggestion
# The sensitivity map is defined as the back-projection of a sinogram of ones :math:`s = A^\top \mathbf{1}`, which corresponds to the number of rays intersecting each voxel.
#
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,483 @@
...
+    .. math::
+
+        y \sim \gamma \mathcal{P}(\frac{c \circ H(g*x) + s}{\gamma})
+
+    where :math:`H \in \mathbb{R}_{+}^{m \times n}` is the projection operator,
+    :math:`g \in \mathbb{R}_{+}^{n}` is a Gaussian blur kernel, :math:`x\in\mathbb{R}_{+}^{n}`
+    is the emission image, :math:`s \in \mathbb{R}_{+}^{m}` is the (expected) background,
+    :math:`\mathcal{P}` denotes Poisson noise with gain :math:`\gamma > 0`,
+    :math:`c=\exp(-H\mu)\in \mathbb{R}_{+}^{m}` is an (optional) attenuation term
+    with :math:`\mu \in \mathbb{R}_{+}^{n}` an attenuation map (typically obtained through an auxiliary CT scan).
```
Tmodrzyk, line 30: The character `s` is often reserved for the sensitivity of the system (column sums of the system matrix, A^\top 1). Having `b` for the background is more common.

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,223 @@
...
+# to obtain a reconstructed emission volume.
+#
+# The algorithm can be seen as a preconditioned gradient descent on the negative log-likelihood of the Poisson model:
+#
+# .. math::
+#
+#   x^{(k+1)} = x^{(k)} - P \nabla f(Ax^{(k)}+b,y)
+#
+# where :math:`f` is the Poisson data-fidelity term, :math:`P=\mathrm{diag}(\frac{x}{A^T\mathbf{1}})` is a preconditioner
+# and :math:`b` is the background.
```
Tmodrzyk, line 192: Here you used `b` for background, but `s` everywhere else. `b` everywhere should be preferred.

examples/physics/demo_pet3d.py
```diff
@@ -0,0 +1,235 @@
...
+#
+#       You can play with different geometries and voxel sizes to get a good grasp of
+#       the scanner geometry.
+#
+
+device = "cuda" if torch.cuda.is_available() else "cpu"
+img_size = (128, 128, 24)
+voxel_size = (3, 3, 3)
+
+# number of sides of the polygone approximating a circle
```
Tmodrzyk, line 71:
```suggestion
# number of sides of the polygon approximating a circle
```

examples/physics/demo_pet3d.py
```diff
@@ -0,0 +1,235 @@
...
+num_sides = 32
+
+# number of detectors per polygone side
```
Tmodrzyk, line 74:
```suggestion
# number of detectors per polygon side
```

examples/physics/demo_pet3d.py
```diff
@@ -0,0 +1,235 @@
...
+    to denormalize the reconstructions and obtain physical units.
+
+.. tip::
+
+    If you prefer to get started with PET on a simpler 2D problem, please check out :ref:`the 2D PET demo <sphx_glr_auto_examples_physics_demo_pet2d.py>`.
+
+.. note::
+
+    PET physics relies on the `parallelproj` library.
+    Please `download it <https://parallelproj.readthedocs.io/en/stable/installation.html>`_ to run this example.
```
Tmodrzyk, line 36: Same comment as the other demo file, providing the actual install command would probably be better for beginners

examples/physics/demo_pet3d.py
```diff
@@ -0,0 +1,235 @@
...
+# ------------------------------------
+#
+# We define a 3D phantom and attenuation map, whose shape is the same as the phantom.
+#
+# In practice, the attenuation is typically obtained with an auxiliary CT scan of the patient.
+
+x, attenuation = generate_pet_phantom(img_size, device=device)
+mid_slice = img_size[-1] // 2
+
+# longer acquisition times -> more counts -> easier reconstruction
```
Tmodrzyk, line 115: For consistency, we should put the same comment as the 2D exemple. See my specific comment in `demo_pet2d.py`.

deepinv/tests/test_physics.py
```diff
@@ -237,6 +239,38 @@ def find_operator(name, device, imsize=None, get_physics_param=False):
...
+        img_size = (1, 16, 16, 16) if imsize is None else imsize  # C,H,W
+        p = dinv.physics.PET(
+            img_size,
+            normalize=True,
+            device=device,
+            projected_attenuation=True,
+        )
+        p.noise_model = dinv.physics.ZeroNoise()
+        p.normalize = False  # stop auto-normalize to compute gradients wrt to attn
+        params = ["attenuation", "background"]
```
Tmodrzyk, line 273: We should add a test case where `projected_attenuation` is `False` to ensure that we handle both cases correctly

deepinv/utils/phantoms.py
```diff
@@ -128,3 +128,125 @@ def __getitem__(self, index):
...
+    device: str = "cpu",
+    mu_value: float = 0.01,
+    add_spheres: bool = True,
+    add_inner_cylinder: bool = True,
+    r0: float = 0.45,
+    r1: float = 0.28,
+    oversampling_factor=4,
+):
+    """
+    Generate a 3D PET phantom and attenuation.
```
Tmodrzyk, line 144:
```suggestion
    Generate a 2D or 3D PET phantom and attenuation.
```

deepinv/utils/phantoms.py
```diff
@@ -128,3 +128,125 @@ def __getitem__(self, index):
...
+
+    :param tuple img_shape: Shape of the input image `(H, W)` or volume `(H, W, D)`
```
Tmodrzyk, line 146:
```suggestion
    :param tuple img_shape: Shape of the output image `(H, W)` or volume `(H, W, D)`
```

deepinv/utils/phantoms.py
```diff
@@ -128,3 +128,125 @@ def __getitem__(self, index):
...
+    :param tuple device: Device to use
+    :param float mu_value: Initial mu value
+    :param bool add_spheres: Whether to add spheres
+    :param bool add_inner_cylinder: Whether to add cylinders
+    :param float r0: Radius of the sphere
+    :param float r1: Radius of the cylinder
```
Tmodrzyk, line 152: What happens if the user selects huge radii ? 
We should handle this with an error or a warning + clipping the values

Tmodrzyk (commented):
Thanks for the awesome PR @tachella !!
The two demos are amazing, and the visualization tool of `parallelproj` for the scanner geometry is also awesome. 
This will make PET experiments much easier. 

I have a few minor comments here and there. 
My main issue is with the configuration for the `parallelproj` install with `pixi` right now. With the current `pyproject.toml` file, I was unable to run the examples. It seems by default we install the CPU binaries of `parallelproj` but at runtime if a GPU is detected through `nvidia-smi`,  parallelproj  will try and load the CUDA libs and fail to do so. 
I forced `pixi` to install the CUDA libs for parallelproj with  
```
[tool.pixi.system-requirements]
cuda = "12"
```
and
```
[tool.pixi.feature.physics.dependencies]
parallelproj = "*"
libparallelproj = "*"
cuda-version = "12.*"
```
but I don't think this fix is correct. 
When the CUDA binaries are installed, we can run the code on CPU (no more error about loading CUDA libs that are not installed) however trying to use the GPU will result in other errors. 
Namely `cupy` seems to be required but is not installed. 
I did not try to install deepinv + parallelproj through `conda`.
We should make sure the install is working with both managers before merging.

Steps to reproduce my issues, on a CUDA-capable computer:
``pixi install -e full``
Then run the `demo_pet2d.py` example.

Edit: after several trials, I could not get `deepinv` to work with CUDA if installed through `conda`, even when installing the CUDA binaries for pytorch through conda beforehand

deepinv/physics/pet.py
tachella: not sure I understand your point. The code should work well when an attenuation sinogram is provided (as long as projected_attenuation=False)?

deepinv/physics/pet.py
tachella: ok, I've done the change

examples/physics/demo_pet2d.py
tachella: you're right, done

Tmodrzyk:
Hi @tachella thank you again for this awesome PR.
I was able to run the examples directly after installing the venv through pixi, so I can confirm everything is working perfectly now ! 
Let me know when you want a second round of review. :smile:

deepinv/physics/pet.py
Tmodrzyk: Sorry, my comment was unclear. 
I meant that we should not rely on the boolean ``projected_attenuation`` at inference, because this attribute is set at initialization. 
Users could pass a map at init and set ``projected_attenuation = False``, and then decide they want to pass a sinogram at inference. In this case the code silently breaks without raising any error or warnings. 
edit: it will break down the line with shape mismatch, but we might as well catch the error beforehand and print a more relevant message.

Tmodrzyk:
Apart from what's already in this amazing PR, it would be awesome to have other common phantoms for PET imaging:
- the NEMA IEC spheres: basically a bunch of radioactive spheres in an attenuated box (see the picture). Best would be to have an option for user-defined activity levels in each spheres.
image
- the Brainweb phantoms (e.g this https://github.com/casperdcl/brainweb), with the possibility to add spherical lesions 
image

These are both very common in experiments. It can of course be part of another PR. :smile:

docs/source/user_guide/physics/physics.rst
tachella: agreed

deepinv/physics/pet.py
tachella: thanks for spotting this, I've added a 'by'

pyproject.toml
tachella: agreed, I've added 1.10.0

deepinv/physics/pet.py
tachella: this has been fixed

deepinv/physics/pet.py
tachella: ok, I see now. I changed the implementation such that it automatically decides whether the attenuation is on the image or sinogram space based on img_size

examples/physics/demo_pet3d.py
tachella: done

examples/physics/demo_pet3d.py
tachella: yes, you're right. I added the same as in the other demo

examples/physics/demo_pet2d.py
tachella: I've added an explanation saying that both can be added (and done the same on the 3D example)

deepinv/tests/test_physics.py
tachella: shouldn't be necessary now with the automatic choice

deepinv/utils/phantoms.py
tachella: I added a much better description of the phantom + some example + raise errors if the radii are out of bounds

tachella:
> * the NEMA IEC spheres: basically a bunch of radioactive spheres in an attenuated box (see the picture). Best would be to have an option for user-defined activity levels in each spheres.

great idea! we could add them as feature requests

deepinv/physics/pet.py
```diff
@@ -0,0 +1,508 @@
...
+
+        batch_size = x.shape[0]
+        y = torch.zeros(
+            (batch_size, 1) + operator.in_shape, dtype=x.dtype, device=x.device
+        )
+
+        # loop over all samples in the batch and apply linear operator
+        # to the first channel
+        for i in range(batch_size):
+            y[i, 0, ...] = operator.adjoint(x[i, ...].detach())
```
Andrewwango:
```suggestion
            y[i, 0, ...] = operator.adjoint(x[i, 0, ...].detach())
```
seems to be a bug here.

deepinv/physics/pet.py
```diff
@@ -0,0 +1,508 @@
...
+        # since forward takes two input arguments (x, operator)
+        # we have to return two arguments (the latter is None)
+        if grad_output is None:
+            return None, None
+        else:
+            operator = ctx.operator
+
+            batch_size = grad_output.shape[0]
+            x = torch.zeros(
+                (batch_size,) + operator.out_shape,
```
Andrewwango:
```suggestion
                (batch_size, 1,) + operator.out_shape,
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,508 @@
...
+                dtype=grad_output.dtype,
+                device=grad_output.device,
+            )
+
+            # loop over all samples in the batch and apply linear operator
+            # to the first channel
+            for i in range(batch_size):
+                x[i, ...] = operator(grad_output[i, 0, ...].detach())
```
Andrewwango:
```suggestion
                x[i, 0, ...] = operator(grad_output[i, 0, ...].detach())
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,508 @@
...
+
+        :param ctx: context object
+            Context object that can be used to store information for the backward pass.
+        :param x: torch.Tensor
+            Mini-batch of 3D images with dimension (batch_size, 1, num_voxels_x, num_voxels_y, num_voxels_z).
+        :param operator: parallelproj.LinearOperator
+            Linear operator that can act on a single 3D image.
+
+        :return: torch.Tensor
+            Mini-batch of 3D images with dimension (batch_size, operator.out_shape).
```
Andrewwango:
```suggestion
            Mini-batch of 3D images with dimension (batch_size, 1, operator.out_shape).
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,508 @@
...
+        ctx.operator = operator
+
+        batch_size = x.shape[0]
+        c = x.shape[1]
+        y = torch.zeros(
+            (batch_size, c) + operator.out_shape, dtype=x.dtype, device=x.device
+        )
+
+        for i in range(batch_size):
+            for j in range(c):
```
Andrewwango: Always single channel data -> harcode c = 1

deepinv/physics/pet.py
```diff
@@ -0,0 +1,508 @@
...
+
+    @staticmethod
+    def backward(ctx, grad_output: torch.Tensor) -> tuple[torch.Tensor, None]:
+        r"""
+        Backward pass of the forward operation.
+
+        :param ctx: context object
+            Context object that can be used to obtain information from the forward pass.
+        :param grad_output: torch.Tensor
+            Mini-batch of gradients with dimension (batch_size, operator.out_shape).
```
Andrewwango:
```suggestion
            Mini-batch of gradients with dimension (batch_size, 1, operator.out_shape).
```

Andrewwango (changes requested):
some bugs

deepinv/physics/pet.py
tachella: done, I've also added some raise errors on the forward

deepinv/physics/pet.py
tachella: done

deepinv/physics/pet.py
tachella: done

deepinv/physics/pet.py
tachella: done

deepinv/physics/pet.py
tachella: done

deepinv/physics/pet.py
```diff
@@ -0,0 +1,512 @@
...
+    r"""
+    Non time-of-flight Positron emission tomography (PET) physics model.
+
+    This operator relies on the `parallelproj` library by :footcite:t:`schramm2024parallelproj`.
+
+    The PET forward model is defined as
+
+    .. math::
+
+        y \sim \gamma \mathcal{P}(\frac{c \circ H(g*x) + b}{\gamma})
```
Tmodrzyk:
```suggestion
        y \sim \gamma \mathcal{P}\left(\frac{c \circ H(g*x) + b}{\gamma}\right)
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,512 @@
...
+
+    where :math:`H \in \mathbb{R}_{+}^{m \times n}` is the projection operator,
+    :math:`g \in \mathbb{R}_{+}^{n}` is a Gaussian blur kernel, :math:`x\in\mathbb{R}_{+}^{n}`
+    is the emission image, :math:`b \in \mathbb{R}_{+}^{m}` is the (expected) background,
+    :math:`\mathcal{P}` denotes Poisson noise with gain :math:`\gamma > 0`,
+    :math:`c=\exp(-H\mu)\in \mathbb{R}_{+}^{m}` is an (optional) attenuation term
```
Tmodrzyk, line 29: The projection operator ``H`` of the PET scanner and the projection opetor ``H`` of the CT scanner are not equal.
We could use H_{CT} to avoid confusion.

deepinv/physics/pet.py
```diff
@@ -0,0 +1,512 @@
...
+    :param tuple img_size: shape of the input 2D `(H, W)` or 3D volumes `(D, H, W)`.
+    :param tuple voxel_size: voxel size in mm. Default is 2 x 2 x 2 mm.
+    :param float fwhm_data_mm: full width at half maximum (FWHM) of the Gaussian blur :math:`g`. It has a crucial impact on the maximum achievable resolution,
+        which is typically a fraction of the FWHM.
+    :param None, parallelproj.pet_scanners.ModularizedPETScannerGeometry scanner: Scanner configuration. If None, the default scanner from parallelproj is used.
+    :param int radial_trim: radial trim of rays on the sides of the volume to improve efficiency.
+    :param float gain: gain factor :math:`\gamma` for the Poisson noise model.
+    :param bool normalize: If `True` the forward operator is normalized such that :math:`\|A\|=1`.
+    :param bool normalize_counts: If `False` the :math:`\gamma` term in front of the Poisson noise is removed,
+        that is the measurements are true counts.
```
Tmodrzyk: The phrasing "If `False` the :math:`\gamma` term in front of the Poisson noise is removed, that is the measurements are true counts." seems odd

deepinv/physics/pet.py
```diff
@@ -0,0 +1,512 @@
...
+        background: torch.Tensor | None = None,
+        attenuation: torch.Tensor | None = None,
+        **kwargs,
+    ) -> torch.Tensor:
+        r"""
+        Apply the linear operator :math:`Ax=c \circ H(g*x)` to a signal :math:`x`
+
+        :param torch.Tensor x: input image or volume
+        :param torch.Tensor add_background: whether to add background :math:`s`. By default, no background is added.
+        :param torch.Tensor background: If not `None`, update the background :math:`s` of the operator.
```
Tmodrzyk:
```suggestion
        :param torch.Tensor add_background: whether to add background :math:`b`. By default, no background is added.
        :param torch.Tensor background: If not `None`, update the background :math:`b` of the operator.
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,512 @@
...
+        :param torch.Tensor attenuation: If not `None`, update the attenuation of the operator.
+            The space (image or sinogram) is inferred automatically from the tensor shape.
```
Tmodrzyk:
```suggestion
        :param torch.Tensor attenuation: If not `None`, update the attenuation :math:`c` of the operator.
            The space (image or sinogram) is inferred automatically from the tensor shape.
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,512 @@
...
+
+    def A_adjoint(
+        self, y: torch.Tensor, attenuation=None, background=None, **kwargs
+    ) -> torch.Tensor:
+        r"""
+        Apply the adjoint of the linear operator :math:`A^{\top}y` where :math:`A=c \circ H(g*\cdot)` to a sinogram :math:`y`
+
+        :param torch.Tensor y: input sinogram
+        :param torch.Tensor attenuation: If not `None`, update the attenuation
+        :param torch.Tensor background: If not `None`, update the background
```
Tmodrzyk:
```suggestion
        :param torch.Tensor attenuation: If not `None`, update the attenuation :math:`c` of the operator
        :param torch.Tensor background: If not `None`, update the background :math:`b` of the operator
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,512 @@
...
+
+    def forward(
+        self, x: torch.Tensor, attenuation=None, background=None, **kwargs
+    ) -> torch.Tensor:
+        r"""
+        Generate PET measurements.
+
+        :param torch.Tensor x: input image or volume
+        :param torch.Tensor attenuation: If not `None`, update the attenuation
+        :param torch.Tensor background: If not `None`, update the background
```
Tmodrzyk:
```suggestion
        :param torch.Tensor attenuation: If not `None`, update the attenuation :math:`c` of the operator
        :param torch.Tensor background: If not `None`, update the background :math:`b` of the operator
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,512 @@
...
+        r"""
+        Update the background and/or attenuation parameters.
+
+        The space of the attenuation tensor is inferred automatically: if the last
+        ``len(img_size)`` dimensions match ``img_size``, the tensor is treated as an
+        image-space attenuation map :math:`\mu` and projected; otherwise it is treated as
+        a sinogram-space attenuation :math:`c=\exp(-H\mu)` and used directly.
+
+        :param torch.Tensor attenuation: If not `None`, update the attenuation. Can be in
+            image space (shape matching `img_size`) or sinogram space.
```
Tmodrzyk:
```suggestion
        :param torch.Tensor attenuation: If not `None`, update the attenuation :math:`c` of the operator. Can be in
            image space (shape matching `img_size`) or sinogram space.
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,512 @@
...
+        which installs all optional dependencies.
+
+        Check the `parallelproj` documentation for more details: https://parallelproj.readthedocs.io/en/stable/.
+
+
+    .. tip::
+
+        Check out the :ref:`2D <sphx_glr_auto_examples_physics_demo_pet2d.py>` and
+        :ref:`3D <sphx_glr_auto_examples_physics_demo_pet3d.py>` examples to get started with this operator.
+
```
Tmodrzyk, line 66: Could it be possible to also add a link to your amazing phantom generating code ? 
It would be a shame if users missed it, even if it is included in the longer example

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,246 @@
...
+========================================
+
+This demo shows how to define a non time-of-flight PET scanner, simulate measurements
+and reconstruct an image from them.
+
+The PET forward model is defined as
+
+.. math::
+
+    y \sim \gamma \mathcal{P}(\frac{c \circ H(g*x) + b}{\gamma})
```
Tmodrzyk:
```suggestion
    y \sim \gamma \mathcal{P}\left(\frac{c \circ H(g*x) + b}{\gamma}\right)
```

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,246 @@
...
+
+where :math:`H \in \mathbb{R}_{+}^{m \times n}` is the projection operator,
+:math:`g \in \mathbb{R}_{+}^{n}` is a Gaussian blur kernel, :math:`x\in\mathbb{R}_{+}^{n}`
+is the emission image, :math:`b \in \mathbb{R}_{+}^{m}` is the (expected) background,
+:math:`\mathcal{P}` denotes Poisson noise with gain :math:`\gamma > 0`,
+:math:`c=\exp(-H\mu)\in \mathbb{R}_{+}^{m}` is an (optional) attenuation term
+with :math:`\mu \in \mathbb{R}_{+}^{n}` an attenuation map (typically obtained through an auxiliary CT scan).
```
Tmodrzyk, line 19: same comment as ``PET.py``

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,246 @@
...
+#       You can play with different geometries and voxel sizes to get a good grasp of
+#       the scanner geometry, and visualize it with `physics.plot_geometry()`
+#
+# .. note::
+#
+#      In this example, we normalize the forward operator :math:`\|A\|=1` (`normalize=True`) and
+#      the Poisson counts to be between 0 and 1 (`normalize_counts=True`), to simplify reconstruction.
+#      If you want to use the operator with real PET measurements, you will need to
+#      carefully handle the normalization. See also the :ref:`normalized 2D PET example <sphx_glr_auto_examples_physics_demo_pet2d.py>`.
+
```
Tmodrzyk: This is probably an error: this is the normalized PET example

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,246 @@
...
+# In practice, the attenuation is typically obtained with an auxiliary CT scan of the patient.
+
+x, attenuation = generate_pet_phantom(img_size, device=device)
+
+dinv.utils.plot([x, attenuation], titles=["Emission image", "Attenuation image"])
+
+# %%
+# Simulating measurements
+# -----------------------
+# We can generate measurements
```
Tmodrzyk: Either the sentence has been cut out, or we can remove this entirely.

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,246 @@
...
+# The shape of measurements is approximately `(B, 1, N, N/2)` where
+# `N=num_lor_endpoints_per_side*num_sides` is the number of detectors per ring.
+# This provides one measurement for every possible Line of Response (LOR), or in other words 'rays', connecting
+# two detectors in the scanner, which are arranged in a sinogram format, with the first axis
+# corresponding to the angle of the ray and the second axis corresponding to the distance of the ray to the center of the scanner.
```
Tmodrzyk:
```suggestion
# corresponding to the angle of the ray and the second axis corresponding to the distance of the ray to the center of the field of view.
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,512 @@
...
+import torch
+from typing import TYPE_CHECKING
+import contextlib
+import io
+
+if TYPE_CHECKING:
+    import parallelproj
+
+
+class PET(LinearPhysics):
```
Tmodrzyk, line 13: Could we overload the ``A_dagger`` to enable using filtered back-projection, just like in CT (``deepinv.physics.Tomography`` or ``deepinv.physics.Tomography``) ?
People rarely use the least-square pseudo inverse.

examples/physics/demo_pet3d.py
```diff
@@ -0,0 +1,259 @@
...
+.. math::
+
+    y \sim \gamma \mathcal{P}(c \circ H(g*x) + b)
+
+where :math:`H \in \mathbb{R}_{+}^{m \times n}` is the projection operator,
+:math:`g \in \mathbb{R}_{+}^{n}` is a Gaussian blur kernel, :math:`x\in\mathbb{R}_{+}^{n}`
+is the emission image, :math:`b \in \mathbb{R}_{+}^{m}` is the (expected) background,
+:math:`\mathcal{P}` denotes Poisson noise,
+:math:`c=\exp(-H\mu)\in \mathbb{R}_{+}^{m}` is an (optional) attenuation term
+with :math:`\mu \in \mathbb{R}_{+}^{n}` an attenuation map (typically obtained through an auxiliary CT scan).
```
Tmodrzyk, line 19: same comment as ``PET.py``, use another notation for this ``H``

examples/physics/demo_pet3d.py
```diff
@@ -0,0 +1,259 @@
...
+
+.. note::
+
+    In this example, we consider the unnormalized case, which allows to obtain quantitative reconstructions (i.e., :math:`x` has real
+    physical units). The operator also can be used in a normalized setting (forcing :math:`\|A\|_2=1` and normalizing counts to be between 0 and 1).
+    See also the :ref:`normalized 2D PET example <sphx_glr_auto_examples_physics_demo_pet2d.py>`.
+    When using deep learning-based reconstruction methods, is often easier to consider the normalized case, but a special attention is required
+    to denormalize the reconstructions and obtain physical units.
```
Tmodrzyk:
```suggestion
    When using deep learning-based reconstruction methods, it is often easier to consider the normalized case, but a special attention is required
    to denormalize the reconstructions and obtain physical units.
```

examples/physics/demo_pet3d.py
```diff
@@ -0,0 +1,259 @@
...
+dinv.utils.plot(
+    [x[..., mid_slice], attenuation[..., mid_slice]],
+    titles=["Emission image", "Attenuation image"],
+)
+
+# %%
+# Simulating measurements
+# -----------------------
+# We can generate measurements
+# The shape of measurements is approximately `(B, 1, N, N/2, R^2)` where
```
Tmodrzyk: same as 2D example, either remove or complete the sentence

examples/physics/demo_pet3d.py
```diff
@@ -0,0 +1,259 @@
...
+# `N=num_lor_endpoints_per_side*num_sides` is the number of detectors per ring
+# and `R` is the number of rings.
+# This provides one measurement for every possible Line of Response (LOR), or in other words 'rays', connecting
+# two detectors in the scanner, which are arranged in a sinogram format, with the first axis
+# corresponding to the angle of the ray, the second axis corresponding to the distance of the ray to the center of the scanner,
+# and the last axis corresponding to the depth of the ray (i.e., which rings of detectors are connected by the ray).
```
Tmodrzyk:
```suggestion
# This provides one measurement for every possible Line of Response (LOR), or in other words 'rays', connecting
# two detectors in the scanner, which are arranged in a sinogram format, with the first axis
# corresponding to the angle of the ray, the second axis corresponding to the distance of the ray to the center of the field of view
# and the last axis corresponding to the depth of the ray (i.e., which rings of detectors are connected by the ray).
```

examples/physics/demo_pet3d.py
```diff
@@ -0,0 +1,259 @@
...
+    ["sino. atten.", "meas.", "corrected meas."],
+    figsize=(6, 6),
+)
+
+# %%
+# Backprojection and sensitivities
+# --------------------------------
+# We backproject the data to visualize the sensitivity map of the scanner.
+# The sensitivity map is defined as the back-projection of a sinogram of ones, which corresponds to the number of rays intersecting each voxel.
+#
```
Tmodrzyk:
```suggestion
# The sensitivity map is defined as the back-projection of a sinogram of ones :math:`s = A^\top \mathbf{1}`, which corresponds to the number of rays intersecting each voxel.
#
```

Tmodrzyk (changes requested):
Thanks again for all the modifications @tachella ! 
Just a few errors / typos to fix in the docstrings and examples before merging.

deepinv/physics/pet.py
tachella: I've rephrased, hopefully is more clear now!

deepinv/physics/pet.py
tachella: mm here I think it's better to point at the example to avoid repeating too much

deepinv/physics/pet.py
tachella: I am applying the same operators to be honest - should I apply different ones?

deepinv/physics/pet.py
tachella: what would be the filter? the same ramp filter that we use in CT?

examples/physics/demo_pet2d.py
tachella: thanks for spotting it! solved it

examples/physics/demo_pet2d.py
tachella: thanks, I removed the sentence which wasn't necessary

examples/physics/demo_pet2d.py
tachella: done

examples/physics/demo_pet3d.py
tachella: done!

deepinv/physics/pet.py
Tmodrzyk: No no it's totally ok for a toy example to use the same geometry in the code. 
But in most cases the CT geometry would be different from the PET one: better resolution, more detectors, more angles etc

deepinv/physics/pet.py
Tmodrzyk: Yes ramp filter is enough, the real baseline is OSEM anyway in this field (vs FBP in CT).

tachella: I think this is not super necessary, and might complicate a future extension to ToF data (where there's no clear FBP),  I would keep it as is if that's fine for you

Tmodrzyk (approved):

deepinv/physics/pet.py
```diff
@@ -0,0 +1,516 @@
...
+        ::
+
+            conda install -c conda-forge parallelproj
+
+
+        If you are working on a pixi installation, simply do
+
+        ::
+
+            pixi install -e full
```
Andrewwango, line 55: there isn't "parallelproj" specified here?

Tmodrzyk:
@tachella it seems the doc is not building anymore because of the deleted entries in `refs.bib` from the merge.
Is there a reason for removing these ?

deepinv/physics/pet.py
```diff
@@ -0,0 +1,516 @@
...
+
+    .. tip::
+
+        Check out the :ref:`2D <sphx_glr_auto_examples_physics_demo_pet2d.py>` and
+        :ref:`3D <sphx_glr_auto_examples_physics_demo_pet3d.py>` examples to get started with this operator.
+
+    .. note::
+
+        This operator is defined to work on sinogram (binned) data.
+        List-mode data is not supported yet, but it can be pre-binned into a sinogram using `parallelproj`.
```
Andrewwango: I suggest we frame this more positively. (also note that parallelproj doesn't have the mentioned functionality)

```suggestion
        This operator currently only supports sinogram non-ToF data.
        To use this operator with listmode data and/or ToF data, you can easily swap out the projector `self.proj` for the appropriate listmode or ToF projector. See `parallelproj` `docs <https://parallelproj.readthedocs.io/>`_ for more details.
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,516 @@
...
+
+    :param tuple img_size: shape of the input 2D `(H, W)` or 3D volumes `(D, H, W)`.
+    :param tuple voxel_size: voxel size in mm. Default is 2 x 2 x 2 mm.
+    :param float fwhm_data_mm: full width at half maximum (FWHM) of the Gaussian blur :math:`g`. It has a crucial impact on the maximum achievable resolution,
+        which is typically a fraction of the FWHM.
+    :param None, parallelproj.pet_scanners.ModularizedPETScannerGeometry scanner: Scanner configuration. If None, the default scanner from parallelproj is used.
```
Andrewwango:
```suggestion
    :param None, parallelproj.pet_scanners.ModularizedPETScannerGeometry scanner: Scanner configuration. If None, a default demo scanner from parallelproj is used.
```

deepinv/physics/pet.py
```diff
@@ -0,0 +1,516 @@
...
+    ) -> torch.Tensor:
+        r"""
+        Apply the linear operator :math:`Ax=c \circ H(g*x)` to a signal :math:`x`
+
+        :param torch.Tensor x: input image or volume
+        :param torch.Tensor add_background: whether to add background :math:`b`. By default, no background is added.
+        :param torch.Tensor background: If not `None`, update the background :math:`b` of the operator.
+        :param torch.Tensor attenuation: If not `None`, update the attenuation :math:`c` of the operator.
+            The space (image or sinogram) is inferred automatically from the tensor shape.
+
```
Andrewwango, line 225: add a return annotation: returns sinogram `y`, and carefully detail what the dimensions mean i.e. shape (1,1,...,...) where ... is etc.

deepinv/physics/pet.py
```diff
@@ -0,0 +1,516 @@
...
+            out = out + self.background
+        return out
+
+    def A_adjoint(
+        self, y: torch.Tensor, attenuation=None, background=None, **kwargs
+    ) -> torch.Tensor:
+        r"""
+        Apply the adjoint of the linear operator :math:`A^{\top}y` where :math:`A=c \circ H(g*\cdot)` to a sinogram :math:`y`
+
+        :param torch.Tensor y: input sinogram
```
Andrewwango: again detail input / output shapes/axes descriptions

deepinv/physics/pet.py
```diff
@@ -0,0 +1,516 @@
...
+        x: torch.Tensor,
+        add_background: bool = False,
+        background: torch.Tensor | None = None,
+        attenuation: torch.Tensor | None = None,
+        **kwargs,
+    ) -> torch.Tensor:
+        r"""
+        Apply the linear operator :math:`Ax=c \circ H(g*x)` to a signal :math:`x`
+
+        :param torch.Tensor x: input image or volume
```
Andrewwango: describe shape

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,245 @@
+r"""
```
Andrewwango, line 1: I would put these two examples in external-libraries, no? Since it's an end-to-end demo (i.e. simulation + recon) showcasing a wrapper.

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,245 @@
...
+    Check the `parallelproj` documentation for more details: https://parallelproj.readthedocs.io/en/stable/.
+
+"""
+
+import deepinv as dinv
+from deepinv.physics import PET
+from deepinv.utils.phantoms import generate_pet_phantom
+import torch
+import parallelproj
+from array_api_compat import torch as torch_compat
```
Andrewwango, line 51: why's this needed?

examples/physics/demo_pet2d.py
```diff
@@ -0,0 +1,245 @@
...
+# --------------------------------
+# We backproject the data to visualize the sensitivity map of the scanner.
+# The sensitivity map is defined as the back-projection of a sinogram of ones :math:`s = A^\top \mathbf{1}`,
+# which corresponds to the number of rays intersecting each voxel.
+#
+# Here we also obtain a simple linear least-squares reconstruction by using
+# :meth:`A_dagger <deepinv.physics.LinearPhysics.A_dagger>`.
+
+with torch.no_grad():
+    x_dag = physics.A_dagger(y - background)
```
Andrewwango, line 193: why not visualising the dag too?

Andrewwango (changes requested):
Thanks @tachella @Tmodrzyk  for the hard work! Some minor comments but otehrwise looks good!

deepinv/physics/pet.py
tachella: the full environment installs parallelproj (and all other optional dependencies)

examples/physics/demo_pet2d.py
tachella: mm not a fan of "external libraries" it's a bit too abstract - I would keep it here and refactor the external libraries in the future

examples/physics/demo_pet2d.py
tachella: you're right, done

examples/physics/demo_pet2d.py
tachella: parallelproj requires this for interfacing with torch

deepinv/physics/pet.py
tachella: done

deepinv/physics/pet.py
tachella: done

deepinv/physics/pet.py
tachella: done

deepinv/physics/pet.py
tachella: thanks, done

deepinv/physics/pet.py
tachella: thanks, done

tachella:
hey @Andrewwango, many thanks for the comments, I applied the changes - let me know if you want any other changes :)

Andrewwango (approved):
lgtm

PR #1102, opened 2026-03-12
Directly pass astra geometries to "TomographyWithAstra" physics operator 

MargaretDuff:
This is useful for Core Imaging Library compatibility - we can easily extract the astra geometries from our data containers so this is the easiest way to pass our users to deep inverse. 

To do list: 
- [x] Local tests for 3D conebeam, 2D fanbeam and 2D and 3D parallel beam geometries 
- [x] Implement unit tests 
- [x] Update changelog 
- [x] Check docstrings have rendered correctly 


Andrewwango:
Thanks for working on this @MargaretDuff !

Please let me know when it's ready for review, even if it's just an initial one.

It would also be great to link to an example using this, e.g. in your user showcase!

Andrewwango:
I think this is a nice addition to allow different tomography geometries, opening up DeepInverse to more tomography users, e.g. CIL users.

@romainvo @askorikov do you think this is sufficient, or do you see a cleaner solution? Any suggestions or review is welcome!

askorikov:
Hi @Andrewwango, @MargaretDuff! I think the implementation makes sense. It terms of API, we could consider inferring the `is_2d` parameter using something like `is_2d = astra_object_geom.get('GridSliceCount') is not None`. And in terms of names, I would maybe use `astra_vol_geom`, because that's the terminology we use with Astra-native objects (which we consume here).

deepinv/physics/tomography.py
```diff
@@ -489,52 +494,69 @@ def __init__(
...
+        
+        if astra_object_geom is not None or astra_projection_geom is not None:
+            if astra_object_geom is None or astra_projection_geom is None:
+                raise ValueError("Both `astra_object_geom` and `astra_projection_geom` must be specified together.")
+            if is_2d is None:
+                raise ValueError("When `astra_object_geom` and `astra_projection_geom` are specified, `is_2d` must also be specified to indicate whether the geometry is 2D or 3D.")
+            self.object_geometry = astra_object_geom
+            self.projection_geometry = astra_projection_geom
+            self.is_2d = is_2d
+            self.img_size = (self.object_geometry['GridRowCount'], self.object_geometry['GridColCount']) if self.is_2d else (self.object_geometry['GridRowCount'], self.object_geometry['GridColCount'], self.object_geometry['GridSliceCount']) # TODO: Check this for 3D
```
askorikov: This seems correct. Just note that the shape of the input will need to be [GridSliceCount, GridRowCount, GridColCount], as usual with Astra.

romainvo:
Hi @MargaretDuff! I agree with @askorikov, the contribution increases interoperability, and the implementation is good as is :)!

Small point: I would not make `img_size` an optional kwarg; it could lead to ambiguities in how to instantiate the operator if the user does not want to pass pre-created Astra geometries. This way, everything is explicit.

This is important as it makes the difference clear with other operators in the library, which can adapt the size of the processed signal at runtime (e.g. for Downsampling `img_size can be None at init), whereas the domain of instantiated Tomography(WithAstra) physics is fixed.

Andrewwango:
Thank you @romainvo @askorikov ! Excited for this PR, I think it opens up the physics for more tomography community including of course CIL users @MargaretDuff

deepinv/physics/tomography.py
MargaretDuff: Yes, I think that is now correct - thanks

MargaretDuff:
@Andrewwango, @askorikov and @romainvo - thank-you for your review and sorry it has sat so long. I have: 
- Merged in master so it is up to date 
- Added tests - comparing building operator with astra directly and passing to the new functionality and building with deepinv, for all 4 geometry types. Also tested the forward pass of native astra and the deepinv implementation agrees for all 4 geometry types. 
- I made a design change, which I can change back - I created a new constructor (as a classmethod) for the direct building of the class from astra geometries

deepinv/physics/tomography.py
```diff
@@ -609,6 +631,107 @@ def __init__(
...
+            or projection_geometry["DetectorRowCount"] != 1
+        ):
+            raise ValueError(
+                "`is_2d=True` describes a flat acquisition, but `object_geometry` has "
+                f"{object_geometry['GridSliceCount']} slices and `projection_geometry` has "
+                f"{projection_geometry['DetectorRowCount']} detector rows, which must both "
+                "be 1. Pass `is_2d=False` to keep the third dimension."
+            )
+
+        if torch.device(device).type != "cuda":
```
```suggestion
        if not dinv.utils.devices_equal(device, "cuda"):
```

deepinv/physics/tomography.py
```diff
@@ -578,6 +586,20 @@ def __init__(
...
+        self._build_transform(normalize=normalize, device=device)
+
+    def _build_transform(
+        self, normalize: bool | None, device: torch.device | str
+    ) -> None:
+        """Build the X-ray transform from the geometries and normalize the operator.
+
+        Shared by :func:`__init__` and :func:`from_astra_geometry`, which are only
+        responsible for setting ``object_geometry``, ``projection_geometry``,
+        ``img_size`` and ``is_2d``.
```
Andrewwango: remove

deepinv/physics/tomography.py
```diff
@@ -609,6 +631,107 @@ def __init__(
...
+        r"""Build the operator from pre-created ``astra`` geometries.
+
+        Alternative constructor for the cases where ``astra`` geometries are already available.
+
+        .. note::
+
+            Both geometries must be 3D, even for a 2D acquisition.
+            Describe a 2D acquisition as a flat 3D one instead — a single slice imaged by
+            a single detector row — and set ``is_2d=True``, as it cannot be inferred from
+            the geometries.
```
```suggestion
For 2D, use one slice and set `is_2d=True`.
```

deepinv/physics/tomography.py
```diff
@@ -609,6 +631,107 @@ def __init__(
...
+        :param dict projection_geometry: An ``astra`` projection geometry, as returned by ``astra.create_proj_geom``.
+        :param bool is_2d: Whether the geometries describe a 2D slice or a 3D volume.
+        :param bool normalize: If ``True`` :func:`A` and :func:`A_adjoint` are normalized so that the operator has unit norm. (default: ``True``)
+        :param torch.device | str device: The operator only supports CUDA computation. (default: ``torch.device('cuda')``)
+        :return: (:class:`deepinv.physics.TomographyWithAstra`) the tomography operator.
+        """
+        if object_geometry is None or projection_geometry is None:
+            raise ValueError(
+                "`object_geometry` and `projection_geometry` must both be specified."
+            )
```
Andrewwango: unnecessary since they are required arguments anyway

deepinv/physics/tomography.py
```diff
@@ -609,6 +631,107 @@ def __init__(
...
+                f"`object_geometry` is missing the keys {sorted(missing)}. It must be a 3D "
+                "astra volume geometry, even when `is_2d=True`."
+            )
+
+        missing = {"DetectorRowCount", "DetectorColCount"} - set(projection_geometry)
+        if missing:
+            raise ValueError(
+                f"`projection_geometry` is missing the keys {sorted(missing)}. It must be a 3D "
+                "astra projection geometry, even when `is_2d=True`."
+            )
```
Andrewwango: feels unnecessary - we make this clear already in the docstring, which we assume the user reads. these missing ones will be caught below anyway

deepinv/physics/tomography.py
```diff
@@ -609,6 +631,107 @@ def __init__(
...
+
+        if is_2d and (
+            object_geometry["GridSliceCount"] != 1
+            or projection_geometry["DetectorRowCount"] != 1
+        ):
+            raise ValueError(
+                "`is_2d=True` describes a flat acquisition, but `object_geometry` has "
+                f"{object_geometry['GridSliceCount']} slices and `projection_geometry` has "
+                f"{projection_geometry['DetectorRowCount']} detector rows, which must both "
+                "be 1. Pass `is_2d=False` to keep the third dimension."
```
```suggestion
                "`is_2d=True` but `object_geometry` has "
                f"{object_geometry['GridSliceCount']} slices and `projection_geometry` has "
                f"{projection_geometry['DetectorRowCount']} detector rows, which must both "
                "be 1. Use `is_2d=False` instead."
```

deepinv/physics/tomography.py
```diff
@@ -609,6 +631,107 @@ def __init__(
...
+            )
+
+        if torch.device(device).type != "cuda":
+            warn(
+                f"TomographyWithAstra only supports CUDA Tensors and CUDA operations, got device={device}",
+                RuntimeWarning,
+            )
+
+        self = cls.__new__(cls)
+        LinearPhysics.__init__(self, device=device, **kwargs)
```
Andrewwango: TODO Andrew/Romain: is this standard?

deepinv/tests/test_external_libraries.py
```diff
@@ -1,8 +1,23 @@
 import deepinv as dinv
+import numpy as np
 import torch
 import pytest
 
 
+def assert_geometries_close(actual: dict, expected: dict) -> None:
```
Andrewwango:
```suggestion
def assert_astra_geometries_close(actual: dict, expected: dict) -> None:
```

deepinv/tests/test_external_libraries.py
```diff
@@ -1,8 +1,23 @@
...
+
+    :param dict actual: The geometry to check.
+    :param dict expected: The geometry it should describe.
+    """
+    assert set(actual) == set(expected)
+    for key, value in expected.items():
+        if isinstance(value, str):  # the geometry "type"
+            assert actual[key] == value
+        else:  # counts, spacings, the "option" sub-dict and the angles
+            assert actual[key] == pytest.approx(value)
```
Andrewwango: tbh maybe this does not need to be a separate function and can be just instead replaced by e.g.

assert all(actual[k] == (v if isinstance(v, str) else pytest.approx(v)) for (k,v) in expected.items()) and set(actual) == set(expected)

deepinv/tests/test_external_libraries.py
```diff
@@ -240,3 +255,180 @@ def test_tomography_with_astra_logic(
...
+            (True, "parallel"),
+            (True, "fanbeam"),
+            (False, "parallel"),
+            (False, "conebeam"),
+        ],
+    )
+    def test_tomography_with_astra_from_astra_geometry(
+        self, is_2d, geometry_type, device
+    ):
+        r"""
```
Andrewwango: don't need docstrings for unit tests

deepinv/tests/test_external_libraries.py
```diff
@@ -240,3 +255,180 @@ def test_tomography_with_astra_logic(
...
+            img_size=img_size,
+            angles=num_angles,
+            angular_range=angular_range,
+            n_detector_pixels=n_detector_pixels,
+            geometry_type=geometry_type,
+            normalize=False,
+            device=device,
+        )
+
+        ## --- Describe the same acquisition with the astra API ---
```
Andrewwango: unnecessary comments

deepinv/tests/test_external_libraries.py
```diff
@@ -240,3 +255,180 @@ def test_tomography_with_astra_logic(
...
+
+        physics_from_geometry = dinv.physics.TomographyWithAstra.from_astra_geometry(
+            object_geometry,
+            projection_geometry,
+            is_2d=is_2d,
+            normalize=False,
+            device=device,
+        )
+
+        ## --- Test both operators describe the same geometry ---
```
Andrewwango: unnecessary comments

deepinv/tests/test_external_libraries.py
```diff
@@ -240,3 +255,180 @@ def test_tomography_with_astra_logic(
...
+            object_geometry,
+        )
+        astra.data3d.delete(sinogram_id)
+        assert torch.allclose(
+            y[0, 0].reshape(sinogram.shape).cpu(), torch.from_numpy(sinogram)
+        )
+
+        ## --- Test the reconstruction ---
+        assert torch.allclose(physics_from_geometry.A_adjoint(y), physics.A_adjoint(y))
+        assert torch.allclose(physics_from_geometry.fbp(y), physics.fbp(y))
```
Andrewwango: hmm these are going a bit beyond the feature it's testing

Andrewwango (changes requested):
Thanks @MargaretDuff !

deepinv/physics/tomography.py
```diff
@@ -609,6 +631,107 @@ def __init__(
...
+    @classmethod
+    def from_astra_geometry(
+        cls,
+        object_geometry: dict[str, Any],
+        projection_geometry: dict[str, Any],
+        is_2d: bool,
+        normalize: bool | None = None,
+        device: torch.device | str = torch.device("cuda"),
+        **kwargs,
+    ) -> TomographyWithAstra:
```
romainvo: I really like this pattern! After discussing with @jscanvic and @Andrewwango I think we can make this alternative constructor slightly more robust. 

I failed to find "official" best practices, PEP, on how properly implement this alternative constructor but one alternative would be to put everything in the `__init__`, in the sense that the main constructor handles the routing of each alternative way of constructing it (as it already does somehow) and the alternative constructor provides the "clear" interface for the alternative signature.

MargaretDuff:
Hi developers! I have done some minor updates from Andrew's review. I am happy now for you to take this on to get it merged - it is unlikely I will have much time ( and the time I have, I would like focus on a CIL to DeepInv demo!)

deepinv/physics/tomography.py
```diff
@@ -659,39 +657,19 @@ def from_astra_geometry(
...
-                "`is_2d=True` describes a flat acquisition, but `object_geometry` has "
+                "`is_2d=True` but `object_geometry` has "
                 f"{object_geometry['GridSliceCount']} slices and `projection_geometry` has "
                 f"{projection_geometry['DetectorRowCount']} detector rows, which must both "
-                "be 1. Pass `is_2d=False` to keep the third dimension."
+                "be 1. Use `is_2d=False` instead."
             )
 
-        if torch.device(device).type != "cuda":
+        if devices_equal(device, "cuda"):
```
Andrewwango: not?

Andrewwango (approved):
thanks @MargaretDuff , will merge soon

Andrewwango (approved):

PR #1104, opened 2026-03-15
Unify patching / tiling 

mh-nguyen712:
Toward closing #1039

In this PR:
- refactor the TiledMixin2d to have functional API, which can be used to define `deepinv.utils.image_to_patches` and `deepinv.utils.patches_to_image`, the function `deepinv.utils.patchify` now becomes an alias of `deepinv.utils.image_to_patches`. Both the mixin and the functional API use the same implementation under the hood.
- This new implementation is then used in:
    + `PatchDataset` with `TiledMixin2d` 
    + `PatchCovarianceNoiseEstimator` with  `deepinv.utils.image_to_patches`

Remaining issues in #1039: 
- PatchSampler: the current TiledMixin2d and functions are not optimal for this, since we need random patches (different than fixed stride patches)
- DistributedSampler: this would require more works, since this module is still in active dev, eg #1088 

 


Andrewwango:
Really nice contribution @mh-nguyen712  thank you! Please can you add a PR description with what you did?

mh-nguyen712:
Thanks @Andrewwango, added!

deepinv/utils/patch_extractor.py
```diff
@@ -56,3 +74,100 @@ def patch_extractor(
...
+        stride=stride_2d,
+        pad_if_needed=pad_if_needed,
+    )
+
+
+def patches_to_image(
+    patches: Tensor,
+    stride: int | tuple[int, int],
+    img_size: tuple[int, int] | None = None,
+    reduce_overlap: str = "sum",
```
romainvo: I would set the default value to "mean" otherwise it changes the pixel range of the reconstructed image

deepinv/utils/_tiling.py
```diff
@@ -0,0 +1,142 @@
...
+        3, patch_size[1], stride[1]
+    )
+    return patches.contiguous()
+
+
+def _patches_to_image_impl(
+    patches: Tensor,
+    stride: tuple[int, int],
+    img_size: tuple[int, int] | None = None,
+    reduce_overlap: str = "sum",
```
romainvo: I would set the default value to "mean" otherwise it changes the pixel range of the reconstructed image. (and set reduce_overlap="mean" across all the methods calling it)

deepinv/utils/patch_extractor.py
```diff
@@ -56,3 +74,100 @@ def patch_extractor(
...
+    The behavior mirrors :class:`deepinv.utils.TiledMixin2d` while exposing a
+    functional API.
+
+    :param torch.Tensor image: Input image tensor of shape ``(B, C, H, W)``.
+    :param int | tuple[int, int] patch_size: Patch size ``(patch_h, patch_w)``.
+        If an ``int`` is provided, the same value is used for both dimensions.
+    :param int | tuple[int, int] stride: Stride between adjacent patches as
+        ``(stride_h, stride_w)``. If ``None``, defaults to half the patch size.
+    :param bool pad_if_needed: If ``True``, adds extra right/bottom padding so
+        that all extracted patches have the requested ``patch_size``.
```
romainvo: I would be a little more specific here. It seems that in practice, when `pad_if_need=False`, extracted patches still have the requested `patch_size` but the method crops `(img_size - patch_size) % stride` number of pixels from the right and the bottom of the original image, so that the produced patches cover exactly the cropped area.
- thus reconstruction of the original image is not possible when `pad_if_needed=False` and img_size is not compatible with stride and patch_size.

deepinv/models/utils.py
```diff
@@ -68,56 +69,18 @@ def test_pad(model, L, modulo=16):
 
 
 def patchify(
```
romainvo: Should we deprecate the alias ? I seems confusing to have two different functions which do exactly the same thing.

Also the same method is also defined in deepinv.utils, I guess it makes more sense to have it there in the global utils than here

deepinv/utils/patch_extractor.py
```diff
@@ -56,3 +74,100 @@ def patch_extractor(
...
+    Input shape: torch.Size([1, 3, 256, 256]), patchified shape: torch.Size([1, 3, 7, 7, 64, 64])
+    >>> list_patch = [patches[0,:, i, j, ...] for i in range(patches.shape[2]) for j in range(patches.shape[3])]
+    >>> dinv.utils.plot([x, make_grid(list_patch, nrow=patches.shape[2])], titles=["Original", "Overlapping patches"])  # doctest: +SKIP
+
+    .. plot::
+
+        import deepinv as dinv
+        from torchvision.utils import make_grid
+        x = dinv.utils.load_example('butterfly.png')
+        patches = dinv.utils.patchify(x, patch_size=64, stride=32)
```
romainvo: Use `image_to_patches` instead of patchify

deepinv/utils/patch_extractor.py
```diff
@@ -56,3 +74,100 @@ def patch_extractor(
...
+    :return: Patches of shape ``(B, C, n_rows, n_cols, patch_h, patch_w)``.
+
+    |sep|
+
+    :Examples:
+
+    >>> import deepinv as dinv
+    >>> from torchvision.utils import make_grid
+    >>> x = dinv.utils.load_example('butterfly.png')
+    >>> patches = dinv.utils.patchify(x, patch_size=64, stride=32)
```
romainvo: Use image_to_patches instead of patchify

deepinv/utils/patch_extractor.py
```diff
@@ -56,3 +74,100 @@ def patch_extractor(
...
+    _, stride_2d = _resolve_tiling_params(patch_size=patches.shape[-2:], stride=stride)
+    return _patches_to_image_impl(
+        patches=patches,
+        stride=stride_2d,
+        img_size=img_size,
+        reduce_overlap=reduce_overlap,
+    )
+
+
+def patchify(
```
romainvo, line 177: Same comment that in models/utils.py, should we deprecate the alias ? 

Also we should remove the `patchify` in method/utils.py and only keep this one.

deepinv/models/noise_level_estimation.py
```diff
@@ -1,7 +1,8 @@
+from __future__ import annotations
 import torch
 import torch.nn as nn
 
-from deepinv.models.utils import patchify
+from deepinv.utils.patch_extractor import image_to_patches
```
romainvo, line 5: More of a curiosity: what's the best practice when importing modules internally, should we do `from deepinv.utils import image_to_patches` ? would be less sensitive to future refactoring of the code I guess

deepinv/optim/prior.py
```diff
@@ -528,22 +529,8 @@ def __init__(
...
-            x = torch.cat(
-                (
-                    torch.flip(x[:, :, :, -self.patch_size : -1], (3,)),
-                    x,
-                    torch.flip(x[:, :, :, 1 : self.patch_size], (3,)),
-                ),
-                3,
-            )
+            pad = self.patch_size - 1
+            x = F.pad(x, (pad, pad, pad, pad), mode="reflect")
```
romainvo: We could make the padding mode an option to the user for more flexibility, e.g. as an instance attribute ?

deepinv/utils/__init__.py
```diff
@@ -42,7 +42,12 @@
     dirac_comb_like,
 )
 from .phantoms import RandomPhantomDataset, SheppLoganDataset
-from .patch_extractor import patch_extractor
+from .patch_extractor import (
+    patch_extractor,
+    image_to_patches,
+    patches_to_image,
+    patchify,
```
romainvo, line 50: same comment than above about having two methods for the same thing (patchify and image_to_patches)

deepinv/utils/patch_extractor.py
```diff
@@ -56,3 +74,100 @@ def patch_extractor(
     patches = patches.reshape(B, n_patches, C, patch_size, patch_size)
 
     return patches, linear_inds
+
+
+def image_to_patches(
```
romainvo, line 78: Why not exposing the `extra_pad` feature of _image_to_patches_impl here like you did in the Mixin class ?

deepinv/utils/mixins.py
```diff
@@ -378,29 +384,29 @@ def image_to_patches(
...
-        patches = image.unfold(2, patch_size[0] + pad[2] + pad[3], stride[0]).unfold(
-            3, patch_size[1] + pad[0] + pad[1], stride[1]
+        patch_size = _add_tuple(self.patch_size, (pad[2] + pad[3], pad[0] + pad[1]))
+        return _image_to_patches_impl(
+            image=image,
+            patch_size=patch_size,
+            stride=self.stride,
+            pad_if_needed=self.pad_if_needed,
+            extra_pad=pad,
         )
```
romainvo: Should we use the functional call to deepinv.utils.image_to_patches instead, and have the `patch_size = _add_tuple(self.patch_size, (pad[2] + pad[3], pad[0] + pad[1]))` logic only written once in the code ? 
- requires exposing the `extra_pad` feature  in `deepinv.utils.image_to_patches` (with the `pad` kwarg like you do here)

deepinv/utils/mixins.py
```diff
@@ -416,56 +422,16 @@ def patches_to_image(
 
         :param torch.Tensor patches: Patches tensor of shape `(B, C, n_rows, n_cols, patch_h, patch_w)`.
         :param img_size: Target output size (height, width). If provided, output is cropped to this size from the top-left corner.
-        :param reduce_overlap: How to handle overlapping regions. Options are `"sum"` or `"mean"`.
+        :param reduce_overlap: How to handle overlapping regions. Options are `"sum"` or `"mean"`. If `"sum"`, overlapping regions are summed (default). If `"mean"`, overlapping regions are averaged, which can give an identical reconstruction to the original image.
```
romainvo, line 405: same comment, why do `"sum"` by default and not `"mean"`

romainvo (changes requested):
Thanks @mh-nguyen712 for the refactoring :)!  Mainly comments about docs and where to put code/logic to avoid duplicates.

----

Some suggested change that I could not add in the review:
In the `TiledMixin2d` init, we should use `_resolve_tiling_params` to avoid reimplementing the same logic twice

deepinv/utils/patch_extractor.py
romainvo: Just saw that you use summation in TiledSpaceVaryingBlur by default, sorry, I was in inference-by-patch mode, feel free to ignore

deepinv/utils/patch_extractor.py
romainvo: In this snippet for example, patches have shape `torch.Size([1, 3, 5, 5, 64, 64])`

```python
import deepinv as dinv
from torchvision.utils import make_grid

x = dinv.utils.load_example('butterfly.png', img_size=(217, 217))

patches = dinv.utils.image_to_patches(x, patch_size=64, stride=32, pad_if_needed=False)
print(patches.shape)

img_from_patches = dinv.utils.patches_to_image(patches, stride=32, img_size=x.shape[-2:], reduce_overlap="mean")
list_patch = [patches[0, :, i, j, ...] for i in range(patches.shape[2]) for j in range(patches.shape[3])]
dinv.utils.plot([x, make_grid(list_patch, nrow=patches.shape[2]), img_from_patches], titles=["Original", "Overlapping patches", "Reconstructed"])
```

Capture d’écran 2026-04-20 201043

deepinv/models/noise_level_estimation.py
mh-nguyen712: from my experience, this is to avoid potential circular import since a full path import will impact less function/module

deepinv/models/utils.py
mh-nguyen712: Yes I agree about putting it in deepinv.utils. I just kept it for backward-compatibility, but can remove or move to deepinv.utils
I think having alias is not problematic, as some people would prefer the name `patchify` :D 
I move this to deepinv.utils?

deepinv/optim/prior.py
mh-nguyen712: Done!

deepinv/models/utils.py
mh-nguyen712: Removed it!

deepinv/utils/patch_extractor.py
mh-nguyen712: I think we can keep the alias, some people would prefer this naming :D

deepinv/utils/patch_extractor.py
mh-nguyen712: Thanks good idea, set default to `mean`

deepinv/utils/patch_extractor.py
mh-nguyen712: Thanks for the remark!

mh-nguyen712: I clarified a bit more the docs

deepinv/utils/mixins.py
mh-nguyen712: Done

deepinv/utils/patch_extractor.py
mh-nguyen712: Done

deepinv/utils/patch_extractor.py
```diff
@@ -56,3 +73,116 @@ def patch_extractor(
...
+    r"""Split a batch of images into overlapping 2D patches.
+
+    The behavior mirrors :class:`deepinv.utils.TiledMixin2d` while exposing a functional API.
+
+    :param torch.Tensor image: Input image tensor of shape ``(B, C, H, W)``.
+    :param int | tuple[int, int] patch_size: Patch size ``(patch_h, patch_w)``.
+        If an ``int`` is provided, the same value is used for both dimensions.
+    :param int | tuple[int, int] stride: Stride between adjacent patches as
+        ``(stride_h, stride_w)``. If ``None``, defaults to half the patch size.
+    :param bool pad_if_needed: If ``True``, adds extra right/bottom padding so that the patches cover the entire image. Default is ``True``.
```
romainvo, line 94: Add description for  `pad` kwarg

docs/source/changelog.rst
```diff
@@ -21,7 +21,7 @@ Changed
 - Make available every mask used at evaluation for splitting models :class:`deepinv.loss.SplittingLoss.SplittingModel` when ``eval_n_samples > 1`` (:gh:`881` by `Jérémy Scanvic`_)
 - Deprecate `Loss.name` in favor of the class name as done in the trainer (:gh:`881` by `Jérémy Scanvic`_)
 - Update references for single pixel demo (:gh:`1151` by `Laura C. Diaz-Delgado`_)
-- Add changelog section to the contributing guidelines (:gh:`1153` by `Thibaut Modrzyk`_)
+- Add changelog section to the contributing guidelines (:gh:`1153` by `Thibaut Modrzyk`_)- Unify patching / tiling and unpatching / un-tiling logic in physics and utils, with support for padding and non-overlapping patches. Add :func:`deepinv.utils.image_to_patches` and :func:`deepinv.utils.patches_to_image` utility functions, and refactor physics to use them instead of `unfold` (:gh:`1104` by `Minh Hai Nguyen`_)
```
romainvo:
```suggestion
- Add changelog section to the contributing guidelines (:gh:`1153` by `Thibaut Modrzyk`_)
- Unify patching / tiling and unpatching / un-tiling logic in physics and utils, with support for padding and non-overlapping patches. Add :func:`deepinv.utils.image_to_patches` and :func:`deepinv.utils.patches_to_image` utility functions, and refactor physics to use them instead of `unfold` (:gh:`1104` by `Minh Hai Nguyen`_)
```

deepinv/physics/blur.py
```diff
@@ -1070,7 +1070,11 @@ def A_adjoint(
         y = self._pad(y, pad)
 
         # Extract patches
+        print(
+            f"DEBUG -- A_adjoint: y shape after padding: {y.shape}, expected pad: {pad}"
+        )
         patches = self.image_to_patches(y)
+        print(f"DEBUG -- A_adjoint: patches shape after extraction: {patches.shape}")
```
romainvo: Clean debug prints

```suggestion
        patches = self.image_to_patches(y)
```

romainvo (commented):
Thanks for the changes! Last few comments on small doc and remaining debug prints.

A last point about refactoring: I think we should factorize the logic in `TiledSpaceVaryingBlur.num_filters` and `_compute_needed_pad`, as they essentially do the same thing. You could make `_compute_needed_pad` return `n_h` and `n_w` for example and afterwards use `_compute_needed_pad` in the `num_filters` helper

docs/source/changelog.rst
mh-nguyen712: Thanks!

deepinv/utils/patch_extractor.py
mh-nguyen712: done, thanks for the remark!

mh-nguyen712:
Thanks again for the careful review @romainvo! I made the refactoring and removed debugging code

romainvo (approved):
Thanks for the final changes @mh-nguyen712 ! Looks good to me :)

PR #1105, opened 2026-03-17
add DEEPINV_CACHE_DIR

mh-nguyen712:
Add a global cache directory for deepinv: downloaded datasets, images, ... 

### DIFF
--- a/deepinv/utils/__init__.py
+++ b/deepinv/utils/__init__.py
@@ -23,6 +23,7 @@ from .demo import (
     load_dataset,
     load_degradation,
     get_cache_home,
+    get_data_home,
     get_image_url,
     get_degradation_url,
     download_example,
--- a/examples/models/demo_training.py
+++ b/examples/models/demo_training.py
@@ -44,7 +44,7 @@ physics = dinv.physics.Inpainting((1, 64, 64), mask=0.8, device=device, rng=rng)
 from torchvision.transforms import Compose, ToTensor, Resize, CenterCrop, Grayscale
 
 dataset = dinv.datasets.Urban100HR(
-    dinv.utils.get_data_home() / "Urban100",
+    dinv.utils.get_cache_home() / "datasets" / "Urban100",
     download=True,
     transform=Compose([ToTensor(), Grayscale(), Resize(256), CenterCrop(64)]),
 )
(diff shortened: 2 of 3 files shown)


tachella:
hey @mh-nguyen712, I think this is a good idea! We already have a [get_data_home](https://deepinv.github.io/deepinv/api/stubs/deepinv.utils.get_data_home.html#deepinv.utils.get_data_home) that is supposed to do something similar.. Should we keep a single solution?

mh-nguyen712:
> hey @mh-nguyen712, I think this is a good idea! We already have a [get_data_home](https://deepinv.github.io/deepinv/api/stubs/deepinv.utils.get_data_home.html#deepinv.utils.get_data_home) that is supposed to do something similar.. Should we keep a single solution?

Yes indeed they are similar. I didn't pay much attention to this, so I just replace the old variable names (`DEEPINV_DATA` or `XDG_DATA_HOME`) by `DEEPINV_CACHE_DIR` and set a default value. If you prefer, I can change it to `DEEPINV_DATA`. I think this is not really important since the env variable was hidden, now I add a small rst files to explain all env variables.

deepinv/utils/demo.py
```diff
@@ -51,20 +51,17 @@ def get_image_url(file_name: str, dataset: str = "images") -> str:
...
+
     if data_home is not None:
         path = Path(data_home)
     else:
-        data_home = os.environ.get("XDG_DATA_HOME")
-        if data_home is not None:
-            path = Path(data_home) / "deepinv"
-        else:
-            path = Path(".") / "datasets"
+        path = Path.home() / ".cache" / "deepinv"
```
jscanvic: Please fallback to $XDG_CACHE_HOME if it exists, and only then to $HOME/.cache

mh-nguyen712:
Thanks for the comments @jscanvic!
Regarding your comments:
1) I would prefer to deprecate `get_data_home` and replace by `get_cache_home`, which is defined by `DEEPINV_CACHE_DIR` if user set it, then default to `XDG_CACHE_HOME`.
I think we can just use the cache for simplicity (instead of separating into cache / config / data), similar to torch or huggingface libraries. 

2) This can be solved by setting `DEEPINV_CACHE_DIR` default to `XDG_DATA_HOME`. As `DEEPINV_DATA` do not appear in the documentation, I think we can just ignore it. 

Please let me know your thoughts!

jscanvic:
Sounds good to me thanks Hai

PR #1107, opened 2026-03-22
Re-integrate DEAL reconstruction model with improved structure and fixes

Holimmo7:
This PR re-introduces the DEAL reconstruction model into DeepInverse after a previous revert.

Key points:
- Restores the DEAL model implementation with improved structure
- Fixes integration issues in examples and tests
- Ensures compatibility with DeepInverse design

Next step:
- Replace custom CG solver with DeepInverse CG implementation as suggested by maintainers


deepinv/models/deal.py
```diff
@@ -0,0 +1,1194 @@
...
+            r_norm_old = r_norm.clone()
+            r = r - alpha * BTBp
+
+            r_norm = (r**2).sum(dim=(1, 2, 3), keepdim=True)
+            beta = r_norm / r_norm_old
+            p = r + beta * p
+
+        return output, i
+
+    def cg(
```
tachella, line 947: Can you use one of the available least squares solvers instead of redefining CG? In this way, we would give access to more varied solvers which are thoroughly tested

See more details here
https://deepinv.github.io/deepinv/api/stubs/deepinv.optim.linear.least_squares.html#deepinv.optim.linear.least_squares

deepinv/models/deal.py
```diff
@@ -0,0 +1,1194 @@
...
+            c_ks.append(c_k)
+            return torch.clip(c_k, 0, 1), c_ks
+        return torch.clip(c_k, 0, 1)
+
+
+class DEAL(Reconstructor):
+    """
+    Deep Equilibrium Attention Least Squares (DEAL) reconstruction model.
+
+    This model solves linear inverse problems using a learned equilibrium-based
```
tachella: provide a mathematical description of the algorithm with the basic updates

deepinv/models/deal.py
```diff
@@ -0,0 +1,1194 @@
...
+        )
+
+        self.model.load_state_dict(merged_state_dict, strict=True)
+
+    @torch.no_grad()
+    def forward(self, y: torch.Tensor, physics: LinearPhysics) -> torch.Tensor:
+        """
+        Run the DEAL reconstruction.
+
+        Parameters
```
tachella: use :param type name: (torch standard) docstrings, as in the rest of the library

examples/models/demo_deal.py
```diff
@@ -0,0 +1,92 @@
...
+
+This example shows how to use the Deep Equilibrium Attention Least Squares
+(DEAL) reconstruction model in DeepInverse for a simple deblurring problem.
+
+DEAL solves linear inverse problems using a learned equilibrium-based
+regularizer combined with conjugate gradient iterations. It can be used for
+image restoration and reconstruction tasks such as denoising, deblurring,
+and computed tomography reconstruction.
+
+Given measurements :math:`y = Ax + n`, DEAL computes a reconstruction by
```
tachella: The demo should be improved to better illustrate some details about the algorithm:

- provide a math description of the main iterations
- plot the mask to illustrate the adaptivity of the regularizer
- illustrate number of parameters and other salient features of the model

docs/source/user_guide/reconstruction/pretrained-models.rst
```diff
@@ -40,6 +40,10 @@ These models can be set-up in one line and perform inference in another line:
      - :ref:`Plug-and-play <iterative>` w/ pretrained denoiser
      - General
      - Medium
+   * - :class:`DEAL <deepinv.models.DEAL>`
+     - Feedforward
```
tachella:
```suggestion
     - Iterative
```

deepinv/tests/test_models.py
```diff
@@ -1680,3 +1683,82 @@ def test_gaussian_noise_estimators(model_name, mode, channels, sigma, device, rn
     assert torch.allclose(
         model(y), torch.cat([model(_y.unsqueeze(0)) for _y in y]), rtol=1e-3, atol=1e-4
     )
+
+
+class DummyInnerDEAL(nn.Module):
```
tachella: add the model to MODEL_LIST to test in the same way as the other models (for performance, compatibility with input sizes, devices, etc), and only keep standalone tests to check specific features of this algorithm that others do not have

deepinv/models/deal.py
```diff
@@ -0,0 +1,1194 @@
...
+import torch.nn.utils.parametrize as P
+from torch import Tensor, nn
+
+from deepinv.physics import LinearPhysics
+from .base import Reconstructor
+
+ms = torch
+
+
+class LinearSpline_Func(torch.autograd.Function):
```
tachella: put first the most important class (the one accessible through documentation), and then include the helper functions and classes (eg `LinearSpline_Func`)

tachella (changes requested):
hey @Holimmo7 

Many thanks for this contribution! I think we can still improve the PR to better illustrate the model capabilities and leverage pre-existing least square solvers

- The implementation appears to redefine conjugate gradient (CG) logic, even though robust, well-tested CG solvers and other alternatives [already exist](https://deepinv.github.io/deepinv/api/stubs/deepinv.optim.linear.least_squares.html#deepinv.optim.linear.least_squares).

- The description refers to the approach as "feedforward," but it operates iteratively. 

- The current example could be expanded to better illustrate the algorithm’s mechanics, such as detailing its prior, iterations, its strengths and limitations, and perhaps including a visual or step-by-step breakdown of a key component. As it stands, the demonstration focuses on running the model on a single image, which may not fully convey its broader applicability.

- the testing approach diverges from the standard `test_models` framework, introducing new, isolated test functions. Aligning with the existing test suite could improve maintainability and ensure consistency with the project’s testing conventions.

Holimmo7:
Hi @tachella,

Thank you very much for your detailed feedback.

I updated the PR based on your comments. I changed the description from feedforward to iterative and added a clearer mathematical explanation in the DEAL docstring. I also updated the docstrings to follow the :param: and :type: style used in the library.

In the demo, I added more explanation, showed how to use the prior separately, printed the number of parameters, and visualized the mask to better illustrate the behavior. I also added a standalone DEALRegularizer using deepinv.optim.Prior, integrated DEAL into the standard test pipeline, and reorganized the file so the main DEAL class comes first.

For the solver, I replaced the main inverse-problem CG with the DeepInverse conjugate_gradient. I kept the internal CG in the denoising/backward part for now, because replacing it directly changed the behavior and reduced reconstruction quality. This way, the main solver uses the official implementation while keeping the current results stable.

Please let me know if you would like me to further refactor.

deepinv/models/__init__.py
```diff
@@ -37,6 +37,9 @@
 from .precond import EDMPrecond
 from .ram import RAM
 from .client import Client
+from .wrapper import DiffusersDenoiserWrapper
```
mh-nguyen712: duplicate import (L42 too)

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+    r"""
+    Deep Equilibrium Attention Least Squares (DEAL) reconstruction model.
+
+    This model solves linear inverse problems using a learned equilibrium-based
+    regularizer combined with iterative conjugate gradient least-squares updates. It can be used for
+    image restoration and reconstruction tasks such as denoising, deblurring,
+    and computed tomography reconstruction.
+
+    This implementation is adapted from the official DEAL repository:
+    https://github.com/mehrsapo/DEAL
```
mh-nguyen712: I think it's better to use href here
```suggestion
    This implementation is adapted from the official `DEAL repository<https://github.com/mehrsapo/DEAL>`_.
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+
+    For the original method, see :footcite:t:`pourya2025dealing`.
+
+    A pretrained network can be loaded by setting ``pretrained='download'``.
+
+    The reconstruction is obtained by solving a regularized least-squares problem.
+
+    .. math::
+
+        \hat{x} = \arg\min_x \frac{1}{2}\|Ax - y\|^2 + \lambda R_\theta(x)
```
mh-nguyen712: The regularizer term is denoted by g everywhere in the library and R is usually denoted the reconstruction model. 
I think using the same notation is preferable

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+                    "trained_models/deal_color.pth"
+                )
+
+            else:
+                url = (
+                    "https://raw.githubusercontent.com/mehrsapo/DEAL/main/"
+                    "trained_models/deal_gray.pth"
+                )
+
+            state = torch.hub.load_state_dict_from_url(
```
mh-nguyen712: Please use `dinv.models.utils.load_state_dict_from_url` instead (with the same args), so we can handle the downloading verbose easier. Thanks!

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+            {
+                key: value
+                for key, value in raw_state_dict.items()
+                if key in model_state_dict
+            }
+        )
+
+        self.model.load_state_dict(merged_state_dict, strict=True)
+
+    @torch.no_grad()
```
mh-nguyen712: I think we should remove this decorator, unless the model can not be trained.

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+
+        x_init = torch.zeros_like(Ht(y))
+
+        if hasattr(self.model, "max_iter"):
+            self.model.max_iter = max(int(self.max_iter), 1)
+
+        x_hat = self.model.solve_inverse_problem(
+            y,
+            H=H,
+            Ht=Ht,
```
mh-nguyen712: I would use physics.A and physics.A_adjoint directly here instead of defining H and Ht

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+
+        :param physics: forward operator
+        :type physics: deepinv.physics.LinearPhysics
+
+        :return: reconstructed image
+        :rtype: torch.Tensor
+        """
+        y = y.to(self.device)
+
+        if physics.__class__.__name__ == "Denoising":
```
mh-nguyen712:
```suggestion
        if isinstance(physics, Denoising):
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+        ctx: Any,
+        x: torch.Tensor,
+        coefficients: torch.Tensor,
+        x_min: torch.Tensor,
+        x_max: torch.Tensor,
+        num_knots: int,
+        zero_knot_indexes: torch.Tensor,
+    ) -> torch.Tensor:
+        """Evaluate the linear spline activation."""
+        step_size = (x_max - x_min) / (num_knots - 1)
```
mh-nguyen712, line 271: edge case when num_knots=1?

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+        ctx: Any,
+        x: torch.Tensor,
+        coefficients: torch.Tensor,
+        x_min: torch.Tensor,
+        x_max: torch.Tensor,
+        num_knots: int,
+        zero_knot_indexes: torch.Tensor,
+    ) -> torch.Tensor:
+        """Evaluate the derivative of the linear spline activation."""
+        step_size = (x_max - x_min) / (num_knots - 1)
```
mh-nguyen712, line 337: same here

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+            x_init=x_init,
+            verbose=False,
+            path=False,
+        )
+
+        return x_hat.clamp(0.0, 1.0) if self.clamp_output else x_hat
+
+
+class LinearSpline_Func(torch.autograd.Function):
+    """
```
mh-nguyen712, line 247: I think using underscore (_) for class naming in not recommended for classes in python.

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+        )
+        grad_coefficients_vect.scatter_add_(
+            0, indexes.view(-1), ((1 - fracs) * grad_out).view(-1)
+        )
+
+        grad_coefficients = grad_coefficients_vect.view(coefficients.shape)
+        return grad_x, grad_coefficients, None, None, None, None
+
+
+class LinearSplineDerivative_Func(torch.autograd.Function):
```
mh-nguyen712:
```suggestion
class LinearSplineDerivativeFunc(torch.autograd.Function):
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+            0, indexes.view(-1) + 1, torch.ones_like(fracs).view(-1) / step_size
+        )
+        grad_coefficients_vect.scatter_add_(
+            0, indexes.view(-1), -torch.ones_like(fracs).view(-1) / step_size
+        )
+
+        return grad_x, grad_coefficients_vect, None, None, None, None
+
+
+class Quadratic_Spline_Func(torch.autograd.Function):
```
mh-nguyen712:
```suggestion
class QuadraticSplineFunc(torch.autograd.Function):
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+        if self.slope_min is not None or self.slope_max is not None:
+            new_slopes = torch.clamp(new_slopes, self.slope_min, self.slope_max)
+
+        if self.clamp:
+            new_slopes[:, 0] = 0
+            new_slopes[:, -1] = 0
+
+        new_cs = torch.zeros(self.coefficients.shape, device=device, dtype=cs.dtype)
+        new_cs[:, 1:] = torch.cumsum(new_slopes, dim=1) * self.step_size
+        new_cs = new_cs + (cs - new_cs).mean(dim=1).unsqueeze(1)
```
mh-nguyen712:
```suggestion
        new_cs = new_cs + (cs - new_cs).mean(dim=1, keepdim=True)
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+        ).squeeze(1)
+
+    def tv2(self) -> torch.Tensor:
+        """
+        Compute the second-order total-variation regularization.
+        """
+        return self.relu_slopes.norm(1, dim=1).sum()
+
+
+def symmetric_pad(tensor: ms.Tensor, pad_size: tuple[int, int, int, int]) -> ms.Tensor:
```
mh-nguyen712: This looks like torch.nn.functional.pad with `reflect`?

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+        dirac = self.dirac.to(dtype=self.conv_layers[0].weight.dtype)
+        return self.transpose(self.convolution(dirac))
+
+
+class ZeroMean(nn.Module):
+    """Enforce zero-mean kernels for each output channel."""
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """Project kernels to have zero mean per output channel."""
+        return x - torch.mean(x, dim=(1, 2, 3)).unsqueeze(1).unsqueeze(2).unsqueeze(3)
```
mh-nguyen712:
```suggestion
        return x - torch.mean(x, dim=(1, 2, 3), keepdim=True)
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1279 @@
...
+        self.cal_scaling(sigma)
+
+        if self.training:
+            self.c_k_list = []
+            grad_steps = 1
+            n_out = int(torch.randint(14, 59, (1, 1)).item())
+            n_in = 50
+            eps_in = 1e-4
+            eps_out = 1e-4
+            eps_bck = 1e-4
```
mh-nguyen712, line 1041: These values seem a bit tricky. Is there any detail about these parameters?

deepinv/__init__.py
```diff
@@ -68,5 +68,7 @@
 # Check Python version
 if sys.version_info < (3, 10):  # pragma: no cover
     warnings.warn(
-        "You are using a Python version lower than 3.10. deepinv officially supports Python >= 3.10. Running on < 3.10 may work, but it is unsupported, unstable and may lead to unexpected bugs. For more information on updating to Python 3.10, see https://www.python.org/downloads"
+        "You are using a Python version lower than 3.10. deepinv officially supports Python >= 3.10. "
```
mh-nguyen712, line 71: Why changing this?

examples/unfolded/demo_deal.py
```diff
@@ -0,0 +1,116 @@
...
+
+with torch.no_grad():
+    x_lin = physics.A_dagger(y)
+    x_hat = model(y, physics)
+psnr = PSNR()
+psnr_y = psnr(y, x).item()
+psnr_lin = psnr(x_lin, x).item()
+psnr_hat = psnr(x_hat, x).item()
+
+plot(
```
mh-nguyen712, line 175: Can you check again the model? I run a this example and the reconstruction is not working yet

mh-nguyen712 (changes requested):
Many thanks for the new model, that's a lot of hard works! Here is my first batch of high-level comments. Will try to look into details.

deepinv/models/deal.py
Holimmo7: Thanks, good point. These helpers were unused in the current implementation, so I removed them.

deepinv/models/deal.py
Holimmo7: Thanks for pointing this out. I added comments explaining the role of these parameters and that they follow the original DEAL implementation.

examples/unfolded/demo_deal.py
Holimmo7: Thanks for checking this. I updated the demo to use the pretrained denoising setting, which is the native setting of the released checkpoints. I re-ran the example locally and it now works correctly, with DEAL improving the noisy image from about 20.2 dB to about 29.5 dB PSNR on the butterfly example.

Holimmo7:
Hi @mh-nguyen712,
Many thanks for the detailed feedback and all the helpful comments!
I have gone through them one by one and addressed them all. The implementation has been updated, and the demo has been fixed and verified locally. The reconstruction now works correctly with the pretrained setting.

examples/models/demo_deal.py
```diff
@@ -0,0 +1,120 @@
+"""
+DEAL denoising demo.
```
tachella: Can you do a demo on both denoising and reconstruction (eg deblurring or inpainting)?

examples/models/demo_deal.py
```diff
@@ -0,0 +1,120 @@
...
+
+from deepinv.loss.metric import PSNR
+from deepinv.models import DEAL, DEALRegularizer
+from deepinv.physics import Denoising, GaussianNoise
+from deepinv.utils import load_example, plot
+
+device = "cuda" if torch.cuda.is_available() else "cpu"
+
+# We use the grayscale DEAL checkpoint in this demo, so we keep one channel.
+x = load_example("butterfly.png", img_size=128).to(device)[:, 0:1, :, :]
```
tachella:
```suggestion
x = load_example("butterfly.png", img_size=128, device=device, grayscale=True)
```

examples/models/demo_deal.py
```diff
@@ -0,0 +1,120 @@
...
+prior = DEALRegularizer(model.model)
+
+with torch.no_grad():
+    grad_prior = prior.grad(x, sigma=sigma255)
+    mask = model.model.mask.mean(dim=1, keepdim=True)
+
+print(f"Standalone DEAL prior gradient shape: {tuple(grad_prior.shape)}")
+
+# %%
+# Reconstruct the image, compare with a baseline, and display PSNR.
```
tachella: explain in the text what is the mask that it is being plotted

deepinv/models/deal.py
```diff
@@ -0,0 +1,1232 @@
...
+    a linearized least-squares subproblem using conjugate gradient.
+
+    .. math::
+
+        x^{(k+1)} \approx \arg\min_x \frac{1}{2}\|Ax - y\|^2 + \lambda \nabla g_\theta(x^{(k)})^\top x
+
+    The regularizer is parameterized by a neural network which produces
+    spatially varying weights, allowing the model to adapt to local image structure.
+
+    :param pretrained: checkpoint path or ``'download'``.
```
tachella: move all parameters to the default of the library which is :param type name: (without the :type format below)
```suggestion
    :param str pretrained: checkpoint path or ``'download'``
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1232 @@
...
+
+    A pretrained network can be loaded by setting ``pretrained='download'``.
+
+    The reconstruction is obtained by solving a regularized least-squares problem.
+
+    .. math::
+
+        \hat{x} = \arg\min_x \frac{1}{2}\|Ax - y\|^2 + \lambda g_\theta(x)
+
+    where :math:`A` is the forward operator, :math:`y` the measurements,
```
tachella: I would prefer giving the explicit form of g_theta, ie the one provided in the paper

deepinv/models/deal.py
```diff
@@ -0,0 +1,1232 @@
...
+    :type device: str or None
+
+    :param clamp_output: if ``True``, clamp output to ``[0, 1]``.
+    :type clamp_output: bool
+    """
+
+    def __init__(
+        self,
+        pretrained: str,
+        sigma: float = 25.0,
```
tachella: please use values between 0 and 1, which is the standard of the library (ie divide by 255)
```suggestion
        sigma: float = .1,
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1232 @@
...
+    @property
+    def device(self) -> torch.device:
+        """Return the current device of the internal DEAL module."""
+        return next(self.model.parameters()).device
+
+    def forward(self, y: torch.Tensor, physics: LinearPhysics) -> torch.Tensor:
+        """
+        Run the DEAL reconstruction.
+
+        :param y: input measurements (e.g. sinogram)
```
tachella: same here with formatting :param type name:

tachella (changes requested):
Thanks for the new changes! I have a few more comments to address

Holimmo7:
Hi @tachella,
Thanks for the feedback. 
I have addressed all the comments by updating the demo to include both denoising and a reconstruction example, using the suggested call, clarifying the meaning of the plotted mask, and improving the Deal documentation with an explicit regularizer form and the library standard  docstring style. I also adjusted the default sigma to follow the library convention and re ran the demo and tests locally.

deepinv/tests/test_models.py
```diff
@@ -138,13 +142,63 @@ def choose_denoiser(name, imsize):
...
             in_channels=imsize[0],
             out_channels=imsize[0],
             img_resolution=imsize[1],
             pretrained=None,
         )
+
+    elif name == "deal":
+        base_model = dinv.models.DEAL(
+            pretrained="download",
+            sigma=25.0,
```
tachella: This should be in normalized range [0,1]

deepinv/models/deal.py
```diff
@@ -0,0 +1,1228 @@
...
+        )
+
+        self.model.load_state_dict(merged_state_dict, strict=True)
+
+    @property
+    def device(self) -> torch.device:
+        """Return the current device of the internal DEAL module."""
+        return next(self.model.parameters()).device
+
+    def forward(self, y: torch.Tensor, physics: LinearPhysics) -> torch.Tensor:
```
tachella: If the model can work both as a Denoiser or Reconstructor, I think the best would be to allow 
```suggestion
    def forward(self, y: torch.Tensor, physics: LinearPhysics | None, sigma: float = None) -> torch.Tensor:
```
And let the user apply it as denoiser as `x_hat = model(y, sigma=sigma)` or as a reconstructor `x_hat=model(y, physics)`, see for example how it was done [here](https://deepinv.github.io/deepinv/api/stubs/deepinv.models.RAM.html#deepinv.models.RAM.forward)

deepinv/models/deal.py
```diff
@@ -0,0 +1,1228 @@
...
+
+class ZeroMean(nn.Module):
+    """Enforce zero-mean kernels for each output channel."""
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """Project kernels to have zero mean per output channel."""
+        return x - torch.mean(x, dim=(1, 2, 3), keepdim=True)
+
+
+class DEALRegularizer(Prior):
```
tachella: if we don't provide any example using this, I think we can remove this class for this PR (could be re-added on a later PR if it works as a generic regularizer).

examples/models/demo_deal.py
```diff
@@ -0,0 +1,152 @@
...
+`DEAL repository <https://github.com/mehrsapo/DEAL>`_.
+"""
+
+# %%
+# Import packages and load a grayscale example image.
+
+import torch
+
+from deepinv.loss.metric import PSNR
+from deepinv.models import DEAL, DEALRegularizer
```
tachella: remove the DEALRegularizer if not really used

examples/models/demo_deal.py
```diff
@@ -0,0 +1,152 @@
...
+torch.manual_seed(0)
+
+# Load image (grayscale)
+x = load_example("butterfly.png", img_size=128, device=device, grayscale=True)
+
+# %%
+# IMPORTANT: sigma handling
+# physics uses [0,1], model uses [0,255]
+
+sigma255 = 25.0
```
tachella: this shouldn't be necessary if the model's inputs are already following the [0,1] convention (instead of [0,255])

deepinv/models/deal.py
```diff
@@ -0,0 +1,1228 @@
...
+    :param float target_y_std: target std for auto-scaling when enabled
+    :param bool color: if ``True``, use the color DEAL variant; otherwise grayscale
+    :param str | None device: compute device. If ``None``, use CUDA if available
+    :param bool clamp_output: if ``True``, clamp output to ``[0, 1]``
+    """
+
+    def __init__(
+        self,
+        pretrained: str,
+        sigma: float = 0.1,
```
tachella:
```suggestion
        sigma_denoiser: float = 0.1,
```
This is somewhat similar to the sigma used for denoisers in the context of plug-and-play? If yes, then I would rename it to follow the deepinv convention, see [here](file:///C:/Users/julian/Downloads/documentation%20(26)/api/stubs/deepinv.optim.PGD.html#deepinv.optim.PGD)

deepinv/models/deal.py
```diff
@@ -0,0 +1,1228 @@
...
+    a linearized least-squares subproblem using conjugate gradient:
+
+    .. math::
+
+        x^{(k+1)} \approx \arg\min_x \frac{1}{2}\|Ax - y\|^2
+        + \lambda \nabla g_\theta(x^{(k)})^\top x
+
+    :param str pretrained: checkpoint path or ``'download'``
+    :param float sigma: noise-level parameter used by DEAL
+    :param float lam: regularization strength used by the DEAL solver
```
tachella:
```suggestion
    :param float lambda_reg: regularization strength :math:`\lambda` used by the DEAL solver
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1228 @@
...
+
+    where :math:`A` is the forward operator, :math:`y` the measurements, and
+    :math:`g_\theta` is the learned adaptive regularizer.
+
+    In the implementation, the learned regularizer is induced by a masked linear
+    operator of the form
+
+    .. math::
+
+        L_\theta(x) = m_\theta(x) \odot K_\theta x,
```
tachella: I think having the same x for both the mask and the final term is highly confusing, since in the end you take the gradient wrt to the last x but not the one inside the mask. Maybe adding an auxiliary variable u?
We should also add the channel index to indicate that this is not a single mask
```suggestion
        L_{\theta,c}(u, x) = m_{\theta,c}(u) \odot K_{\theta,c} x,
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1228 @@
...
+
+    so that the regularization term can be written as
+
+    .. math::
+
+        g_\theta(x) = \frac{1}{2}\|L_\theta(x)\|_2^2
```
tachella:
```suggestion
        g_{\theta}(u, x) = \sum_{c=1}^{C} \frac{1}{2}\|L_{\theta,c}(u, x)\|_2^2
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1228 @@
...
+        = \frac{1}{2}\|m_\theta(x) \odot K_\theta x\|_2^2,
```
tachella:
```suggestion
        = \sum_{c=1}^{C} \frac{1}{2}\|m_{\theta,c}(u) \odot K_{\theta,c} x\|_2^2,
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1228 @@
...
+
+    where :math:`K_\theta` is a learned linear operator,
+    :math:`m_\theta(x)` is a spatially varying mask predicted by the network,
```
tachella:
```suggestion
    :math:`m_{\theta,c}(u)` are spatially varying masks predicted by the network,
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1228 @@
...
+        L_\theta(x) = m_\theta(x) \odot K_\theta x,
+
+    so that the regularization term can be written as
+
+    .. math::
+
+        g_\theta(x) = \frac{1}{2}\|L_\theta(x)\|_2^2
+        = \frac{1}{2}\|m_\theta(x) \odot K_\theta x\|_2^2,
+
+    where :math:`K_\theta` is a learned linear operator,
```
tachella:
```suggestion
    where :math:`K_{\theta,c}` are learned linear filters,
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1228 @@
...
+    and :math:`\odot` denotes element-wise multiplication.
+
+    The optimization is performed iteratively using a fixed-point scheme.
+    At each outer iteration, the algorithm updates the reconstruction by solving
+    a linearized least-squares subproblem using conjugate gradient:
+
+    .. math::
+
+        x^{(k+1)} \approx \arg\min_x \frac{1}{2}\|Ax - y\|^2
+        + \lambda \nabla g_\theta(x^{(k)})^\top x
```
tachella:
```suggestion
        + \lambda \nabla_{x} g_\theta(u=x^{(k)},x)^\top x
```

examples/models/demo_deal.py
```diff
@@ -0,0 +1,152 @@
+"""
+DEAL denoising and reconstruction demo.
+====================================================================================================
+
+This example shows how to use the Deep Equilibrium Attention Least Squares
+(DEAL) model in DeepInverse for both denoising and a simple reconstruction
+setting.
+
```
tachella, line 8: Please include here (even if it is repeating itself), the description of the model iterations and regularization from the docstring of the main class. Consider any editing to make sure that the narrative of this part can be read nicely

examples/models/demo_deal.py
```diff
@@ -0,0 +1,152 @@
+"""
+DEAL denoising and reconstruction demo.
```
tachella:
```suggestion
DEAL denoising and reconstruction
```

examples/models/demo_deal.py
```diff
@@ -0,0 +1,152 @@
...
+
+sigma255 = 25.0
+sigma01 = sigma255 / 255.0
+
+# %%
+# Load pretrained DEAL model
+
+model = DEAL(
+    pretrained="download",
+    sigma=sigma255,  # IMPORTANT: must be in [0,255]
```
tachella: same as above, there shouldn't be mappings between normalizations here

examples/models/demo_deal.py
```diff
@@ -0,0 +1,152 @@
...
+
+n_params = sum(p.numel() for p in model.parameters())
+print(f"DEAL number of parameters: {n_params:,}")
+
+prior = DEALRegularizer(model.model)
+psnr = PSNR()
+
+# %%
+# ------------------------------------------------------------
+# 1) DENOISING
```
tachella: Please write some text around each section, not a single header with no explanation.
```suggestion
# Denoising with DEAL
We first start illustrating an Gaussian denoising example with DEAL. The model can be applied as a denoiser as `y = model(y, physics)` where `physics` is a denoising operator, or by `y = model(y, sigma=sigma)` where `sigma` is the noise level. DEAL relies on regularizer that is adaptively updated at each iteration. Here we plot the regularizer masks of the last iteration. 
```

examples/models/demo_deal.py
```diff
@@ -0,0 +1,152 @@
...
+        f"PSNR: {psnr_noisy:.2f} dB",
+        f"PSNR: {psnr_denoise:.2f} dB",
+        "Mean over channels",
+    ],
+    figsize=(11, 3),
+)
+
+# %%
+# ------------------------------------------------------------
+# 2) RECONSTRUCTION: INPAINTING
```
tachella: Again, write some text to explain what is going on
```suggestion
# Reconstruction example with inpainting
bla bla bla 
```

examples/models/demo_deal.py
```diff
@@ -0,0 +1,152 @@
...
+# 1) DENOISING
+# ------------------------------------------------------------
+
+physics_denoise = Denoising(GaussianNoise(sigma=sigma01)).to(device)
+y_denoise = physics_denoise(x)
+
+with torch.no_grad():
+    grad_prior = prior.grad(x, sigma=sigma255)
+    x_hat_denoise = model(y_denoise, physics_denoise)
+    mask_denoise = model.model.mask.mean(dim=1, keepdim=True)
```
tachella, line 75: Can you update the class such that this line can be written more compactly as 
```suggestion
    mask_denoise = model.mask.mean(dim=1, keepdim=True)
```

examples/models/demo_deal.py
```diff
@@ -0,0 +1,152 @@
...
+# %%
+# ------------------------------------------------------------
+# 1) DENOISING
+# ------------------------------------------------------------
+
+physics_denoise = Denoising(GaussianNoise(sigma=sigma01)).to(device)
+y_denoise = physics_denoise(x)
+
+with torch.no_grad():
+    grad_prior = prior.grad(x, sigma=sigma255)
```
tachella: remove this if you are not plotting the gradient nor using it for recon

tachella (changes requested):
Many thanks for the last batch of changes! 

I have a few extra corrections to go :)

Holimmo7:
Hi @tachella ,
Thanks again for the detailed feedback.
 I addressed the remaining comments, including the normalized sigma usage, the denoiser/reconstructor API, the demo cleanup , and the documentation updates. I also re ran the demo, the deal tests, and the formatting checks successfully. Please let me know if there is still anything to adjust.

deepinv/models/deal.py
```diff
@@ -0,0 +1,1215 @@
...
+import torch.nn.functional as F
+import torch.nn.utils.parametrize as P
+from torch import Tensor, nn
+
+from deepinv.physics import Denoising, LinearPhysics
+from deepinv.optim.linear import conjugate_gradient
+from .base import Reconstructor
+from .utils import load_state_dict_from_url
+
+ms = torch
```
mh-nguyen712: remove this
```suggestion
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1215 @@
...
+        target_y_std: float = 25.0,
+        color: bool = False,
+        device: str | None = None,
+        clamp_output: bool = True,
+    ) -> None:
+        super().__init__()
+
+        init_device = torch.device(
+            device or ("cuda" if torch.cuda.is_available() else "cpu")
+        )
```
mh-nguyen712: I think this is not necessary just use `.to(device)`, it works even `device=None`

deepinv/models/deal.py
```diff
@@ -0,0 +1,1215 @@
...
+
+        self.sigma_denoiser = float(sigma_denoiser)
+        self.lambda_reg = float(lambda_reg)
+        self.max_iter = int(max_iter)
+        self.auto_scale = bool(auto_scale)
+        self.target_y_std = float(target_y_std)
+        self.clamp_output = bool(clamp_output)
+
+        self.model = _DEALImpl(color=color).to(init_device).eval()
```
mh-nguyen712: Maybe drop `.eval()` if the model can be trained

deepinv/models/deal.py
```diff
@@ -0,0 +1,1215 @@
...
+            )
+        else:
+            try:
+                state = torch.load(
+                    pretrained, map_location=self.device, weights_only=True
+                )
+            except TypeError:
+                state = torch.load(pretrained, map_location=self.device)
+
+        raw_state_dict = state.get("state_dict", state)
```
mh-nguyen712: This would fail if `pretrained=None`? Maybe re-structure the state loading.

deepinv/models/deal.py
```diff
@@ -0,0 +1,1215 @@
...
+                (y.size(0), 1, 1, 1),
+                255.0 * float(self.sigma_denoiser),
+                device=self.device,
+                dtype=y.dtype,
+            )
+            x_hat = self.model.denoise(y, sigma_tensor)
+            return x_hat.clamp(0.0, 1.0) if self.clamp_output else x_hat
+
+        if self.auto_scale:
+            y_std = float(y.std().detach().cpu())
```
mh-nguyen712:
```suggestion
            y_std = float(y.detach().std())
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1215 @@
...
+            if 0.0 < y_std < 5.0:
+                scale = self.target_y_std / (y_std + 1e-12)
+                y = y * scale
+
+        x_init = physics.A_adjoint(y)
+
+        if hasattr(self.model, "max_iter"):
```
mh-nguyen712, line 202: Should it be called in __init__ instead?

mh-nguyen712 (changes requested):
Some more comments!

Holimmo7:
Hi @mh-nguyen712 ,
Thanks for the new feedback!
I Removed the ms = torch alias, simplified device handling by relying on .to(device) , dropped .eval() from the constructor and instead set it explicitly in the demo. Also made checkpoint loading safe when pretrained=None and updated the y_std computation as suggested.
For max_iter, I kept the update in forward so that changes to model.max_iter after initialization are still respected without recreating the wrapper.
Please let me know if there is any other suggestion.

deepinv/models/deal.py
```diff
@@ -0,0 +1,1207 @@
...
+        with torch.no_grad():
+            for _ in range(n_steps):
+                u = self.transpose(self.convolution(u))
+                u = u / torch.linalg.norm(u)
+
+            sn = torch.linalg.norm(self.transpose(self.convolution(u)))
+            self.L.copy_(sn)
+            return sn
+
+    def check_tranpose(self) -> None:
```
mh-nguyen712: it seems that this is a test function? Would it be more logic to put it in test?

deepinv/models/deal.py
```diff
@@ -0,0 +1,1207 @@
...
+        Ht: Callable[[torch.Tensor], torch.Tensor],
+        idx: list[int] | None = None,
+    ) -> torch.Tensor:
+        """Apply the linear system operator solved inside conjugate gradients."""
+        BtBD = (Ht(H(x)) + self.lmbda[idx] * self.Lt(self.L(x, idx), idx)) / (
+            1 + self.lmbda[idx]
+        )
+        return BtBD
+
+    def cg_sample(
```
mh-nguyen712, line 918: Is it possible to use the `least_squares` solver of deepinverse?

Holimmo7: Thanks again, 
The public reconstruction wrapper is already using the DeepInverse least-squares tooling more consistently, but this remaining BtB / cg_sample logic is still part of the internal original DEAL implementation. I kept it unchanged here because replacing that internal path can affect the pretrained model behavior.

deepinv/models/deal.py
Holimmo7: Good point, this was only a validation/debug helper, so I removed it from the model implementation.

Holimmo7:
Hi @tachella @mh-nguyen712,
This is just a follow up.
Thnak you in advance.

examples/unfolded/demo_deal.py
```diff
@@ -0,0 +1,152 @@
+"""
+DEAL denoising and reconstruction demo.
+====================================================================================================
+
+This example shows how to use the Deep Equilibrium Attention Least Squares
+(DEAL) model in DeepInverse for both denoising and a simple reconstruction
+setting.
+
```
tachella: I think my previous comment was not fully addressed - the main iterations of the algorithms in ..math:: mode are missing. It can be the same equations used in the docstring of the DEAL class (even if it repeats itself, it's good to keep the mathematical description on both parts)

deepinv/tests/test_models.py
```diff
@@ -1689,3 +1743,138 @@ def test_gaussian_noise_estimators(
...
+        max_iter=1,
+        auto_scale=False,
+        clamp_output=True,
+        color=False,
+        device=device,
+    )
+    model.eval()
+
+    # Simple DeepInverse physics (denoising)
+    physics = Denoising().to(device)
```
tachella:
```suggestion
    physics = Denoising()
```
device here is not necessary

deepinv/tests/test_models.py
```diff
@@ -138,13 +142,63 @@ def choose_denoiser(name, imsize):
...
+        base_model = dinv.models.DEAL(
+            pretrained="download",
+            sigma_denoiser=0.1,
+            lambda_reg=10.0,
+            max_iter=5,
+            auto_scale=False,
+            color=(imsize[0] == 3),
+        )
+
+        class DEALDenoiserWrapper(nn.Module):
```
tachella: why do you need this wrapper? the different sigma cases cannot be handled inside the standard DEAL class?

deepinv/tests/test_models.py
tachella: now that is added to the MODEL_LIST, I don't think this class and the tests below are really necessary anymore

deepinv/models/deal.py
```diff
@@ -0,0 +1,1189 @@
...
+    :param bool auto_scale: if ``True``, rescales measurements based on their std
+    :param float target_y_std: target std for auto-scaling when enabled
+    :param bool color: if ``True``, use the color DEAL variant; otherwise grayscale
+    :param str | None device: compute device. If ``None``, use CUDA if available
+    :param bool clamp_output: if ``True``, clamp output to ``[0, 1]``
+    """
+
+    def __init__(
+        self,
+        pretrained: str,
```
tachella:
```suggestion
        pretrained: str = "pretrained",
```
set to pretrained by default

tachella: can you also please set this param in the end of the list (after `clamp_output`)? Other models in the library typically have pretrained at the end of the input list, not at the very beginning

examples/unfolded/demo_deal.py
```diff
@@ -0,0 +1,151 @@
+"""
```
tachella: Please put this example in `examples/unfolded` instead of `models`

tachella (changes requested):
Thanks for the changes, I have a new set of comments which should be the last ones if you address them :)

deepinv/models/deal.py
```diff
@@ -0,0 +1,1189 @@
...
+                map_location=self.device,
+                file_name=url.split("/")[-1],
+            )
+        else:
+            try:
+                state = torch.load(
+                    pretrained, map_location=self.device, weights_only=True
+                )
+            except TypeError:
+                state = torch.load(pretrained, map_location=self.device)
```
mh-nguyen712: These two line do the same thing (torch.load has weights_only=True by default)

deepinv/models/deal.py
```diff
@@ -0,0 +1,1189 @@
...
+            merged_state_dict = model_state_dict.copy()
+            merged_state_dict.update(
+                {
+                    key: value
+                    for key, value in raw_state_dict.items()
+                    if key in model_state_dict
+                }
+            )
+
+            self.model.load_state_dict(merged_state_dict, strict=True)
```
mh-nguyen712: I would prefer use a simple `load_state_dict`. This is a good safety filter but can silently ignore unmatched weights. 
```suggestion
            self.model.load_state_dict(state, strict=True)
```

deepinv/models/deal.py
```diff
@@ -0,0 +1,1189 @@
...
+
+        if isinstance(physics, Denoising):
+            sigma_tensor = torch.full(
+                (y.size(0), 1, 1, 1),
+                255.0 * float(self.sigma_denoiser),
+                device=self.device,
+                dtype=y.dtype,
+            )
+            x_hat = self.model.denoise(y, sigma_tensor)
+            return x_hat.clamp(0.0, 1.0) if self.clamp_output else x_hat
```
mh-nguyen712: These two branches can be merged into one.

deepinv/models/deal.py
```diff
@@ -0,0 +1,1189 @@
...
+
+        if self.auto_scale:
+            y_std = float(y.detach().std())
+            if 0.0 < y_std < 5.0:
```
mh-nguyen712, line 223: This is tricky. If it's really important, should be stated clearly in the docstring (on top of __init__)

deepinv/models/deal.py
```diff
@@ -0,0 +1,1189 @@
...
+        )
+        return BtBD
+
+    def cg_sample(
+        self,
+        b: torch.Tensor,
+        x0: torch.Tensor,
+        max_iter: int,
+        eps: float = 1e-5,
+    ) -> tuple[torch.Tensor, int]:
```
mh-nguyen712: It seems that this function is not used. Please remove it

deepinv/models/deal.py
```diff
@@ -0,0 +1,1189 @@
...
+            eps_in = 1e-6
+            eps_out = 1e-5
+
+        c_k = torch.zeros_like(y)
+        c_k_old = c_k.clone()
+
+        with torch.no_grad():
+            for i in range(n_out):
+                self.cal_mask(c_k)
+                c_k, self.last_cg_iter = self.cg(y, c_k_old, n_in, eps=eps_in)
```
mh-nguyen712: In `solve_inverse_problem`, you can now use deepinv's `conjugate_gradient` function. It would be really nice to also use that function here. I think they are quite similar and I'm happy to help if needed.

docs/source/changelog.rst
```diff
@@ -8,8 +8,11 @@ Current
 
 New Features
 ^^^^^^^^^^^^
+New Features
+^^^^^^^^^^^^
```
mh-nguyen712:
```suggestion

```

mh-nguyen712 (changes requested):
Many thanks for the updates, some more comments then I think we're good!

Holimmo7:
Hi @tachella,
Thank you very much for the comments. I addressed them and pushed the changes. I also ran Black, the full test_models.py test file, and the DEAL demo successfully.

deepinv/models/deal.py
Holimmo7: Thanks, I removed the previous silent filtering. The official DEAL checkpoint is wrapped in a state_dict, and it also does not store some deterministic model buffers. I now extract the state_dict, complete only missing buffers from the initialized model, and then call load_state_dict(.., strict=True).

deepinv/models/deal.py
Holimmo7: Thanks, I replaced the no-grad denoising fixed-point CG call with DeepInverse conjugate_gradient. I kept the custom CG in the training/backward hook path for now to avoid changing the implicit differentiation behavior.
Please let me know if you think this part still needs any further adjustment.

Holimmo7:
Hi @mh-nguyen712,
Thank you very much for the helpful comments. I addressed the requested changes: simplified the loading logic where possible, merged the denoising branches, documented auto_scale, removed the unused function, replaced the no-grad denoising CG call with DeepInverse's conjugate gradient, and cleaned the changelog heading. The Deal tests pass locally.
Please let me know if any further adjustment is needed.

examples/unfolded/demo_deal.py
mh-nguyen712: @Holimmo7 can you check that this is addressed please?

docs/source/changelog.rst
```diff
@@ -8,6 +8,7 @@ Current
 
 New Features
 ^^^^^^^^^^^^
+- Add :class:`deepinv.models.DEAL` model (:gh:`939` by `Hossein Alimohammadi`_)
```
mh-nguyen712: Wrong PR number
```suggestion
- Add :class:`deepinv.models.DEAL` model (:gh:`1107` by `Hossein Alimohammadi`_)
```

mh-nguyen712 (commented):
Many thanks for the updates, looks really good now @Holimmo7! 
Happy to approve once the gpu-tests are passing! @tachella can you make a quick double check?

Holimmo7:
Hi @mh-nguyen712,
Many thanks for the review and all the helpful comments.

examples/unfolded/demo_deal.py
```diff
@@ -0,0 +1,190 @@
+r"""
+DEAL denoising and reconstruction
+====================================================================================================
+
+This example shows how to use the Deep Equilibrium Attention Least Squares
+(DEAL) model in DeepInverse for both denoising and a simple reconstruction
+setting.
```
tachella:
```suggestion
settings.
```

examples/unfolded/demo_deal.py
```diff
@@ -0,0 +1,190 @@
...
+where :math:`A` is the forward operator, :math:`y` are the measurements, and
+:math:`g_{\theta}` is the learned adaptive regularizer.
+
+In DEAL, the regularizer is induced by a masked linear operator
+
+.. math::
+
+    L_{\theta,c}(u, x) = m_{\theta,c}(u) \odot K_{\theta,c}x,
+
+which gives
```
tachella:
```suggestion
which depends on an auxiliary image :math:`u`, resulting in
```

examples/unfolded/demo_deal.py
```diff
@@ -0,0 +1,190 @@
...
+        "",
+        f"PSNR: {psnr_noisy:.2f} dB",
+        f"PSNR: {psnr_denoise:.2f} dB",
+        "Mean over channels",
+    ],
+    figsize=(11, 3),
+)
+
+# %%
+# Reconstruction example with inpainting
```
tachella, line 144: same here add the line:
```suggestion
# Reconstruction example with inpainting
----------------------------------------------
```

examples/unfolded/demo_deal.py
```diff
@@ -0,0 +1,190 @@
...
+)
+model.eval()
+
+n_params = sum(p.numel() for p in model.parameters())
+print(f"DEAL number of parameters: {n_params:,}")
+
+psnr = PSNR()
+
+# %%
+# Denoising with DEAL
```
tachella, line 104: Add this line to have a proper subsection with title
```suggestion
# Denoising with DEAL
-------------------------
```

examples/unfolded/demo_deal.py
```diff
@@ -0,0 +1,152 @@
...
+# 1) DENOISING
+# ------------------------------------------------------------
+
+physics_denoise = Denoising(GaussianNoise(sigma=sigma01)).to(device)
+y_denoise = physics_denoise(x)
+
+with torch.no_grad():
+    grad_prior = prior.grad(x, sigma=sigma255)
+    x_hat_denoise = model(y_denoise, physics_denoise)
+    mask_denoise = model.model.mask.mean(dim=1, keepdim=True)
```
tachella: Can you please address this comment? This hasn't been resolved in the latest fixes :)

tachella (changes requested):
Hey @Holimmo7, thanks for the PR, I think we are quite close to finishing, please address the final comments :)

Holimmo7:
Dear @tachella,
Thanks for the final comments. I addressed the wording and subsection title suggestions in the DEAL demo. The old DEALRegularizer/grad_prior code is no longer present in the current version of the example. The demo now directly runs DEAL denoising and inpainting and only plots the output and mask.
I also updated the demo to use the public model.mask property instead of accessing model.model.mask directly.

examples/unfolded/demo_deal.py
Holimmo7: Thanks, this is now addressed in the current version of the demo. I added the mathematical description of the reconstruction problem, the masked regularizer, and the fixed-point iteration in .. math:: mode near the beginning of the example.

examples/unfolded/demo_deal.py
Holimmo7: Thanks, I updated the demo to use the public model.mask property instead of accessing model.model.mask directly.

tachella (changes requested):
checking again (I think my previous comment was incorrect)

Holimmo7:
Hi @tachella, @mh-nguyen712,
Thanks again for your helpful comments. I believe I have addressed all requested changes in the latest version. Could you please let me know if there is anything still preventing the PR from being merged?

tachella (approved):
LGTM, thanks for adding this method to the library!

mh-nguyen712 (approved):
LGTM! Many thanks for the addition!

Holimmo7:
Dear @mh-nguyen712 , @tachella,
Thank you very much for the review and for merging the PR. I really appreciate your time and helpful feedback.

PR #1117, opened 2026-03-27
Typing in deepinv/optim

### LINKED ISSUE #1070: Add typing to deepinv/optim
It seems there is very few typing in this part of the library. This is just a subissue of #536 for me to remember to do it.

Tmodrzyk:
Static typing + docstring fixes in `deepinv/optim`.
Most changes are located in `optim_iterators`, there was almost no static typing there.
Towards #536, closes #1070 .
I tried checking that I didn't miss any untyped definition using `mypy`, but I can't seem to make it work properly so I'm not sure that I covered 100% of the cases...

### DIFF
--- a/deepinv/optim/fixed_point.py
+++ b/deepinv/optim/fixed_point.py
@@ -7,7 +7,6 @@ import warnings
 from tqdm import tqdm
 
 if TYPE_CHECKING:
-    from deepinv.physics import Physics
     import deepinv
 
 
--- a/deepinv/optim/linear/utils.py
+++ b/deepinv/optim/linear/utils.py
@@ -1,5 +1,4 @@
 from __future__ import annotations
-from typing import TYPE_CHECKING
 from deepinv.utils.tensorlist import TensorList
 import torch
 


deepinv/optim/linear/bicgstab.py
```diff
@@ -2,21 +2,20 @@
 from .utils import dot
 from deepinv.utils.tensorlist import TensorList, zeros_like
 import torch
-from torch import Tensor
 from typing import Callable
 
 
 def bicgstab(
     A: Callable,
-    b: Tensor,
```
mh-nguyen712, line 11: Why changing to torch.Tensor? It seems that Tensor was working here (even sometimes it does not and I don't totally understand why)

deepinv/optim/linear/conjugate_gradient.py
```diff
@@ -1,18 +1,19 @@
...
+    A: Callable,
+    b: torch.Tensor,
     max_iter: int = 1e2,
     tol: float = 1e-5,
     eps: float = 1e-8,
     parallel_dim: None | int | list[int] = 0,
     init=None,
     verbose: bool = False,
-):
+) -> torch.Tensor:
```
mh-nguyen712, line 16: add typing for init too

deepinv/optim/fixed_point.py
```diff
@@ -75,9 +75,9 @@ class FixedPoint(nn.Module):
     def __init__(
         self,
         iterator: deepinv.optim.OptimIterator = None,
-        update_params_fn: Callable[int, dict[str, float | Iterable]] = None,
-        update_data_fidelity_fn: Callable[int, deepinv.optim.DataFidelity] = None,
-        update_prior_fn: Callable[int, deepinv.optim.Prior] = None,
+        update_params_fn: Callable[[int], dict[str, float | Iterable]] = None,
```
mh-nguyen712, line 77: why adding [] ?

deepinv/optim/optimizers.py
```diff
@@ -279,8 +280,8 @@ def __init__(
         has_cost: bool = False,
         backtracking: BacktrackingConfig | bool = None,
         custom_metrics: dict[str, Metric] = None,
-        custom_init: Callable[[torch.Tensor, Physics], dict] = None,
-        get_output: Callable[[dict], torch.Tensor] = lambda X: X["est"][0],
+        custom_init: Callable[[Tensor, Physics], dict] = None,
```
mh-nguyen712: why here changing torch.Tensor to Tensor?

Tmodrzyk:
Hi @mh-nguyen712 thanks for the review ! Initially I wanted to have Tensor instead of torch.Tensor everywhere but I had errors on some of the files and I couldn't figure why, so I reverted to torch.Tensor which works everywhere.
I think I forgot to reverse to torch.Tensor in some places though !

deepinv/optim/linear/bicgstab.py
Tmodrzyk: I wanted to have ``torch.Tensor`` everywhere to be consistent accross the folder. Should I keep it as ``Tensor`` ?

deepinv/optim/optimizers.py
Tmodrzyk: I revert this to ``torch.Tensor`` !

deepinv/optim/fixed_point.py
Tmodrzyk: Mypy was throwing an error saying that Callable should have a list of argument types, and not a single one.

mh-nguyen712:
Many thanks for the updates! I think using both `torch.Tensor` or `Tensor` is fine, since there is no ambiguity here. In some cases, with heavy annotations, I think `Tensor` is preferable. The main issue (that I didn't fully understand) is that in some files, sphinx raises warning of using `Tensor`, but in other files, it works. Do you have an idea?

Tmodrzyk:
Unfortunately no I still don't really now why it raises warnings on some files but not others. :disappointed: 
Should I restore the changes where I swapped `Tensor` and `torch.Tensor` and conversely ?

mh-nguyen712:
I think it's ok. Maybe in some heavy anntations, eg 
```python
X: (
        torch.Tensor
        | dict[str, torch.Tensor]
        | dict[str, torch.Tensor]
        | tuple[torch.Tensor, ...]
    )
```
we can just use `Tensor` if there is no warning. But I'm not sure if it's a good idea or we just use `torch.Tensor` everywhere?

Tmodrzyk:
I think this is a good idea to lighten the notations where possible, I'll do my best :smile:

Andrewwango:
on the Tensor vs torch.Tensor, I honestly have no idea too. I try to be laconic but ultimately I just do what sphinx tells me to do 🤷

Andrewwango (approved):
LGTM

Tmodrzyk:
Ah I didn't see that sphinx threw warnings again. I'm trying to fix it correctly but no success for now.
Rolling back to ``torch.Tensor`` would probably be easier ? 
Another dirty fix could be to ignore these specific warnings in the sphinx config but I fear it might cause more issues down the line.

mh-nguyen712:
Agree @Tmodrzyk, we can fall back to torch.Tensor can try to understand later

Tmodrzyk:
It is a bit frustrating though aha, the type hints do become harder to read without the alias `Tensor`... 
I think I solved the issue locally (with the help of gpt) but the fix is very hacky.

I'm not sure I fully understand the issue, here's what I got:
- we use intersphinx to link to torch's ``torch.Tensor`` doc page
- we use postponed annotations (with ``from __future__ import annotations``) so type hints are stored as strings
- we also use sphinx's ``autodoc_typehints`` meaning we do use the type hints present in the function's code to generate the doc. Sphinx shows them in the description box. I think the parameter type indicated in the docstring just gets thrown out, but the description of the parameter is used to generate the doc ? 

The issue arises because:
- intersphinx can only resolve fully qualified names (torch.Tensor and not the alias)
- the postponed annotations means sphinx only get the string "Tensor" from the type hints and not the fully qualified name
- sphinx decides later on that ``Tensor`` is a python class (a ``py:class``) in its own right and does not take it from Torch's inventory
- it outputs a warning (treated as error) because we don't define it anywhere

I thought using `autodoc_type_aliases` would solve the issue. But it only seems to change rendered type hints, so it doesn't solve anything.


Edit: the aforementioned hacky fix

```python


def _rewrite_type_xrefs(app: Any, env: Any, node: Any, contnode: Any):
    if node.get("refdomain") != "py" or node.get("reftype") != "class":
        return None

    target = node.get("reftarget")
    if target == "Tensor":
        node["reftarget"] = "torch.Tensor"
        return None
    if target == "'torch.Tensor'":
        node["reftarget"] = "torch.Tensor"
        return None
    # Typing internals that sometimes leak into rendered annotations.
    if target == "TypeAliasForwardRef":
        node["reftarget"] = "typing.ForwardRef"
        return None

    return None
```

mh-nguyen712:
But I still do not understand. Indeed, there are many place we use `Tensor` and it works. It seems to me that, if we change those `Tensor` into `torch.Tensor` then back to `Tensor`, it does not work anymore... 

So you need to add your hacky function to `conf.py`? If it works, I would be happy to add this

jscanvic:
(Sorry if this is noise)

We might have encountered that in the past with InterpolationMode:

https://github.com/deepinv/deepinv/blob/17a836043c8a8439052608af7b6e4a2eb13a362b/deepinv/transform/rotate.py#L39

Doing a quick research I found 

https://github.com/deepinv/deepinv/blob/17a836043c8a8439052608af7b6e4a2eb13a362b/docs/source/conf.py#L437-L441

Maybe we just need to append torch.Tensor to it?

Tmodrzyk:
Thanks for taking a look @jscanvic. This definitely already happened but I think the proposed fix is ok only for these types that appear only in a few places in the lib. ``torch.Tensor`` (or the alias) will probably appear in most files. 
This is just ignoring the warnings. I'm almost sure it will cause silent failing of the docs build down the line.

Tmodrzyk:
In the end I reverted to ``torch.Tensor`` everywhere, the behavior of the alias ``Tensor`` is too unpredictable.
I'll try to come up with a clean fix later, as this was not the initial goal of this PR.
@Andrewwango @mh-nguyen712 let me know if there's anything else I should do before merging. :smile:

PR #1123, opened 2026-03-30
Fix ComposedPhysics and StackedPhysics update_parameters logic

### LINKED ISSUE #999: Updating logic in Composed and Stacked Physics is not well behaved
### Environment

All

### Describe the bug

The `update_parameters` function in `ComposedPhysics` assumes that the composed Physics have different attribute names, if by any chance both physics have an attribute with the same name, e.g. `"filter"`, the function does not allow to update each of these independently:
https://github.com/deepinv/deepinv/blob/114d1efc640ba8e2cda554b5d2f74b8e28843c22/deepinv/physics/forward.py#L852-L859

It's exactly the same pattern for `StackedPhysics`, which might be even more prone to error if let's say I want to have a super operator that stacks different "views" of the same class of Physics.

---

Somewhat related, the `buffer` tree is not catched properly by some tests, e.g. `test_operators_differentiability`. For the "composition" operator defined here
https://github.com/deepinv/deepinv/blob/114d1efc640ba8e2cda554b5d2f74b8e28843c22/deepinv/tests/test_physics.py#L247-L259

the test does not properly process the Physics buffer structure
https://github.com/deepinv/deepinv/blob/114d1efc640ba8e2cda554b5d2f74b8e28843c22/deepinv/tests/test_physics.py#L1680-L1693
we have 
- `params=['filter']`
- `physics._buffers` which is empty
- `buffers.keys()` equals to `dict_keys(['physics_list.0.mask', 'physics_list.0.filter', 'physics_list.0.angle'])`
- `parameters` which is empty

instead we should have `parameters.keys() = ["filter"]` or `["physics_list.0.filter"]`.

### Steps to reproduce

None.

### Expected behavior

An easy fix would be to remove the `update_parameters` function at the `Composed` and `Stacked` level, and only handle the update on sub-physics, though it would definitely be breaking.

Another option, is to give a list of **kwargs to `update_parameters`, with the same structure as the `Composed` or `Stacked` operator ?

Please tell me if I'm missing something or if you have some ideas ? 

PS: @mh-nguyen712 @jscanvic the PR in #989 needs this issue to be resolved for the tests to pass properly.

_No response_

### Additional context

_No response_

romainvo:
This PR fixes #999. Currently, the `StackedPhysics` and `ComposedPhysics` classes handle the parameter update in the same manner `Physics` does, by passing kwargs to `physics.A`/`physics.A_adjoint` or `physics.update` / `physics.update_parameters`. 

This design poses a major issue as it does not handle the case where combined operators belong to the same family of physics, and may have parameters with the same name, e.g. `filter` in the case of conv-like physics. For example, with the current implementation, a user cannot implement robust multi-view operators.

I've come up with the following update logic: in `update_parameters`, instead of passing `**kwargs` where `kwargs` is a dict whose keys correspond to buffer names, we pass `params=params` where `params` can have two formats:
1. `params` is dict of dicts: the goal is to target the update on specific physics in the list (if not all the physics should be updated). In the next snippet, only the physics of indices 0 and 2 are updated
```python
stacked_physics = StackedPhysics(...)
params = {0: {"filter": filter_0}, 2: {"filter": filter_2}}
stacked_physics.update_parameters(params=params)
```

2. `params` is a list of dicts: the user can update multiple physics; it updates physics with equivalent indices in `self.physics_list`. In the following snippet, only the physics of indices 0 and 1 are updated (if the physics of index 2 exists, it is not updated).
```python
stacked_physics = StackedPhysics(...)
params = [{"filter": filter_0}, {"filter": filter_1}]
stacked_physics.update_parameters(params=params)
```

The update logic is quite different from how it works with standard physics, so passing `**params`, coming out of `Generators`, to do `physics.update(**params)` won't work anymore (it was working before, but only in specific cases, no problem with `Pansharpen` for example). I'm thinking about it right now, but we could overload the update logic with a third use case that is closer to the standard one.

3. in addition to `**kwargs` the user needs to pass an additional `index` kwarg, to specify which physics should be updated, e.g.
```python
stacked_physics = StackedPhysics(...)
params = {"filter": filter_0} # same format than usal
stacked_physics.update_parameters(index=1, **params) # only the physics of index 1 is updated
```

Everything I've described here also applies to `ComposedPhysics`. To be honest, I am not super satisfied with the overall solution, but I could not think of something better. I think we could keep options 1 and 3 and drop the overload in option 2 to keep complexity as low as possible. Please tell me what you think, and if you have better solutions in mind :) ?  

### DIFF
--- a/deepinv/tests/test_physics.py
+++ b/deepinv/tests/test_physics.py
@@ -266,7 +266,7 @@ def find_operator(name, device, imsize=None, get_physics_param=False):
         )
         p = p1 * p2
         norm = 1 / 2**2
-        params = ["filter"]
+        params = {0: ["filter"]} # only the first physics has optimizable buffer
     elif name == "composition2":
         img_size = (3, 16, 16) if imsize is None else imsize
         p1 = dinv.physics.Downsampling(
@@ -278,7 +278,7 @@ def find_operator(name, device, imsize=None, get_physics_param=False):
             filter=dinv.physics.blur.gaussian_blur(sigma=(0.5)),
         )
         p = p2 * p1
-        params = ["filter"]
+        params = {1: ["filter"]} # only the second physics has optimizable buffer
     elif name == "denoising":
         p = dinv.physics.Denoising(dinv.physics.GaussianNoise(0.1, rng=rng))
         params = []
@@ -1710,8 +1710,8 @@ def test_operators_differentiability(name, device):
     if name == "radio":
         dtype = torch.cfloat
 
-    if "composition" in name:
-        pytest.skip("Skip composition operators for differentiability test.")
+    # if "composition" in name:
+    #     pytest.skip("Skip composition operators for differentiability test.")
 
     # Only test for floating point tensor
     valid_dtype = [torch.float16, torch.float32, torch.float64]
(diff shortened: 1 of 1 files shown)


jscanvic:
Thanks Romain!

Some thoughts:
- I agree the current behavior is a major design flaw
- I think it's a sensible approach to pass in a mapping from subphysics to associated params
- I tend to think it's better if the mapping is of the same type as the one used to store physics in the first place
  -  ... so for StackedPhysics I'd favor passing in a list over a dict
  - ... and for ComposedPhysics a length 2 tuple rather than a list or a dict
- Partial updates scare me a little, it might be error-prone and tedious to maintain
  -  ... so I'd avoid allowing mappings with missing keys and I'd prefer having empty dicts as explicit placeholders
- Option 3 doesn't seem very PEP20 to me, I'd do something like `physics.nth(1).update_parameters(**params)` instead (or any variant along the lines of `physics.physics_list[1]`)

To me the biggest question to answer is, can we do something closer to `Module.load_state_dict`?

One could imagine doing something like:
```
physics.update_parameters({
   "0.mask": mask1,
   "1.mask": mask2
})
```

I'm not 100% confident this is the way to go but I think we should consider it

jscanvic:
I'm also not sure we can fix that for StackedPhysics / ComposedPhysics in isolation - the solution we come up with should probably compatible with the default implementation of `update_parameters` otherwise it's going to be a nightmare to maintain e.g. the trainer

Issue #1025, opened 2026-01-24
Replace assert-based user input validation with explicit exceptions (ValueError/TypeError/...)


nucli-vicky:
## Motivation
The codebase still uses `assert` statements in quite a few places to validate user-provided inputs. This is fragile in library code because assert statements are intended for debugging and are removed when Python is run with optimizations (`python -O`), which can silently disable these checks.

More importantly: even without `-O`, using `AssertionError` for invalid user input is non-idiomatic and makes downstream error handling less predictable (users expect `ValueError` / `TypeError` for bad inputs).

## Example
In `deepinv/transform/temporal.py`;

*Current:*
```python
    def __init__(self, *args, padding="reflect", **kwargs):
        super().__init__(*args, **kwargs)
        self.flatten_video_input = False
        self.padding = padding
        assert self.padding in ("reflect", "wrap"), "padding must be reflect or wrap."
```

*Proposed change:*
```python
    def __init__(self, *args, padding="reflect", **kwargs):
        super().__init__(*args, **kwargs)
        self.flatten_video_input = False
        if padding not in ("reflect", "wrap"):
             raise ValueError(f"padding must be one of ('reflect', 'wrap'), got {padding}")
        self.padding = padding
```

## Proposed Plan
All `assert` statements validating user inputs should be replaced with explicit exceptions. A list can be compiled quickly, by going into the deepinv src folder and using `grep -R --exclude-dir=tests "assert "`. From there, each instance of `assert` use is to be checked and converted to a `raise XError` when it is used to validate user-input

After making this change, we should note using explicit errors over asserts when validating user inputs in the contributing guidelines (something short like, "Do not use `assert` for user input validation.", and inappropriate use of `assert` in new PRs should be pointed out during review.

nucli-vicky:
All affected files + number of occurrences of `assert `; 
```
     25 ./models/restormer.py
     15 ./models/swinir.py
     12 ./physics/functional/astra.py
     11 ./physics/functional/convolution.py
      8 ./models/diffunet.py
      7 ./models/drunet.py
      5 ./physics/noise.py
      5 ./physics/generator/base.py
      5 ./models/ram.py
      4 ./training/trainer.py
      4 ./physics/structured_random.py
      4 ./physics/singlepixel.py
      4 ./physics/functional/interp.py
      3 ./physics/blur.py
      3 ./optim/phase_retrieval.py
      3 ./models/wrapper.py
      3 ./models/ncsnpp.py
[...]

PR #1132, opened 2026-04-04
Assert -> error raises part 1

nucli-vicky:
Towards #1025

**Note:** `DPSDataFidelity`'s docstring said 
```
:param float weight: Weighting factor for the data fidelity term. Default to 100.
```
as it defaults to 1 in practice, I also changed this docstring


nucli-vicky:
Remaining to-do after this PR:

```
     15 ./models/swinir.py
     12 ./physics/functional/astra.py
      8 ./models/diffunet.py
      7 ./models/drunet.py
      5 ./physics/noise.py
      5 ./physics/generator/base.py
      5 ./physics/blur.py
      5 ./models/ram.py
      4 ./training/trainer.py
      4 ./physics/structured_random.py
      4 ./physics/functional/interp.py
      4 ./models/ncsnpp.py
      2 ./utils/logger.py
      2 ./physics/phase_retrieval.py
      2 ./physics/functional/convolution.py
      2 ./optim/utils.py
      2 ./optim/epll.py
      2 ./models/wavdict.py
      2 ./models/scunet.py
      2 ./loss/mri/measplit.py
      2 ./loss/measplit.py
      1 ./utils/signals.py
      1 ./transform/shift.py
      1 ./transform/base.py
      1 ./physics/tomography.py
      1 ./physics/range.py
      1 ./physics/compressed_sensing.py
      1 ./models/precond.py
      1 ./distributed/strategies.py
      1 ./datasets/satellite.py
      1 ./datasets/cmrxrecon.py
```

Tmodrzyk:
Hi @nucli-vicky thank you for this PR ! Don't hesitate to ping me and the other reviewers when you need a review :smile:

nucli-vicky:
Thanks @Tmodrzyk ! I'd like a review for this now, as I don't have the time to finish the remaining currently and don't want to let this PR go stale.

Tmodrzyk:
Hi @nucli-vicky thanks for this PR, I'll review it and finish up replacing the other asserts in another PR. 😃

deepinv/physics/remote_sensing.py
```diff
@@ -70,7 +70,8 @@ def __init__(
         eps=1e-6,
         **kwargs,
     ):
-        assert len(img_size) == 3, "img_size must be of shape (C,H,W)"
+        if len(img_size) != 3:  # pragma: no cover
+            raise ValueError("img_size must be of shape (C,H,W)")
```
Tmodrzyk, line 74:
```suggestion
        if len(img_size) != 3:  # pragma: no cover
            raise ValueError(f"img_size must be of shape (C,H,W) but got {img_size}")
```

deepinv/transform/base.py
```diff
@@ -217,7 +217,8 @@ def inverse(self, x: torch.Tensor, batchwise=True, **params) -> torch.Tensor:
         if batchwise:
             return self.transform(x, **inv_params)
 
-        assert len(x) % self.n_trans == 0, "batchwise must be True"
+        if len(x) % self.n_trans != 0:  # pragma: no cover
+            raise ValueError("batchwise must be True")
```
Tmodrzyk: I know this was the original error message but I think we should make it slightly clearer.

Tmodrzyk (commented):
Thanks for the hard work @nucli-vicky !
I love how you improved most of the error messages !
I have a few minor specific comments and a more general question: why do you ``# pragma: no cover`` on each of the tests ? 
Shouldn't we cover these lines in our tests with incorrect inputs etc to see if they are handled correctly ?

deepinv/transform/base.py
nucli-vicky:
```python
            raise ValueError(
                f"batchwise=False requires len(x) to be divisible by n_trans, but got len(x)={len(x)} and n_trans={self.n_trans}. Set batchwise=True or adjust the batch size."
            )
```

nucli-vicky:
Thanks @Tmodrzyk , made changes based on your comments!

I agree that, ideally, we should test everywhere with inputs that should trigger the Error raises.

I added the `pragma: no cover` because

(i) I believe `assert` doesn't get counted as an uncovered line if not triggered. This means not adding `pragma` to the new `if` statements would result in these changes to all be treated as not covered and could cause CI/CD not allowing the PR because of not enough coverage (I seem to remember something like this being a thing?)

(ii) Adding `pragma: no cover` is done extensively throughout the codebase.

(iii) Adding tests for every error raise would take a long time + a huge amount of code. I'm also not sure if this would add too much bloat to the tests? Maybe we should have some file which tests specifically error raising behavior on wrong inputs, such that the core "logic testing" test files don't get too complexified?

Tmodrzyk:
Thanks for your detailed answer @nucli-vicky ! I think all your points are fair, thank you again for the hard work on this PR. 😃

Tmodrzyk (approved):
LGTM

Tmodrzyk:
I don't think the GPU tests failing are related to this PR, I'll wait for #1158 to be merged and run the tests again.

Tmodrzyk:
The tests are passing on my local GPU-enabled laptop, so I'm just checking again that it passes on the CI before merging.

nucli-vicky:
@Tmodrzyk the current failure I would think is some bad luck with randomness or at least not related to the changes here.

On the other hand, since we set `torch.manual_seed(10)`, the test should pass each run.

I think this may be related to the multithreading done within torch when doing ops on cpu, which can influence the order of some sums/products etc, and floating point rounding being influenced by this could lead to such a test failure (especially as we have `max_iter=100` with the sampler, this rounding non-determinism can add up).

Tmodrzyk:
Yes indeed, we can probably safely merge. 
@Andrewwango what do you think ?

nucli-vicky:
@Tmodrzyk we can rerun this specific test, to ensure it is purely due to randomness

Andrewwango (approved):
Happy to merge

Tmodrzyk (approved):
LGTM

Issue #1005, opened 2025-12-19
`dtype` and `device` attributes in nn.Module inherited classes


romainvo:
### Environment

All

### Describe the bug

Currently, many objects in the library inheriting from `nn.Module` have a `self.dtype` and `self.device` attributes that is not correctly updated when calling the `self.to(device)` method. This can create inconsistencies and bug like the one reported in  #968 .

The PR in #989 should fix the `self.device` bug in `LinearPhysics` instances and enable smooth depreciation.

However, some physics still rely on `self.dtype`:
- [ ] CompressedSensing
- [ ] Tomography (and filter computation)
- [ ] PhaseRetrieval

this could be resolved by the same fix applied in #989 .

 ---

Some models also rely on `self.dtype` and `self.device` when in practice it should not be required:
- [ ] `diffunet.py`
- [ ] `diffusion_sde.py`
- [ ] `guided_diffusion.py`
- [ ] `icnn.py`
- [ ] `multispectral.py`
- [ ] `ncsnpp.py`
- [ ] `wavdict.py` (--> `WaveletPrior` in  `prior.py`)

---

More tricky, `PhysicsGenerator` (which also inherits `nn.Module`) also rely on `self.device`. We expect it to generate "params" directly on a targeted device for efficiency. The question here is to decide if it's ok to keep a `self.device` attribute for the generators (even though they're nn.Module) or ask the users (and internal logic) to explicitly pass a chosen device. 

Another solution is to not make `PhysicsGenerator` a `nn.Module`, in practice they hold no buffers/ learnable parameters, except their random state. And if we want a `state_dict` method, we could implement one, like it's done for torch optimizers and lr schedulers.

### Steps to reproduce

None.

### Expected behavior

None.

PR #1140, opened 2026-04-10
Priors from the NMAPG PR

samuro95:
This is an extraction of the priors (with some  re-factoring) of the original #829 from @johertrich .

In particular it contains two new priors (Ridge Regularizer and Least Square Denoiser)
- the Ridge Regularizer is now both a Model and a Prior
- The LS Denoiser Prior is more general by default (with any denoiser) and then if 'download = pretrained' it uses the proposed DRUNET pretrained weights with color or grayscale + pretrained input and output scaling parameters.


samuro95:
TODO: add test for nonconvex prior convergence

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from .utils import get_weights_url
+
+
+class RidgeRegularizer(nn.Module):
+    r"""
+    (Weakly) Convex Ridge Regularizer model :math:`\reg{x}=\sum_{c} \psi_c(W_c x)`.
```
Tmodrzyk, line 13: Is the formula matching their paper ? It seems slightly different

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+
+    for filters :math:`W_c` and potentials :math:`\psi_c`. The filters :math:`W_c` are realized by a concatenation multiple convolution
+    layers without nonlinearity. The potentials :math:`\psi_c` are given by scaled versions smoothed absolute values,
```
Tmodrzyk, line 16:
```suggestion
    layers without nonlinearity. The potentials :math:`\psi_c` are given by scaled and smoothed versions of the absolute value,
```

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+    see :footcite:t:`hertrich2025learning` for a precise description.
+
+    To allow the automatic tuning of the regularization parameter, we parameterize the regularizer with two additional scalings, i.e.,
+    we implement :math:`\frac{\alpha}{\sigma^2}\reg{\sigma x}` instead of :math:`\reg{x}` where :math:`\alpha` and :math:`\sigma` are learnable parameters of the regularizer.
+    If the weak CRR is used, :math:`\alpha` is fixed per default, since it changes the weak convexity constant.
```
Tmodrzyk, line 21: Introduce the acronym "CRR" here or even before

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+
+    The (W)CRR was introduced by :footcite:t:`goujon2023neural` and :footcite:t:`goujon2024learning`.
+    The specific implementation is taken from :footcite:t:`hertrich2025learning`.
+
+    This model can be used as a prior through :class:`deepinv.optim.RidgeRegularizer`.
+
+    :param int in_channels: Number of input channels (`1` for gray valued images, `3` for color images). Default: `3`
+    :param float weak_convexity: Weak convexity of the regularizer. Set to `0.0` for a convex regularizer and to `1.0` for a 1-weakly convex regularizer.
+        Default: `0.0`
+    :param list of int nb_channels: List of ints taking the hidden number of channels in the multiconvolution. Default: `[4, 8, 64]`
```
Tmodrzyk, line 31:
```suggestion
    :param list[int] nb_channels: List of ints taking the hidden number of channels in the multiconvolution. Default: `[4, 8, 64]`
```

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+    :param list of int filter_sizes: List of ints taking the kernel sizes of the convolution. Default: `[5,5,5]`
```
Tmodrzyk, line 32:
```suggestion
    :param list[int] filter_sizes: List of ints taking the kernel sizes of the convolution. Default: `[5,5,5]`
```

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+    :param str device: Device for the weights. Default: `"cpu"`
```
Tmodrzyk, line 33: This should not be an attribute, instead we can make the internal states registered buffers.

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+    :param str, None pretrained: use pretrained weights. If ``pretrained=None``, the weights will be initialized at random
+        using Pytorch's default initialization. If ``pretrained='download'``, the weights will be downloaded from an
+        online repository (only available for the default architecture with 3 or 1 input/output channels).
```
Tmodrzyk, line 36: Can we put the link to these weights ?

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+    :param bool warn_output_scaling: warn if `weak_convexity>0` and the output scaling (:math:`\log(\alpha)` in the above description) is not zero. This case
+        destroys the weak convexity constant defined by teh `weak_convexity` argument. Default: `True`
+    """
+
+    def __init__(
+        self,
+        in_channels: int = 3,
+        weak_convexity: float = 0.0,
+        nb_channels: tuple[int, ...] = (4, 8, 64),
+        filter_sizes: tuple[int, ...] = (5, 5, 5),
```
Tmodrzyk, line 48: Both of these types do not match the docstring

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+        x = x / torch.sqrt(self.__get_conv_lip())
+        return self.filters(x)
+
+    def __conv_transpose(self, x):
+        x = x / torch.sqrt(self.__get_conv_lip())
+        for filt in reversed(self.filters):
+            x = F.conv_transpose2d(x, filt.weight, padding=filt.padding)
+        return x
+
+    def grad(self, x, *args, get_energy=False, **kwargs):
```
Tmodrzyk, line 147:
```suggestion
    def grad(self, x, *args, get_energy=False, **kwargs) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
```

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+            self.__grad_smooth_l1(torch.exp(self.beta) * grad)
+            - self.__grad_smooth_l1(grad) * self.weak_cvx
+        )
+        grad = grad * torch.exp(self.output_scaling - self.scaling - self.input_scaling)
+        grad = self.__conv_transpose(grad)
+        if get_energy:
+            return reg, grad
+        return grad
+
+    def forward(self, x, *args, **kwargs):
```
Tmodrzyk, line 176:
```suggestion
    def forward(self, x, *args, **kwargs) -> torch.Tensor:
```

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+        grad = grad * torch.exp(self.scaling + self.input_scaling)
+        if get_energy:
+            reg = (
+                self.__smooth_l1(torch.exp(self.beta) * grad) * torch.exp(-self.beta)
+                - self.__smooth_l1(grad) * self.weak_cvx
+            )
+            reg = reg * torch.exp(
+                self.output_scaling - 2 * self.scaling - 2 * self.input_scaling
+            )
+            reg = reg.sum(dim=(1, 2, 3))
```
Tmodrzyk, line 165: This seems duplicated from the `.forward()` so we should call it instead.

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+        )
+        reg = reg * torch.exp(
+            self.output_scaling - 2 * self.scaling - 2 * self.input_scaling
+        )
+        reg = reg.sum(dim=(1, 2, 3))
+        return reg
+
+    def _apply(self, fn):
+        self.dirac = fn(self.dirac)
+        return super()._apply(fn)
```
Tmodrzyk, line 208: We should register diract as a buffer instead

deepinv/optim/prior.py
```diff
@@ -963,3 +966,214 @@ def prox(
...
+    The (W)CRR was introduced by :footcite:t:`goujon2023neural` and :footcite:t:`goujon2024learning`.
+    The specific implementation is taken from :footcite:t:`hertrich2025learning`.
+
+    This prior wraps the corresponding model :class:`deepinv.models.RidgeRegularizer`
+    and uses it to evaluate the regularizer, its gradient or proximity operator.
+
+    :param int in_channels: Number of input channels (`1` for gray valued images, `3` for color images). Default: `3`
+    :param float weak_convexity: Weak convexity of the regularizer. Set to `0.0` for a convex regularizer and to `1.0` for a 1-weakly convex regularizer.
+        Default: `0.0`
+    :param list of int nb_channels: List of ints taking the hidden number of channels in the multiconvolution. Default: `[4, 8, 64]`
```
Tmodrzyk, line 993:
```suggestion
    :param list[int] nb_channels: List of ints taking the hidden number of channels in the multiconvolution. Default: `[4, 8, 64]`
```

deepinv/optim/prior.py
```diff
@@ -963,3 +966,214 @@ def prox(
...
+    :param list of int filter_sizes: List of ints taking the kernel sizes of the convolution. Default: `[5,5,5]`
```
Tmodrzyk, line 994:
```suggestion
    :param list[int] filter_sizes: List of ints taking the kernel sizes of the convolution. Default: `[5,5,5]`
```

deepinv/optim/prior.py
```diff
@@ -963,3 +966,214 @@ def prox(
...
+        destroys the weak convexity constant defined by teh `weak_convexity` argument. Default: `True`
+    """
+
+    def __init__(
+        self,
+        in_channels: int = 3,
+        weak_convexity: float = 0.0,
+        nb_channels: tuple[int, ...] = (4, 8, 64),
+        filter_sizes: tuple[int, ...] = (5, 5, 5),
+        device: str = "cpu",
```
Tmodrzyk, line 1011: Same comment as for the model, this should not be an attribute anymore: the preferred syntax is `.to(device)`

deepinv/optim/prior.py
```diff
@@ -963,3 +966,214 @@ def prox(
...
+        :param float sigma: Noise level.
+        :param bool get_energy: Optional flag. If set to True, the function additionally returns the objective value at :math:`x`. Dafault: False.
+        :return: (:class:`torch.Tensor`) gradient at :math:`x`.
+        """
+        grad = torch.exp(self.output_scaling) * self.model.potential_grad(
+            torch.exp(self.input_scaling) * x.contiguous(), sigma
+        )
+        if get_energy:
+            reg = self(x, sigma, *args, **kwargs)
+            return reg, grad
```
Tmodrzyk, line 1163: With this we are breaking the common API of `Potential` which says that `.grad` returns a tensor. 
Is this really needed ?

deepinv/optim/prior.py
```diff
@@ -963,3 +966,214 @@ def prox(
...
+        Calculates the gradient of the regularizer at :math:`x`.
+
+        :param torch.Tensor x: Variable :math:`x` at which the gradient is computed.
+        :param float sigma: Noise level.
+        :param bool get_energy: Optional flag. If set to True, the function additionally returns the objective value at :math:`x`. Dafault: False.
+        :return: (:class:`torch.Tensor`) gradient at :math:`x`.
+        """
+        grad = torch.exp(self.output_scaling) * self.model.potential_grad(
+            torch.exp(self.input_scaling) * x.contiguous(), sigma
+        )
```
Tmodrzyk, line 1160: Shouldn't this be:
```python
torch.exp(self.output_scaling + self.input_scaling) * self.model.potential_grad(torch.exp(self.input_scaling) * x.contiguous(), sigma)
``` 
?

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+            ]
+        )
+
+        class ZeroMean(nn.Module):
+            """Enforces zero mean on the filters"""
+
+            def forward(self, x):
+                return x - torch.mean(x, dim=(1, 2, 3), keepdim=True)
+
+        torch.nn.utils.parametrize.register_parametrization(
```
jscanvic, line 78: Neat! I never came across it before

deepinv/optim/prior.py
```diff
@@ -51,34 +56,36 @@ class ZeroPrior(Prior):
 
     def __init__(self):
         super().__init__()
+
+        def forward(x, *args, **kwargs):
+            return torch.tensor(0.0)
+
+        self._g = forward
```
Tmodrzyk, line 63: I don't understand this part, we already define `.fn()` so I believe this code will never be executed. Also, it does not do the same thing as `.fn()` which returns a batch of zeroes: this returns a scalar zero.

examples/optimization/demo_learned_regularization.py
```diff
@@ -0,0 +1,282 @@
...
+====================================================================================================
+
+In this example, we show how to solve inverse problems using a learned regularizer.
+We consider denoising, computed tomography and inpainting with the convex ridge regularizer (CRR) (:footcite:t:`goujon2023neural`),
+weakly convex ridge regularizer (WCRR) (:footcite:t:`goujon2024learning`) and least squares regularizer (LSR) (see, e.g., :footcite:t:`hurault2021gradient` or :footcite:t:`zou2023deep`).
+This example only covers the reconstruction with these regularizers. For the training them in a bilevel regime we refer to :footcite:t:`hertrich2025learning`.
+
+For the reconstructions we solve the variational problem
+
+    .. math::
```
jscanvic, line 12: It renders weird I think we need to remove one level of indentation here

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+                ckpt = torch.hub.load_state_dict_from_url(
+                    url,
+                    map_location=lambda storage, loc: storage,
+                    file_name=file_name,
+                )
+                self.load_state_dict(ckpt, strict=True)
+            else:
+                self.load_state_dict(torch.load(pretrained, map_location=device))
+
+    def __smooth_l1(self, x):
```
jscanvic, line 125: Can you change the double underscores to single underscores? Just for the sake of homogeneity with the rest of the library

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+        return torch.clip(x**2, 0.0, 1.0) / 2 + torch.clip(torch.abs(x), 1.0) - 1.0
+
+    def __grad_smooth_l1(self, x):
+        return torch.clip(x, -1.0, 1.0)
+
+    def __get_conv_lip(self):
+        impulse = self.filters(self.dirac)
+        for filt in reversed(self.filters):
+            impulse = F.conv_transpose2d(impulse, filt.weight, padding=filt.padding)
+        return torch.fft.rfft2(impulse, s=[256, 256]).abs().max()
```
jscanvic, line 135: Are we happy with hardcoding 256 here?

examples/optimization/demo_learned_regularization.py
```diff
@@ -0,0 +1,282 @@
+r"""
+Learned Regularization Functionals
+====================================================================================================
+
+In this example, we show how to solve inverse problems using a learned regularizer.
+We consider denoising, computed tomography and inpainting with the convex ridge regularizer (CRR) (:footcite:t:`goujon2023neural`),
+weakly convex ridge regularizer (WCRR) (:footcite:t:`goujon2024learning`) and least squares regularizer (LSR) (see, e.g., :footcite:t:`hurault2021gradient` or :footcite:t:`zou2023deep`).
+This example only covers the reconstruction with these regularizers. For the training them in a bilevel regime we refer to :footcite:t:`hertrich2025learning`.
```
jscanvic, line 8: I find the last 2 sentences a bit confusing

Can you merge them into a single clearer sentence?

Maybe something along the lines of: The regularizers are pre-trained here but they could also be optimized jointly with the reconstruction in a bilevel way, see eg ref

examples/optimization/demo_learned_regularization.py
```diff
@@ -0,0 +1,282 @@
...
+    use_input_output_scaling=True,
+    pretrained="download_gray",
+    device=device,
+)
+lsr_sigma = 0.03
+
+# set all parameters to be not trainable
+for reg in [crr, wcrr, lsr, crr_gray, wcrr_gray, lsr_gray]:
+    for p in reg.parameters():
+        p.requires_grad_(False)
```
jscanvic, line 86: Should this be done (optionally?) in RidgeRegularizer and LeastSquaresResidual directly?

deepinv/optim/prior.py
```diff
@@ -963,3 +966,214 @@ def prox(
...
+            weak_convexity=weak_convexity,
+            nb_channels=nb_channels,
+            filter_sizes=filter_sizes,
+            device=device,
+            pretrained=pretrained,
+            warn_output_scaling=warn_output_scaling,
+        )
+        self.explicit_prior = True
+
+    def __getattr__(self, name):
```
jscanvic, line 1027: What is this method used for?

It adds complexity so I'd prefer if we didn't ship it

deepinv/optim/prior.py
```diff
@@ -963,3 +966,214 @@ def prox(
...
+        try:
+            return super().__getattr__(name)
+        except AttributeError:
+            model = super().__getattr__("model")
+            return getattr(model, name)
+
+    def grad(self, x, *args, get_energy=False, **kwargs):
```
jscanvic, line 1034: Can you type hint the output here and in other methods (fn, prox)?

deepinv/optim/prior.py
```diff
@@ -336,7 +343,6 @@ class WaveletPrior(Prior):
     :param float p: :math:`p`-norm of the prior. Default is 1.
     :param str device: device on which the wavelet transform is computed. Default is "cpu".
     :param int wvdim: dimension of the wavelet transform, can be either 2 or 3. Default is 2.
-    :param bool is_complex: whether the input is complex-valued. Default is False.
```
jscanvic, line 339: Why removing it?

deepinv/optim/prior.py
```diff
@@ -51,34 +56,36 @@ class ZeroPrior(Prior):
...
         self.explicit_prior = True
 
     def fn(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
         r"""
-        Computes the zero prior :math:`\reg{x} = 0` at :math:`x`.
+        Computes the zero prior :math:`\reg(x) = 0` at :math:`x`.
 
-        :param torch.Tensor x: Variable :math:`x` at which the prior is computed.
-        :return: (:class:`torch.Tensor`) prior :math:`\reg{x}`.
+        It returns a tensor of zeros of the same shape as :math:`x`.
```
jscanvic, line 70: I'd leave in the :param: and :return: lines below the 2nd sentence and same for grad and prox

deepinv/optim/fixed_point.py
```diff
@@ -354,7 +354,7 @@ def forward(
                     if self.verbose:
                         print(
                             f"[Stopping] Reached maximum number of failed backtracking checks "
-                            f"({self.max_iter_backtracking})."
+                            f"({self.backtracking_config.max_iter})."
```
jscanvic, line 357: Good catch!

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+        reg = self.__conv(x)
+        reg = reg * torch.exp(self.scaling + self.input_scaling)
+        reg = (
+            self.__smooth_l1(torch.exp(self.beta) * reg) * torch.exp(-self.beta)
+            - self.__smooth_l1(reg) * self.weak_cvx
+        )
+        reg = reg * torch.exp(
+            self.output_scaling - 2 * self.scaling - 2 * self.input_scaling
+        )
+        reg = reg.sum(dim=(1, 2, 3))
```
jscanvic, line 203: Future-proofing for 1d and 3d signals

```suggestion
        reg = reg.flatten(1).sum(1)
```

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+                    filter_sizes[i],
+                    padding=filter_sizes[i] // 2,
+                    bias=False,
+                    device=device,
+                )
+                for i in range(len(filter_sizes))
+            ]
+        )
+
+        class ZeroMean(nn.Module):
```
jscanvic, line 72: Put in inside the class (at the end) but outside the the `__init__` and access it via RidgeRegularizer.ZeroMean for homogeneity with eg SplittingModel in SplittingLoss

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+
+        torch.nn.utils.parametrize.register_parametrization(
+            self.filters[0], "weight", ZeroMean()
+        )
+
+        self.dirac = torch.zeros(
+            (1, in_channels, 2 * self.filter_size - 1, 2 * self.filter_size - 1),
+            device=device,
+        )
+        self.dirac[0, 0, self.filter_size - 1, self.filter_size - 1] = 1.0
```
jscanvic, line 86: Just wondering is it on purpose that the Dirac is centered at the last tap instead of the first?

jscanvic (changes requested):
Thanks Samuel, it looks great so far! A few comments / changes

deepinv/optim/prior.py
```diff
@@ -963,3 +966,214 @@ def prox(
...
+        pretrained: str = "download",
+        device: str = "cpu",
+    ):
+        super(LeastSquaresResidual, self).__init__()
+        self.explicit_prior = True
+        self.use_input_output_scaling = use_input_output_scaling
+
+        weights = None
+
+        if pretrained is not None and self.use_input_output_scaling:
```
Tmodrzyk, line 1096: If we set `self.use_input_output_scaling` to `False` we won't read the pretrained paths from the users.
If this was intended this should be mentioned in the docstring.

deepinv/optim/prior.py
```diff
@@ -3,12 +3,17 @@
 import numpy as np
 import torch
 import torch.nn as nn
+
+from deepinv.models.utils import get_weights_url
```
Tmodrzyk, line 7:
```suggestion
```

examples/optimization/demo_learned_regularization.py
```diff
@@ -0,0 +1,282 @@
...
+psnr_lsr = psnr(recon_lsr, test_img).item()
+
+print(
+    f"Resulting PSNRs:\nNoisy image: {psnr_noisy:.2f}, CRR: {psnr_crr:.2f}, WCRR: {psnr_wcrr:.2f}, LSR: {psnr_lsr:.2f}"
+)
+
+plot(
+    [test_img, y, recon_crr, recon_wcrr, recon_lsr],
+    ["ground truth", "noisy", "CRR", "WCRR", "LSR"],
+)
```
Tmodrzyk, line 130: On the rendered doc the images are fairly small. 
Could we maybe crop them to better see the details ? 
We could also use the `subtitles` option to display PSNR to be consistent across the library.

examples/optimization/demo_learned_regularization.py
```diff
@@ -0,0 +1,282 @@
...
+psnr_wcrr = psnr_dyn(recon_wcrr, test_img_ct).item()
+psnr_lsr = psnr_dyn(recon_lsr, test_img_ct).item()
+
+print(
+    f"Resulting PSNRs:\nFBP: {psnr_fbp:.2f}, CRR: {psnr_crr:.2f}, WCRR: {psnr_wcrr:.2f}, LSR: {psnr_lsr:.2f}"
+)
+plot(
+    [test_img_ct, fbp, recon_crr, recon_wcrr, recon_lsr],
+    ["ground truth", "FBP", "CRR", "WCRR", "LSR"],
+)
```
Tmodrzyk, line 200: Is the WCRR reconstruction correct here ? It seems very degraded compared to the denoising example and to the other methods shown here.

examples/optimization/demo_learned_regularization.py
```diff
@@ -0,0 +1,282 @@
...
+psnr_lsr = psnr(recon_lsr, test_img).item()
+
+print(
+    f"Resulting PSNRs:\nMasked: {psnr_masked:.2f}, x_init: {psnr_init:.2f}, CRR: {psnr_crr:.2f}, WCRR: {psnr_wcrr:.2f}, LSR: {psnr_lsr:.2f}"
+)
+
+plot(
+    [test_img, y, x_init, recon_crr, recon_wcrr, recon_lsr],
+    ["ground truth", "masked", "init", "CRR", "WCRR", "LSR"],
+)
```
Tmodrzyk, line 277: Here also the displayed image are fairly small. We may also just increase `figsize`.

deepinv/models/ridge_regularizer.py
```diff
@@ -0,0 +1,208 @@
...
+    def __smooth_l1(self, x):
+        return torch.clip(x**2, 0.0, 1.0) / 2 + torch.clip(torch.abs(x), 1.0) - 1.0
+
+    def __grad_smooth_l1(self, x):
+        return torch.clip(x, -1.0, 1.0)
+
+    def __get_conv_lip(self):
+        impulse = self.filters(self.dirac)
+        for filt in reversed(self.filters):
+            impulse = F.conv_transpose2d(impulse, filt.weight, padding=filt.padding)
```
Tmodrzyk, line 134: Two comments here.
1. Since it does not depend on the input (from what I'm getting it's a global Lipschitz constant and not a local estimate), instead of computing it every time we need it in `__conv` and `__conv_transpose` we might just compute it once and store it as a registered buffer ? 
2. Is this constant still correct for RGB images ?

Tmodrzyk (changes requested):
Hi @samuro95 thank you for this huge PR ! This is a great addition to the library.

I think we should double-check some of the computations (the gradient and the Lipschitz constant one).
Otherwise most comments are related to the use of registered buffers so that the new priors combine well with `.to(device)` (see #968 and #1005 for more details) which should be a quick fix !

samuro95:
Thanks Thibaut and Jeremy. @johertrich if you have time could you have a look at the comment left by the reviewers ? I did not change the code, it is simply copy-pasted from your PR.

johertrich:
@samuro95 Thanks! I can have a look, but currently I cannot commit to this PR.

examples/optimization/demo_learned_regularization.py
tachella, line 8:
```suggestion
.. note::

    This example only covers the reconstruction with these regularizers. For training them using bilevel optimization, please refer to :footcite:t:`hertrich2025learning`.
```

examples/optimization/demo_learned_regularization.py
```diff
@@ -0,0 +1,282 @@
+r"""
+Learned Regularization Functionals
+====================================================================================================
+
+In this example, we show how to solve inverse problems using a learned regularizer.
+We consider denoising, computed tomography and inpainting with the convex ridge regularizer (CRR) (:footcite:t:`goujon2023neural`),
+weakly convex ridge regularizer (WCRR) (:footcite:t:`goujon2024learning`) and least squares regularizer (LSR) (see, e.g., :footcite:t:`hurault2021gradient` or :footcite:t:`zou2023deep`).
+This example only covers the reconstruction with these regularizers. For the training them in a bilevel regime we refer to :footcite:t:`hertrich2025learning`.
+
+For the reconstructions we solve the variational problem
```
tachella, line 10:
```suggestion
To perform reconstruction, we solve the variational problem
```

examples/optimization/demo_learned_regularization.py
```diff
@@ -0,0 +1,282 @@
...
+# --------------------------------------------------------------------
+# In this example we use the CRR, WCRR and LSR. First, we load the models with the pretrained weights (trained on BSDS500) in deepinv.
+# Moreover we load the data fidelity terms and the three different physics operators (Denoising, CT and Inpainting).
+#
+
+# color versions
+# CRR and WCRR are the RidgeRegularizer with weak convexity 0.0 and 1.0
+crr = RidgeRegularizer(weak_convexity=0.0, device=device)
+wcrr = RidgeRegularizer(weak_convexity=1.0, device=device)
+lsr = LeastSquaresResidual(
```
tachella, line 64: I'm not sure this name is super clear, is this a prior? I'm afraid that this gets confused with L2 errors

examples/optimization/demo_learned_regularization.py
tachella, line 12:
```suggestion
.. math::
```

examples/optimization/demo_learned_regularization.py
```diff
@@ -0,0 +1,282 @@
...
+psnr_wcrr = psnr(recon_wcrr, test_img).item()
+psnr_lsr = psnr(recon_lsr, test_img).item()
+
+print(
+    f"Resulting PSNRs:\nNoisy image: {psnr_noisy:.2f}, CRR: {psnr_crr:.2f}, WCRR: {psnr_wcrr:.2f}, LSR: {psnr_lsr:.2f}"
+)
+
+plot(
+    [test_img, y, recon_crr, recon_wcrr, recon_lsr],
+    ["ground truth", "noisy", "CRR", "WCRR", "LSR"],
```
tachella, line 129: Increase the plot size
```suggestion
    ["ground truth", "noisy", "CRR", "WCRR", "LSR"], figsize=(10,10)
```

examples/optimization/demo_learned_regularization.py
```diff
@@ -0,0 +1,282 @@
...
+    recon_wcrr = wcrr.prox(y, gamma=0.5, stepsize_inter=0.1)
+    recon_lsr = lsr.prox(y, lsr_sigma, gamma=0.5, stepsize_inter=0.05)
+
+psnr_noisy = psnr(y, test_img).item()
+psnr_crr = psnr(recon_crr, test_img).item()
+psnr_wcrr = psnr(recon_wcrr, test_img).item()
+psnr_lsr = psnr(recon_lsr, test_img).item()
+
+print(
+    f"Resulting PSNRs:\nNoisy image: {psnr_noisy:.2f}, CRR: {psnr_crr:.2f}, WCRR: {psnr_wcrr:.2f}, LSR: {psnr_lsr:.2f}"
```
tachella, line 124: The results aren't amazing, is this expected?

Image

examples/optimization/demo_learned_regularization.py
```diff
@@ -0,0 +1,282 @@
...
+
+    .. math::
+        \begin{equation}
+        \label{eq:min_prob}
+        \tag{1}
+        \underset{x}{\arg\min} \quad  \datafid{Ax}{y} + \lambda \reg{x},
+        \end{equation}
+
+for a data-fidelity term :math:`\datafid{Ax}{y}`, forward operator :math:`A` and a learned regularizer :math:`\reg{x}`.
+Depending on whether the terms are smooth, we solve the problem with gradient descent (GD) or proximal gradient descent (PGD).
```
tachella, line 20: It would be nice here to include a summary of the definition of the different regularizer functions presented in the example

tachella (changes requested):
Many thanks for preparing this PR! 
A few comments/suggestions on the example :)

johertrich:
I had a look at the PR, in particular to the demo. While the architectures are working well, the gradient descent called in `regularizer.prox` and in the demo does not find at all a minimizer of the variational problem, not even for the convex case with the CRR. This is basically the reason why we had added the nmAPG in the first place and this is also the reason for the bad results in the demo observed by @tachella and @Tmodrzyk . (The difference to the nmAPG minimization is massive, more than 5dB in the PSNR...)

To go forward with this PR, I think we have to use a different optimization routine (both in the `prox` method as well as in the demo), here we would have a couple of options:
1. use the nmAPG (then we are more or less back at #829 )
2. try to use the methods from `deepinv.optim`, something like gradient descent with momentum and backtracking should also be doable with this. However, this still leaves a dependence on plenty of hyperparameters (particularly the step size). Thus this would be lots of work to get this to a satisfactory state.
3. we experimented a bit to use some version of LBFGS for that and this worked quite well. We could try whether this also works with the PyTorch standard LBFGS implementation. In this case we could only consider smooth data terms, but I think this would be good enough for the beginning.

If you don't like the nmAPG, I think the third option might be the easiest one. In any case, I don't think that it makes sense to add the architectures if there is no easy-to-use tool for optimizing the variational problem and if the prox is not working well.

PR #1139, opened 2026-04-09
various small fixes + 3D extension of deblurring/downsampling

tachella:
### Mini-fixes
- download pretrained weights for `deepinv.models.GSDRUNet` by default, just like all other denoisers 
- remove mentions to `pyiqa` and fix range check 
- set `deepinv.optim.L2` data fidelity to `deepinv.sampling.DiffDPIR` by default
- raise error when using `deepinv.sampling.DPS` with > 1000 iterations
- Fixes inpainting bug with non-binary masks #1167 
- removes unused variables in `minres` solver. Adds missing mentions to `minres` in docstrings
 
###  Extends blur/downsampling operators
- extends definition of `sinc`, `gaussian_blur`, `bilinear`, `bicubic` filters to n-dimensions.

### Checks to be done before submitting your PR

- [ ] `python3 -m pytest deepinv/tests` runs successfully.
- [ ] `black .` and `ruff check .` run successfully.
- [ ] `make html` runs successfully (in the `docs/` directory).
- [ ] Updated docstrings related to the changes (as applicable).
- [ ] Added an entry to the [changelog.rst](https://github.com/deepinv/deepinv/blob/main/docs/source/changelog.rst).

PR #1152, opened 2026-04-14
Add extensions of gaussian_blur to 1 and 3 dimensions and add GaussianBlurGenerator

romainvo:
The PR
- generalizes the existing `deepinv.physics.blur.gaussian_blur` method to the 1D and 3d cases
- provides the possibility to generate batches of Gaussian blur kernels, with a different kernel for each batch element
 
It also adds a `GaussianBlurGenerator` leveraging the `gaussian_blur` functional.

I haven't done it yet, but if we want to remove the duplicate code, the easy option is to make `deepinv.physics.blur.gaussian_blur` point to the new functional
- they share the same base signature, passing the same set of params to both functions should generate the same tensor, modulo some approximation error due to interpolation in the original `gaussian_blur` method

--

Also pushing a quick fix to #1093 --> in GeneratorMixture, we removed the synchronization on the sampled p value in self.step (so p on self.device), but self.cum_probs was always on cpu.


romainvo:
@tachella the piece of code I told you about, does it overlap with what you started?

tachella (commented):
Many thanks for this PR - the generator is quite cool

I was wondering if `gaussian_blur_nd` could just be an extension of `gaussian_blur` where we pass an `n_dim` or similar input, which by default is set to 2 (so we keep backwards compatibility)?

deepinv/physics/functional/blur.py
```diff
@@ -0,0 +1,187 @@
+from __future__ import annotations
+import torch
+from math import sqrt, pi
+
+
+def gaussian_blur_nd(
```
mh-nguyen712: similar to the comment of Julian, I would keep the name `gaussian_blur` and add a `n_dim` argument

deepinv/physics/functional/blur.py
```diff
@@ -0,0 +1,187 @@
...
+    psf_size: tuple[int, ...] | None = None,
+    sigma: int | float | tuple[float, ...] | torch.Tensor = (1.0, 1.0),
+    angle: int | float | tuple[float, ...] | torch.Tensor = 0.0,
+    batch_size: None | int = 1,
```
mh-nguyen712: I would remove the batch_size argument and infer directly from sigma / angle. If sigma, angle are float, only return with batch_size = 1

deepinv/physics/functional/blur.py
```diff
@@ -0,0 +1,187 @@
...
+        sigma = torch.tensor(
+            [sigma] * batch_size, device=device, dtype=dtype
+        )  # Shape: (batch_size, dim)
+
+    B = sigma.shape[0]  # batch size inferred from sigma shape
+
+    # Rotation angles
+    # ---------------
+    # For 3D, angles is a list of three angles (alpha, beta, gamma)
+    if dim == 2:
```
mh-nguyen712, line 87: a small helper function for handling int / float / tensor for angle / sigma with a `dim` parameter could simplify a bit the implementation

deepinv/physics/functional/blur.py
```diff
@@ -0,0 +1,187 @@
...
+        # Rotation matrix for 2D
+        cos_theta = torch.cos(angle).view(-1)
+        sin_theta = torch.sin(angle).view(-1)
+        rot_mat = torch.zeros((B, 2, 2), device=device)
+        rot_mat[:, 0, 0] = cos_theta.squeeze()
+        rot_mat[:, 0, 1] = -sin_theta.squeeze()
+        rot_mat[:, 1, 0] = sin_theta.squeeze()
+        rot_mat[:, 1, 1] = cos_theta.squeeze()
+
```
mh-nguyen712: we can avoid `squeeze()` and preallocation
```suggestion
        rot_mat = torch.stack([cos_theta, -sin_theta, 
                               sin_theta,  cos_theta], 
                               dim=1).view(B, 2, 2)
```

deepinv/physics/functional/blur.py
```diff
@@ -0,0 +1,187 @@
...
+        # Rotation around z-axis
+        Rz = torch.zeros((B, 3, 3), device=device)
+        Rz[:, 0, 0] = torch.cos(gamma)
+        Rz[:, 0, 1] = -torch.sin(gamma)
+        Rz[:, 1, 0] = torch.sin(gamma)
+        Rz[:, 1, 1] = torch.cos(gamma)
+        Rz[:, 2, 2] = 1.0
+
+        # Combined rotation matrix: R = Rz @ Ry @ Rx
+        R = torch.bmm(Rz, torch.bmm(Ry, Rx))
```
mh-nguyen712: I think it's could be faster if we build the matrix using the closed-form (https://en.wikipedia.org/wiki/Rotation_matrix#General_3D_rotations) (3x3 matrix with precomputed sin / cos), instead of calling two bmm

deepinv/physics/functional/blur.py
```diff
@@ -0,0 +1,187 @@
...
+        # Apply rotation: (batch_size, *psf_size, 3)
+        coords = torch.einsum("bij,b...j->b...i", R, coords)
+
+    # Compute the N-dimensional Gaussian
+    kernel = torch.ones((B, *psf_size), device=device, dtype=dtype)
+    for d in range(dim):
+        kernel *= torch.exp(
+            -0.5 * (coords[..., d] ** 2) / (sigma[:, d].view(-1, *[1] * dim) ** 2)
```
mh-nguyen712: add the extra dim in sigma directly into the previous step of handling sigma / angle

deepinv/physics/generator/blur.py
```diff
@@ -37,6 +38,189 @@ def __init__(
         self.num_channels = num_channels
 
 
+class GaussianBlurGenerator(PSFGenerator):
```
mh-nguyen712, line 42: Very nice generator!

deepinv/physics/generator/blur.py
```diff
@@ -37,6 +38,189 @@ def __init__(
...
+        angle_max: float | tuple[float, ...] = 360.0,
+        num_channels: int = 1,
+        rng: torch.Generator = None,
+        device: str = "cpu",
+        dtype: type = torch.float32,
+    ):
+        r"""
+        Random Gaussian blur generator. Generates 1D, 2D, or 3D Gaussian kernels with random standard deviations and rotation angles.
+
+        :param tuple[int, ...] psf_size: the shape of the generated point spread function (PSF).
```
mh-nguyen712: maybe precise that it will define the dimension (1, 2 or 3)

deepinv/physics/generator/blur.py
```diff
@@ -37,6 +38,189 @@ def __init__(
...
+        :param torch.Generator rng: PyTorch random number generator for reproducibility. If ``None``, a torch.Generator will be created on the specified device.
+        :param str device: the device to create the tensors on. Defaults to "cpu".
+        :param type dtype: the data type of the generated tensors. Defaults to torch.float32.
+        """
+
+        dim = len(psf_size)
+        if dim not in {1, 2, 3}:
+            raise ValueError("Only 1D, 2D, and 3D kernels are supported.")
+
+        if isinstance(sigma_min, (int, float)):
```
mh-nguyen712: it seems to me that the check for sigma_min / sigma_max and angle_min / angle_max is similar. Can we refactor a bit to make it more readable?

deepinv/physics/generator/blur.py
```diff
@@ -37,6 +38,189 @@ def __init__(
...
+        angle: torch.Tensor = None,
+        seed: int = None,
+        **kwargs,
+    ):
+        self.rng_manual_seed(seed)
+        dim = len(self.psf_size)
+
+        if sigma is None:
+            if self.isotropic:
+                sigma = torch.stack(
```
mh-nguyen712: A `_generate_param()` that can generate either sigma or angle can be useful here to avoid repetition.

mh-nguyen712 (changes requested):
Thanks for this new feature! Some comments.

romainvo:
Thanks @tachella @mh-nguyen712  for the comments! I cleaned the functionals from deepinv/physics/blur.py and moved them into deepinv/physics/functional/blur.py for clarity. I've added a level-module __getattr__ in `deepinv/physics/blur.py` for backward compatibility, which fetches functionals in `functional/blur.py` with a deprecation warning (seemed like the pythonic way https://peps.python.org/pep-0562/ but I don't mind doing something else)

Normally, everything should be backward compatible.

I've tried to clean/factorize the code wrt your comments @mh-nguyen712 :) !

romainvo:
I added a non-regression test on the new implementation, to check that it matches the old implementation (modulo the interpolation errors due to rotation in the previous implementation).

The `test_least_squares_implicit_backward` in `test_optim.py` seems to fail with the `"deblur_valid"` config when the blur kernel is a "non-pure" dirac, e.g. with `filter = gaussian_blur(sigma=(0.25, 0.1), angle=45.0)` which gives something like that:
```python
tensor([[[[1.1254e-07, 2.5437e-13, 3.7835e-44],
          [2.5437e-13, 1.0000e+00, 2.5437e-13],
          [3.7835e-44, 2.5437e-13, 1.1254e-07]]]])
```

The old implementation did not trigger the error because the error interpolation was flattening the kernel. I switch to `angle=0.0`, it's sufficient for our test purposes. Is this expected behaviour of the implicit solver ? @mh-nguyen712 

I cleaned all the files using the old call to `physics.blur.gaussian_blur` and deprecated it.

deepinv/physics/functional/blur.py
```diff
@@ -0,0 +1,400 @@
...
+            psf_size[d],
+            device=device,
+            dtype=dtype,
+        )
+        grids.append(ax)
+
+    # Create a meshgrid for the coordinates using 'ij' indexing so that
+    # the i-th grid corresponds to the i-th dimension in psf_size.
+    # mesh[-1] will correspond to the x-coordinates, mesh[-2] to the y-coordinates, and mesh[-3] to the z-coordinates (if 3D).
+    mesh = torch.meshgrid(*[grids[d] for d in range(dim)], indexing="ij")
```
mh-nguyen712: grids is already a list, no?
```suggestion
    mesh = torch.meshgrid(*grids, indexing="ij")
```

deepinv/physics/functional/blur.py
```diff
@@ -0,0 +1,400 @@
...
+
+    .. math::
+
+        \begin{equation*}
+            w(x, y) = \begin{cases}
+                (a + 2)|x|^3 - (a + 3)|x|^2 + 1 & \text{if } |x| \leq 1 \\
+                a|x|^3 - 5a|x|^2 + 8a|x| - 4a & \text{if } 1 < |x| < 2 \\
+                0 & \text{otherwise}
+            \end{cases}
+        \end{equation*}
```
mh-nguyen712: not your code but this latex is not rendered
```suggestion
            w(x, y) = \begin{cases}
                (a + 2)|x|^3 - (a + 3)|x|^2 + 1 & \text{if } |x| \leq 1 \\
                a|x|^3 - 5a|x|^2 + 8a|x| - 4a & \text{if } 1 < |x| < 2 \\
                0 & \text{otherwise}
            \end{cases}
```

deepinv/physics/generator/blur.py
```diff
@@ -37,6 +38,239 @@ def __init__(
...
+
+        |sep|
+
+        :Examples:
+
+        >>> from deepinv.physics.generator import GaussianBlurGenerator
+        >>> rng = torch.Generator(device="cpu").manual_seed(0)
+        >>> generator = GaussianBlurGenerator((7, 7), device="cpu", rng=rng, isotropic=False)
+        >>> params = generator.step(batch_size=1)  # dict_keys(['filter'])
+        >>> print(params['filter'])
```
mh-nguyen712: I would make a plot here.
We can catch the plot by adding 
.. plot::
    the same code 

See here for an example: https://deepinv.github.io/deepinv/api/stubs/deepinv.utils.image_to_patches.html#deepinv.utils.image_to_patches

deepinv/physics/generator/blur.py
```diff
@@ -37,6 +38,239 @@ def __init__(
...
+            "num_channels": num_channels,
+            "isotropic": isotropic,
+            "sigma_min": sigma_min,
+            "sigma_max": sigma_max,
+            "angle_min": angle_min,
+            "angle_max": angle_max,
+        }
+        super().__init__(device=device, dtype=dtype, rng=rng, **kwargs)
+
+    def _resolve_angles(
```
mh-nguyen712: Maybe I'm wrong but I feel like we can simplify or combine these 2 functions

deepinv/physics/generator/blur.py
```diff
@@ -37,6 +38,239 @@ def __init__(
...
+                            batch_size, generator=self.rng, **self.factory_kwargs
+                        )
+                        * (smax - smin)
+                        + smin
+                        for smin, smax in zip(
+                            self.sigma_min, self.sigma_max, strict=True
+                        )
+                    ],
+                    dim=-1,
+                )  # Shape: (batch_size, dim)
```
mh-nguyen712: this generation of uniform random number between a min and a max value is used several times, it could be nice to have a helper function for this then call it everywhere.

mh-nguyen712 (changes requested):

mh-nguyen712:
> I added a non-regression test on the new implementation, to check that it matches the old implementation (modulo the interpolation errors due to rotation in the previous implementation).
> 
> The `test_least_squares_implicit_backward` in `test_optim.py` seems to fail with the `"deblur_valid"` config when the blur kernel is a "non-pure" dirac, e.g. with `filter = gaussian_blur(sigma=(0.25, 0.1), angle=45.0)` which gives something like that:
> 
> ```python
> tensor([[[[1.1254e-07, 2.5437e-13, 3.7835e-44],
>           [2.5437e-13, 1.0000e+00, 2.5437e-13],
>           [3.7835e-44, 2.5437e-13, 1.1254e-07]]]])
> ```
> 
> The old implementation did not trigger the error because the error interpolation was flattening the kernel. I switch to `angle=0.0`, it's sufficient for our test purposes. Is this expected behaviour of the implicit solver ? @mh-nguyen712
> 
> I cleaned all the files using the old call to `physics.blur.gaussian_blur` and deprecated it.

The error of `test_least_squares_implicit_backward` is a bit weird to me. As what matters for the solver is the final kernel as a tensor, not really how the kernel is generated. I'll double check on my side

deepinv/physics/generator/blur.py
```diff
@@ -37,6 +39,175 @@ def __init__(
...
+        sigma: torch.Tensor = None,
+        angle: torch.Tensor = None,
+        seed: int = None,
+        **kwargs,
+    ):
+        self.rng_manual_seed(seed)
+        dim = len(self.psf_size)
+
+        # sigma = self._generate_sigma(dim, batch_size, sigma)
+        # angle = self._generate_angle(dim, batch_size, angle)
```
mh-nguyen712:
```suggestion

```

mh-nguyen712 (commented):
LGTM, thanks Romain! Happy to merge once the CI is passing

deepinv/physics/generator/blur.py
```diff
@@ -37,6 +39,175 @@ def __init__(
         self.num_channels = num_channels
 
 
+class GaussianBlurGenerator(PSFGenerator):
+    def __init__(
```
mh-nguyen712, line 78: Add this class to the rst file

deepinv/physics/generator/blur.py
```diff
@@ -37,6 +39,173 @@ def __init__(
...
+
+        :Examples:
+
+        >>> from deepinv.physics.generator import GaussianBlurGenerator
+        >>> rng = torch.Generator(device="cpu").manual_seed(0)
+        >>> generator = GaussianBlurGenerator((7, 7), device="cpu", rng=rng, isotropic=False)
+        >>> params = generator.step(batch_size=4)  # dict_keys(['filter'])
+        >>> dinv.utils.plot(params['filter'])  # doctest: +SKIP
+
+        .. plot::
```
tachella: wow really cool

deepinv/physics/generator/blur.py
```diff
@@ -37,6 +39,173 @@ def __init__(
...
+        sigma_max: float | tuple[float, ...] = 5.0,
+        isotropic: bool = True,
+        angle_min: float | tuple[float, ...] = 0.0,
+        angle_max: float | tuple[float, ...] = 360.0,
+        num_channels: int = 1,
+        rng: torch.Generator = None,
+        device: str = "cpu",
+        dtype: type = torch.float32,
+    ):
+        r"""
```
tachella: the docstring should go above the init (not below) to render properly

tachella: The docstring should go above the init (not below) to render properly

docs/source/user_guide/physics/functional.rst
```diff
@@ -104,4 +104,16 @@ Similar to the PyTorch structure, they are available within ``deepinv.physics.fu
 
    * - :func:`deepinv.physics.functional.random_choice`
      - Randomly selects elements from a given input tensor based on specified probabilities, useful for stochastic sampling in various applications.
+  
+   * - :func:`deepinv.physics.functional.gaussian_blur`
+     - Generates a Gaussian blur kernel, commonly used for image smoothing and noise reduction.
```
tachella:
```suggestion
     - Generates a Gaussian blur kernel in 1D, 2D or 3D, commonly used to model point spread functions.
```

tachella:
```suggestion
     - Generates a Gaussian blur kernel in 1D, 2D or 3D, commonly used to model point spread functions.
```

tachella:
```suggestion
     - Generates a Gaussian blur kernel in 1D, 2D or 3D, commonly used to model point spread functions.
```

tachella:
```suggestion
     - Generates a Gaussian blur kernel in 1D, 2D or 3D, commonly used to model point spread functions.
```

tachella:
```suggestion
     - Generates a Gaussian blur kernel in 1D, 2D or 3D, commonly used to model point spread functions.
```

tachella:
```suggestion
     - Generates a Gaussian blur kernel in 1D, 2D or 3D, commonly used to model point spread functions.
```

tachella:
```suggestion
     - Generates a Gaussian blur kernel in 1D, 2D or 3D, commonly used to model point spread functions.
```

tachella:
```suggestion
     - Generates a Gaussian blur kernel in 1D, 2D or 3D, commonly used to model point spread functions.
```

tachella (commented):
The PR looks really cool, many thanks for this! I only have very minor comments

tachella:
github is bugging and sent my comments multiple times..

By the way, I think it would be great to have also 1D and 3D extensions of `bilinear_filter` `bicubic_filter` and `sinc_filter`. My plan is to add these in #1139

romainvo:
Thanks @tachella for your comments! I've applied the changes :)

tachella (approved):
LGTM, thanks a lot!

PR #1156, opened 2026-04-18
HOTFIX GeneratorMixture

romainvo:
The recent PR #1093 supposed to fix GeneratorMixture actually broke it:
- adding `register_buffer` for `self.probs` and `self.cum_probs` to avoid device mismatch in `self.step`
- add pairs of compatible generators in `MIXTURES` in `test_generators.py` to effectively test the new `use_batch_sampling` feature of GeneratorMixture
- fix logic in `self.step` so that GeneratorMixture functions properly when all batch elements are sampled from the same generator.

### DIFF
--- a/deepinv/physics/generator/base.py
+++ b/deepinv/physics/generator/base.py
@@ -223,11 +223,14 @@ class GeneratorMixture(PhysicsGenerator):
         verbose: bool = False,
     ) -> None:
         super().__init__(device=device, rng=rng)
-        probs = torch.tensor(probs)
+
+        probs = torch.tensor(probs, device=device)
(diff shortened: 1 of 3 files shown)


mh-nguyen712 (approved):
Thanks for the fix! There is a warning of sphinx when building the docs, since the `get_data_home` is deprecated recently in #1105 but it still appears in `changelog.rst`. Can you also fix that please so that the doc is passing (sorry, this is not related to your PR)

romainvo:
> Thanks for the fix! There is a warning of sphinx when building the docs, since the `get_data_home` is deprecated recently in #1105 but it still appears in `changelog.rst`. Can you also fix that please so that the doc is passing (sorry, this is not related to your PR)

I removed the :func: markup 👍

PR #1164, opened 2026-04-25
Add SRResNet

nucli-vicky:
Adds the SRResNet architecture for single-image super-resolution, typically used in supervised SISR training.

SRResNet is the generator network in [SRGAN](https://openaccess.thecvf.com/content_cvpr_2017/papers/Ledig_Photo-Realistic_Single_Image_CVPR_2017_paper.pdf), but the same paper also trains it under L2 and perceptual loss.

While it is a relatively simple network, and not necessarily SOTA, it can still provide a decent baseline, and opens the possibility of adding more supervised SISR networks to deepinv, which will be useful for benchmarking between distortion-loss single-step, GANs, and diffusion models for SISR.


nucli-vicky:
Trained a quick example model (default config) for RGB 4x bicubic super-resolution (`physics =dinv.physics.DownsamplingMatlab(factor=4)`) under L1 loss. 

DIV2K train and val set. Random 128x128 HR crops during training. Eval during train on 128x128 centercrops from each DIV2k val image. ADAM with lr `5e-4` at batch-size 16.

val_psnr_srresnet

Example from DIV2k val:

srresnet_sample

(left-to-right: LR input, HR Output, HR GT)

Performance can certainly be further improved, but this was a quick test training run to verify model is able to train and has behavior which makes sense.

Andrewwango:
Thanks for this Vicky!

Andrewwango (approved):
Looks great thanks Vicky! If you want, you could add an example with your div2k pretrained model (I can upload it to the deepinv HF for you), perform a couple of steps of finetuning on a couple patches from the deepinv HF, to show that it works and it's trainable (and "blind"!).

nucli-vicky:
Hi @Andrewwango , thanks and sounds good! I'll take a look this weekend :)

examples/models/demo_super_resolution.py
```diff
@@ -0,0 +1,208 @@
...
+#     larger and more diverse dataset.
+#
+
+trainer = dinv.Trainer(
+    model=model,
+    physics=physics,
+    optimizer=torch.optim.Adam(model.parameters(), lr=1e-4),
+    train_dataloader=train_dataloader,
+    eval_dataloader=test_dataloader,
+    epochs=10,
```
Andrewwango: can you add a if-else similar to other training examples that on CPU epochs = 1 and the dataset is small and on GPU epochs = 10; and add a small note saying this.
this is mainly to save time on the CPU CI workflow

examples/models/demo_super_resolution.py
```diff
@@ -0,0 +1,208 @@
...
+best_model = trainer.load_best_model()
+
+# %%
+# Plot a reconstruction from the best checkpoint:
+#
+
+x, y = x.to(device), y.to(device)
+with torch.no_grad():
+    x_ft = best_model(y)
+    x_bic_ft = torch.nn.functional.interpolate(
```
Andrewwango, line 192: I guess this is ok but the true adjoint would be using DownsamplingMatlab.A_adjoint with their proprietary filter?

Andrewwango (changes requested):
Looks good to me! Just tiny comments

examples/models/demo_super_resolution.py
nucli-vicky: I agree. However, `DownsamplingMatlab.A_adjoint` gives structurally plausible results, with pixel intensities downscaled to a very small factor. This is expected, but I don't think we should go into how to get this factor in this demo?

nucli-vicky: We could ofcourse use `imresize_matlab(scale=4, ...)`, but this is also not the true adjoint.

examples/models/demo_super_resolution.py
nucli-vicky: done!

Andrewwango (approved):

PR #1165, opened 2026-04-26
Doc + Error + Typing fixes perceptual

nucli-vicky:
- [x] All metrics, even those not using pyiqa, had a docstring part ":param bool check_input_range: if True, ``pyiqa`` will ....". This is not present for those classes, and therefore docstring is misleading. Removed except for `LPIPS`, where I (readded) the `check_input_range`.
- [x] Typing everywhere (NIQE skipped due to #911)
- [x] Added a check that LPIPS' `net_type` has one of the correct values. If wrong `net_type` is passed, `torchmetrics` would raise an uninformative error (`AttributeError: '_NoTrainLpips' object has no attribute 'chns'`)


mh-nguyen712 (approved):
Thanks for the fixes!

PR #1170, opened 2026-04-28
Fix cone_vec geom with Astra FDK

Andrewwango:
`cone_vec` geom type of an Astra proj geom should also count as a conebeam geometry. Therefore, it should get the FDK weighting in the FBP weight calc in the Astra wrapper.

### DIFF
--- a/deepinv/physics/tomography.py
+++ b/deepinv/physics/tomography.py
@@ -604,7 +604,7 @@ class TomographyWithAstra(LinearPhysics):
         sinogram_scaled = torch.clone(sinogram)
         is_3d = len(sinogram.shape) == 5
 
-        if self.geometry_type == "conebeam" and is_3d:
+        if self.geometry_type in ("conebeam", "cone_vec") and is_3d:
             # dimensions (V,N) are (row,col) of the 2D detector
             # A is the number of angles
             B, C, V, A, N = sinogram.shape


romainvo:
Thanks @Andrewwango for pushing this! 

If I'm not mistaken the `"cone_vec"` geometry_type is the naming used by astra. Currently, the `TomographyWithAstra` should raise an error if setting a `geometry_type` different than `"parallel"`, `"fanbeam"` or `"conebeam"`, did you sucessfully initialize the operator when passing "cone_vec" ?

I think your issue is somewhat linked to #1102 , it's a nice addition and it reduces friction for astra users. 

I think the proper fix would be to also overload the accepted `geometry_type` argument in  `astra.py:create_projection_geometry`, e.g. something like

```python
if geometry_type in ("parallel", "parallel3d", "parallel3d_vec"):
    ...
```

not sure how we should manage [2d astra geometries](https://astra-toolbox.com/docs/geom2d.html#projection-geometries) (e.g. "fanflat", "parallel", "parallel_vec") though, as we don't use them in practice.

romainvo:
Throwing some idea here but we can deal with this specificity later.

> not sure how we should manage [2d astra geometries](https://astra-toolbox.com/docs/geom2d.html#projection-geometries) (e.g. "fanflat", "parallel", "parallel_vec") though, as we don't use them in practice.

To make things transparent for astra-users we could have a method that translates [2D astra geometries](https://astra-toolbox.com/docs/geom2d.html#fanflat) e.g a "fanflat" geometry (2D fanbeam) created with:
```python
proj_geom = astra.create_proj_geom('fanflat', det_spacing, det_count, angles, source_origin, origin_det)
```
 to **our** "fanbeam" geometry which is in practice a _flat_  [3D "conebeam" geometry](https://astra-toolbox.com/docs/geom3d.html#cone), i.e. 

```python
projection_geometry = astra.create_proj_geom(
    "cone",
    detector_spacing,  # horizontal spacing
    1.0,  # vertical spacing is unit length
    detector_row_count,
    detector_col_count,
    angles,
    source_radius,
    detector_radius,
)
```

Andrewwango:
Oops sorry I was jumping ahead a bit and using the wrapper with custom geometries folowing #1102

PR #1174, opened 2026-05-02
Add FFDNet + bugfix DRUNet forward dim==3

nucli-vicky:
- Add FFDNet , a non-blind (Gaussian) denoiser, which can be seen DRUNet's predecessor
- Fix a bug in DRUNet regarding sigma shaping when dim is 3d

### DIFF
--- a/deepinv/models/drunet.py
+++ b/deepinv/models/drunet.py
@@ -237,19 +237,25 @@ class DRUNet(Denoiser):
                         f"Incorrect shape, sigma should be of shape (1,), (batch_size,) or (batch_size, 1, height, width, (depth)), got {tuple(sigma.shape)}"
                     )
             else:
-                noise_level_map = torch.ones(
-                    (x.size(0), 1, *x.shape[2:]), device=x.device
-                ) * sigma[None, None, None, None].to(x.device)
+                noise_level_map = torch.full(
+                    (x.size(0), 1, *x.shape[2:]),
+                    sigma,
+                    device=x.device,
+                    dtype=x.dtype,
+                )
         else:
-            noise_level_map = (
-                torch.ones((x.size(0), 1, *x.shape[2:]), device=x.device) * sigma
+            noise_level_map = torch.full(
+                (x.size(0), 1, *x.shape[2:]),
+                sigma,
+                device=x.device,
+                dtype=x.dtype,
             )
 
         x = torch.cat((x, noise_level_map), 1)
         shape_is_safe = all((s % 8 == 0 and s > 31) for s in x.shape[2:])
         if shape_is_safe:
             x = self.forward_unet(x)
-        elif self.training or (x.size(2) < 32 or x.size(3) < 32):
+        elif self.training or any([x.size(2 + i) < 32 for i in range(self.dim)]):
             x = test_pad(self.forward_unet, x, modulo=16)
         else:
             if self.dim == 3:  # pragma: no cover
--- a/deepinv/models/drunet.py
+++ b/deepinv/models/drunet.py
@@ -217,12 +217,23 @@ class DRUNet(Denoiser):
         """
         if isinstance(sigma, torch.Tensor):
             if sigma.ndim > 0:
-                noise_level_map = sigma.view(
-                    x.size(0), 1, *[1 for _ in range(self.dim)]
-                )
-                noise_level_map = noise_level_map.expand(
-                    -1, 1, *[x.size(2 + i) for i in range(self.dim)]
-                )
+                if sigma.shape == (x.size(0), 1, *x.shape[2:]):
+                    noise_level_map = sigma
+                elif sigma.shape in [
+                    (x.size(0)),
+                    (x.size(0), 1, *[1 for _ in range(self.dim)]),
+                ]:
+
+                    noise_level_map = sigma.view(
+                        x.size(0), 1, *[1 for _ in range(self.dim)]
+                    )
+                    noise_level_map = noise_level_map.expand(
+                        -1, 1, *[x.size(2 + i) for i in range(self.dim)]
+                    )
+                else:
+                    raise ValueError(
+                        "Incorrect shape, sigma should be of shape (1,), (batch_size,) or (batch_size, 1, height, width, (depth))"
+                    )
             else:
                 noise_level_map = torch.ones(
                     (x.size(0), 1, *x.shape[2:]), device=x.device
--- a/deepinv/models/drunet.py
+++ b/deepinv/models/drunet.py
@@ -222,7 +222,7 @@ class DRUNet(Denoiser):
                 if sigma.shape == (x.size(0), 1, *x.shape[2:]):
                     noise_level_map = sigma
                 elif sigma.shape in [
-                    (x.size(0)),
+                    (x.size(0),),
                     (x.size(0), 1, *[1 for _ in range(self.dim)]),
                 ]:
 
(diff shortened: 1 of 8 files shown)


deepinv/models/ffdnet.py
```diff
@@ -0,0 +1,116 @@
...
+    def __init__(
+        self,
+        n_conv_layers: int = 15,
+        nf: int = 64,
+        img_channels: int = 1,
+        residual_denoising: bool = False,
+        norm: str | None = "batch_norm",
+        orthogonal_init: bool = True,
+        last_conv_bias: bool = False,
+        device: str | torch.device = "cpu",
```
mh-nguyen712, line 40: can you add `pretrained: str | None = None` similarly to all other denoisers, please? Even it seems that the original paper does not have pretrained weight, just to be compatible with other models

deepinv/models/ffdnet.py
```diff
@@ -0,0 +1,116 @@
...
+        self.upsampler = nn.PixelShuffle(downsample_factor)
+        if norm not in ["instance_norm", "batch_norm", None]:  # pragma: no cover
+            raise ValueError(
+                f"norm must be one of (instance_norm, batch_norm, None), but got {norm}"
+            )
+        norm = {
+            "instance_norm": nn.InstanceNorm2d,
+            "batch_norm": nn.BatchNorm2d,
+            None: nn.Identity(),
+        }[norm]
```
mh-nguyen712, line 54: I don't mind, but it could be cool to have this in models/utils.py, similar to `conv_nd(dim)`

deepinv/models/ffdnet.py
```diff
@@ -0,0 +1,116 @@
...
+                f"FFDNet requires H,W both divisble by 2. Got tensor of shape {tuple(x.shape)}"
+            )
+        if isinstance(sigma, torch.Tensor):
+            if sigma.ndim > 0:
+                noise_level_map = sigma.view(x.size(0), 1, 1, 1)
+                noise_level_map = noise_level_map.expand(-1, 1, x.size(2), x.size(3))
+            else:
+                noise_level_map = torch.ones(
+                    (x.size(0), 1, x.size(2), x.size(3)), device=x.device
+                ) * sigma[None, None, None, None].to(x.device)
```
mh-nguyen712:
```suggestion
                noise_level_map = torch.full(
                    (x.size(0), 1, x.size(2), x.size(3)), sigma, device=x.device, dtype=x.dtype
                )

**Review by mh-nguyen712 — CHANGES_REQUESTED** [review]
Thanks for the addition!

File: deepinv/models/ffdnet.py (hunk 2 above)
Comment (nucli-vicky) [reply deepinv/models/ffdnet.py:54]: Agree, I'd keep this for another PR personally, as we'd also want to add the 3D variants + vision-style LayerNorm

File: deepinv/models/ffdnet.py (hunk 1 above)
Comment (nucli-vicky) [reply deepinv/models/ffdnet.py:40]: sure! I do think adding pretrained weights for `FFDNet` might be a good future to-do

**Comment by nucli-vicky** [comment]
Thanks for the comments @mh-nguyen712 ! I tried to address them :)

File: docs/source/changelog.rst [hunk 4]
```diff
@@ -13,6 +13,7 @@ New Features
 - Add :class:`deepinv.optim.MLEM` algorithm for Poisson inverse problems (:gh:`1051` by `Thibaut Modrzyk`_)
 - Add the equivariant splitting loss :class:`deepinv.loss.EquivariantSplittingLoss` with equivariant reconstructors :class:`deepinv.models.EquivariantReconstructor` and virtual physics :class:`deepinv.physics.VirtualLinearPhysics` (:gh:`881` by `Jérémy Scanvic`_)
 - Add `DEEPINV_CACHE_DIR` environment variable to set the cache directory for datasets and pretrained weights (:gh:`1105` by `Minh Hai Nguyen`_)
+- Add :class:`deepinv.models.FFDNet` for non-blind Gaussian denoising (:gh:`976` by `Vicky De Ridder`_)
```
romainvo: wrong PR identifier ?

deepinv/models/drunet.py
```diff
@@ -221,14 +221,20 @@ def forward(self, x: torch.Tensor, sigma: torch.Tensor | float) -> torch.Tensor:
...
+                    (x.size(0),),
+                    (x.size(0), 1, *[1 for _ in range(self.dim)]),
+                ]:
+
+                    noise_level_map = sigma.view(
+                        x.size(0), 1, *[1 for _ in range(self.dim)]
+                    )
                     noise_level_map = noise_level_map.expand(
-                        -1, 1, x.size(2), x.size(3)
+                        -1, 1, *[x.size(2 + i) for i in range(self.dim)]
```
romainvo, line 233: I feel like we should test this, e.g., in `test_drunet_options`, as this is already the second recent PR playing with the drunet noise handling. What do you think ?

romainvo (changes requested):
Hi @nucli-vicky, thanks for the fix and the great addition! Minor comments about testing, otherwise the rest looks good to me :)

nucli-vicky:
Thanks for the comments @romainvo !

I improved the tests for the different types of sigma input (+ possible image shapes), and proper handling of both 2d and 3d.

This did uncover that, right now, `DRUNet` errors when given an image where spatial shapes `s` are `32 < s < 64`, but `s % 8 != 0` , because `test_onesplit` is used with `refield=64`. 

For now I addressed this by using `test_pad` any time `shape_is_safe == False` and some `s < 64`.  This does not change any existing behavior, as inputs affected would right now raise `The size of tensor a (8) must match the size of tensor b (9) at non-singleton dimension 3`.

Another option is using `test_one_split` for the failure cases discussed, but with `refield=32`.

deepinv/models/ffdnet.py
mh-nguyen712: What do you think about this suggestion? I think it's a bit cleaner

mh-nguyen712 (changes requested):

deepinv/models/ffdnet.py
nucli-vicky: I agree, thought I added it but I misremembered

nucli-vicky: I did add it, but not for the case where `sigma` is a float, only for a scalar tensor

mh-nguyen712: It should also work for scalar tensor too.

nucli-vicky: added for float too

mh-nguyen712 (approved):

romainvo (approved):
Thanks for the fixes! Looks good to me :)

romainvo:
> Thanks for the comments @romainvo !
> 
> I improved the tests for the different types of sigma input (+ possible image shapes), and proper handling of both 2d and 3d.
> 
> This did uncover that, right now, `DRUNet` errors when given an image where spatial shapes `s` are `32 < s < 64`, but `s % 8 != 0` , because `test_onesplit` is used with `refield=64`.
> 
> For now I addressed this by using `test_pad` any time `shape_is_safe == False` and some `s < 64`. This does not change any existing behavior, as inputs affected would right now raise `The size of tensor a (8) must match the size of tensor b (9) at non-singleton dimension 3`.
> 
> Another option is using `test_one_split` for the failure cases discussed, but with `refield=32`.

I think it's okay for now, we can address that in another PR if needed 👍

PR #1188, opened 2026-05-11
Migrate the domain name

jscanvic:
The goal of this PR is to move away from deepinv.github.io to deepinv.org

**Changes**
- [x] Change URLs pointing to deepinv.github.io in the code base
- [x] Get rid of deepinv.github.io in the documentation build

**Absolute to relative URLs**

In ec34fe934b6daa34846b2d4bd60e555db2d57d3d I bypass cache issues resulting in links pointing to the previous domain by switching from absolute (cache-dependent) to relative (always up-to-date) URLs for internal links. I believe it shouldn't mess with anything, which is why I preferred that solution instead of dealing with cache invalidation stuff.

**Plan for Google Search migration**

1. [x] List all relevant Google Search Console properties
    - https://deepinv.github.io
    - https://deepinv.github.io/deepinv
2. [ ] ~~Submit a change of address for each of the top-level ones~~
    - [ ] ~~For https://deepinv.github.io~~
3. [x] Register the sitemap for deepinv.org

**Update - Jun'22, 2026**

For technical reasons related to there being 2 distinct GitHub pages websites at deepinv/deepinv and deepinv/deepinv.github.io we didn't formally migrate the website using the Google Search Console form. The current redirections should suffice for a proper transfer on their end, but if we notice poor indexation here's an alternate plan we can put into place.

1. Find a way to deal with the deepinv/ prefix for proper redirections & put the solution in place
2. Change CI so it builds the docs into `deepinv/deepinv.github.io:main` instead of `deepinv/deepinv:gh-pages`
3. Get rid of the GitHub Pages website at deepinv/deepinv
4. Setup the GitHub pages at deepinv/deepinv.github.io to point to deepinv.org

**SEO**

- [How to move a site - Google Search](https://developers.google.com/search/docs/crawling-indexing/site-move-with-url-changes)
- [Change of Address tool - Google Search](https://support.google.com/webmasters/answer/9370220)

**Subdomains / subdirectories**

Instead of moving the docs from https://deepinv.github.io to https://deepinv.org, we could instead move it to a subdomain https://docs.deepinv.org or a subdirectory https://deepinv.org/docs. Upon consideration, I believe we should directly migrate to the top-level URL.

Below pros & cons of moving to the top-level URL.

Pros
- Easier and better for SEO
- Neat homepage URL without awkward redirections

Cons
- We might eventually have to migrate stuff to have a non-technical landing page at https://deepinv.org for media, funding agencies, and so on.
- It's a bit more tedious to migrate from a generic URL to a specific URL than doing the converse.

**TODO**
- [x] Setup a Google Search Console account for deepinv.org
- [x] Plan out the Google Search migration
- [x] List pros and cons between https://deepinv.org/ and https://deepinv.org/docs
- [x] Check how to do the DNS change through Cloudflare
- [x] Point DNS to GitHub pages for deepinv.org
- [x] Merge this PR
- [x] Setup redirects from deepinv.github.io to deepinv.org
  - [x] A GitHub Pages redirect in https://github.com/deepinv/deepinv
  - [x] A GitHub Pages (ideally) or client-side redirect in https://github.com/deepinv/deepinv.github.io
- [x] Setup deepinv.org with GitHub pages
- [x] Apply the migration plan for Google Search

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -74,7 +74,7 @@ intersphinx_mapping = {
     "torch": ("https://pytorch.org/docs/stable/", None),
     "torchvision": ("https://pytorch.org/vision/stable/", None),
     "python": ("https://docs.python.org/3.9/", None),
-    "deepinv": ("https://deepinv.github.io/deepinv/", None),
+    "deepinv": ("https://deepinv.org/", None),
     "matplotlib": ("https://matplotlib.org/stable/", None),
 }
 
@@ -94,7 +94,7 @@ autodoc_inherit_docstrings = False
 # For bibtex
 bibtex_footbibliography_backrefs = True
 # for sitemap
-html_baseurl = "https://deepinv.github.io/deepinv/"
+html_baseurl = "https://deepinv.org/"
 html_extra_path = ["robots.txt"]
 # Include reStructuredText sources
 html_copy_source = True
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -74,7 +74,6 @@ intersphinx_mapping = {
     "torch": ("https://pytorch.org/docs/stable/", None),
     "torchvision": ("https://pytorch.org/vision/stable/", None),
     "python": ("https://docs.python.org/3.9/", None),
-    "deepinv": ("https://deepinv.org/", None),
     "matplotlib": ("https://matplotlib.org/stable/", None),
 }
 
@@ -311,7 +310,8 @@ sphinx_gallery_conf = {
     "ignore_pattern": ignore_pattern,
     "reference_url": {
         # The module you locally document uses None
-        "sphinx_gallery": None
+        "sphinx_gallery": None,
+        "deepinv": None,
     },
     # directory where function/class granular galleries are stored
     "backreferences_dir": "gen_modules/backreferences",
--- a/examples/sampling/demo_diffusers.py
+++ b/examples/sampling/demo_diffusers.py
@@ -5,7 +5,7 @@ Using state-of-the-art diffusion models from HuggingFace Diffusers with DeepInve
 This demo shows you how to use our wrapper
 :class:`deepinv.models.DiffusersDenoiserWrapper` to turn any SOTA models from the HuggingFace Hub to an image denoiser. It also can be used to perform unconditional image generation or for posterior sampling.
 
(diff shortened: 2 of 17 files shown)


examples/sampling/demo_diffusers.py
```diff
@@ -5,7 +5,7 @@
 This demo shows you how to use our wrapper
 :class:`deepinv.models.DiffusersDenoiserWrapper` to turn any SOTA models from the HuggingFace Hub to an image denoiser. It also can be used to perform unconditional image generation or for posterior sampling.
 
-See more about the `diffusers pipeline <https://huggingface.co/docs/diffusers/index>`_ and our posterior sampling `user guide <https://deepinv.github.io/deepinv/auto_examples/sampling/demo_diffusion_sde.html>`_.
+See more about the `diffusers pipeline <https://huggingface.co/docs/diffusers/index>`_ and our posterior sampling `user guide <https://deepinv.org/auto_examples/sampling/demo_diffusion_sde.html>`_.
```
Andrewwango: this should use :ref: no?

examples/sampling/demo_diffusion_sde.py
```diff
@@ -36,7 +36,7 @@
 
 The first term is the score function of the unconditional SDE, which is typically approximated by an MMSE denoiser (`denoiser`) using the well-known Tweedie's formula, while the
 second term is approximated by the (noisy) data-fidelity term (`data_fidelity`).
-We implement various data-fidelity terms in `the user guide <https://deepinv.github.io/deepinv/user_guide/reconstruction/sampling.html#id2>`_.
+We implement various data-fidelity terms in `the user guide <https://deepinv.org/user_guide/reconstruction/sampling.html#id2>`_.
```
Andrewwango: this should use :ref: no?

Andrewwango (changes requested):
Minor comment

docs/source/conf.py
```diff
@@ -74,7 +74,6 @@
     "torch": ("https://pytorch.org/docs/stable/", None),
     "torchvision": ("https://pytorch.org/vision/stable/", None),
     "python": ("https://docs.python.org/3.9/", None),
-    "deepinv": ("https://deepinv.github.io/deepinv/", None),
```
tachella, line 77: I remember needing this to have the links from the deepinv classes inside sphinx gallery examples, do you think this will still work in the new doc?

tachella (approved):
this looks great to me, I only have a minor question :)

docs/source/conf.py
jscanvic: Thanks for the review! It seems to work without it

Andrewwango (approved):

PR #1192, opened 2026-05-18
Fix BM3D silently breaks for multi-channel images when number of channels is not 3

### LINKED ISSUE #1191: BM3D silently breaks for multi-channel images when number of channels is not 3
### Environment

Python 3.12.13, DeepInverse 0.4.0

### Describe the bug

Currently, `deepinv.models.BM3D` is kind of hard-coded to take images of shape `(B, C, W, H)` as input. It then for-loops the batch dimension. When `C == 1`, everything works fine. When `C == 3`, everything works fine too. In all other cases, each tensor of shape `(C, W, H)` is **supposedly** reshaped to `(W, H, C)` first (which they are **not**) before applying `bm3d.bm3d`. There is no error because `bm3d.bm3d` thinks `(C, W, H)` is `(W, H, C)`. 

It breaks because [this](https://github.com/deepinv/deepinv/blob/main/deepinv/models/utils.py):
```
def tensor2array(img):
    img = img.cpu().detach().numpy()
    img = np.transpose(img, (1, 2, 0))
    return img


def array2tensor(img):
    return torch.from_numpy(img).permute(2, 0, 1)


def tensor2array(img: Tensor) -> np.ndarray:
    img = img.cpu().detach().numpy()
    if img.shape[0] == 3:  # Color case: cast to numpy format (W,H,C)
        img = np.transpose(img, (1, 2, 0))
    else:  # Grayscale case: cast to numpy format (W,H)
        img = img[0]
    return img


def array2tensor(img: np.ndarray) -> Tensor:
    if len(img.shape) == 3:  # Color case: back to (C,W,H)
        out = torch.from_numpy(img).permute(2, 0, 1)
    else:  # Grayscale case: back to (1,W,H)
        out = torch.from_numpy(img).unsqueeze(0)
    return out
```
First of all, there are two `tensor2array` and two `array2tensor` whereas only the second `tensor2array` and `array2tensor` are used. Second of all, this causes `deepinv.models.BM3D` to break silently because the transpose is **not** performed unless `C == 3`. This can be fixed by changing the condition, e.g.,
```
def tensor2array(img: Tensor) -> np.ndarray:
    img = img.cpu().detach().numpy()
    if img.shape[0] == 1:  # Grayscale case: cast to numpy format (W,H)
        img = img[0]
    else:  # All other cases: cast to numpy format (W,H,C)
        img = np.transpose(img, (1, 2, 0))
    return img


def array2tensor(img: np.ndarray) -> Tensor:
    if len(img.shape) == 1:  # Grayscale case: back to (1,W,H)
        out = torch.from_numpy(img).unsqueeze(0)
    else:  # All other cases: back to (C,W,H)
        out = torch.from_numpy(img).permute(2, 0, 1)
    return out
```

### Steps to reproduce

Run the following
```
import deepinv as dinv
import torch
x = torch.rand(1, 2, 32, 32)
sigma = torch.tensor([0.1])
denoiser = dinv.models.BM3D()
out = denoiser(x, sigma)
```
Use a debugger to inspect the shape of each variable.

### Expected behavior

Running the following
```
import deepinv as dinv
import torch
x = torch.rand(2, 32, 32)
y = tensor2array(x)
print(y.shape)
```
show give `(32, 32, 2)` rather than `(2, 32, 32)`.

Not applicable because it breaks silently.

### Additional context

_No response_

kvttt:
Fix #1191.

### DIFF
--- a/deepinv/models/utils.py
+++ b/deepinv/models/utils.py
@@ -15,30 +15,20 @@ import io
 import contextlib
 
 
-def tensor2array(img):
-    img = img.cpu().detach().numpy()
-    img = np.transpose(img, (1, 2, 0))
-    return img
-
-
-def array2tensor(img):
-    return torch.from_numpy(img).permute(2, 0, 1)
-
-
 def tensor2array(img: Tensor) -> np.ndarray:
     img = img.cpu().detach().numpy()
-    if img.shape[0] == 3:  # Color case: cast to numpy format (W,H,C)
-        img = np.transpose(img, (1, 2, 0))
-    else:  # Grayscale case: cast to numpy format (W,H)
+    if img.shape[0] == 1:  # Grayscale case: cast to numpy format (W,H)
         img = img[0]
+    else:  # All other cases: cast to numpy format (W,H,C)
+        img = np.transpose(img, (1, 2, 0))
     return img
 
 
 def array2tensor(img: np.ndarray) -> Tensor:
-    if len(img.shape) == 3:  # Color case: back to (C,W,H)
-        out = torch.from_numpy(img).permute(2, 0, 1)
-    else:  # Grayscale case: back to (1,W,H)
+    if len(img.shape) == 1:  # Grayscale case: back to (1,W,H)
         out = torch.from_numpy(img).unsqueeze(0)
+    else:  # All other cases: back to (C,W,H)
+        out = torch.from_numpy(img).permute(2, 0, 1)
     return out
 
 
--- a/deepinv/models/utils.py
+++ b/deepinv/models/utils.py
@@ -25,7 +25,7 @@ def tensor2array(img: Tensor) -> np.ndarray:
 
 
 def array2tensor(img: np.ndarray) -> Tensor:
-    if len(img.shape) == 1:  # Grayscale case: back to (1,W,H)
+    if len(img.shape) == 2:  # Grayscale case: back to (1,W,H)
         out = torch.from_numpy(img).unsqueeze(0)
     else:  # All other cases: back to (C,W,H)
         out = torch.from_numpy(img).permute(2, 0, 1)
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -35,6 +35,7 @@ Fixed
 - Fix dimensions mismatch in :class:`deepinv.physics.TomographyWithAstra` with 3D phantoms (:gh:`1137` by `Baptiste Legouix`_)
 - Add warning and error handling for negative inputs in BlurFFT and Poisson noise (:gh:`1155` by `Thibaut Modrzyk`_)
 - Fix a bug in the custom backward of the least-squares solvers for non-leaf tensors (:gh:`1146` by `Minh Hai Nguyen`_)
+- Fix a bug that caused :class:`deepinv.models.BM3D` to silently break for multi-channel images when the number of channels is not 3 (:gh:`1192` by `Kaibo Tang`_)
 
 
 v0.4.0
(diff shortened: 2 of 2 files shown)


deepinv/models/utils.py
```diff
@@ -15,30 +15,20 @@
...
+    else:  # All other cases: cast to numpy format (W,H,C)
+        img = np.transpose(img, (1, 2, 0))
     return img
 
 
 def array2tensor(img: np.ndarray) -> Tensor:
-    if len(img.shape) == 3:  # Color case: back to (C,W,H)
-        out = torch.from_numpy(img).permute(2, 0, 1)
-    else:  # Grayscale case: back to (1,W,H)
+    if len(img.shape) == 1:  # Grayscale case: back to (1,W,H)
```
jscanvic: It should be 2 instead of 1, right?

jscanvic (changes requested):
Thank you for identifying this bug & for fixing it!

Just a typo I believe and it should be good to go

deepinv/models/utils.py
kvttt: Yea right. My brain wasn't braining when I wrote that line. :(

kvttt:
Fixed now. I have also run the previously failed test locally and it passed so it should be good to go.

kvttt:
While we are here, I have some additional thoughts/suggestions. 

1. I saw in `pyproject.toml` that `bm4d` is listed as an optional dependency. However, BM4D is not yet available as a denoiser at the moment. Similar to how `bm3d.bm3d` supports multi-channel 2D images, `bm4d.bm4d_multichannel` also supports multi-channel 3D images, except that the channel dimension is assumed to be in the front. One reason for including BM4D is that one can use BM4D as a plug-and-play prior for 3D imaging applications. 
2. Currently BM3D is done on the CPU, which is time-consuming, especially for larger images. In addition, there is overhead for transferring between GPU and CPU when the tensor is originally on the GPU. I have recently re-implemented BM3D using CuPy (see [bm3d-cupy](https://github.com/kvttt/bm3d_cupy)). The only dependencies are PyWavelets and CuPy. However, the original `bm3d` contains hard-coded wavelets, whereas all wavelets I used are from PyWavelets, so the results might not be fully reproducible. Nevertheless, the performance is similar and speedup is ~16x on my machine. The actual speedup is probably more considering the transferring overhead. Let me know if you are interested in having a faster BM3D. There are two options here. I can port my code from CuPy to PyTorch. Alternatively, CuPy and PyTorch support zero-copy conversion in both directions through the DLPack protocol.

Let me know what you think!

jscanvic:
Ty for pointing this out!

1. The bm4d line is merely a bug fix for bm3d (it's one of its dependencies somehow) #275 I think it would be nice to support it! I'd leave that for another PR though
2. That's really neat! Also really impressive that you made your own implementation! Can you make a new issue to discuss that with other users/maintainers? I think there are a few design decisions we need to make before rolling out a new implementation.

jscanvic (changes requested):
LGTM! Forgot to ask sorry - can you please fill in `docs/source/changelog.rst` so we keep a record of your contribution?

See https://deepinv.github.io/deepinv/contributing.html#log-your-changes for more details

Then I'll approve the PR

jscanvic (approved):

jscanvic:
(CI failure is unrelated - seems to be caused by this update https://pypi.org/project/yarl/1.24.0/ pushed 30 min ago)

romainvo (approved):
Hi @kvttt, thanks for spotting this bug and fixing it! Everything looks good to me :)

PR #1195, opened 2026-05-20
Re-implement BM3D in PyTorch

### LINKED ISSUE #1194: Improve BM3D speed
### Environment (optional)

_No response_

### Describe the feature

Currently, `deepinv.models.BM3D` relies on `deepinv.models.utils.tensor2array` and `deepinv.models.utils.array2tensor` to transfer between PyTorch and NumPy. This is because `bm3d.bm3d` runs on the CPU and requires the input to be `numpy.ndarray`. There are two main consequences as a result of this. 
1. When the tensor is originally on the GPU, doing so requires transferring between the GPU and the CPU, which introduces overhead. The effect of this overhead is more significant when BM3D is used as a plug-and-play prior since it is repeatedly called at each iteration.
2. The algorithm behind BM3D is well-suited for parallelization since each "reference patch" is processed independently. The original `bm3d.bm3d` already takes advantage of parallelization but further acceleration is possible by porting everything to the GPU. I have recently re-implemented BM3D using CuPy (see [bm3d-cupy](https://github.com/kvttt/bm3d_cupy)). I also included a benchmarking script so that one can compare and see the speedup on their own machine. The only dependencies are PyWavelets and CuPy. However, the original `bm3d` contains hard-coded wavelets, whereas all wavelets I used are from PyWavelets, so the results might not be fully reproducible. Nevertheless, the performance is similar and speedup is ~16x on my machine (see below). The actual speedup is potentially even higher if the transferring overhead is taken into consideration. 

Image

```
BM3D (CPU) time: 3.0155 (0.0226) s
  PSNR: 30.85 dB
  SSIM: 0.8194


BM3D (GPU) time: 0.1881 (0.0004) s
  Speedup: 16.03x
  PSNR: 30.90 dB
  SSIM: 0.8228
```

Following @jscanvic's suggestion, I would like to ask the community for opinions. To start this off, here are some of my initial thoughts.

Implementation-wise, there are two options:
- Option 1: Port my code (written with CuPy) to PyTorch. 
- Option 2: Keep my code in CuPy and add helper functions to handle shape and zero-copy conversion between CuPy and PyTorch.

As I mentioned above, my re-implementation uses wavelets from PyWavelets whereas the original `bm3d` hard-codes some of the wavelets. In order to enforce consistency between the wavelets that are used, it is possible to use the function `bm4d._get_transf_matrix` to get the exact wavelets that the original `bm3d` and `bm4d` are using. One benefit of using `bm4d._get_transf_matrix` is better reproducibility. However, it also makes `bm4d` a dependency.

On the user-facing side, my initial thought is to have a keyword argument like `use_legacy=True` to keep the old behavior (use `bm3d.bm3d`) for backward compatibility. The user can then set this keyword argument to `False` to use the GPU-accelerated version.

### Motivation

To have a faster BM3D.

### Alternatives considered

_No response_

### Additional context

Request for this new feature is originally brought up in #1192.

kvttt:
Fix #1194. 

### DIFF
--- a/deepinv/models/bm3d.py
+++ b/deepinv/models/bm3d.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
-import numpy as np
 import torch
 from torch import Tensor
 from .base import Denoiser
 from .utils import array2tensor, tensor2array
+from deepinv.physics.functional import dct
+from deepinv.utils import image_to_patches
 
 EPS = 2 ** (-52)
 
@@ -32,10 +33,10 @@ class BM3D(Denoiser):
         - ``chunk_size``: number of groups to process in parallel. Default: 2048
         - ``ht_group_size``: group size for stage 1 (hard-thresholding). Default: 16
         - ``wiener_group_size``: group size for stage 2 (Wiener filtering). Default: 32
-        - ``spatial_ht_transform``: spatial transform for stage 1. Default: "bior1.5"
-        - ``spatial_wiener_transform``: spatial transform for stage 2. Default: "dct"
-        - ``group_ht_transform``: group transform for stage 1. Default: "haar"
-        - ``group_wiener_transform``: group transform for stage 2. Default: "haar"
+        - ``spatial_ht_transform``: spatial transform for stage 1. Default: `"bior1.5"`
+        - ``spatial_wiener_transform``: spatial transform for stage 2. Default: `"dct"`
+        - ``group_ht_transform``: group transform for stage 1. Default: `"haar"`
+        - ``group_wiener_transform``: group transform for stage 2. Default: `"haar"`
         - ``hard_threshold``: hard-thresholding parameter for stage 1. Default: 3.0
         - ``wiener_mu2``: Wiener filtering parameter for stage 2. Default: 0.4
 
@@ -55,7 +56,6 @@ class BM3D(Denoiser):
     ):
         super(BM3D, self).__init__()
         self.use_legacy = use_legacy
-        self.device = device
         if not self.use_legacy:
             # patch_size
             self.p = int(kwargs.get("patch_size", 8))
@@ -79,45 +79,49 @@ class BM3D(Denoiser):
                 min(max(1, int(kwargs.get("wiener_group_size", 32))), self.n_candidates)
             )  # stage 2 (Wiener filtering) group size
             # 2d transforms (along spatial dimensions)
-            self.spatial_ht, self.spatial_ht_inv = self._transform_matrices(
-                self.p, kwargs.get("spatial_ht_transform", "bior1.5"), self.device
+            spatial_ht, spatial_ht_inv = self._transform_matrices(
+                self.p, kwargs.get("spatial_ht_transform", "bior1.5")
             )  # biorthogonal WT, (p, p)
-            self.spatial_wiener, self.spatial_wiener_inv = self._transform_matrices(
-                self.p, kwargs.get("spatial_wiener_transform", "dct"), self.device
+            self.register_buffer("spatial_ht", spatial_ht)
+            self.register_buffer("spatial_ht_inv", spatial_ht_inv)
+            spatial_wiener, spatial_wiener_inv = self._transform_matrices(
+                self.p, kwargs.get("spatial_wiener_transform", "dct")
             )  # DCT, (p, p)
+            self.register_buffer("spatial_wiener", spatial_wiener)
+            self.register_buffer("spatial_wiener_inv", spatial_wiener_inv)
             # 1d transforms (along group dimension)
-            self.group_ht, self.group_ht_inv = self._transform_matrices(
+            group_ht, group_ht_inv = self._transform_matrices(
                 self.ht_group_size,
                 kwargs.get("group_ht_transform", "haar"),
-                self.device,
             )  # Haar, (k_ht, k_ht)
-            self.group_wiener, self.group_wiener_inv = self._transform_matrices(
+            self.register_buffer("group_ht", group_ht)
+            self.register_buffer("group_ht_inv", group_ht_inv)
+            group_wiener, group_wiener_inv = self._transform_matrices(
                 self.wiener_group_size,
                 kwargs.get("group_wiener_transform", "haar"),
-                self.device,
             )  # Haar, (k_w, k_w)
+            self.register_buffer("group_wiener", group_wiener)
+            self.register_buffer("group_wiener_inv", group_wiener_inv)
             # Kaiser window, applied to weights during aggregation to reduce boundary artifact
             win_1d = torch.kaiser_window(
                 self.p,
                 beta=2.0,
                 periodic=False,
-                device=self.device,
-                dtype=torch.float32,
             )  # (p,)
-            self.win = win_1d[:, None] * win_1d[None, :]  # (p, p)
+            self.register_buffer("win", win_1d[:, None] * win_1d[None, :])  # (p, p)
             # helper arrays for aggregation
-            self.patch_y = torch.arange(self.p, dtype=torch.int64, device=self.device)[
-                None, None, None, :, None
-            ]  # (1, 1, 1, p, 1)
-            self.patch_x = torch.arange(self.p, dtype=torch.int64, device=self.device)[
-                None, None, None, None, :
-            ]  # (1, 1, 1, 1, p)
-            self.patch_c = torch.arange(3, dtype=torch.int64, device=self.device)[
-                None, None, :, None, None
-            ]  # (1, 1, c, 1, 1)
+            self.register_buffer(
+                "patch_y",
+                torch.arange(self.p, dtype=torch.int64)[None, None, None, :, None],
+            )  # (1, 1, 1, p, 1)
+            self.register_buffer(
+                "patch_x",
+                torch.arange(self.p, dtype=torch.int64)[None, None, None, None, :],
+            )  # (1, 1, 1, 1, p)
             # other configs
             self.hard_threshold = float(kwargs.get("hard_threshold", 3.0))
             self.wiener_mu2 = float(kwargs.get("wiener_mu2", 0.4))
+            self.to(device)
 
     def _get_wavelet_matrix_torch(
         self,
@@ -145,21 +149,11 @@ class BM3D(Denoiser):
         return matrix, torch.linalg.inv(matrix)
 
     def _get_dct_matrix(self, n: int, norm: str = "ortho") -> tuple[Tensor, Tensor]:
-        try:
-            from scipy.fft import dct
-        except ImportError:  # pragma: no cover
-            raise ImportError(
-                "scipy package not found. Please install it with `pip install scipy`."
-            )
-        matrix = dct(np.eye(n, dtype=np.float32), norm=norm, axis=0)
-        inverse = matrix.T if norm == "ortho" else np.linalg.inv(matrix)
-        return torch.as_tensor(matrix, dtype=torch.float32), torch.as_tensor(
-            inverse, dtype=torch.float32
-        )
+        matrix = dct(torch.eye(n, dtype=torch.float32), norm=norm).T
+        inverse = matrix.T if norm == "ortho" else torch.linalg.inv(matrix)
+        return matrix, inverse
 
-    def _get_transform_matrices(
-        self, n: int, transform_name: str
-    ) -> tuple[Tensor, Tensor]:
+    def _transform_matrices(self, n: int, transform_name: str) -> tuple[Tensor, Tensor]:
         n = int(n)
         transform_name = transform_name.lower()
         if transform_name == "dct":
@@ -168,49 +162,48 @@ class BM3D(Denoiser):
             return self._get_wavelet_matrix_torch(n, transform_name)
         raise ValueError(f"Unsupported transform '{transform_name}' for size n={n}.")
 
-    def _transform_matrices(
-        self, n: int, transform_name: str, device: torch.device
-    ) -> tuple[Tensor, Tensor]:  # get transform matrices
-        forward, inverse = self._get_transform_matrices(n, transform_name)
-        return forward.to(device=device, dtype=torch.float32), inverse.to(
-            device=device, dtype=torch.float32
-        )
-
     def _patch_tensor(
         self,
         img: Tensor,
     ) -> Tensor:  # extract all `hp*wp` patches
-        patches = img.unfold(0, self.p, 1).unfold(
-            1, self.p, 1
-        )  # Result: (hp, wp, c, p, p)
-        return patches.contiguous().reshape(
-            -1, img.shape[2], self.p, self.p
+        patches = image_to_patches(
+            img.permute(2, 0, 1).unsqueeze(0),  # (1, c, h, w)
+            self.p,
+            1,
+            False,
+        )  # (1, c, hp, wp, p, p)
+        return (
+            patches.permute(2, 3, 0, 1, 4, 5)
+            .contiguous()
+            .reshape(-1, img.shape[2], self.p, self.p)
         )  # (hp*wp, c, p, p)
 
     def _reference_grid(
-        self, hp: int, wp: int
+        self, hp: int, wp: int, device: torch.device
     ) -> tuple[Tensor, Tensor]:  # get locations of all reference patches
-        ys = torch.arange(0, hp, self.ref_stride, dtype=torch.int64, device=self.device)
-        xs = torch.arange(0, wp, self.ref_stride, dtype=torch.int64, device=self.device)
+        ys = torch.arange(0, hp, self.ref_stride, dtype=torch.int64, device=device)
+        xs = torch.arange(0, wp, self.ref_stride, dtype=torch.int64, device=device)
         # always include the last patch to avoid boundary artifact
         if ((hp - 1) // self.ref_stride) * self.ref_stride != hp - 1:
             ys = torch.cat(
-                (ys, torch.as_tensor([hp - 1], dtype=torch.int64, device=self.device))
+                (ys, torch.as_tensor([hp - 1], dtype=torch.int64, device=device))
             )
         if ((wp - 1) // self.ref_stride) * self.ref_stride != wp - 1:
             xs = torch.cat(
-                (xs, torch.as_tensor([wp - 1], dtype=torch.int64, device=self.device))
+                (xs, torch.as_tensor([wp - 1], dtype=torch.int64, device=device))
             )
         yy, xx = torch.meshgrid(ys, xs, indexing="ij")
         return yy.reshape(-1), xx.reshape(-1)  # (n_refs,), (n_refs,)
 
-    def _search_offsets(self) -> tuple[Tensor, Tensor]:  # get candidate offsets
+    def _search_offsets(
+        self, device: torch.device
+    ) -> tuple[Tensor, Tensor]:  # get candidate offsets
         offsets = torch.arange(
             -self.offset_radius * self.search_step,
             self.offset_radius * self.search_step + 1,
             self.search_step,
             dtype=torch.int64,
-            device=self.device,
+            device=device,
         )  # (2*offset_radius + 1,)
         yy, xx = torch.meshgrid(offsets, offsets, indexing="ij")
         return yy.reshape(-1), xx.reshape(-1)  # (n_candidates,), (n_candidates,)
@@ -237,13 +230,14 @@ class BM3D(Denoiser):
         weight: Tensor,
         pos_y: Tensor,
         pos_x: Tensor,
+        patch_c: Tensor,
         patches: Tensor,
         patch_weight: Tensor,
     ) -> None:
         # all `p` pixel indices within all `k` patches for all `n` groups
         iy = pos_y[:, :, None, None, None] + self.patch_y  # (n, k, 1, p, 1)
         ix = pos_x[:, :, None, None, None] + self.patch_x  # (n, k, 1, 1, p)
-        idx = ((iy * w + ix) * c + self.patch_c).reshape(-1)
+        idx = ((iy * w + ix) * c + patch_c).reshape(-1)
         weighted_window = (
             self.win[None, None, None, :, :] * patch_weight
         )  # (n, 1, c, p, p)
@@ -262,10 +256,11 @@ class BM3D(Denoiser):
         ref_x: Tensor,
         off_y: Tensor,
         off_x: Tensor,
+        patch_c: Tensor,
         hp: int,
         wp: int,
-        sigma_ch: float | Tensor,
-        sigma2_ch: float | Tensor,
+        sigma: float,
+        sigma2: float,
         match_patches: Tensor,
         noisy_patches: Tensor,
         group_size: int,
@@ -274,18 +269,15 @@ class BM3D(Denoiser):
         group_transform: Tensor,
         group_inverse: Tensor,
         stage: str,
+        device: torch.device,
     ) -> Tensor:
         # patch features
         match_metric = match_patches[:, 0, :, :].reshape(
             -1, self.p * self.p
         )  # (hp*wp, p*p)
-        patch_norm = torch.tensor(
-            self.p * self.p, dtype=torch.float32, device=self.device
-        )
+        patch_norm = torch.tensor(self.p * self.p, dtype=torch.float32, device=device)
         # initialize accumulators
-        accum = torch.zeros(
-            h * w * c, dtype=torch.float32, device=self.device
-        )  # (h*w*c,)
+        accum = torch.zeros(h * w * c, dtype=torch.float32, device=device)  # (h*w*c,)
         weight = torch.zeros_like(accum)  # (h*w*c,)
         # process `chunk_size` groups in parallel
         for start in range(0, n_refs, self.chunk_size):
@@ -312,7 +304,9 @@ class BM3D(Denoiser):
                 torch.sum((cand - ref[:, None, :]) ** 2, dim=2) / patch_norm
             )  # (chunk_size, n_candidates)
             dist = torch.where(
-                valid, dist, torch.full_like(dist, torch.inf)
+                valid,
+                dist,
+                torch.inf,
             )  # (chunk_size, n_candidates)
             nearest = torch.topk(
                 dist, k=group_size, dim=1, largest=False
@@ -335,9 +329,9 @@ class BM3D(Denoiser):
             if stage == "hard":
                 threshold = (
                     torch.tensor(
-                        self.hard_threshold, dtype=torch.float32, device=self.device
+                        self.hard_threshold, dtype=torch.float32, device=device
                     )
-                    * sigma_ch[None, None, :, None, None]
+                    * sigma
                 )  # (1, 1, c, 1, 1)
                 mask = (
                     torch.abs(noisy_coeff) >= threshold
@@ -345,7 +339,7 @@ class BM3D(Denoiser):
                 coeff = noisy_coeff * mask
                 denom = torch.maximum(
                     mask.sum(dim=(1, 3, 4)).to(dtype=torch.float32),
-                    torch.tensor(1.0, dtype=torch.float32, device=self.device),
+                    torch.tensor(1.0, dtype=torch.float32, device=device),
                 )  # (chunk_size, group_size)
             # stage 2 (Wiener filtering)
             else:
@@ -361,10 +355,8 @@ class BM3D(Denoiser):
                     pilot_coeff * pilot_coeff
                 )  # (chunk_size, group_size, c, p, p)
                 noise_power = (
-                    torch.tensor(
-                        self.wiener_mu2, dtype=torch.float32, device=self.device
-                    )
-                    * sigma2_ch[None, None, :, None, None]
+                    torch.tensor(self.wiener_mu2, dtype=torch.float32, device=device)
+                    * sigma2
                 )  # (1, 1, c, 1, 1)
                 wiener = pilot_power / (
                     pilot_power + noise_power
@@ -373,53 +365,45 @@ class BM3D(Denoiser):
                 coeff = noisy_coeff * wiener
                 denom = torch.maximum(
                     torch.sum(wiener * wiener, dim=(1, 3, 4)),
-                    torch.tensor(1.0, dtype=torch.float32, device=self.device),
+                    torch.tensor(1.0, dtype=torch.float32, device=device),
                 )  # (chunk_size, group_size)
             # inverse 3d transform
             filtered = self._spatial_inverse(
                 self._group_inverse(coeff, group_inverse), spatial_inverse
             )  # (chunk_size, group_size, c, p, p)
             # compute aggregation weights
-            patch_weight = (1.0 / (sigma2_ch[None, :] * denom))[
+            patch_weight = (1.0 / (sigma2 * denom))[
                 :, None, :, None, None
             ]  # (chunk_size, 1, c, 1, 1)
             # aggregate
             self._aggregate(
-                w, c, accum, weight, group_y, group_x, filtered, patch_weight
+                w, c, accum, weight, group_y, group_x, patch_c, filtered, patch_weight
             )
         # normalize
         out = accum.reshape(h, w, c) / torch.maximum(
             weight.reshape(h, w, c),
-            torch.tensor(EPS, dtype=torch.float32, device=self.device),
+            torch.tensor(EPS, dtype=torch.float32, device=device),
         )  # (h, w, c)
         return out
 
     @torch.inference_mode()
-    def _bm3d_fast(self, y: Tensor, sigma: float | Tensor) -> Tensor:
+    def _bm3d_fast(self, y: Tensor, sigma: float) -> Tensor:
         # y
-        y = torch.as_tensor(y, dtype=torch.float32, device=self.device)  # (h, w, c)
+        device = y.device
         squeeze = y.ndim == 2
         if squeeze:
             y = y[..., None]  # (h, w, c)
         h, w, c = y.shape
         # sigma
-        sigma = torch.as_tensor(sigma, dtype=torch.float32, device=self.device)
-        if sigma.ndim == 0:
-            sigma = torch.full(
-                (c,), float(sigma), dtype=torch.float32, device=self.device
-            )  # (c,)
-        elif sigma.numel() == c:
-            sigma = sigma.reshape(c).to(dtype=torch.float32)  # (c,)
-        if torch.all(sigma <= 0).item():  # edge case: no denoising
-            return y[..., 0] if squeeze else y
-        sigma_ch = torch.as_tensor(
-            sigma, dtype=torch.float32, device=self.device
-        )  # (c,)
-        sigma2_ch = sigma_ch * sigma_ch  # (c,)
+        sigma2 = sigma * sigma
+        # patches
         hp, wp = h - self.p + 1, w - self.p + 1  # hp*wp = total number of patches
-        ref_y, ref_x = self._reference_grid(hp, wp)  # (n_refs,), (n_refs,)
-        off_y, off_x = self._search_offsets()  # (n_candidates,), (n_candidates,)
+        ref_y, ref_x = self._reference_grid(hp, wp, device)  # (n_refs,), (n_refs,)
+        off_y, off_x = self._search_offsets(device)  # (n_candidates,), (n_candidates,)
         n_refs = int(ref_y.numel())  # total number of reference patches
+        patch_c = torch.arange(c, dtype=torch.int64, device=device)[
+            None, None, :, None, None
+        ]  # (1, 1, c, 1, 1)
         # stage 1 (hard-threshold)
         y_patches = self._patch_tensor(y)  # (hp*wp, c, p, p)
         basic = self._estimate(
@@ -431,10 +415,11 @@ class BM3D(Denoiser):
             ref_x,
             off_y,
             off_x,
+            patch_c,
             hp,
             wp,
-            sigma_ch,
-            sigma2_ch,
+            sigma,
+            sigma2,
             y_patches,
             y_patches,
             self.ht_group_size,
@@ -443,6 +428,7 @@ class BM3D(Denoiser):
             self.group_ht,
             self.group_ht_inv,
             "hard",
+            device,
         )  # (h, w, c)
         # stage 2 (Wiener filtering)
         basic_patches = self._patch_tensor(basic)  # (hp*wp, c, p, p)
@@ -455,10 +441,11 @@ class BM3D(Denoiser):
             ref_x,
             off_y,
             off_x,
+            patch_c,
             hp,
             wp,
-            sigma_ch,
-            sigma2_ch,
+            sigma,
+            sigma2,
             basic_patches,
             y_patches,
             self.wiener_group_size,
@@ -467,6 +454,7 @@ class BM3D(Denoiser):
             self.group_wiener,
             self.group_wiener_inv,
             "wiener",
+            device,
         )  # (h, w, c)
         return denoised[..., 0] if squeeze else denoised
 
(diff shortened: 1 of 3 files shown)


kvttt:
This PR is ready to be reviewed. Things I have done so far in this PR:
1. Re-implement BM3D in PyTorch,
2. Drop dependency on `pywt` and switch to `ptwt`,
3. Add and locally run test for `use_legacy=False` (and loosen one of the expected PSNR improvement by 0.05 dB),
4. Add to the warning in the docstring to let the users know the new behavior, and
5. Chore: `black .`, `ruff check`, `make html-fast`, add to `changelog.rst`.

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,407 @@
...
+    transform_name = transform_name.lower()
+    if transform_name == "dct":
+        return get_dct_matrix(n, norm="ortho")
+    elif (n & (n - 1)) == 0:
+        return get_wavelet_matrix_torch(n, transform_name)
+    raise ValueError(f"Unsupported transform '{transform_name}' for size n={n}.")
+
+
+@torch.no_grad()
+def bm3d_fast(
```
mh-nguyen712: Thanks for this very cool implementation. 
I think that we can refactor a bit this function. For example we can define `bm3d_fast` as a private method of the `BM3D` class, and move functions defined inside `bm3d_fast` as methods of the class too, rather than having many functions defined inside a function. So some variables can be initialized once in `__init__` if `use_legacy=False` (eg kaiser window). 

OPTINAL: I'm not sure if it's possible, but a quick look seems that the current implementation uses standard torch functions, so it could be great to implement a batched version of `bm3d_fast`

mh-nguyen712 (commented):
Thanks for the nice addition!

deepinv/models/bm3d.py
kvttt: I agree. I think having all the functions that are currently on the outside become private methods is a great idea. And you are right, a lot of variables can be initialized as soon as `__init__` is called with `use_legacy=False`. This include the Kaiser window, and also the wavelet transformation matrices and their inverses. And this will be shared across all images within the batch instead of having to re-initialized for every call of `bm3d_fast`. 

> OPTINAL: I'm not sure if it's possible, but a quick look seems that the current implementation uses standard torch functions, so it could be great to implement a batched version of `bm3d_fast`

I'm leaning toward not implementing a batched version of `bm3d_fast` although it is entirely possible. Here's my reason:

b

In this figure, I am trying different chunk sizes. A chunk size of 2048 means that 2048 reference patches are processed in parallel. It looks like a chunk size of 1024 is the fastest on my machine for the cat image that I tested. Smaller chunk size is slower because of less parallelization and larger chunk size is also slower because of I'm guessing kernel launch overhead/chunk size exceeds the number of tensor cores? I did not test chunk size beyond 16384 because CUDA OOM. Therefore, I believe `bm3d_fast`'s speed will not benefit from a batched version as long as an appropriate chunk size is chosen. However, the batched version does sound like a more elegant solution.

mh-nguyen712: Thanks for the detailed benchmark! So it should be good to keep the current implementation per image, and a refactoring should be enough.

kvttt:
As discussed, I made all functions private methods of the `BM3D` class. And calling `BM3D.__init__()` now also sets up some constants that are used repeatedly across different `BM3D.forward()` calls. In addition, I improved the docstring to explain the optional `kwargs` for the fast implementation and added typing. `pytest`, `ruff`, `black`, `make html` succeeded locally.

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,496 @@
...
+EPS = 2 ** (-52)
+
 
 class BM3D(Denoiser):
     r"""
     BM3D denoiser.
 
     The BM3D denoiser was introduced by :footcite:t:`dabov2007image`.
 
+    :param bool use_legacy: Whether to use the legacy implementation of BM3D. Default: True
```
mh-nguyen712:
```suggestion
    :param bool use_legacy: Whether to use the legacy implementation of BM3D. Default: `True`
```

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,496 @@
...
+    :param device: Device to run the fast implementation of BM3D on. Default: "cpu"
```
mh-nguyen712:
```suggestion
    :param device: Device to run the fast implementation of BM3D on. Default: `"cpu"`
```

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,496 @@
...
+        - ``ref_stride``: stride for selecting reference patches. Default: 3
+        - ``chunk_size``: number of groups to process in parallel. Default: 2048
+        - ``ht_group_size``: group size for stage 1 (hard-thresholding). Default: 16
+        - ``wiener_group_size``: group size for stage 2 (Wiener filtering). Default: 32
+        - ``spatial_ht_transform``: spatial transform for stage 1. Default: "bior1.5"
+        - ``spatial_wiener_transform``: spatial transform for stage 2. Default: "dct"
+        - ``group_ht_transform``: group transform for stage 1. Default: "haar"
+        - ``group_wiener_transform``: group transform for stage 2. Default: "haar"
+        - ``hard_threshold``: hard-thresholding parameter for stage 1. Default: 3.0
+        - ``wiener_mu2``: Wiener filtering parameter for stage 2. Default: 0.4
```
mh-nguyen712, line 39: We should put string inside `` for better documentation. 
```suggestion

        - ``patch_size``: size of each image patch. Default: 8
        - ``search_radius``: search window radius for block matching. Default: 19
        - ``search_step``: step size for block matching. Default: 1
        - ``ref_stride``: stride for selecting reference patches. Default: 3
        - ``chunk_size``: number of groups to process in parallel. Default: 2048
        - ``ht_group_size``: group size for stage 1 (hard-thresholding). Default: 16
        - ``wiener_group_size``: group size for stage 2 (Wiener filtering). Default: 32
        - ``spatial_ht_transform``: spatial transform for stage 1. Default: `"bior1.5"`
        - ``spatial_wiener_transform``: spatial transform for stage 2. Default: `"dct"`
        - ``group_ht_transform``: group transform for stage 1. Default: `"haar"`
        - ``group_wiener_transform``: group transform for stage 2. Default: `"haar"`
        - ``hard_threshold``: hard-thresholding parameter for stage 1. Default: 3.0
        - ``wiener_mu2``: Wiener filtering parameter for stage 2. Default: 0.4
```

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,496 @@
...
+        However, it may produce slightly different results than the legacy implementation.
     """
 
-    def forward(self, x: Tensor, sigma: float | Tensor, **kwargs) -> Tensor:
+    def __init__(
+        self, use_legacy: bool = True, device: torch.device | str = "cpu", **kwargs
+    ):
+        super(BM3D, self).__init__()
+        self.use_legacy = use_legacy
+        self.device = device
```
mh-nguyen712: In our library, we're moving towards drop the `self.device` attribute, and using `.to(device)` instead. So the code should not call `self.device` ideally

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,496 @@
...
+            )  # (p,)
+            self.win = win_1d[:, None] * win_1d[None, :]  # (p, p)
+            # helper arrays for aggregation
+            self.patch_y = torch.arange(self.p, dtype=torch.int64, device=self.device)[
+                None, None, None, :, None
+            ]  # (1, 1, 1, p, 1)
+            self.patch_x = torch.arange(self.p, dtype=torch.int64, device=self.device)[
+                None, None, None, None, :
+            ]  # (1, 1, 1, 1, p)
+            self.patch_c = torch.arange(3, dtype=torch.int64, device=self.device)[
```
mh-nguyen712: For device compatibility, tensor attribute should be registered as buffer, by using self.register_buffer(name, tensor) so that `.to(device, dtype)` works

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,496 @@
...
+                periodic=False,
+                device=self.device,
+                dtype=torch.float32,
+            )  # (p,)
+            self.win = win_1d[:, None] * win_1d[None, :]  # (p, p)
+            # helper arrays for aggregation
+            self.patch_y = torch.arange(self.p, dtype=torch.int64, device=self.device)[
+                None, None, None, :, None
+            ]  # (1, 1, 1, p, 1)
+            self.patch_x = torch.arange(self.p, dtype=torch.int64, device=self.device)[
```
mh-nguyen712:
```suggestion
            self.register_buffer(patch_x, torch.arange(self.p, dtype=torch.int64, device=self.device))
```

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,496 @@
...
+                None, None, None, None, :
+            ]  # (1, 1, 1, 1, p)
+            self.patch_c = torch.arange(3, dtype=torch.int64, device=self.device)[
+                None, None, :, None, None
+            ]  # (1, 1, c, 1, 1)
+            # other configs
+            self.hard_threshold = float(kwargs.get("hard_threshold", 3.0))
+            self.wiener_mu2 = float(kwargs.get("wiener_mu2", 0.4))
+
```
mh-nguyen712, line 141: Then at the end we can add safeguard: `self.to(device)`

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,496 @@
...
+            from scipy.fft import dct
+        except ImportError:  # pragma: no cover
+            raise ImportError(
+                "scipy package not found. Please install it with `pip install scipy`."
+            )
+        matrix = dct(np.eye(n, dtype=np.float32), norm=norm, axis=0)
+        inverse = matrix.T if norm == "ortho" else np.linalg.inv(matrix)
+        return torch.as_tensor(matrix, dtype=torch.float32), torch.as_tensor(
+            inverse, dtype=torch.float32
+        )
```
mh-nguyen712:
```suggestion
        return torch.from_numpy(matrix, dtype=torch.float32), torch.from_numpy(
            inverse, dtype=torch.float32
        )
```

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,496 @@
...
+    ) -> tuple[Tensor, Tensor]:
+        n = int(n)
+        transform_name = transform_name.lower()
+        if transform_name == "dct":
+            return self._get_dct_matrix(n, norm="ortho")
+        elif (n & (n - 1)) == 0:
+            return self._get_wavelet_matrix_torch(n, transform_name)
+        raise ValueError(f"Unsupported transform '{transform_name}' for size n={n}.")
+
+    def _transform_matrices(
```
mh-nguyen712: I think it's cleaner if we add `device` as an argument to `_get_trainsform_matrices` directly

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,496 @@
...
+    def _patch_tensor(
+        self,
+        img: Tensor,
+    ) -> Tensor:  # extract all `hp*wp` patches
+        patches = img.unfold(0, self.p, 1).unfold(
+            1, self.p, 1
+        )  # Result: (hp, wp, c, p, p)
+        return patches.contiguous().reshape(
+            -1, img.shape[2], self.p, self.p
+        )  # (hp*wp, c, p, p)
```
mh-nguyen712, line 200: There is `deepinv.utils.image_to_patches` which does the similar thing (need to check the shape). I think it's cool to re-use the function

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,496 @@
...
+            cand_idx = cand_y * wp + cand_x  # (chunk_size, n_candidates)
+            # block matching
+            ref = match_metric[ref_idx]  # (chunk_size, p*p)
+            cand = match_metric[cand_idx]  # (chunk_size, n_candidates, p*p)
+            dist = (
+                torch.sum((cand - ref[:, None, :]) ** 2, dim=2) / patch_norm
+            )  # (chunk_size, n_candidates)
+            dist = torch.where(
+                valid, dist, torch.full_like(dist, torch.inf)
+            )  # (chunk_size, n_candidates)
```
mh-nguyen712, line 336: we can avoid allocating a tensor
```suggestion
            dist = torch.where(
                valid, dist, float('int')
            )  # (chunk_size, n_candidates)
```

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,496 @@
...
+            noisy_coeff = self._group_forward(
+                self._spatial_forward(noisy_group, spatial_transform), group_transform
+            )  # (chunk_size, group_size, c, p, p)
+            # stage 1 (hard-threshold)
+            if stage == "hard":
+                threshold = (
+                    torch.tensor(
+                        self.hard_threshold, dtype=torch.float32, device=self.device
+                    )
+                    * sigma_ch[None, None, :, None, None]
```
mh-nguyen712: Here it's a scalar, right? (since you called `.item()` in `forward`)

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,496 @@
...
+            sigma = torch.full(
+                (c,), float(sigma), dtype=torch.float32, device=self.device
+            )  # (c,)
+        elif sigma.numel() == c:
+            sigma = sigma.reshape(c).to(dtype=torch.float32)  # (c,)
+        if torch.all(sigma <= 0).item():  # edge case: no denoising
+            return y[..., 0] if squeeze else y
+        sigma_ch = torch.as_tensor(
+            sigma, dtype=torch.float32, device=self.device
+        )  # (c,)
```
mh-nguyen712: similar to previous comment, this is not necessary, since we have the same sigma for all channels normally

mh-nguyen712 (changes requested):
Thanks for the updates! Here is my first batch of comments. 
The main issue is the current implementation uses `self.device` in many place, which is not recommended, since it's not compatible with the standard `.to(device)` (see my comments).

deepinv/models/bm3d.py
kvttt: Just so I understand you correctly: the preferred way is to do
```
bm3d = deepinv.models.BM3D().to("cuda")
```
rather than
```
bm3d = deepinv.models.BM3D(device="cuda")
```
Is this correct?

If this is the case, do we assume the input tensors passed to `forward()` are already on that device? And we don't check or silently move input tensors to that device?

deepinv/models/bm3d.py
```diff
@@ -1,43 +1,496 @@
...
+        return matrix, torch.linalg.inv(matrix)
+
+    def _get_dct_matrix(self, n: int, norm: str = "ortho") -> tuple[Tensor, Tensor]:
+        try:
+            from scipy.fft import dct
+        except ImportError:  # pragma: no cover
+            raise ImportError(
+                "scipy package not found. Please install it with `pip install scipy`."
+            )
+        matrix = dct(np.eye(n, dtype=np.float32), norm=norm, axis=0)
```
kvttt: By the way, should I use `deepinv.physics.functional.dct` here instead of `scipy.fft.dct`?

deepinv/models/bm3d.py
kvttt: I agree. I think maybe I can also combine `_get_transform_matrices` and `_transform_matrices` into one and then add `device` as an argument.

deepinv/models/bm3d.py
kvttt: Yes. I think the original behavior of `forward()` assumes the same scalar `sigma` for all channels.

kvttt: I will try to simplify `_estimate()` based on the fact that sigma is a scalar. Things that I can think of are: changing the type of `sigma_ch` and `sigma2_ch` here from `float | Tensor` to `float`, remove unnecessary broadcasting, remove unnecessary `torch.as_tensor()`, etc.

deepinv/models/bm3d.py
mh-nguyen712: The standard usage would be: 
```python
bm3d = deepinv.models.BM3D(device="cuda")
``` 
or simply
```python
bm3d = deepinv.models.BM3D()
```

but we should be able to call 
```python
bm3d.to("cuda")
```

That's why having `self.device` is not a good idea. And when a tensor attribute is registered as buffer, torch will handle the device transfer properly

mh-nguyen712: Just like standard `nn.Module`

deepinv/models/bm3d.py
mh-nguyen712: Yes I think it's better!

deepinv/models/bm3d.py
kvttt: So we still want `__init__()` to take `device` as an argument, but we don't want to have an attribute `self.device` right?

But I still want to ask what the standard/preferred way to handle the device of an input tensor is. Do I need to explicitly check the input tensor is on the same device as the model (and move to that device if necessary)? I guess this is not possible since there is no `self.device` to check against? Or is the preferred way to just let `torch` throw its own error?

kvttt:
For future reference:

```
matrix_scipy = scipy.fft.dct(np.eye(n, dtype=np.float32), norm=norm, axis=0)
```
and 
```
matrix_deepinv = deepinv.physics.functional.dct(torch.eye(n, dtype=torch.float32), norm=norm)
```
do not give the same matrix. They differ by a transpose since `deepinv.physics.functional.dct` is hardcoded to perform DCT along the last axis.

deepinv/models/bm3d.py
```diff
@@ -11,33 +15,470 @@ class BM3D(Denoiser):
...
+        return denoised[..., 0] if squeeze else denoised
+
+    def forward(self, x: Tensor, sigma: float | Tensor) -> Tensor:
+        if self.use_legacy:
+            try:
+                import bm3d
+            except ImportError:  # pragma: no cover
+                raise ImportError(
+                    "bm3d package not found. Please install it with `pip install bm3d`."
+                )
```
mh-nguyen712: maybe move this to __init__

deepinv/models/bm3d.py
```diff
@@ -11,33 +15,470 @@ class BM3D(Denoiser):
...
-            )
+            if self.use_legacy:
+                out[i, :, :, :] = array2tensor(
+                    bm3d.bm3d(tensor2array(x[i, :, :, :]), sigma[i].item())
+                )
+            else:
+                out[i, :, :, :] = self._bm3d_fast(
+                    x[i, :, :, :].permute(1, 2, 0),
+                    sigma[i].item(),
+                ).permute(2, 0, 1)
```
mh-nguyen712: Call a `permute()` on x and another after processing the whole batch (to prevent python overhead when calling too many torch functions)

deepinv/models/bm3d.py
```diff
@@ -11,33 +15,470 @@ class BM3D(Denoiser):
...
+        )  # (h, w, c)
+        return out
+
+    @torch.inference_mode()
+    def _bm3d_fast(self, y: Tensor, sigma: float) -> Tensor:
+        # y
+        device = y.device
+        squeeze = y.ndim == 2
+        if squeeze:
+            y = y[..., None]  # (h, w, c)
```
mh-nguyen712: I think we can skip this check, the input of `forward` should always (B, C, H, W)

deepinv/models/bm3d.py
```diff
@@ -11,33 +15,470 @@ class BM3D(Denoiser):
...
+        spatial_inverse: Tensor,
+        group_transform: Tensor,
+        group_inverse: Tensor,
+        stage: str,
+        device: torch.device,
+    ) -> Tensor:
+        # patch features
+        match_metric = match_patches[:, 0, :, :].reshape(
+            -1, self.p * self.p
+        )  # (hp*wp, p*p)
```
mh-nguyen712, line 297: Why use only the first channel here? Is it the same in the original implementation?

deepinv/models/bm3d.py
```diff
@@ -11,33 +15,470 @@ class BM3D(Denoiser):
...
+            )  # (chunk_size, n_candidates)
+            cand_y = torch.clamp(cand_y_raw, 0, hp - 1)  # (chunk_size, n_candidates)
+            cand_x = torch.clamp(cand_x_raw, 0, wp - 1)  # (chunk_size, n_candidates)
+            cand_idx = cand_y * wp + cand_x  # (chunk_size, n_candidates)
+            # block matching
+            ref = match_metric[ref_idx]  # (chunk_size, p*p)
+            cand = match_metric[cand_idx]  # (chunk_size, n_candidates, p*p)
+            dist = (
+                torch.sum((cand - ref[:, None, :]) ** 2, dim=2) / patch_norm
+            )  # (chunk_size, n_candidates)
```
mh-nguyen712, line 331: I was wandering if a torch.cdist can be faster here? (and especially memory friendly)

deepinv/models/bm3d.py
```diff
@@ -11,33 +15,470 @@ class BM3D(Denoiser):
...
+            if stage == "hard":
+                threshold = (
+                    torch.tensor(
+                        self.hard_threshold, dtype=torch.float32, device=device
+                    )
+                    * sigma
+                )  # (1, 1, c, 1, 1)
+                mask = (
+                    torch.abs(noisy_coeff) >= threshold
+                )  # (chunk_size, group_size, c, p, p)
```
mh-nguyen712:
```suggestion
                mask = (
                    torch.abs(noisy_coeff) >= self.hard_threshold * sigma
                )  # (chunk_size, group_size, c, p, p)
```

deepinv/models/bm3d.py
```diff
@@ -11,33 +15,470 @@ class BM3D(Denoiser):
...
+                coeff = noisy_coeff * mask
+                denom = torch.maximum(
+                    mask.sum(dim=(1, 3, 4)).to(dtype=torch.float32),
+                    torch.tensor(1.0, dtype=torch.float32, device=device),
+                )  # (chunk_size, group_size)
```
mh-nguyen712:
```suggestion
                denom = torch.clamp_min(
                    mask.sum(dim=(1, 3, 4)).to(dtype=torch.float32),
                    1,
                )  # (chunk_size, group_size)
```

deepinv/models/bm3d.py
```diff
@@ -11,33 +15,470 @@ class BM3D(Denoiser):
...
+                noise_power = (
+                    torch.tensor(self.wiener_mu2, dtype=torch.float32, device=device)
+                    * sigma2
+                )  # (1, 1, c, 1, 1)
+                wiener = pilot_power / (
+                    pilot_power + noise_power
+                )  # (chunk_size, group_size, c, p, p)
+                # apply gain
+                coeff = noisy_coeff * wiener
+                denom = torch.maximum(
```
mh-nguyen712: same here clamp_min

deepinv/models/bm3d.py
```diff
@@ -11,33 +15,470 @@ class BM3D(Denoiser):
...
+            # compute aggregation weights
+            patch_weight = (1.0 / (sigma2 * denom))[
+                :, None, :, None, None
+            ]  # (chunk_size, 1, c, 1, 1)
+            # aggregate
+            self._aggregate(
+                w, c, accum, weight, group_y, group_x, patch_c, filtered, patch_weight
+            )
+        # normalize
+        out = accum.reshape(h, w, c) / torch.maximum(
```
mh-nguyen712: use clamp_min instead

deepinv/models/bm3d.py
```diff
@@ -3,6 +3,10 @@
 from torch import Tensor
 from .base import Denoiser
 from .utils import array2tensor, tensor2array
+from deepinv.physics.functional import dct
+from deepinv.utils import image_to_patches
+
+EPS = 2 ** (-52)
```
mh-nguyen712: use dtype-awave instead of a fixed value: `torch.finfo(dtype).eps`

deepinv/models/bm3d.py
```diff
@@ -11,33 +15,470 @@ class BM3D(Denoiser):
...
+            )  # truncate output to match the behavior of `pywt.wavedec` when `model='periodization'`
+        matrix = matrix / torch.linalg.norm(matrix, dim=0)
+        return matrix, torch.linalg.inv(matrix)
+
+    def _get_dct_matrix(self, n: int, norm: str = "ortho") -> tuple[Tensor, Tensor]:
+        matrix = dct(torch.eye(n, dtype=torch.float32), norm=norm).T
+        inverse = matrix.T if norm == "ortho" else torch.linalg.inv(matrix)
+        return matrix, inverse
+
+    def _transform_matrices(self, n: int, transform_name: str) -> tuple[Tensor, Tensor]:
```
mh-nguyen712: we can pass `device` argument here the construct the tensors directly on GPU if specified

deepinv/models/bm3d.py
```diff
@@ -11,33 +15,470 @@ class BM3D(Denoiser):
...
+        # y
+        device = y.device
+        squeeze = y.ndim == 2
+        if squeeze:
+            y = y[..., None]  # (h, w, c)
+        h, w, c = y.shape
+        # sigma
+        sigma2 = sigma * sigma
+        # patches
+        hp, wp = h - self.p + 1, w - self.p + 1  # hp*wp = total number of patches
```
mh-nguyen712, line 415: Maybe add a check that h and w should be larger than patch size and raise value errors if not

deepinv/models/bm3d.py
kvttt: Yes. In Section V. Extension to color-image denoising, there is this line.

image

For color-image denoising, the authors propose to first convert RGB to YCbCr, then use the first channel, i.e., Y, for block matching.

deepinv/models/bm3d.py
```diff
@@ -11,33 +15,470 @@ class BM3D(Denoiser):
...
+        group_transform: Tensor,
+        group_inverse: Tensor,
+        stage: str,
+        device: torch.device,
+    ) -> Tensor:
+        # patch features
+        match_metric = match_patches[:, 0, :, :].reshape(
+            -1, self.p * self.p
+        )  # (hp*wp, p*p)
+        patch_norm = torch.tensor(self.p * self.p, dtype=torch.float32, device=device)
```
mh-nguyen712: I think we can use a python scalar here, which makes the code much more simple (since we have for loop and torch tracing or jit is not very interesting here)

mh-nguyen712: and in some other place

deepinv/models/bm3d.py
mh-nguyen712: Thanks for the clarification! So we should convert to YCbCr first, right?

mh-nguyen712 (changes requested):
Thanks a lot for the updates! Some more comments and should be good!

deepinv/models/bm3d.py
kvttt: Moving this to `__init__()` naively gives `NameError: name 'bm3d' is not defined`. Therefore, I will try to move this to `__init__()` and add `self.bm3d_lib` as an attribute.

deepinv/models/bm3d.py
kvttt: I think this is something worth discussing. I will post a comment on the outside.

kvttt:
There is a design choice that is worth discussing, i.e., regarding how block matching is performed for multi-channel images. 

There are mainly three use cases for BM3D.

1. For grayscale images, there is no ambiguity. Group matching is performed using the only channel available.
2. For multi-channel images in general, `bm3d.bm3d` works, except that block matching is performed on the first channel.
3. For RGB images, the original implementation claims the following. For `bm3d.bm3d`:
```
Help on function bm3d in module bm3d:

bm3d(z: numpy.ndarray, sigma_psd: Union[numpy.ndarray, list, float], profile: Union[bm3d.profiles.BM3DProfile, str] = 'np', stage_arg: Union[bm3d.profiles.BM3DStages, numpy.ndarray] = <BM3DStages.ALL_STAGES: 3>, blockmatches: tuple = (False, False)) -> Union[numpy.ndarray, Tuple[numpy.ndarray, Tuple[numpy.ndarray, numpy.ndarray]]]
    Perform BM3D denoising on z: either hard-thresholding, Wiener filtering or both.

    :param z: Noisy image. either MxN or MxNxC where C is the channel count.
              For multichannel images, blockmatching is performed on the first channel.
    ...
```
The last line here says "blockmatching is performed on the first channel". The is the default behavior of `deepinv.models.BM3D` before this PR: block matching is performed on the first channel, i.e., the red channel when input is RGB image.

Alternatively, `bm3d` provides `bm3d.bm3d_rgb`:
```
Help on function bm3d_rgb in module bm3d:

bm3d_rgb(z: numpy.ndarray, sigma_psd: Union[numpy.ndarray, list, float], profile: Union[bm3d.profiles.BM3DProfile, str] = 'np', colorspace: str = 'opp') -> numpy.ndarray
    BM3D For color images. Performs color transform to do block-matching in luminance domain.
    :param z: Noisy image, 3 channels (MxNx3)
    :param sigma_psd: Noise PSD, either MxN or MxNx3 (different PSDs for different channels)
                        or
                      Noise standard deviation, either float, or [float, float, float] for 3 different stds.
    :param profile: Settings for BM3D: BM3DProfile object or a string.
                    ('np', 'refilter', 'vn', 'vn_old', 'high', 'deb')
    :param colorspace: 'YCbCr' or 'opp' for choosing the color transform
    :return: denoised color image, same size as z
```
There is a kwarg named `colorspace`. The original paper by Dabov et al. recommends transforming from RGB to YCbCr for color image denoising. However, the default behavior of `bm3d.bm3d_rgb` is to transform to `'opp'`.

Do we want to switch from `bm3d.bm3d` to `bm3d.bm3d_rgb` for color images? 

I want to get people's opinion on the following two options.

1. Using a combination of both `bm3d.bm3d` and `bm3d.bm3d_rgb`. Maybe we need to provide the users with the option to specify something like `rgb=True` and `colorspace='opp'` as kwargs. Then, there are three situations:
    1. If `rgb == True` and number of channels is not 3, raise error.
    2. If `rgb == True` and number of channels is 3, call `bm3d.bm3d_rgb` with the user-specified `colorspace`.
    3. If `rgb == False`, then we are dealing with either grayscale images or general (non-RGB) multi-channel images. Call `bm3d.bm3d` with defaults.
For the customized re-implementation of this PR. The only component that needs to be added is the transformation between RGB and other color spaces. `bm3d` has its own customized function for this purpose. The easiest way is to reproduce it in `torch`. But we need to make sure the default kwargs give the old behavior.
2. Alternatively, we can leave everything as it is right now. Performing block matching on the red channel works too (as I have demonstrated earlier using the cat image). 

Edit: I'm just gonna tag some of y'all to get more feedback/ideas. :) @mh-nguyen712 @jscanvic @romainvo @Tmodrzyk @matthieutrs

mh-nguyen712:
> There is a design choice that is worth discussing, i.e., regarding how block matching is performed for multi-channel images.
> 
> There are mainly three use cases for BM3D.
> 
> 1. For grayscale images, there is no ambiguity. Group matching is performed using the only channel available.
> 2. For multi-channel images in general, `bm3d.bm3d` works, except that block matching is performed on the first channel.
> 3. For RGB images, the original implementation claims the following. For `bm3d.bm3d`:
> 
> ```
> Help on function bm3d in module bm3d:
> 
> bm3d(z: numpy.ndarray, sigma_psd: Union[numpy.ndarray, list, float], profile: Union[bm3d.profiles.BM3DProfile, str] = 'np', stage_arg: Union[bm3d.profiles.BM3DStages, numpy.ndarray] = <BM3DStages.ALL_STAGES: 3>, blockmatches: tuple = (False, False)) -> Union[numpy.ndarray, Tuple[numpy.ndarray, Tuple[numpy.ndarray, numpy.ndarray]]]
>     Perform BM3D denoising on z: either hard-thresholding, Wiener filtering or both.
> 
>     :param z: Noisy image. either MxN or MxNxC where C is the channel count.
>               For multichannel images, blockmatching is performed on the first channel.
>     ...
> ```
> 
> The last line here says "blockmatching is performed on the first channel". The is the default behavior of `deepinv.models.BM3D` before this PR: block matching is performed on the first channel, i.e., the red channel when input is RGB image.
> 
> Alternatively, `bm3d` provides `bm3d.bm3d_rgb`:
> 
> ```
> Help on function bm3d_rgb in module bm3d:
> 
> bm3d_rgb(z: numpy.ndarray, sigma_psd: Union[numpy.ndarray, list, float], profile: Union[bm3d.profiles.BM3DProfile, str] = 'np', colorspace: str = 'opp') -> numpy.ndarray
>     BM3D For color images. Performs color transform to do block-matching in luminance domain.
>     :param z: Noisy image, 3 channels (MxNx3)
>     :param sigma_psd: Noise PSD, either MxN or MxNx3 (different PSDs for different channels)
>                         or
>                       Noise standard deviation, either float, or [float, float, float] for 3 different stds.
>     :param profile: Settings for BM3D: BM3DProfile object or a string.
>                     ('np', 'refilter', 'vn', 'vn_old', 'high', 'deb')
>     :param colorspace: 'YCbCr' or 'opp' for choosing the color transform
>     :return: denoised color image, same size as z
> ```
> 
> There is a kwarg named `colorspace`. The original paper by Dabov et al. recommends transforming from RGB to YCbCr for color image denoising. However, the default behavior of `bm3d.bm3d_rgb` is to transform to `'opp'`.
> 
> Do we want to switch from `bm3d.bm3d` to `bm3d.bm3d_rgb` for color images?
> 
> I want to get people's opinion on the following two options.
> 
> 1. Using a combination of both `bm3d.bm3d` and `bm3d.bm3d_rgb`. Maybe we need to provide the users with the option to specify something like `rgb=True` and `colorspace='opp'` as kwargs. Then, there are three situations:
>    
>    1. If `rgb == True` and number of channels is not 3, raise error.
>    2. If `rgb == True` and number of channels is 3, call `bm3d.bm3d_rgb` with the user-specified `colorspace`.
>    3. If `rgb == False`, then we are dealing with either grayscale images or general (non-RGB) multi-channel images. Call `bm3d.bm3d` with defaults.
>       For the customized re-implementation of this PR. The only component that needs to be added is the transformation between RGB and other color spaces. `bm3d` has its own customized function for this purpose. The easiest way is to reproduce it in `torch`. But we need to make sure the default kwargs give the old behavior.
> 2. Alternatively, we can leave everything as it is right now. Performing block matching on the red channel works too (as I have demonstrated earlier using the cat image).
> 
> Edit: I'm just gonna tag some of y'all to get more feedback/ideas. :) @mh-nguyen712 @jscanvic @romainvo @Tmodrzyk @matthieutrs

Many thanks for these detailed notes!
I think in this PR we're happy with block matching with only the red color! But having full options as in the orignial bm3d is a good update in the future! (But if you can and you want, feel free to add those options here)

deepinv/models/bm3d.py
```diff
@@ -11,33 +13,485 @@ class BM3D(Denoiser):
...
+        n_refs: int,
+        ref_y: Tensor,
+        ref_x: Tensor,
+        off_y: Tensor,
+        off_x: Tensor,
+        patch_c: Tensor,
+        hp: int,
+        wp: int,
+        sigma: float,
+        sigma2: float,
```
mh-nguyen712: remove this argument and compute sigma2 from sigma inside the function

kvttt:
> (But if you can and you want, feel free to add those options here)

I will save this for a future PR because I think it makes more sense to keep the original behavior.

deepinv/models/bm3d.py
```diff
@@ -11,33 +13,480 @@ class BM3D(Denoiser):
...
+            # forward 3d transform
+            noisy_coeff = self._group_forward(
+                self._spatial_forward(noisy_group, spatial_transform), group_transform
+            )  # (chunk_size, group_size, c, p, p)
+            # stage 1 (hard-threshold)
+            if stage == "hard":
+                mask = (
+                    torch.abs(noisy_coeff) >= self.hard_threshold * sigma
+                )  # (chunk_size, group_size, c, p, p)
+                coeff = noisy_coeff * mask
```
mh-nguyen712, line 359: Sorry for the late comment
```suggestion
                coeff = torch.nn.functional.hardshrink(coeff, self.hard_threshold * sigma)
```

deepinv/tests/test_models.py
```diff
@@ -1333,7 +1333,8 @@ def test_denoiser_perf(device, load_example_image):
 
     # Classical denoisers
     classical_denoisers = [
-        (dinv.models.BM3D(), (2.75, 7.5, 6.5)),
+        (dinv.models.BM3D(use_legacy=True), (2.7, 7.5, 6.5)),
```
mh-nguyen712: can you please also add a small test for output between `use_legacy=True` and `use_legacy=False`, possibly with large tolerence if not passing?

deepinv/models/bm3d.py
```diff
@@ -11,33 +13,480 @@ class BM3D(Denoiser):
...
+                w, c, accum, weight, group_y, group_x, patch_c, filtered, patch_weight
+            )
+        # normalize
+        out = accum.reshape(h, w, c) / torch.clamp_min(
+            weight.reshape(h, w, c),
+            torch.finfo(torch.float32).eps,
+        )  # (h, w, c)
+        return out
+
+    @torch.inference_mode()
```
mh-nguyen712: I'm not sure if we want to put inference_mode here, although there is no model weights, someone might want to differentiate w.r.t the input x. 
I would remove it, and leave user the choice!

deepinv/models/bm3d.py
kvttt: `mask` is necessary here for computing `denom`, which determines the `patch_weight` used for aggregation across patches later. Therefore, I think we can keep the original version.

deepinv/tests/test_models.py
kvttt: Will do. But I think there are a lot of implementation details that could potentially make `use_legacy=True` and `use_legacy=False` not match exactly with each other. In my local test, I have a use a large `atol=1e-1` to pass.

However, the lines you highlighted here test the PSNR for each method, which I think can be used to argue that since both legacy and new implementations give good performance in terms of PSNR, we should be good.

deepinv/models/bm3d.py
kvttt: I originally put `@torch.no_grad()` and later `@torch.inference_mode()` here for speed. I will do a quick check locally to see if there is a difference in terms of speed.

kvttt: No difference in speed. Without `@torch.inference_mode()`: 64.94s. With `@torch.inference_mode()`: 63.31s.

kvttt:
The no optional dependencies test fails with the following error:
```
FAILED deepinv/tests/test_models.py::test_bm3d_consistency@main - ImportError: bm3d package not found. Please install it with `pip install bm3d`.
== 1 failed, 20823 passed, 942 skipped, 18347 warnings in 3358.68s (0:55:58) ===
```

I probably forgot to declare somewhere that this test depends on `bm3d` and needs to be skipped when no optional dependencies test is run. However, I'm not familiar with CI/CD. Is there anything I need to add in the test?

jscanvic:
Thanks for this PR! On my phone right now but if you look up "pytest.importorskip" in the tests you should find examples you can adapt to make CI pass

deepinv/tests/test_models.py
```diff
@@ -350,6 +350,31 @@ def test_TV_models_identity():
     assert torch.allclose(x, x_hat, atol=1e-5)  # The model should be the identity
 
 
+def test_bm3d_consistency(load_example_image):
+    # Load 2 example images
```
mh-nguyen712, line 365: You can add this: 
```
    try:
        import bm3d, ptwt, pywt  # noqa: F401
    except ImportError:
        pytest.skip("bm3d, ptwt, pywt are required for this test.")

kvttt:
> ❌ GPU tests failed
> 
> 🔗 [View run details](https://github.com/deepinv/deepinv/actions/runs/26683236604)

The fix in the previous commit works. The only error left is not related to BM3D.

mh-nguyen712 (approved):
LGTM! Many thanks again for this addition @kvttt !

PR #1207, opened 2026-05-28
Add Distributional Consistency Loss

GeorgeWebber:
At the request of @nucli-vicky and @jscanvic (who I had the pleasure of meeting in Rio), I'm submitting a PR to add Distributional Consistency loss to DeepInv.

DC loss is a new loss term that I proposed in a recent paper for solving inverse problems with many independent measurements and known noise models. It may be of interest to the community to explore its use further and in other scenarios to those shown in the paper (https://arxiv.org/abs/2510.13972).

This is my first time contributing to DeepInv; I hope I've followed the guidelines!

### Checks

- [ X ] `python3 -m pytest deepinv/tests` runs successfully.
- [ X ] `black .` and `ruff check .` run successfully.
- [ X ] `make html` runs successfully (in the `docs/` directory).
- [ X ] Updated docstrings related to the changes (as applicable).
- [ X ] Added an entry to the [changelog.rst](https://github.com/deepinv/deepinv/blob/main/docs/source/changelog.rst).


deepinv/loss/dc.py
```diff
@@ -0,0 +1,311 @@
...
+        if self.distribution == "poisson":
+            return self._poisson_logit_pit(q, y)
+        return self._clipped_gaussian_logit_pit(q, y, sigma)
+
+    def forward(
+        self,
+        y: Tensor,
+        x_net: Tensor,
+        physics,
+        model=None,
```
nucli-vicky: I would type physics and model too

docs/source/changelog.rst
```diff
@@ -8,6 +8,7 @@ Current
 
 New Features
 ^^^^^^^^^^^^
+- Add :class:`deepinv.loss.DCLoss`, a distributional consistency loss for Gaussian, clipped-Gaussian, and Poisson noise with deterministic quantile sampling
```
nucli-vicky: After the statement, the PR number contributor name are typically added,
 e.g.:

- Add :class:`deepinv.loss.DCLoss`, a distributional consistency loss for Gaussian, clipped-Gaussian, and Poisson noise with deterministic quantile sampling (:gh:`1207` by `George Webber`_)

examples/self-supervised-learning/demo_dc_loss.py
```diff
@@ -0,0 +1,155 @@
...
+            history["image_mse"].append(image_mse(x_gt, x_est).item())
+            history["measurement_mse"].append(F.mse_loss(q_est, y).item())
+            history["dc_loss"].append(dc_loss(y=y, x_net=x_est, physics=physics).item())
+
+    x_est = x_est.detach()
+    return physics.A(x_est), x_est, history
+
+
+torch.manual_seed(0)
+device = dinv.utils.get_device()
```
nucli-vicky, line 78: How long would this run on a CPU? Might consider making this a GPU-only doc, incase the CPU version would require too much time. 

This is done by adding it to `gpu_dependent_files = [".*demo_astra_tomography.py", ".*demo_custom_niqe.py"]` at `docs/source/conf.py`

If it doesn't take that long, making it run on CPU is fine ofcourse :)

deepinv/tests/test_loss.py
```diff
@@ -42,6 +42,54 @@
...
+
+def create_clipped_gaussian_dc_problem(device, sigma=0.05, size=16):
+    coords = torch.linspace(0.0, 1.0, size, device=device)
+    yy, xx = torch.meshgrid(coords, coords, indexing="ij")
+    x = (0.1 + 0.8 * (xx + yy) / 2.0).unsqueeze(0).unsqueeze(0)
+    y = torch.clamp(x + sigma * torch.randn_like(x), 0.0, 1.0)
+    y[..., 0, 0] = 0.0
+    y[..., -1, -1] = 1.0
+    physics = dinv.physics.Denoising(device=device)
+    return x, y, physics
```
nucli-vicky, line 99: I would set seeding here, to ensure CI doesnt break too often due to unlucky random seed. Especially when the `Physics` is used throughout a lot of iterations, it might be worth it to pass a deterministic Generator to the noise object.

deepinv/loss/dc.py
```diff
@@ -0,0 +1,311 @@
...
+        self.n_points = n_points
+        self.eps = eps
+        self.boundary_eps = boundary_eps
+
+        if self.n_points is not None and self.n_points <= 0:
+            raise ValueError("n_points must be strictly positive when provided.")
+        if not 0 < self.eps < 0.5:
+            raise ValueError("eps must lie strictly between 0 and 0.5.")
+        if not 0 < self.boundary_eps < 0.5:
+            raise ValueError("boundary_eps must lie strictly between 0 and 0.5.")
```
nucli-vicky, line 108: `deepinv` measures code coverage, meaning that any code path not executed during tests gets flagged as 'untested' and decreases our coverage.

Best practice when code contains error raises on wrong inputs is to test that these get triggered and raise the correct error message on the wrong inputs.

Another option is to add `# pragma: no cover` , e.g.

```python
if not 0 < self.eps < 0.5: # pragma: no cover
```

to communicate to coverage that this can be excluded from the %.

In principle, this should be used sparingly, but otherwise tests might get a bit bloated, see [my comment here](https://github.com/deepinv/deepinv/pull/1132#issuecomment-4298042899).

nucli-vicky (commented):
Thanks a lot for this contribution @GeorgeWebber ! Some small initial comments from my side

deepinv/loss/dc.py
GeorgeWebber: Resolved by writing additional tests

examples/self-supervised-learning/demo_dc_loss.py
GeorgeWebber: It takes under a minute on the CPU :)

deepinv/tests/test_loss.py
GeorgeWebber: Added in a generator to fix the seed

GeorgeWebber (commented):
Thanks for the review comments @nucli-vicky! I've addressed each one now, so I think the PR is ready? Over to you :)

docs/source/refs.bib
```diff
@@ -1341,6 +1341,14 @@ @article{lyu2023m4raw
...
 }
 
+@inproceedings{webber2026distributional,
+  title={Distributional Consistency Loss: Beyond Pointwise Data Terms in Inverse Problems},
+  author={Webber, George and Reader, Andrew J.},
+  booktitle={International Conference on Learning Representations},
+  year={2026},
+  eprint={2510.13972},
+  archivePrefix={arXiv},
+  doi={10.48550/arXiv.2510.13972}
```
jscanvic, line 1351:
```suggestion
  doi={10.48550/arXiv.2510.13972}}
```

jscanvic (changes requested):
Thank you for taking the time to contribute your loss George! And sorry for coming back to you so late, this summer's been busy for me

Some comments:
- New losses should generally be added to `LOSSES` (or one of the 2 other lists below it) in `test_loss.py` and implemented in `choose_loss`. Also, we generally try to avoid having more than one new test function for new losses. I'd recommend you 1) merge all of your test functions into a large, single one, 2) that you skim out less important checks / cases to make it as minimal as possible.
- The demo is not very idiomatic, it uses a custom training / optimizing loop. Could you re-write it using `deepinv.Trainer`? Maybe I misunderstood your method but I thought that, like DIP, you'd be training a neural network to give you the output image instead of directly optimizing over the output. Could you re-write the demo so it is closer to that? Feel free to take inspiration from our DIP demos for that

Once it's done and CI passes I'll make a final check to the compiled docs, maybe suggest last few minor changes over the docs / demos, then I'll approve your PR

Let me know if I can help you with any kind of guidance!

Issue #1205, opened 2026-05-28
mixed-precision support in trainer


nucli-vicky:
Mixed precision greatly speeds up training and lowers memory requirements. Compared to a few years ago, implementing it in `pytorch` is also relatively easy.
This has caused mixed precision training to become relatively standard practice in most domains.

Additionally, the performance of GPUs is (partially) locked behind [tensor cores](https://developer.nvidia.com/blog/optimizing-gpu-performance-tensor-cores/), which cannot be used in pure `float32` precision. 

I propose we add `mixed-precision` as an option to the `Trainer`, which would benefit a variety of users;

- When on limited compute budget, it improves the number and scale of experiments people are able to perform.
- When training on cloud infra, it reduces costs.
- When training on GPUs with tensor cores (everything starting from Volta architecture), the current `Trainer` is not able to utilize the full compute capability offered.

**Considerations**

Some `physics` operators need higher precision for stability. Using mixed precision only for the model forward - backward - step pass seems safest from that PoV. Cases such as denoiser training, or anything which does im2im without actively using physics within the gradient computation or model forward pass seems safe.

I think the implementation would be quite simple, but would like some perspectives from others, especially w.r.t the interaction with `Physics`.

Additionally, it may be best for the progress on #728 to be completed and merged first? @Tmodrzyk , what do u think?

Tmodrzyk:
Hi @nucli-vicky, honestly the trainer refactor turned out to be more than we could realistically tackle, and the PR has gone stale.
I think you can safely work on this without stepping on #728.
I still think the `Trainer` needs a major refactor, and we identified logging as the main source of clutter in the current code.

nucli-vicky:
Thanks for the confirmation @Tmodrzyk , and agree that logging and other things in the trainer do need a refactor; maybe it can be tackled piece by piece

jscanvic:
Just a suggestion but it might be worth doing a quick and dirty PoC without worrying too much about how best to do it in deepinv clean and neat so we can discuss it more easily before going with a specific solution

nucli-vicky:
Agree @jscanvic ! See #1208 for a rough PoC

nucli-vicky:
Closed in #1208

Issue #1096, opened 2026-03-08
Wasteful `model.train()` at eval time during `compute_eval_losses=True` training


Andrewwango:
If I want to do self-supervised early-stopping, I need to set `compute_eval_losses=True` in `deepinv.Trainer` to have the loss values for early stopping. In `Trainer.compute_loss`, I don't know why we set `train=True` at eval time:

```python
        if train or self.compute_eval_losses:
            # Evaluate reconstruction network
            x_net = self.model_inference(y=y, physics=physics, x=x, train=True)
```

It makes sense during training, but during eval, this calls model inference with grads, which is wasteful if we only want the loss values (for computing at eval time). But we do need `model` to be in training mode (for loss-adapted models that have different performance at train vs eval time). So we need to tell `model_inference` somehow to put the model in training mode but to do `no_grad`. Maybe we wrap `no_grad` in `compute_loss` if `compute_eval_losses=True`?

PR #1208, opened 2026-05-28
Support training with mixed precision (`float16`, `bfloat16`)

nucli-vicky:
See #1205 

Also fixes bug wrt model being used in `.train()` mode during eval when `compute_eval_losses=True`


nucli-vicky:
### To show this implementation trains properly, and gives comparable runs between `use_amp`  `True` and `False`. 

Small test training run for 10 epochs on DIV2K 2x SR with `SRResNet`:

Test with `ckp_9.pth.tar` on DIV2K val: 

Trained without AMP:

- PSNR: 27.514 +- 3.858
- SSIM: 0.818 +- 0.106


Trained with AMP (`use_amp=True`):

- PSNR: 28.139 +- 4.200
- SSIM: 0.819 +- 0.104

nucli-vicky:
**Training curve when not using AMP:**

image

**Training curve when using AMP:**

image

nucli-vicky:
While writing this, I came across the following:

`compute_loss` is called both when doing eval, and when doing inference with `compute_eval_losses`.

However, in there, it will *always* call `self.model_inference(` with `train=True`. 

This causes 

(i) no `torch.no_grad` to speed up eval.

(ii) when `compute_eval_losses`, both the eval losses and metrics will use an `x_net` obtained under `net.train()`: with dropout active and `batchnorm` layers not in eval mode either.

deepinv/training/trainer.py
```diff
@@ -68,6 +69,7 @@ class Trainer:
     :param bool online_measurements: Generate new measurements `y` in an online manner at each iteration by calling
         `y=physics(x)`. If `False` (default), the measurements are loaded from the training dataset.
     :param str, torch.device device: Device on which to run the training (e.g., 'cuda', 'mps' or 'cpu'). Default is first 'cuda' and second 'mps' if available, otherwise 'cpu'.
+    :param bool use_amp: Whether to use automatic mixed precision (float16). Currently, will only be used during training steps, not eval or test.
```
mh-nguyen712: I would use a general parameter here, for example `mixed_precision: str | None = "no"` with choice between "no" (or "fp32"), "fp16" or "bf16". "bf16" is interesting since it has the same dynamic range as fp32 so no need for GradScaler

mh-nguyen712 (changes requested):
Thanks for the initiative! I think it's more flexible the leave user chose the compute dtype between fp32, fp16 or bfp16 rather than hardcoded it

deepinv/training/trainer.py
```diff
@@ -862,19 +878,30 @@ def compute_loss(self, physics, x, y, train=True, epoch: int = None, step=False)
 
         if train or self.compute_eval_losses:
             # Evaluate reconstruction network
-            x_net = self.model_inference(y=y, physics=physics, x=x, train=True)
+            x_net = self.model_inference(y=y, physics=physics, x=x, train=train)
```
nucli-vicky: Turns out this is what's breaking the tests. I will revert it, and let this be handled by a future PR, as this one has been quite stale for a while.

nucli-vicky:
@mh-nguyen712 thanks for the suggestion! I incorporated it here now :) [finally]

Andrewwango:
re compute_eval_losses, this is reported in #1096, but I don't think there is a trivial fix because some losses can only be calculated in train mode eg ReducedResolutionLoss

Andrewwango:
Hey @nucli-vicky , thanks for the contribution, and except for fp16, I think it's neat and concise addition.

Re. fp16, I think the fp16 support with gradscaler is quite risky as it potentially introduces lots of uncaught silent PyTorch bugs to a new user (e.g. https://github.com/pytorch/pytorch/issues/58228 ) - it would be unfortunate for such a user to see our neatly wrapped fp16 support and trust it 100%. I think for fp16, if we support it at all, it should come with clear warnings that it can lead to hidden bugs, and is included for legacy, and users should really use bf16 instead.

Secondly, it would be good to have a similar plot to the above using bf16 where the training should be pretty much identical.

nucli-vicky:
Thanks @Andrewwango 

Re `compute_eval_losses`: Seems I missed that issue, thanks! But I think it's about more than efficiency, for nets with batchnorm, the eval results (and also metrics)  will be way different when `compute_eval_losses` is on, as the batch statistics will be used for validation instead of the population statistics tracked during training. Even mathematical properties can be different (e.g. for certain convnets with batchnorm, the scale-equivariance property may hold in val mode, but not in train mode)

Re float16: I think the issues relate only/mostly to the training of complex dtype networks / data. Maybe we can narrow the warning down to that?

Andrewwango:
@nucli-vicky re. compute_eval_losses: it makes sense and I agree with you. Just need to find a way to deal with losses that can only be evaluated in train mode... perhaps we need to fix those losses directly, more specifically their adapted model wrappers (e.g. the model wrapper forward pass branches train vs eval mode as before, but in the loss calculation, there is no branch and only train mode behaviour is used)

re. float16: I don't know much about the various pitfalls (outside complex data). Maybe to mitigate the risk, it would be good to have some tests or examples on various common deepinv training scenarios to check that there are no performance shocks.

nucli-vicky:
@Andrewwango Thanks for your thoughts! Finally got to finalizing this PR 😅 

Re. `compute_eval_losses`: agree with you here. I'm not sure if we should wait for the full Trainer refactor to address this, or might be a good thing to take care of during the Hackathon?

Re. float16: From my experience, I've never encountered issues due to `float16` AMP across a variety of tasks and model types. While I personally don't expect any issues (aside from complex dtype models), we can't really be sure across the wide range of models, tasks, and data that `deepinv`could get used for. Therefore I've added a warning that notes AMP as an experimental feature, and encouraging users to report any issues/bugs they may encounter.

deepinv/training/trainer.py
```diff
@@ -69,6 +71,8 @@ class Trainer:
     :param bool online_measurements: Generate new measurements `y` in an online manner at each iteration by calling
         `y=physics(x)`. If `False` (default), the measurements are loaded from the training dataset.
     :param str, torch.device device: Device on which to run the training (e.g., 'cuda', 'mps' or 'cpu'). Default is first 'cuda' and second 'mps' if available, otherwise 'cpu'.
+    :param bool, str mixed_precision: Mixed precision to use. If False, standard float32 training is performed. If True, defaults to float16.
```
mh-nguyen712, line 75: I would allow any string in this list: `float32`, `fp32`, `float16`, `fp16`, `bfloat16`, `bf16`

nucli-vicky: thanks, good point to also include the `bf16` and `fp16` abbreviations :)

Im not sure as to including float32: in that case, there is no mixed precision, its just regular, single-precision training. So I think being able to specify float32 it for `mixed_precision` is wrong semantically. What do u think?

deepinv/training/trainer.py
```diff
@@ -653,6 +683,11 @@ def check_clip_grad(self):
 
         return grad_norm
 
+    def get_autocast_context(self):
+        if self.mixed_precision in [torch.float16, torch.bfloat16]:
+            return autocast(device_type=self.device.type, dtype=self.mixed_precision)
```
mh-nguyen712, line 715: Make sure that `self.device` is a `torch.device`, since user might pass a string as device. I think it's safer to make sure this in `__post_init__`

deepinv/training/trainer.py
```diff
@@ -872,14 +922,22 @@ def compute_loss(self, physics, x, y, train=True, epoch: int = None, step=False)
...
-            norm = self.check_clip_grad()
-            if norm is not None:
-                logs["gradient_norm"] = self.check_grad_val.avg
+            # GradScaler forbids calling unscale_() more than once per optimizer between step() calls.
+            defer_clip = (
+                self.scaler.is_enabled()
+                and self.optimizer_step_multi_dataset
+                and not step
+            )
+            if not defer_clip:
```
mh-nguyen712, line 960: I think there is an inconsistent gradient clipping behavior between AMP and non-AMP if `optimizer_step_multi_dataset=True`:
- If AMP: clipping is deferred until gradients across all datasets accumulate (clip(g1 + g2)).
- No AMP: defer_clip is False, so check_clip_grad() runs for each dataset individually before accumulation (clip(g1) + clip(g2)).
  
I think we should be clear on that

deepinv/training/trainer.py
```diff
@@ -1335,7 +1399,17 @@ def train(
         :returns: The trained model.
         """
         self.setup_train()
-
+        if self.mixed_precision and self.device.type != "cuda":
```
mh-nguyen712: I think it's cleaner to move these warnings to `__post_init__`

mh-nguyen712 (changes requested):
Thanks for the updates! Here are my last comments and hope that we can merge this before the hackathon!

deepinv/training/trainer.py
```diff
@@ -559,6 +585,8 @@ def setup_train(self, train: bool = True, **kwargs):
         except (ValueError, TypeError, AttributeError):
             self._model_accepts_update_parameters = False
 
+        self.scaler = GradScaler(enabled=self.mixed_precision == torch.float16)
```
mh-nguyen712: I think it's better to precise the `device` in the GradScaler. 
```suggestion
        self.scaler = GradScaler(
            device=self.device.type, 
            enabled=self.mixed_precision == torch.float16 
        )
```

deepinv/training/trainer.py
```diff
@@ -1335,7 +1399,17 @@ def train(
         :returns: The trained model.
         """
         self.setup_train()
-
+        if self.mixed_precision and self.device.type != "cuda":
+            warnings.warn(
```
mh-nguyen712: I think `bfloat16` is natively supported by modern CPU too, eg: https://intel.github.io/intel-extension-for-pytorch/cpu/latest/tutorials/features/amp.html

deepinv/training/trainer.py
mh-nguyen712: Yes agree!

nucli-vicky: added fp16 and bf16

deepinv/training/trainer.py
nucli-vicky: fair! Unfortunately there is right now not really a way to do `clip(g1) + clip(g2)` for `float16` as it would require calling unscale twice which is supported. 

Wondering where best to document this discrepancy?

deepinv/training/trainer.py
nucli-vicky: I think prior to this PR, code passing as device a string would not break. So are u proposing we convert `device` variable to a `torch.device` in `__post_init__`? 

Will do

deepinv/training/trainer.py
nucli-vicky: thanks!

deepinv/training/trainer.py
nucli-vicky: true, but this warning also mainly serves for other accelerators (MPS for instance). 

As mixed precision is currently only implemented for training (which users will typically not do on CPU) and a lot of CPUs do not have native `bfloat16`, I think it's okay to keep the warning as is?

nucli-vicky:
Thanks @mh-nguyen712 for the comments, tried to address most of them.

deepinv/training/trainer.py
mh-nguyen712: I think we can add a note to the main docstring of the Trainer class. Can you add this? Then I think the PR is ready to be merged! Thanks for you works!

nucli-vicky: note added!

mh-nguyen712 (approved):

PR #1211, opened 2026-05-30
fix(transform): fix Reflect.inverse and Shift axis swap

### LINKED ISSUE #1197: Bugs in transforms
### Environment

Latest

### Describe the bug

- Reflect.inverse is incorrect - it is left unimplemented & the fallback is invalid

```python
import torch
import deepinv as dinv

x = dinv.utils.load_example("butterfly.png")

transform = dinv.transform.Reflect()
x_flipped = transform.transform(x, dims=torch.tensor([-1]))
# The inverse flip is mistakenly applied to dims=-(-1)=1, i.e., on the channels.
x2 = transform.inverse(x_flipped, dims=torch.tensor([-1]))

# Flipped back is blue and flipped instead of having the right color and being unflipped.
dinv.utils.plot([x, x_flipped, x2], titles=["Original", "Flipped", "Flipped back"])

g_params = { "dims": torch.tensor([-1]) }
g_params_inv = transform.invert_params(g_params)

print(g_params_inv)  # {'dims': tensor([1])}
assert torch.equal(g_params["dims"], g_params_inv["dims"]), "Inverse params should be the same as original params for this transform"

```

- Shift parameters are inverted: x_shift is applied vertically and y_shift horizontally

https://github.com/deepinv/deepinv/blob/3831fb4864779c8160bd0120d882e97f77b15a90/deepinv/transform/shift.py#L67-L73

Thanks Sarra for spotting those

### Steps to reproduce

/

### Expected behavior

/

_No response_

### Additional context

_No response_

SteliosGee:
Closes #1197

## Bug 1: Reflect.inverse

The base class `invert_params` negates all parameters (`{k: -v}`), which incorrectly transforms `dims=[-1]` to `dims=[1]`. Since reflecting (flipping) twice along the same axis is the identity operation, the inverse of a reflection is itself. Fixed by overriding `invert_params` in `Reflect` to return params unchanged.

## Bug 2: Shift axis swap

`x_shift` was generated from the height range (`H_max`) and applied to dim `-2` (height), while `y_shift` was generated from the width range (`W_max`) and applied to dim `-1` (width). In image convention, x=horizontal=width, y=vertical=height. Fixed by swapping the roll arguments so `x_shift` maps to width and `y_shift` maps to height.

### DIFF
--- a/deepinv/transform/reflect.py
+++ b/deepinv/transform/reflect.py
@@ -28,6 +28,10 @@ class Reflect(Transform):
         super().__init__(*args, **kwargs)
         self.dim = dim
 
+    def invert_params(self, params: dict) -> dict:
+        """Inverse of a reflection is itself (flipping twice returns the original)."""
+        return params
+
     def _get_params(self, x: torch.Tensor) -> dict:
         """Randomly generate sets of reflection axes without replacement.
 
--- a/deepinv/transform/shift.py
+++ b/deepinv/transform/shift.py
@@ -66,7 +66,7 @@ class Shift(Transform):
         """
         return torch.cat(
             [
-                torch.roll(x, [sx, sy], [-2, -1])
+                torch.roll(x, [sy, sx], [-2, -1])
                 for sx, sy in zip_longest(x_shift, y_shift, fillvalue=0)
             ],
             dim=0,


jscanvic (changes requested):
Welcome and thank you for handling this

The fix for reflections is good to go

The fix for shifts is good but I'd make it opt-in with a deprecation warning for the default behavior. Something like a new parameter & attribute `standard_xy` of type bool defaulting to False (current behavior) resulting in a warning, and if giving the corrected behavior if set to True.

Add a regression test for both values of the parameter

Also, please fill in the change log https://deepinv.github.io/deepinv/contributing.html#log-your-changes

PR #1222, opened 2026-06-07
Fix small typo in the docs of DecomposablePhysics

jscanvic:
If A is in R^(m times n), U should be in R^(m times m) and V in R^(n times n) while currently it's the opposite

### DIFF
--- a/deepinv/physics/forward.py
+++ b/deepinv/physics/forward.py
@@ -996,7 +996,7 @@ class DecomposablePhysics(LinearPhysics):
         A = U\text{diag}(s)V^{\top} \in \mathbb{R}^{m\times n}
 
     where :math:`U\in\mathbb{C}^{m\times m}` and :math:`V\in\mathbb{C}^{n\times n}`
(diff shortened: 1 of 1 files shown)


deepinv/physics/forward.py
```diff
@@ -995,7 +995,7 @@ class DecomposablePhysics(LinearPhysics):
 
         A = U\text{diag}(s)V^{\top} \in \mathbb{R}^{m\times n}
 
-    where :math:`U\in\mathbb{C}^{n\times n}` and :math:`V\in\mathbb{C}^{m\times m}`
+    where :math:`U\in\mathbb{C}^{m\times m}` and :math:`V\in\mathbb{C}^{n\times n}`
```
mh-nguyen712, line 998: Can you also fix that `\text{diag}(s)` is a rectangular diagonal matrix of size m x n

jscanvic:
Thanks for the review Hai

Should be good now

mh-nguyen712 (approved):

PR #1228, opened 2026-06-11
WIP RecoveryCoefficient

### LINKED ISSUE #1180: Add RecoveryCoefficient metric / utilities
### Environment (optional)

_No response_

### Describe the feature

A common metric used in emission tomography is the "Recovery Coefficient" (RC), which is defined as the sum of voxel activities in the reconstructed image over the sum of activities in the ground-truth image:
$RC = \frac{a_{recon}}{a_{gt}}$
The metric should be provided with a reconstruction, a ground-truth, and a mask, and compute this ratio.

### Motivation

Usual natural image metrics (PSNR, SSIM) are not used in emission tomography.
Users working on PET will want to have access to common but domain-specific metrics. 

### Alternatives considered

_No response_

### Additional context

_No response_

Kushagra481:

#1180 

### DIFF
--- a/deepinv/loss/metric/distortion.py
+++ b/deepinv/loss/metric/distortion.py
@@ -60,36 +60,34 @@ class MAE(Metric):
 
 class MSE(Metric):
     r"""deepinv.metric.MSE(complex_abs, reduction, norm_inputs, center_crop, ``kwargs``)
-    Mean Squared Error metric.
+        Mean Squared Error metric.
 
-    Calculates :math:`\|\hat{x}-x\|_2^2` where :math:`\hat{x}=\inverse{y}`.
-distortion
-    .. note::
+        Calculates :math:`\|\hat{x}-x\|_2^2` where :math:`\hat{x}=\inverse{y}`.
+        .. note::
 
-        By default, no reduction is performed in the batch dimension.
+            By default, no reduction is performed in the batch dimension.
 
-    .. note::
-distortion
-        :class:`deepinv.loss.metric.MSE` is functionally equivalent to :class:`torch.nn.MSELoss` when ``reduction='mean'`` or ``reduction='sum'``,
-        but when ``reduction=None`` our MSE reduces over all dims except batch dim (same behavior as ``torchmetrics``) whereas ``MSELoss`` does not perform any reduction.
+        .. note::
+            :class:`deepinv.loss.metric.MSE` is functionally equivalent to :class:`torch.nn.MSELoss` when ``reduction='mean'`` or ``reduction='sum'``,
+            but when ``reduction=None`` our MSE reduces over all dims except batch dim (same behavior as ``torchmetrics``) whereas ``MSELoss`` does not perform any reduction.
 
-    :Example:
+        :Example:
 
-    >>> import torch
-    >>> from deepinv.loss.metric import MSE
-    >>> m = MSE()
-    >>> x_net = x = torch.ones(3, 2, 8, 8) # B,C,H,W
-    >>> m(x_net, x)
-    tensor([0., 0., 0.])
+        >>> import torch
+        >>> from deepinv.loss.metric import MSE
+        >>> m = MSE()
+        >>> x_net = x = torch.ones(3, 2, 8, 8) # B,C,H,W
+        >>> m(x_net, x)
+        tensor([0., 0., 0.])
 
-    :param bool complex_abs: perform complex magnitude before passing data to metric function. If ``True``,
-        the data must either be of complex dtype or have size 2 in the channel dimension (usually the second dimension after batch).
-    :param str reduction: a method to reduce metric score over individual batch scores. ``mean``: takes the mean, ``sum`` takes the sum, ``none`` or None no reduction will be applied (default).
-    :param str norm_inputs: normalize images before passing to metric. ``l2`` normalizes by :math:`\ell_2` spatial norm, ``min_max`` normalizes by min and max of each input.
-    :param int, tuple[int], None center_crop: If not `None` (default), center crop the tensor(s) before computing the metrics.
-        If an `int` is provided, the cropping is applied equally on all spatial dimensions (by default, all dimensions except the first two).
-        If `tuple` of `int`, cropping is performed over the last `len(center_crop)` dimensions. If positive values are provided, a standard center crop is applied.
-        If negative (or zero) values are passed, cropping will be done by removing `center_crop` pixels from the borders (useful when tensors vary in size across the dataset).
+        :param bool complex_abs: perform complex magnitude before passing data to metric function. If ``True``,
+            the data must either be of complex dtype or have size 2 in the channel dimension (usually the second dimension after batch).
+        :param str reduction: a method to reduce metric score over individual batch scores. ``mean``: takes the mean, ``sum`` takes the sum, ``none`` or None no reduction will be applied (default).
+        :param str norm_inputs: normalize images before passing to metric. ``l2`` normalizes by :math:`\ell_2` spatial norm, ``min_max`` normalizes by min and max of each input.
+        :param int, tuple[int], None center_crop: If not `None` (default), center crop the tensor(s) before computing the metrics.
+            If an `int` is provided, the cropping is applied equally on all spatial dimensions (by default, all dimensions except the first two).
+            If `tuple` of `int`, cropping is performed over the last `len(center_crop)` dimensions. If positive values are provided, a standard center crop is applied.
+            If negative (or zero) values are passed, cropping will be done by removing `center_crop` pixels from the borders (useful when tensors vary in size across the dataset).
     """
 
     def metric(self, x_net: Tensor, x: Tensor, *args, **kwargs) -> Tensor:
@@ -1096,23 +1094,26 @@ class GMSD(Metric):
         gmsd_per_channel = gms.std(dim=(-2, -1), correction=0)
 
         return gmsd_per_channel.mean(dim=-1)
+
+
 class RecoveryCoefficient(Metric):
     r"""
     Recovery Cofficient metric used in emission tomography.
     Compute the ratio between the total Reconstructed activity and the total ground truth activity inside a given mask.
     """
-    def __init__(self,eps:float = 1e-12,**kwargs):
+
+    def __init__(self, eps: float = 1e-12, **kwargs):
         super().__init__(**kwargs)
         self.eps = eps
         self.lower_better = False
-    def metric(self,x_net,x,*args,**kwargs):
-        mask = kwargs.get("mask",None)
+
+    def metric(self, x_net, x, *args, **kwargs):
+        mask = kwargs.get("mask", None)
         if mask is None:
             raise ValueError("Recovery Coefficient requires a mask argument.")
         mask = mask.to(x.dtype)
-        dims = tuple(range(1,x.ndim))
-        recon_activity = (x_net*mask).sum(dim=dims)
-        gt_activity = (x*mask).sum(dim=dims)
+        dims = tuple(range(1, x.ndim))
+        recon_activity = (x_net * mask).sum(dim=dims)
+        gt_activity = (x * mask).sum(dim=dims)
 
-        return recon_activity/(gt_activity+self.eps)
-
+        return recon_activity / (gt_activity + self.eps)
--- a/deepinv/loss/metric/distortion.py
+++ b/deepinv/loss/metric/distortion.py
@@ -60,34 +60,34 @@ class MAE(Metric):
 
 class MSE(Metric):
     r"""deepinv.metric.MSE(complex_abs, reduction, norm_inputs, center_crop, ``kwargs``)
-        Mean Squared Error metric.
+    Mean Squared Error metric.
 
-        Calculates :math:`\|\hat{x}-x\|_2^2` where :math:`\hat{x}=\inverse{y}`.
-        .. note::
+    Calculates :math:`\|\hat{x}-x\|_2^2` where :math:`\hat{x}=\inverse{y}`.
+    .. note::
 
-            By default, no reduction is performed in the batch dimension.
+        By default, no reduction is performed in the batch dimension.
 
-        .. note::
-            :class:`deepinv.loss.metric.MSE` is functionally equivalent to :class:`torch.nn.MSELoss` when ``reduction='mean'`` or ``reduction='sum'``,
-            but when ``reduction=None`` our MSE reduces over all dims except batch dim (same behavior as ``torchmetrics``) whereas ``MSELoss`` does not perform any reduction.
+    .. note::
+        :class:`deepinv.loss.metric.MSE` is functionally equivalent to :class:`torch.nn.MSELoss` when ``reduction='mean'`` or ``reduction='sum'``,
+        but when ``reduction=None`` our MSE reduces over all dims except batch dim (same behavior as ``torchmetrics``) whereas ``MSELoss`` does not perform any reduction.
 
-        :Example:
+    :Example:
 
-        >>> import torch
-        >>> from deepinv.loss.metric import MSE
-        >>> m = MSE()
-        >>> x_net = x = torch.ones(3, 2, 8, 8) # B,C,H,W
-        >>> m(x_net, x)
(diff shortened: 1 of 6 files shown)


Tmodrzyk:
Hi @Kushagra481 thank you for this PR ! 
There are still a few things for you to do before we can review it.
First you should add a small description of the PR's goal (thanks for linking the issue).
Second you should try to pass the linting, this can be done locally (see Section 4 of the [contributing guidelines](https://deepinv.github.io/deepinv/contributing.html)

After that, we will have to wait for the CI test steps to be fixed, because as you correctly identified in #1180, the broken tests do not seem to be caused by your PR.

Tmodrzyk:
Hi again @Kushagra481, it is better if we keep the discussions related to this PR on this thread, and not the thread of #1180.
Now that the linting step of the CI is passing, we'll have to wait that the test step of the CI is fixed, I'll ping you again and make a review when this is the case. :smile:

Tmodrzyk:
Hi @Kushagra481 I think the CI is fixed, at least the one running on CPU. 
You can pull from main and see if the docs build, and the CPU tests are passing. 
If this fails, you should take a look at the logs to try and fix what's been broken by your changes.
Don't worry this is part of the process and CI rarely passes on the first try. :smile:

Kushagra481:
@Tmodrzyk Yea I will do it till this weekend I have a interview friday so mb cant do it rn

Kushagra481:
@Tmodrzyk welp i did it today was free

deepinv/loss/metric/distortion.py
```diff
@@ -63,13 +63,11 @@ class MSE(Metric):
     Mean Squared Error metric.
 
     Calculates :math:`\|\hat{x}-x\|_2^2` where :math:`\hat{x}=\inverse{y}`.
-
     .. note::
 
         By default, no reduction is performed in the batch dimension.
 
     .. note::
-
```
Tmodrzyk, line 72: I don't think we need to change these lines in the MSE docstrings, it is unrelated to your contribution

deepinv/loss/metric/distortion.py
```diff
@@ -1096,3 +1094,26 @@ def metric(self, x_net: Tensor, x: Tensor, *args, **kwargs) -> Tensor:
         gmsd_per_channel = gms.std(dim=(-2, -1), correction=0)
 
         return gmsd_per_channel.mean(dim=-1)
+
+
+class RecoveryCoefficient(Metric):
+    r"""
+    Recovery Cofficient metric used in emission tomography.
+    Compute the ratio between the total Reconstructed activity and the total ground truth activity inside a given mask.
+    """
```
Tmodrzyk, line 1145: We generally have more complete docstrings, you can take inspiration from other metrics. 
For instance a formula describing what is computed, and the different parameters.

deepinv/loss/metric/distortion.py
```diff
@@ -1096,3 +1094,26 @@ def metric(self, x_net: Tensor, x: Tensor, *args, **kwargs) -> Tensor:
...
+    def metric(self, x_net, x, *args, **kwargs):
+        mask = kwargs.get("mask", None)
+        if mask is None:
+            raise ValueError("Recovery Coefficient requires a mask argument.")
+        mask = mask.to(x.dtype)
+        dims = tuple(range(1, x.ndim))
+        recon_activity = (x_net * mask).sum(dim=dims)
+        gt_activity = (x * mask).sum(dim=dims)
+
+        return recon_activity / (gt_activity + self.eps)
```
Tmodrzyk: We usually type every arguments in the functions, here `x_net` and `x` have no types.
The mask should also probably be typed and an optional argument instead of a kwarg.

deepinv/tests/test_metric.py
```diff
@@ -591,3 +595,27 @@ def test_gmsd():
...
+    mask[:, :, :2, :2] = 1
+
+    out = rc(x_net, x, mask=mask)
+    assert torch.equal(out, 2 * torch.ones((2,)))
+
+    # missing mask should raise
+    with pytest.raises(ValueError) as exc_info:
+        rc(x_net, x)
+
+    assert str(exc_info.value) == "Recovery Coefficient requires a mask argument."
```
Tmodrzyk: We could add this test to the `full_reference_metrics` test. 
Actually, all full reference metrics could support this masking mechanism.

Tmodrzyk (changes requested):
Hi @Kushagra481 thank you for this PR !
There are a number of things to change in order to comply with the coding style / practices of the library, see my specific comments. 😄

deepinv/loss/metric/distortion.py
Kushagra481: I don't remember changing this part but I'll look into it

deepinv/tests/test_metric.py
Kushagra481: Do you mean adding RecoveryCoefficient to FULL_REFERENCE_METRICS and extending this test to pass a mask when the metric is RecoveryCoefficient, or do you mean adding generic mask support to all full-reference metrics?

deepinv/loss/metric/distortion.py
Kushagra481: def metric(
    self,
    x_net: Tensor,
    x: Tensor,
    mask: Tensor,
    *args,
    **kwargs,
) -> Tensor:
    mask = mask.to(x.dtype)
    dims = tuple(range(1, x.ndim))

    recon_activity = (x_net * mask).sum(dim=dims)
    gt_activity = (x * mask).sum(dim=dims)

    return recon_activity / (gt_activity + self.eps)
something like this?

Tmodrzyk: Yes indeed ! You can also check docstring guidelines here: https://deepinv.org/contributing.html#docstring-guidelines

deepinv/tests/test_metric.py
Tmodrzyk: Adding RecoveryCoefficient to FULL_REFERENCE_METRICS
Adding the masking mechanism to all metrics will be dealt with in another PR, but it's good to have in mind that ultimately they will probably all support this.

Kushagra481: Yeah, I’m not quite following how this should be integrated into the existing test. I understand that RecoveryCoefficient is being added to `FULL_REFERENCE_METRICS`, but I’m unsure how you envision handling the mask within the already parameterized `test_full_reference_metrics` function.
Could you explain the intended approach in a bit more detail or with a small example? I think I may be missing where the mask should be introduced and how the test should be structured relative to the other full-reference metrics.

Kushagra481:
@Tmodrzyk yeah i made the changes as requested

deepinv/tests/test_metric.py
Tmodrzyk: You can have a small test (like``if metric_name is "RecoveryCoefficient"``) in the ``test_full_reference_metrics`` in order to test the mask as you did in your snippet (raising an error if it is not provided).
If this passes, the rest of the test should follow as with other full reference metrics. 
In particular, the metric should be summable with other full reference metrics (see the test), so you should take care of the shapes returned etc.
Let me know if you need more details ! :smile:

Kushagra481:
I was looking into the failing `train_loss=True` test for `RecoveryCoefficient` and noticed something that I wanted to confirm before making any changes.

`Metric.invert_metric()` currently applies `-m`, which works for metrics where "higher is always better." However, `RecoveryCoefficient` seems different because its optimum is at `RC = 1.0`, and values on either side of 1 represent degradation.

For example:

* `RC = 1.0` (ideal) → `-1.0`
* `RC = 2.0` (worse) → `-2.0`

This makes the worse reconstruction produce a smaller loss value than the ideal reconstruction.

Would it make sense for `RecoveryCoefficient` to override `invert_metric()` and use something like `abs(m - 1)` (distance from the ideal value) when `train_loss=True`, or is the current behavior intentional and I'm misunderstanding the expected semantics?

deepinv/loss/metric/distortion.py
```diff
@@ -1096,3 +1096,77 @@ def metric(self, x_net: Tensor, x: Tensor, *args, **kwargs) -> Tensor:
         gmsd_per_channel = gms.std(dim=(-2, -1), correction=0)
 
         return gmsd_per_channel.mean(dim=-1)
+
+
+class RecoveryCoefficient(Metric):
+    r"""deepinv.metric.RecoveryCoefficient(eps, reduction, **kwargs)
```
Tmodrzyk:
```suggestion
    r"""
```

deepinv/loss/metric/distortion.py
```diff
@@ -1096,3 +1096,77 @@ def metric(self, x_net: Tensor, x: Tensor, *args, **kwargs) -> Tensor:
...
+    Recovery Coefficient metric used in emission tomography.
+
+    Computes the ratio between the total reconstructed activity and the total
+    ground-truth activity inside a given mask:
```
Tmodrzyk:
```suggestion
    ground-truth activity inside a given region of interest, defined by a mask:
```

deepinv/loss/metric/distortion.py
```diff
@@ -1096,3 +1096,77 @@ def metric(self, x_net: Tensor, x: Tensor, *args, **kwargs) -> Tensor:
...
+    >>> x = torch.ones(2, 1, 4, 4)
+    >>> x_net = torch.ones(2, 1, 4, 4)
+    >>> mask = torch.ones_like(x)
+    >>> m(x_net, x, mask=mask)
+    tensor([1., 1.])
+
+    :param float eps: Small constant added to the denominator for numerical stability.
+    :param str reduction: Method used to reduce metric values across the batch.
+        ``"mean"`` computes the mean, ``"sum"`` computes the sum, and
+        ``None`` or ``"none"`` returns one score per sample.
```
Tmodrzyk: This parameter is not used / not defined in init, we should remove it from the docstring

deepinv/tests/test_metric.py
```diff
@@ -591,3 +605,27 @@ def test_gmsd():
...
```
Tmodrzyk: Now that we are using the `test_full_reference_metrics` function, we can safely remove these extra tests

deepinv/tests/test_metric.py
```diff
@@ -155,20 +158,31 @@ def test_full_reference_metrics(
         if channels != 3:
             pytest.skip("LPIPS requires 3 channel input.")
 
+    def get_mask(t):
+        return (
+            {"mask": torch.ones_like(t)} if metric_name == "RecoveryCoefficient" else {}
+        )
```
Tmodrzyk: I'm fine with this solution for now, we can extend this mechanism to handle a several masked metrics later on.

Tmodrzyk (changes requested):
Hi @Kushagra481 thank you again for implementing the metric and the modifying the tests.
We are almost there, there are just a few clean-ups you need to do before we can merge (see my specific comments). 
I'd also like to have a second set of eyes on the test part. Your current solution is not very scalable but I think this is fine for now, we can handle more masked metrics in another PR. 
@nucli-vicky could you take a look ?

deepinv/tests/test_metric.py
Kushagra481: so any changes @nucli-vicky  @Tmodrzyk like I haven't study much about scalability of code(I'm learning basic dsa) so if you have any reccomendations or similar test cases implemented should I study from the codebase

Tmodrzyk: Scalability can mean very different things depending on the context. Here I just meant that the test ```if metric_name == "RecoveryCoefficient"` won't be adapted to handle, per say, 10 different masked metrics.  
However as I mentioned in the review, in my opinion, we can easily fix that later on when needed. :smile:

Kushagra481:
@Tmodrzyk I don't quite understand the error in this one like sometimes it appears sometimes it doesnt anscomb and the logs when I put in AI are telling me to change values now I am not going to do that but can you please explain this problem also very thankful for you being so patient with me

Tmodrzyk:
Hi @Kushagra481 no problem, I'm very happy that you are contributing and taking into account the feedback ! :smiley:  
I can confirm that this random CI failure was not caused by your PR.
We better wait again that this failure is fixed on the main branch, and then merge from main.

Kushagra481:
should I try now or will u ping me when its fixed (also thanks again for being so patient)

Tmodrzyk:
The CI's down again unfortunately. I'll ping you when it's ready, but you can also check with the badge "Test CPU" on the landing page of the repo :)

deepinv/loss/metric/distortion.py
```diff
@@ -1096,3 +1096,74 @@ def metric(self, x_net: Tensor, x: Tensor, *args, **kwargs) -> Tensor:
...
+    >>> mask = torch.ones_like(x)
+    >>> m(x_net, x, mask=mask)
+    tensor([1., 1.])
+
+    :param float eps: Small constant added to the denominator for numerical stability.
+    :param kwargs: Additional keyword arguments passed to the parent
+        :class:`deepinv.loss.metric.Metric` class.
+    """
+
+    def __init__(self, eps: float = 1e-12, **kwargs):
```
nucli-vicky: `eps = 1e-12` might be too small for float32 and float16. adam optimizer for instance uses `1e-8` for float32. 

Something like `1e-7` for float32 and `1e-4` for float16 seems perfectly safe.

deepinv/tests/test_metric.py
```diff
@@ -155,20 +158,31 @@ def test_full_reference_metrics(
         if channels != 3:
             pytest.skip("LPIPS requires 3 channel input.")
 
+    def get_mask(t):
```
nucli-vicky: maybe we can rename this to something like `mybe_get_mask` or `get_metric_kwargs`?

The latter is also more adaptable for other new metrics (although I agree with @Tmodrzyk we should probably find a better solution as more such metrics are added)

The first is also fine, I would go for either over `get_mask`, as that is a bit confusing for metrics that do not use a mask.

nucli-vicky (changes requested):
Thank you for this contribution, @Kushagra481 !

deepinv/loss/metric/distortion.py
Kushagra481: Do you think `eps` should be dtype-aware instead of a fixed `1e-12`? At fp32/fp16, `1e-12` is basically zero, so it doesn't do much to protect against near-zero denominators.
I'm happy to switch to a dtype-based default (e.g. `1e-7` for fp32, `1e-4` for fp16) with `eps=None`, while still allowing an explicit `eps` override. Or would you prefer keeping the fixed constant?

nucli-vicky: I think dtype aware default makes sense. Unfortunately, dtype is unknown at `init` time, so the pattern would have to be


1. init tim: `self.eps = eps`, where default `eps=None`
2. at `metric` time, something like:
```python
eps = self.eps
if eps is None:
   eps = eps_per_dtype[x_net.dtype]
```
which might not be the cleanest

Kushagra481:
I dont think any major changes left so can I do this after Monday I'm giving a interview on monday I have to prep for that if thats fine

Tmodrzyk:
Hi @Kushagra481 congrats for your first contribution to Deepinverse and thank you for helping us out with this implementation !
There's a few last things to do before we can give a final review and merge your code:
- you should add this new metric to the documentation: both in the API doc through https://github.com/deepinv/deepinv/blob/main/docs/source/api/deepinv.metric.rst and in the user guid doc through https://github.com/deepinv/deepinv/blob/main/docs/source/user_guide/training/metric.rst
- you should log your contribution in the `changelog.rst`, remember to add your github username at the bottom of the file :smile:

Tmodrzyk (approved):
LGTM, just lacks the changelog entry :)

Kushagra481:
@Tmodrzyk in changelog.rst do I have to write in current new features or under some version?

nucli-vicky (approved):
LGTM. Thanks @Kushagra481 and thanks for your first contribution 🥳

Tmodrzyk:
@Kushagra481 Congrats for your first contribution ! If you'd like to contribute more to the project, don't hesitate to check the other issues, we still have many good first issues. :smile:

PR #1233, opened 2026-06-14
Add RTK backend for Computed Tomography Physics

idrisT11:
## Summary

This PR adds a computed tomography physics operator interfacing with the ITK-RTK external library backend.

The implementation depends on both `itk-cuda-common` and `itk-rtk`, and it can be installed via the `itk-rtk-cuda128` package.

The class `TomographyWithRTK` supports both fan-beam and cone-beam geometries and works exclusively on GPU in its current implementation.

The implementation follows the same overall logic as `TomographyWithAstra` for consistency.
The class `TomographyWithRTK` is implemented in its own source file. This could be changed to be added in the same file as the other tomography operators.

---

### Example

An example has been added showcasing an iterative CBCT reconstruction `"using a simple TV prior.". The execution of this example can take some time.

### Tests

Unit tests were implemented following the same structure as `TomographyWithAstra, and are located in the same test file.

---

#### Small Note

- `changelog.rst` was updated as required by the documentation, but I am not sure if I did it as intended.


romainvo:
Hi @idrisT11! It looks like a really nice addition to Tomography. I will try to take a look as soon as possible! 

In the meantime, do you think you could give a quick intro to what RTK is doing (if you have links to key examples or demos, that would be great)? Just key features that are similar (or not) to other than X-Ray CT libraries (like Astra). 
- It looks like RTK supports different forward/backward ray casting strategies, with optionally exact adjoint? 
- Does it support batching, i.e. parallel forward projection on different samples?  
- What type of floating-point precision does it support? 

Thank you again for the contribution! Very excited to review this :)

SimonRit:
> Hi @idrisT11! It looks like a really nice addition to Tomography. I will try to take a look as soon as possible!
> 
Thanks, we're looking forward to your review!

> In the meantime, do you think you could give a quick intro to what RTK is doing (if you have links to key examples or demos, that would be great)? Just key features that are similar (or not) to other than X-Ray CT libraries (like Astra).
> 
RTK is our research tool which has been expanding over the years, see intro at https://www.openrtk.org/. I think it started before ASTRA and many other [alternatives](https://www.openrtk.org/RTK/resources/alternatives.html). ASTRA has mainly focused on the (back)projectors (speed, multi-GPU, etc.) while we've added many higher level features such as [dynamic](https://docs.openrtk.org/en/latest/documentation/docs/3d_time.html) and [spectral](https://docs.openrtk.org/en/latest/documentation/docs/multi_energy.html) tomography. It's also been popular because it can read quite a few cone-beam file formats from different vendors (Elekta, Varian, IBA, medPhoton, etc.), some of whom we collaborate with.

>     * It looks like RTK supports different forward/backward ray casting strategies, with optionally exact adjoint?
> 
We have different (back)projectors indeed, with adjoint pairs (CPU and Cuda). We use adjoint Cuda ones in this PR. 

>     * Does it support batching, i.e. parallel forward projection on different samples?
> 
I think so (it's thread proof) but I'm not sure we've tried. In any case, we work with a single GPU (see [here](https://github.com/RTKConsortium/ITKCudaCommon/blob/main/src/itkCudaDataManager.cxx#L28)). I don't see why one wouldn't be able to use parallel forward CPU projections, we do it internally on different blocks of the volume.

>     * What type of floating-point precision does it support?
> 
CPU versions are templated and have been tested with float32 and float64. GPU versions are hardcoded in float32 only.

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,359 @@
...
+    def A(self, x: torch.Tensor, **kwargs) -> torch.Tensor:
+        """Forward projection.
+
+        :param torch.Tensor x: input of shape [B,C,...,H,W]
+        :return: projection of shape [B,C,...,A,N]
+        """
+
+        itk, rtk = import_itk_rtk()
+
+        x_stacked = x.squeeze(0).squeeze(0)
```
romainvo: I understand here that `x` is of shape (B,C,..,H,W) and that `rtk` only only support (D,H,W) 
tensors (or H,W,D) ? 

If possible, we should allow the user to pass a batch of inputs and at least handle the computation by looping through the batch elements.

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,359 @@
...
+
+        # Cast from tensor to ITK cuda image
+        imageSource_cuda = itk.cuda_image_from_cuda_array(x_stacked)
+        imageSource_cuda.SetOrigin(self.volume_information["origin"])
+        imageSource_cuda.SetSpacing(self.volume_information["spacing"])
+
+        fp_source = rtk.ConstantImageSource[self._CUDA_IMAGE_TYPE].New()
+        fp_source.SetSize(self.projection_stack_information["size"])
+        fp_source.SetOrigin(self.projection_stack_information["origin"])
+        fp_source.SetSpacing(self.projection_stack_information["spacing"])
```
romainvo: Would this make sense to have `fp_source` be initialized in the `__init__` and re-used across `A` calls ? (and updated later if the user wishes to change the geometry) ?

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,359 @@
...
+
+        # Forward projection: Ax
+        forward_projector = rtk.CudaForwardProjectionImageFilter[
+            self._CUDA_IMAGE_TYPE
+        ].New()
+        forward_projector.SetGeometry(self.geometry)
+        forward_projector.SetInput(fp_source.GetOutput())
+        forward_projector.SetInput(1, imageSource_cuda)
+        forward_projector.SetStepSize(self.ray_step_size)
+        forward_projector.Update()
```
romainvo: I feel like this would make sense to make the `forward_projector` an instance attribute, and re-use it across calls to `A`?

- what is the difference between these two calls to `SetInput`:
    - `forward_projector.SetInput(fp_source.GetOutput())`
    - `forward_projector.SetInput(1, imageSource_cuda))`
- what does the call to `forward_projector.Update` does in practice, does it apply all changes to some underlying object ?

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,359 @@
...
+        Ax = forward_projector.GetOutput()
+        Ax.DisconnectPipeline()
+
+        # Cast back from itk cuda image to tensor
+        projections = torch.as_tensor(
+            Ax, device=x.device
+        ).clone()  # This may not be optimized for memory usage
```
romainvo: What is the `type` of `Ax` in practice, if the underlying array implements the `__cuda_array_interface__` protocol it should be possible to instantiate a torch.Tensor without cloning right?

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,359 @@
...
+            print(self.projection_stack_information["spacing"])
+
+            print("Output volume information: ")
+            print(self.volume_information["size"])
+            print(self.volume_information["origin"])
+            print(self.volume_information["spacing"])
+
+    def _validate_info(
+        self, mode: str, info: dict[str, int | float], name: str
+    ) -> None:
```
romainvo: This method is static

```suggestion
    @staticmethod
    def _validate_info(
        mode: str, info: dict[str, int | float], name: str
    ) -> None:
```

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,359 @@
...
+
+        itk, rtk = import_itk_rtk()
+
+        x_stacked = x.squeeze(0).squeeze(0)
+
+        # Add a dimension to simulate a 2D projection in fanbeam mode
+        if self.mode == "fanbeam":
+            x_stacked = torch.stack([x_stacked.clone()] * self._NB_STACK_VOL, dim=1).to(
+                "cuda:0"
+            )
```
romainvo: Use `torch.expand` instead
```suggestion
            x_stacked = x_stacked.unsqueeze(1).expand(-1,-self._NB_STACK_VOL,-1)

File: deepinv/physics/TomographyWithRTK.py [hunk 7]
```diff
@@ -0,0 +1,359 @@
...
+
+        # Cast back from itk cuda image to tensor
+        backproj = torch.as_tensor(
+            Atx, device=y.device
+        ).clone()  # This may not be optimized for memory usage
+
+        if self.mode == "fanbeam":
+            backproj = backproj.sum(dim=1)
+        else:
+            backproj = backproj.permute(2, 1, 0)
```
romainvo: 1. What does this summation step `backproj.sum(dim=1)` correspond to ? 
2. What is the shape of `backproj` ? (W,H,D) ?

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,359 @@
...
+        if self.mode == "fanbeam":
+            projections = projections[:, 0, :]
+
+        if self.normalize:
+            projections /= self.norm_mat
+
+        return projections.unsqueeze(0).unsqueeze(0)
+
+    def A_adjoint(self, y: torch.Tensor, **kwargs) -> torch.Tensor:
+        """Approximation of the adjoint.
```
romainvo: Is the adjoint exact or approximated ? (you can test this with `LinearPhysics.adjointness_test` in practice)

Similar comments than for `TomographyWithRTK.A`:
- can you add documentations on the tensor shapes for 2D and 3D please ?
- would it make sense to initialize `fp_source` and `back_projector` only once and re-use them? 
- avoid cloning if possible

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,359 @@
...
+        parker_angular_gap_threshold: float = 0,
+        truncation_correction_padding: float = 0,
+        hann_cut_frequency: float = 0,
+        **kwargs,
+    ) -> torch.Tensor:
+        """Reconstruct an image from projection data using the FDK algorithm.
+
+        :param torch.Tensor y: input of shape [B,C,...,A,N]
+        :param parker_angular_gap_threshold: Angular gap threshold (in degrees) at which [Parker, Med Phys, 1982] weighting is used.
+        :param truncation_correction_padding: Padding ratio applied to reduce truncation artefacts via [Ohnesorge, Eur Radiol, 1999].
```
romainvo: Can you please add these refs in `docs/source/refs.bib`? you can then cite the works by using:
- ``:footcite:t:`<key>``` to cite in the format *Author et al. [1]*.
- ``:footcite:p:`<key>``` to cite with only the reference number *[1]*.

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,359 @@
...
+        else:
+            backproj = backproj.permute(2, 1, 0)
+
+        if self.normalize:
+            backproj /= self.norm_mat
+
+        return backproj.unsqueeze(0).unsqueeze(0)
+
+    def fbp(
+        self,
```
romainvo: You also need to define the `A_dagger` method, which either call `self.fbp` or the parent method, e.g. something like that

```python
    def A_dagger(self, y: torch.Tensor, fbp: bool = False, **kwargs) -> torch.Tensor:
        if fbp:
            return self.fbp(y, **kwargs)
        else:
            return super(TomographyWithRTK, self).A_dagger(y, **kwargs)
```

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,359 @@
...
+    Supported geometries:
+
+    * ``fanbeam`` (2D-like using the 3D geometry)
+    * ``conebeam`` (3D trajectory)
+
+    The adjoint is implemented using RTK's ray-cast adjoint.
+    The pseudo-inverse is approximated using filtered backprojection reconstruction.
+
+    This implementation requires ITK with CudaCommon v2.1 and RTK v3.
+    """
```
romainvo, line 139: This is a very good start! Ideally, we want the documentation to be as self-contained as possible, I would add a few details:
- could you add an external link to the RTK project's documentation so that users can access more specific informations if they need to ?
- add a small note indicating the the backprojector implements the exact adjoint, as opposed to `TomographiyWithAstra` (which is a key feature for some algorithms)
- a warning, using `.. warning::` that batch elements are processed sequentially and not in parallel (unless you can make this work easily of course). You 
- wraps this "The pseudo-inverse is approximated using filtered backprojection reconstruction." in a `.. warning::` (to be precise it is the Feldkamp Davis Kress algorithm and not fbp right ?)

Some typos:
- use `\forw` instead of `A`
- I would add this "This implementation requires ITK with CudaCommon v2.1 and RTK v3." At the beginning after explaning that it uses RTK backend.

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,359 @@
...
+        geometry: any,
+        projection_stack_information: dict[str, list],
+        volume_information: dict[str, list],
+        mode: str,
+        verbose: bool = False,
+        normalize: bool = False,
+        ray_step_size: float = 0.0,
+        *args,
+        **kwargs,
+    ):
```
romainvo, line 165: If possible, I think we should stick as much as possible to the signature of `TomographyWithAstra`, so as input:
- `img_size`
- `angles`
- `n_detector_pixels`
- `angular_range`,
- `pixel_spacing`
- `bouding_box`
- `geometry_type` (instead of `mode`)

then some translation methods can convert these to the proper format required by `rtk`

This allow to use both wrappers interchangeanbly with minimal overhead. The descriptions of each argument is in the documentation `TomographyWithAstra`, but please reach out if some of it is unclear.

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,359 @@
...
+        :param geometry: RTK geometry object defining source/detector trajectory.
+        :param projection_stack_information: Dictionary describing the projection stack information
+            (size, spacing, origin).
+        :param volume_information: Dictionary describing reconstruction
+            volume grid (size, spacing, origin).
+        :param str mode: Either ``"fanbeam"`` or ``"conebeam"``.
+        :param bool verbose: If True, print geometry configuration.
+        :param bool normalize: If True, normalize operator to unit norm.
+        :param float ray_step_size: Step size along the ray.
+        """
```
romainvo: For proper documentation rendering, the argument docstring should be placed above the `__init__` with the main class docstring.

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,359 @@
...
+        :return: scaled back-projection of shape [B,C,...,H,W]
+        """
+
+        itk, rtk = import_itk_rtk()
+
+        y_stacked = y.squeeze(0).squeeze(0)
+
+        # Add a dimension to simulate a 2D projection in fanbeam mode
+        if self.mode == "fanbeam":
+            y_stacked = torch.stack([y_stacked.clone()] * 1, dim=1).to("cuda:0")
```
romainvo:
```suggestion
            y_stacked = y_stacked.unsqueeze(1)
```

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,359 @@
...
+                continue
+
+            actual_len = len(element)
+            if actual_len != expected_len:
+                raise ValueError(
+                    f"Expected element length {expected_len} for mode {mode!r}, "
+                    f"got {actual_len} for key {key!r} in {name}"
+                )
+
+    def A(self, x: torch.Tensor, **kwargs) -> torch.Tensor:
```
romainvo, line 271: Curiosity and to make sure, what is the shape of the projections produced by RTK? For astra the projections (in 3D) have shape (B,C,V,A,N) where `V` is the number of rows on the detector, `A` the number of angular positions, and `N` the number of columns.

Can you make the docstring a little more detailed please, with explanations on what each dimension of the tensor means?(you can copy-paste from `TomographyWithAstr` if applicable)

deepinv/physics/TomographyWithRTK.py
romainvo, line 1: General comments:

For readibility, I suggest that generic `LinearPhysics` methods (`A`, `A_adjoint`, `A_dagger`) in `TomographyWithRTK` deals minimally with backend-related operations. To illustrate what I mean: the `A` method should take as input a generic (B,C,...,H,W) tensor `x`, and then all processing steps (formatting `x`, setting-up the projector, etc..) can be implemented in other dedicated methods. In the astra wrapper this is implemented in the following way (`self.x_ray_transform` deals with everything else): 

```python
def A(self, x: torch.Tensor, **kwargs) -> torch.Tensor:
    out = AutogradTransform.apply(x, self.xray_transform)

```

Having all the backend logic in a well compartmentalised object makes it easier to wrap it in a custom `torch.Autograd` function (see my comment below)

--

I assume that `torch` cannot track gradient through RTK computations ? (I could not install rtk properly yet and check it)

When supporting a new `Physics` we try to make sure that differentiability with at least the input, is supported. This is notably useful for unrolled methods but also many others. 

For custom backend and `LinearPhysics` one needs to make sure that `A` and `A_adjoint` computations are wrapped with a custom `torch.autograd.Function` (https://pytorch.org/docs/stable/notes/extending.html#extending-torch-autograd). I refer to the `AutogradTransform` class in `physics/functional/astra.py` for an example. 

This is relatively advanced torch functionalities, I can help you with it if you want to.

romainvo (changes requested):
Hi @idrisT11, thank you again for your contribution! I have taken a look at the main class `TomographyWithRTK` for the moment, and very quickly to the test class. This is a very good start, I think we can converge to something really nice by focusing on the following aspects.

First, in terms of features we expect for LinearPhysics, the following two are essential and would need to be addressed;
- `A`, `A_adjoint` should support batched-tensors of shape (B,C,...) with B potentially larger than 1
- the output of `A` or `A_adjoint` is differentiable with respect to the input (using custom `torch.Autograd.function` 

Second, for consistency with `TomographyWithAstra` and for future refactoring, I would focus on having a specific class (or static methods) that handle all the backend related ops, i.e. intializing the geometry, the projectors, pputting x/y in the proper format, etc. (for astra this is the `XrayTransform` class in `deepinv/physics/functional/astra.py`). In turn `TomographyWithRTK` would only call this class or methods to apply forward/backproj operation on `x`or `y` respectively. For example the following block could be moved to a dedicated method to keep `A` clean and modular. This method would then be called by `A`.

```python
        x_stacked = x.squeeze(0).squeeze(0)

        # Add a dimension to simulate a 2D projection in fanbeam mode
        if self.mode == "fanbeam":
            x_stacked = torch.stack([x_stacked.clone()] * self._NB_STACK_VOL, dim=1).to(
                "cuda:0"
            )

        # Cast from tensor to ITK cuda image
        imageSource_cuda = itk.cuda_image_from_cuda_array(x_stacked)
        imageSource_cuda.SetOrigin(self.volume_information["origin"])
        imageSource_cuda.SetSpacing(self.volume_information["spacing"])

        fp_source = rtk.ConstantImageSource[self._CUDA_IMAGE_TYPE].New()
        fp_source.SetSize(self.projection_stack_information["size"])
        fp_source.SetOrigin(self.projection_stack_information["origin"])
        fp_source.SetSpacing(self.projection_stack_information["spacing"])

        # Forward projection: Ax
        forward_projector = rtk.CudaForwardProjectionImageFilter[
            self._CUDA_IMAGE_TYPE
        ].New()
        forward_projector.SetGeometry(self.geometry)
        forward_projector.SetInput(fp_source.GetOutput())
        forward_projector.SetInput(1, imageSource_cuda)
        forward_projector.SetStepSize(self.ray_step_size)
        forward_projector.Update()
        Ax = forward_projector.GetOutput()
        Ax.DisconnectPipeline()

        # Cast back from itk cuda image to tensor
        projections = torch.as_tensor(
            Ax, device=x.device
        ).clone()  # This may not be optimized for memory usage
```

There is more details in my comments. The rest of my remarks is mainly curiosity to understand how rtk should be used at runtime. Please, do not hesitate to reach out if something is not clear, thank you!

romainvo:
I took the liberty of pushing updates to the `pyproject.toml` to handle the new `itk-rtk` dependency. We use [pixi](https://github.com/prefix-dev/pixi) to handle joint `conda` and `pip` dependencies. I've restricted the platform to Linux and `>=2.6.0`. Please update to what seems optimal; for example, I was not sure whether Windows was supported. 

Locally, you can launch the test as ran on the CI with the following `pixi run -e full pytest -n 2 --dist=loadgroup -k "TestTomographyWithRTK"` or if you have your own env setup `python -m pytest -n 2 --dist=loadgroup -k "TestTomographyWithRTK"`.

idrisT11:
Hi @romainvo, thank you for the review ! I'm glad to hear that you are interested.
 
I will answer point by point you message.
 
 
> * `A`, `A_adjoint` should support batched-tensors of shape (B,C,...) with B potentially larger than 1

That's a very good point, this clearly seems important here for the PR. I can update A and A_adjoint to handle inputs with batches. As you said in your comment we could allows the user to fix an arbitary number of batch, but since RTK does not directly accept a fourth batch dimension, I can atleast do it through a loop. I'll also check whether there's a more efficient approach.
 
> * the output of `A` or `A_adjoint` is differentiable with respect to the input (using custom `torch.Autograd.function`)

Noted. I'll add support for autograd by wrapping A/A_adjoint in a custom `torch.autograd.Function`.
 
> Second, for consistency with `TomographyWithAstra` and for future refactoring, I would focus on having a specific class (or static methods) that handle all the backend related ops, i.e. intializing the geometry, the projectors, pputting x/y in the proper format, etc. (for astra this is the `XrayTransform` class in `deepinv/physics/functional/astra.py`). In turn `TomographyWithRTK` would only call this class or methods to apply forward/backproj operation on `x`or `y` respectively. For example the following block could be moved to a dedicated method to keep `A` clean and modular. This method would then be called by `A`.

I agree, I could add in the same file a class `XRayTransformRTK` in the same manner. This will require some refactoring on my part.
 
> There is more details in my comments. The rest of my remarks is mainly curiosity to understand how rtk should be used at runtime. Please, do not hesitate to reach out if something is not clear, thank you!
 
I'll go through your detailed comments and address the remaining points as well. Some of the questions regarding RTK runtime behavior may require a bit more investigation on my side, but I'll follow up on those as I work through the refactor.
 
Thank you for the detailed review and suggestions !

deepinv/physics/TomographyWithRTK.py
idrisT11: That's a good point, the RTK rtk.CudaForwardProjectionImageFilter filter (the projector) can only data of the shape [D, H, W] (tensor shape format). So, as you mentionned, we could allow an arbitrary number of batch dimensions, but since RTK does not natively support a fourth batch dimension, I’ll handle them by looping over the batch elements internally. I will start by looking into that.

deepinv/physics/TomographyWithRTK.py
idrisT11: **For the `SetInput` call:**
`SetInput`  defines the arguments of an ITK filter (here the projector). It takes two parameters,  the argument position (first argument, second argument, ...) and the argument value.
 
As a shorthand we usually write write `forward_projector.SetInput(fp_source.GetOutput())` which is equivalent to `forward_projector.SetInput(0, fp_source.GetOutput())`. This might be the confusing part.
 
**For the `forward_projector.Update()` call:**
 
This is related to the underlying workings of ITK. In ITK, we define a pipeline, where a set of filters are connected from inputs to outputs:
- Once the pipeline is fully defined, it is executed by calling `Update()`. In this case here, this is when the buffer for the output image is allocated, and the computing happens.
- In this specific case, the pipeline contains only one filter, `CudaForwardProjectionImageFilter`. But we could imagine a pipeline composed of two filters, for example `CudaForwardProjectionImageFilter` followed by `CudaBackProjectionImageFilter`, to compute `AA^T`. In that case, the entire pipeline would still be executed with a single call to `Update()`

deepinv/physics/TomographyWithRTK.py
idrisT11: `Ax` here is be of `type`  `itk.CudaImage[itk.F, 3]`. `itk.F` is the itk float type, and `3` is the dimension.
 
If I remember correctly, I did that because the buffer allocated by the Projection filter seems to be freed at the end of the function (I assume this is related to python's scoping rules and ITK's buffer pointer model). Cloning it creates a new buffer and avoids that issue. Another  dirty workaround would be to transfer the image to the CPU and then back to the GPU, which would avoid cloning into the GPU but is still inefficient. But I will check deeper if there is a more proper way to do so with ITK.

deepinv/physics/TomographyWithRTK.py
idrisT11: **For the `backproj.sum(dim=1)`**
The 2D `fanbeam` operator is actually implemented internally in 3D, using the 3D backprojection onto the source–detector plane (as described in the documentation: *"`fanbeam` (2D-like using the 3D geometry)"*).

For interpolation reasons we do not reconstruct onto a volume with a depth of 1 pixel. Instead, we use a volume with a depth of 2 pixels (`_NB_STACK_VOL = 2`), which allows the backprojector to compute the interpolation correctly. We then sum the two slices to recover a volume that is effectively 1 pixel deep.


 **For the shape of `backproj`**
ITK and numpy/torch use different axis conventions. ITK indexes images in x–y–z order, while numpy/torch use z–y–x order. As a result, when converting arrays to or from ITK, the array dimensions are flipped. So here , backproj is represented in ITK as (W, H, D) indeed, so we need to convert it back to the origingal torch format (D, H, W).

Here is a reference if this helps: https://discourse.itk.org/t/convert-from-3d-numpy-array-to-simple-itk/4468/2

deepinv/physics/TomographyWithRTK.py
idrisT11: Sorry, this was an oversight on my part. This is actually the exact adjoint, not the approximate one. We originally intended to implement both versions (exact and approximate) and let the user choose between them, but this is not yet implemented in this PR.

For reusing `fp_source` and `back_projector`, I need to do some checks, I just need to see how the ITK pipelines would need to flow in this case.

I'll also look into the documentation and the cloning parts.

idrisT11:
Hi again,

Here is an update on the RTK backend. I refactored the implementation to address your suggestions and incorporated several of your recommendations. The main changes include:

- Added batch processing by looping over the batch, following the same approach as the Astra backend.

- Removed unnecessary cloning operations, as well as tensor permutations.

- Wrapped the backend around a torch.autograd.Function to integrate with PyTorch's autograd system.

I also made several additional modications, which are described in more detail in my replies to your review comments.

At this point, the `TomographyWithRTK` class should follow the same API, and overall architecture as `TomographyWithAstra`, with only a few backend-specific differences.

The remaining work is to update the tests. They need to be modified to reflect the changes introduced in this refactoring. I’ll aim to do this soon. I will keep you in touch when this gets done.

PS:

> Please update to what seems optimal; for example, I was not sure whether Windows was supported.

I will take a look into that. But for the OS, windows should be supported with RTK, and I’ll confirm the details and update accordingly.

Let me know if you have any further feedback or questions!

deepinv/physics/TomographyWithRTK.py
idrisT11: I needed to check a bit for this, and yes it would be possible.The purpose of this filter is to create a new image buffer for use by subsequent filters. 

In the current implementation, I simplified this by making the filter (projection and backprojection) operate in-place on an already instantiated image, so there is no need yo call `rtk.ConstantImageSource`.

deepinv/physics/TomographyWithRTK.py
idrisT11: For the cloning, the bug was actually fixed in RTK some versions ago, but I didn’t account for that. 

In any case, in the current implementation, the filters (projection and backprojection) operate in-place on a given buffer, so this shouldn't be a problem. The only requirement is that the tensor must be contiguous, as RTK expects contiguous tensors.

deepinv/physics/TomographyWithRTK.py
idrisT11: I have to make some precisions regarding the shape of `backproj` here. In the documentation of TomographyWithRTK, I pointed that D represents the slices of the volume parallel to the rotation plane of the source-detector system. Therefore, the shape is (W, D, H) rather than (W, H, D) in ITK representation. If it seems consistent to you.

The flipping issue during the conversion from ITK to PyTorch tensor still applies, but in my latest commit, I removed the need for permutations by ensuring consistency across the calls.

deepinv/physics/TomographyWithRTK.py
idrisT11: I took these points into account, specifically noting that TomographyWithAstra uses the exact adjoint.

Regarding the FDK/FBP question, Yes, FDK is indeed the algorithm used for cone-beam geometry here. For fan-beam geometry, we apply FDK to the central slice (the slice on the source-detector plane), which is equivalent to FBP.

deepinv/physics/TomographyWithRTK.py
idrisT11: I implemented this specific point, but I also added a few equivalent parameters that are more intuitive to use in RTK. Notably, the geometry parameters that can now be loaded directly from an RTK geometry file, which is more natural when working with RTK.

deepinv/physics/TomographyWithRTK.py
idrisT11: RTK follows a different convention from Astra. In 3D, the shape of the projections is organized as [A, V, N], where the number of angles comes first. This can be interpreted as a stack of individual projections, each with dimensions (V, N), whereas Astra organizes the measurements as a stack of sinograms.

I’ve included the details in the docstring.

deepinv/physics/TomographyWithRTK.py
idrisT11: I managed to get this point done. The backend computations are done in the rtk.py file from physics.functional, specifically in the `XrayTransformRTK` class. I also implemented the `torch.autograd.Function`. Initially, I intended for `TomographyWithRTK` to use the astra `AutogradTransform`, but due to differences in the dimension ordering conventions for 2D projections, I added the logic to a dedicated `torch.autograd.Function` class called `AutogradTransformRTK` to handle the 2D case.

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,337 @@
...
+        *args,
+        **kwargs,
+    ):
+
+        itk, rtk = import_itk_rtk()
+
+        super().__init__(*args, **kwargs)
+
+        self._NB_STACK_VOL = 2
+        self._CUDA_IMAGE_TYPE = itk.CudaImage[itk.F, 3]
```
romainvo, line 172: I assume that RTK only supports CUDA type operations? I would add a `device` argument to the __init__ signature, and raise a warning if the provided device is different than cuda (like we do already in TomographyWithAstra)

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,337 @@
...
+        geometry_parameters: dict[str, float] = MappingProxyType(
+            {
+                "source_radius": 80.0,
+                "detector_radius": 20.0,
+            }
+        ),
+        geometry: any = None,
+        verbose: bool = False,
+        normalize: bool = False,
+        ray_step_size: float = 0.0,
```
romainvo, line 162: I would rather default to None, default to 0.0 might be confusing

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,337 @@
...
+            print(self.volume_information["spacing"])
+
+    def fbp(self, y: torch.Tensor, **kwargs) -> torch.Tensor:
+
+        B, C = y.shape[0:2]
+
+        output = torch.zeros(
+            (B, C) + self.xray_transform.domain_shape, dtype=y.dtype, device=y.device
+        )
+        self.xray_transform.fdk_on_batch(y, output, **kwargs)
```
romainvo, line 264: The output of `fdk_on_batch` will not be differentiable wrt the input, I would add a warning when calling `TomographyWithRTK.fbp` mentioning this. 

The solution to circumvent this is to implement filtering and weighting in pytorch and then call the backprojection operator, I think it would be worth providing the differentiable option to the user, do you think you could built such routing by re-using what we do in `TomographyWithAstra.fbp` ?

deepinv/physics/functional/rtk.py
```diff
@@ -0,0 +1,554 @@
...
+    def _forward_projection(self, x: torch.Tensor, out: torch.Tensor) -> None:
+
+        itk, rtk = import_itk_rtk()
+
+        # Cast from tensor to ITK cuda image
+        imageSource_cuda = itk.cuda_image_from_cuda_array(x)
+        imageSource_cuda.SetOrigin(self.volume_information["origin"])
+        imageSource_cuda.SetSpacing(self.volume_information["spacing"])
+
+        out.mul_(0)  # RTK needs the source to be null
```
romainvo, line 158: You can initialize out with torch.zeros instead of torch.empty in the custom autograd if this is a requirement

deepinv/physics/TomographyWithRTK.py
```diff
@@ -0,0 +1,337 @@
...
+            >>> for i in range(600):
+            ...     geometry.AddProjection(300, 500, i * 360.0 / 600)
+            >>> physics = TomographyWithRTK(
+            ...     geometry=geometry,
+            ...     img_size=(64, 64, 64),
+            ...     n_detector_pixels=(100, 100),
+            ...     geometry_type="conebeam",
+            ...     normalize=False,
+            ...     ray_step_size=0.5,
+            ... )
```
romainvo, line 138: Can you wrap the two examples with a `if torch.cuda.is_available()`? doctests are failing because no gpu is available, i.e. `AttributeError: module 'itk' has no attribute 'CudaImage'`

deepinv/tests/test_external_libraries.py
```diff
@@ -211,3 +211,218 @@ def test_tomography_with_astra_logic(
...
+
+        try:
+            from itk import RTK as rtk
+        except ImportError:
+            pytest.skip("itk-rtk is not installed")
+
+        if not torch.cuda.is_available():
+            pytest.skip("TomographyWithRTK requires a CUDA GPU")
+
+        self.rtk = rtk
```
romainvo, line 232: The test suite won't be skip properly here, "cuda" can be available even if the device is "cpu", the fixture need to take `device` as argument and check its type, i.e if it's cuda or not. No need to define a device fixture, we already do it in conftest.py

```suggestion
    def rtk_env(self, device):
        """Skip the test if itk-rtk or a CUDA GPU is unavailable"""
        itk = pytest.importorskip("itk", reason="itk is not installed")

        try:
            from itk import RTK as rtk
        except ImportError:
            pytest.skip("itk-rtk is not installed")

        if torch.device(device).type != "cuda":
            pytest.skip("TomographyWithRTK requires a CUDA device")

        if not torch.cuda.is_available():
            pytest.skip("TomographyWithRTK requires a CUDA GPU")

        self.rtk = rtk
```

deepinv/tests/test_external_libraries.py
```diff
@@ -211,3 +211,218 @@ def test_tomography_with_astra_logic(
...
+
+        physics = dinv.physics.TomographyWithRTK(
+            geometry=geometry,
+            projection_stack_information=proj_info,
+            volume_information=vol_info,
+            mode=mode,
+            normalize=normalize,
+            ray_step_size=1.0,
+            verbose=False,
+        )
```
romainvo, line 330: It's missing an `img_size` argument which makes the test fail, same comment for the other instances in the tests.

Also it should be `geometry_type` instead of `mode`.

deepinv/tests/test_external_libraries.py
```diff
@@ -211,3 +211,218 @@ def test_tomography_with_astra_logic(
...
+            "size": [D, n_angles],
+            "spacing": [1.0, 1.0],
+            "origin": [-0.5 * (D - 1), -0.5 * (n_angles - 1)],
+        }
+        vol_info = {
+            "size": [W, H],
+            "spacing": [1.0, 1.0],
+            "origin": [-0.5 * (W - 1), -0.5 * (H - 1)],
+        }
+        return proj_info, vol_info
```
romainvo, line 270: This is really nice, I assume this is a util that builds a "proper" geometry from any img size, n_detector_pixel and n_angles? By proper I mean the volume sitting entirely in the field of view, I think it will be very worth it to make them static method of `TomographyWithRTK`

deepinv/tests/test_external_libraries.py
```diff
@@ -211,3 +211,218 @@ def test_tomography_with_astra_logic(
...
+            pytest.skip("itk-rtk is not installed")
+
+        if not torch.cuda.is_available():
+            pytest.skip("TomographyWithRTK requires a CUDA GPU")
+
+        self.rtk = rtk
+
+    @pytest.fixture
+    def device(self):
+        return "cuda:0"
```
romainvo, line 236: We already have a device fixture in `conftest.py` so no need to define it here

deepinv/tests/test_external_libraries.py
```diff
@@ -211,3 +211,218 @@ def test_tomography_with_astra_logic(
...
+
+    @pytest.mark.parametrize("normalize", [True, False])
+    @pytest.mark.parametrize(
+        "mode,img_size,proj_size",
+        [
+            ("fanbeam", (32, 16), 32),
+            ("conebeam", (32, 24, 16), (64, 48)),
+        ],
+    )
+    def test_shapes_and_sanity(self, mode, img_size, proj_size, normalize, device):
```
romainvo, line 300: To describe the type of geometry ("fanbeam" or "conebeam") Can you use `geometry_type` instead of `mode` for consistency

deepinv/physics/functional/rtk.py
```diff
@@ -0,0 +1,554 @@
...
+
+def import_itk_rtk() -> (ModuleType, ModuleType):
+    try:
+        import itk
+        from itk import RTK as rtk
+
+        return (itk, rtk)
+    except ImportError:  # pragma: no cover
+        raise ImportError(
+            "itk-rtk is required to use TomographyWithRTK. It can be installed with itk-rtk-cuda128."
```
romainvo, line 17: I could not find the package, isn't it itk-rtk-cuda124?

deepinv/physics/functional/rtk.py
```diff
@@ -0,0 +1,554 @@
+from __future__ import annotations
+from typing import Any
+from types import ModuleType
+
+import torch
+import math
+
+
+def import_itk_rtk() -> (ModuleType, ModuleType):
```
romainvo, line 9: No need for typing here I think

romainvo (changes requested):
Thank you @idrisT11 for the many improvements, it is very great, I feel like we're converging to something! It looks like the tests are failing, I could not install a proper setup locally either, specifically the `cuda_image_from_cuda_array` utils seems to require `itk_cudacommon_cuda124 >= 2.0.0` but `itk_rtk_cuda124` is pinned on cuda_common 1.1.x ?

SimonRit:
> Thank you @idrisT11 for the many improvements, it is very great, I feel like we're converging to something! It looks like the tests are failing, I could not install a proper setup locally either, specifically the `cuda_image_from_cuda_array` utils seems to require `itk_cudacommon_cuda124 >= 2.0.0` but `itk_rtk_cuda124` is pinned on cuda_common 1.1.x ?

Thanks for the feedback. We've been trying to release [RTK v3](https://github.com/RTKConsortium/RTK/milestone/1) for a long time, which is what you need. The CI currently builds versions for Cuda 12.4, 12.8 and 13.0 as GitHub artifacts which will be published on Pypi when we tag it. You can either download the package [here](https://github.com/RTKConsortium/RTK/actions) for testing or wait for the RTK release in the coming weeks.

idrisT11:
Hi again @romainvo,

Sorry for the lack of response, and thanks for the feedback!

Regarding the `itk_cudacommon_cuda124` version, as @simonrit mentioned, the most recent RTK version currently available on pip is not pinned to `itk_cudacommon==2.0.0` (the version we currently need). Sorry, I didn’t notice this since I was working directly with the development version of RTK. In the meantime, as Simon suggested, while waiting for the release you can download a wheel for testing.

Regarding `test_external_libraries.py`, I haven’t updated the file yet to account for the modified `TomographyWithRTK` interface from my previous commit. I’m still working on it.

I’ll try to update all of this after my vacation.

deepinv/physics/TomographyWithRTK.py
idrisT11: It does support CPU operations. It might just increase the code complexity a bit, since we would need to use different projectors and make some modifications to the ITK pipelines. But, it is something we could implement. I will check what would be required.

deepinv/physics/TomographyWithRTK.py
idrisT11: It surely is possible. I’ll look into that. As you said, we would need to re-implement the filtering and weighting in PyTorch like in the astra functions, I will check if I can just copy the astra implementation.

deepinv/physics/functional/rtk.py
idrisT11: Yes, RTK here needs it to be initialized to zero. We could use torch.zeros to unify the interface of the Autograd between Astra and RTK, at least for the 3D case.

deepinv/physics/TomographyWithRTK.py
idrisT11: I can do this. But regarding GPU support here, as mentioned in the previous comment, I’ll look into implementing the CPU operation as well. This will remove the need for a cuda check here.

deepinv/tests/test_external_libraries.py
idrisT11: Not really. It does not guarantee that the entire object is within the FOV, since it does not fix the source-to-detector or source-to-isocenter distances. But since this function centers the object in the middle of the scene,  it only guarantees that the "center" of the object is within the FOV.

However, we could add this as a static method. I can see how that could be useful.

deepinv/physics/functional/rtk.py
idrisT11: Sorry, you’re right, it is `itk-rtk-cuda124`. But more generally, I think we should change this comment to `itk-rtk-cuda12X`, since the exact package depends on the locally installed CUDA version, and add a link to the available `itk-rtk` CUDA versions. I'll do that.

PR #1234, opened 2026-06-15
CI fixes: add caching to avoid multiple downloads, fix testmon, http errors as warnings

### LINKED ISSUE #1229: Cache data downloaded during tests / doc builds
### Environment (optional)

_No response_

### Describe the feature

We should cache data downloaded during CI steps so that it is reused across several CI runs, probably with the [cache Action](https://github.com/actions/cache) ?
Another possible option directed towards scientific libraries trying to integrate data in their tests / docs is [Pooch](https://www.fatiando.org/pooch/latest/index.html), which seems to be used [by scikit-image](https://scikit-image.org/docs/stable/user_guide/install.html#downloading-all-demo-datasets) and also [by scipy](https://docs.scipy.org/doc/scipy/reference/datasets.html#how-dataset-retrieval-and-storage-works).

### Motivation

Downloading all images for each CI run is time consuming and probably also energy consuming, while the data very rarely, if ever, changes.

### Alternatives considered

_No response_

### Additional context

See [the PR](https://github.com/scikit-image/scikit-image/pull/3945) for Pooch integration in scikit-image
Just adding this quote from [Pooch's doc](\cite{nollConvergenceEMAlgorithm2026})
> When downloading data repeatedly, like in continuous integration, failures can occur due to sporadic network outages or other factors outside of our control. In these cases, it can be frustrating to have entire jobs fail because a single download was not successful.

We could also take inspiration from [nilearn's CI](https://nilearn.github.io/dev/ci.html)
_No response_

### LINKED ISSUE #1230: Monitor amount of downloaded data during CI
### Environment (optional)

_No response_

### Describe the feature

We should monitor the amount of data downloaded by each CI step.
Ideally we could also monitor usage over a time period.
This way we have an idea about the impact of possible solutions such as caching (#1229), which might not be clear over a single CI run.

### Motivation

In order to reduce the amount of redundant downloads, we should first try to measure this volume.

### Alternatives considered

_No response_

### Additional context

Related to #1200 and #1229

tachella:
Recent tests have been failing because some demo images are downloaded multiple times during the test suite, resulting in random HTTP errors. This PR makes a step towards providing a solution, by:

- caching demo downloads in `get_cache_home()`
- fixing `testmon` config to avoid re-running unnecessary examples
- avoiding re-downloads of archives
- skip http errors, and instead print them in the terminal summary (this should be a temporary fix until we move to a better solution - see #1229 
- Increase the import time tests to 10% thresholds (instead of 5%) to avoid random errors


.github/workflows/test_cpu.yml
```diff
@@ -35,6 +35,11 @@ jobs:
     env:
       DEEPINV_MOCK_TESTS: "True"
       DEEPINV_DOWNLOAD_VERBOSE: "False"
+      # Persist deepinv's URL cache to a workspace-local folder so it can be
+      # cached across CI runs (see actions/cache steps below). This avoids
+      # re-downloading example images, models and datasets from HuggingFace
+      # and prevents 429 rate-limit failures.
+      DEEPINV_CACHE_DIR: ${{ github.workspace }}/.deepinv-cache
```
romainvo: I think we should use `${{ github.workspace }}/.cache/deepinv` for consistency (.e.g used in `get_cache_home`, `test_gpu.yml`) ?


```suggestion
      DEEPINV_CACHE_DIR: ${{ github.workspace }}/.cache/deepinv
```

.github/workflows/test_cpu.yml
```diff
@@ -46,24 +51,39 @@ jobs:
...
           path: .testmondata
           key: testmon-cpu-main-${{ matrix.environment }}-${{ matrix.os }}-${{ github.run_id }}
           restore-keys: |
-            testmon-cpu-main-${{ matrix.os }}
+            testmon-cpu-main-${{ matrix.environment }}-${{ matrix.os }}-
+
+      - name: Restore deepinv URL cache # shared across matrix environments
+        uses: actions/cache/restore@v4
+        with:
+          path: ${{ github.workspace }}/.deepinv-cache
```
romainvo: For readibility and flexibility


```suggestion
          path: ${{ env.DEEPINV_CACHE_DIR }}
```

.github/workflows/test_cpu.yml
```diff
@@ -46,24 +51,39 @@ jobs:
...
+        uses: actions/cache/save@v4
         with:
           path: .testmondata
           key: testmon-cpu-main-${{ matrix.environment }}-${{ matrix.os }}-${{ github.run_id }}
 
+      - name: Save deepinv URL cache # only on push to main, mirrors testmondata
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        uses: actions/cache/save@v4
+        with:
+          path: ${{ github.workspace }}/.deepinv-cache
```
romainvo:
```suggestion
          path: ${{ env.DEEPINV_CACHE_DIR }}
```

deepinv/utils/io.py
```diff
@@ -45,19 +49,95 @@ def load_torch(
...
+        return {}
+    with open(index_path, "r", encoding="utf-8") as file:
+        return json.load(file)
+
+
+def _save_url_index(index_path: Path, index: dict[str, str]) -> None:
+    tmp_path = index_path.with_suffix(".json.part")
+    with open(tmp_path, "w", encoding="utf-8") as file:
+        json.dump(index, file, indent=2, sort_keys=True)
+    tmp_path.replace(index_path)
```
romainvo: Not clear to me what is the purpose of `index.json` here (and related utils `_load_url_index` and `_save_url_index`) ?

In practice it looks like `_get_url_cache_path` is sufficient to resolve the proper cache_path ? If `index.json` only serves as passive log I think we can safely remove it, or did I miss something ?

```suggestion
def load_url(url: str, **kwargs) -> BytesIO:
    """Load URL to a buffer.

    This can be used as the argument for other IO functions such as
    :func:`deepinv.utils.load_torch`, :func:`deepinv.utils.load_np` etc.
    to load data directly from a URL.

    Downloaded content is cached under :func:`deepinv.utils.get_cache_home`
    so repeated calls for the same URL do not hit the network again.

    :param str url: URL of the file to load
    :return: `BytesIO` buffer.
    """
    cache_home = get_cache_home()
    cache_dir = cache_home / "url_cache"
    cache_dir.mkdir(parents=True, exist_ok=True)

    cache_path = _get_url_cache_path(url, cache_dir)

    if not cache_path.exists():
        tmp_path = cache_path.with_suffix(cache_path.suffix + ".part")
        try:
            with requests.get(url, stream=True) as response:
                response.raise_for_status()
                with open(tmp_path, "wb") as file:
                    for chunk in response.iter_content(chunk_size=1024 * 1024):
                        if chunk:
                            file.write(chunk)
            tmp_path.replace(cache_path)
        finally:
            if tmp_path.exists():
                tmp_path.unlink()

    return BytesIO(cache_path.read_bytes())


def _get_url_cache_path(url: str, cache_dir: Path) -> Path:
    parsed_url = urlparse(url)
    suffix = "".join(Path(parsed_url.path).suffixes) or ".bin"
    cache_key = hashlib.sha256(url.encode("utf-8")).hexdigest()
    return cache_dir / f"{cache_key}{suffix}"
```

deepinv/tests/conftest.py
romainvo, line 1: Currently digging into this, I feel like we should test this at least minimally, check that it properly skips network related errors, and does nothing for other errors. Otherwise we risk being potentially blind to real network errors (e.g dead links). For the moment I am just checking manually with tests like this

```python
# Test 1: Verify that a ConnectionError is converted to a skip
def test_connection_error_is_skipped():
    """This test should be skipped due to a ConnectionError."""
    raise ConnectionError("Check that ConnectionError is skipped")

# Test 2: Verify that a URLError is converted to a skip
def test_url_error_is_skipped():
    """This test should be skipped due to a URLError."""
    from urllib.error import URLError
    raise URLError("Check that URLError is skipped")

def test_ssl_error_is_skipped():
    """This test should be skipped due to an SSL error."""
    import ssl
    raise ssl.SSLError("Check that SSLError is skipped")
```

we can keep something like for the moment and think of more elaborate testing afterwards.

---

Other point: it looks like the current implem is not compatible with `pytest-xdist`, i.e. when using `-n 2 dist=loadgroup` no summary of failed network errors is printed. This is because worker processes have their own scope, while it looks like `pytest_terminal_summary` is executed on the main process (and does not have access to the skipped recorded test). There are multiple solutions:

1. use the hooks only for the required tests, and use `xdist` elsewhere, the problem is we're loading images and weights pretty much everywhere..
2. we could drop the use of `pytest-xdist` (not much of a solution as the CI is already quite slow)
3. we can also save the recorded skipped tests in some file and have the main process read it. There shouldn't be too much i/o as we expect the number of skipped test to be small
4. we handle the comms manually, so that the main process fetches skipped tests from the workers, don't know if 'xdist` supports this though

romainvo (changes requested):
Thanks @tachella for dealing with this! The caching mechanisms you added look good to me, I have some comments but not related to the logic.

Concerning the added `pytest_runtest_makereport` and `pytest_terminal_summary`, I tested them locally by raising some dummy errors, it looks like the current implementation is not compatible with `pytest-xdist`, the http/network related failures are skipped but the summary of failed tests is not printed (without xdist it works without issue). I proposed some solutions, it's possible there are other workarounds. I can take a closer look when I have time.

.github/workflows/import_time.yml
```diff
@@ -97,7 +97,7 @@ jobs:
             const baseMean = baseResult.mean;
 
             const rel_diff = (headMean - baseMean) / baseMean * 100;
-            const threshold = 5; // 5% regression threshold
+            const threshold = 10; // 10% regression threshold
```
Tmodrzyk, line 100: I'm not sure that this increase is related to this PR ? 
Was it voluntary ?

Tmodrzyk:
Hi @tachella I'm taking a closer look at this, but for now I can't say I like the logic of looking through error messages for keywords related to network. I think this will skip many errors weakly related to the network.

deepinv/tests/conftest.py
```diff
@@ -7,6 +7,89 @@
...
+    "connectionerror",
+    "connectionreseterror",
+    "connectionaborted",
+    "connectionrefused",
+    "connectiontimeout",
+    "sslerror",
+    "PytorchStreamReader",
+    "urlerror",
+    "urllib",
+)
```
Tmodrzyk: This seems *very* risky to me. 
We will probably end up skipping a lot of errors that are only weakly related to the random http errors we aimed for.
I'm also not sure that these keywords alone cover the error types we are aiming for.
So this seems like this will do both too little and too much.

deepinv/utils/io.py
```diff
@@ -45,19 +49,95 @@ def load_torch(
...
     """
-    response = requests.get(url)
-    response.raise_for_status()
-    return BytesIO(response.content)
+    cache_home = get_cache_home()
+    cache_dir = cache_home / "url_cache"
+    cache_dir.mkdir(parents=True, exist_ok=True)
+
+    cache_path = _get_url_cache_path(url, cache_dir)
+    index_path = cache_dir / "index.json"
```
Tmodrzyk: Same comment as @romainvo, this index seems redundant. The paths are already handled by being found or not in the folder by `_get_url_cache_path()`.

deepinv/utils/demo.py
```diff
@@ -11,7 +11,7 @@
 import numpy as np
 import torch
 from torchvision import transforms
-from deepinv.utils.io import load_np, load_torch, load_url as _load_url
+from deepinv.utils.io import get_cache_home, load_np, load_torch, load_url as _load_url
```
Tmodrzyk: I think this renaming is confusing, we should stick to `load_url`

deepinv/utils/io.py
```diff
@@ -45,19 +49,95 @@ def load_torch(
...
+            with requests.get(url, stream=True) as response:
+                response.raise_for_status()
+                with open(tmp_path, "wb") as file:
+                    for chunk in response.iter_content(chunk_size=1024 * 1024):
+                        if chunk:
+                            file.write(chunk)
+            tmp_path.replace(cache_path)
+        finally:
+            if tmp_path.exists():
+                tmp_path.unlink()
```
Tmodrzyk, line 140: I'm not familiar with the logic of this code, but this seems like the place where we should catch the specific http related errors and raise them with a specific type which can then be handled by pytest, i.e skipping them.

Tmodrzyk (changes requested):
Thanks @tachella for taking the time to try to fix this. 
The caching mechanism seem ok overall, except the few things @romainvo pointed out, and a few other minor comments I left.

However, I think the current logic used to handle and skip the network errors is very fragile and risky.
We should definitely not be checking text and types among a list of keywords in the hope to hit the errors we mean to target, and skipping a lot of others that do contain these keywords but are unrelated to our issues.

What we could do instead is focus the part of the code where we make the request inside `load_url()` and catch a list of errors there, raise a specifc error type if the network issues happen. This specific error type could then be caught by `pytest` to skip the affected tests.

.github/workflows/test_cpu.yml
tachella: thanks for spotting this, applied!

.github/workflows/test_cpu.yml
tachella: done

deepinv/utils/io.py
tachella: yes I agree, the json is not necessary - we can just check the path to see if the data has been cached already - I've applied the changes

deepinv/tests/conftest.py
tachella: Great catch on `xdist`. I did a bit of digging (with help of LLMs), and it seems that it's possible to pass the information to the collector process and avoid this silent removal of network errors. Should be fixed now

.github/workflows/import_time.yml
tachella: yes it was me, the test was failing randomly now and then..

deepinv/tests/conftest.py
tachella: I agree! I have proposed a new solution (see comment below)

deepinv/utils/io.py
tachella: yes indeed! I've removed the json

deepinv/utils/demo.py
tachella: this is from a previous PR and necessary for the deprecation message

deepinv/utils/io.py
tachella: indeed! I changed the strategy to raise a custom DownloadException here and collect it later in pytest as a download error

tachella:
Thanks @Tmodrzyk and @romainvo for the careful reviews!! I applied your suggested changes. The three main points are:

- Removing the unnecessary `.json` file for saving the list of cached files. We can get this list directly by checking the paths.
- I removed the 'generic' http error catching, which was risky, as it could mask other errors. Instead, I defined a custom exception `DownloadError` which is used on functions downloading files, and can be used to distinguish tests that fail in these functions.
- Added a fix to avoid having silent download errors when using `xdist`

Let me know if you have any other comment :)

deepinv/models/utils.py
```diff
@@ -525,5 +533,14 @@ def load_state_dict_from_url(*args, **kwargs) -> dict:
...
-    with ctx:
-        return torch.hub.load_state_dict_from_url(*args, **kwargs)
+    try:
+        with ctx:
+            return torch.hub.load_state_dict_from_url(*args, **kwargs)
+    except (
+        urllib.error.URLError,
+        ConnectionError,
+        TimeoutError,
+        socket.timeout,
```
Tmodrzyk: It seems `socket.timeout` is just a [deprecated alias](https://realpython.com/ref/builtin-exceptions/timeouterror/) for `TimeoutError`.
Not an expert on this, but we can probably safely add a few other errors:
- `socket.gaierror` if the server infos change during the download
- `http.client.IncompleteRead` if the download is interrupted

Would like to have inputs from other reviewers that are more confident on these issues though.

deepinv/utils/io.py
```diff
@@ -45,19 +47,104 @@ def load_torch(
...
+    cache_home = get_cache_home()
+    cache_dir = cache_home / "url_cache"
+    cache_path = _get_url_cache_path(url, cache_dir)
+
+    if not cache_path.exists():
+        cache_path.parent.mkdir(parents=True, exist_ok=True)
+        tmp_path = cache_path.with_suffix(cache_path.suffix + ".part")
+        try:
+            with requests.get(url, stream=True) as response:
+                response.raise_for_status()
```
Tmodrzyk: We could add a `timeout` here to prevent waiting indefinitely for server response / errors ? It seems this is common practice [https://3.python-requests.org/user/quickstart/#timeouts](https://3.python-requests.org/user/quickstart/#timeouts)

Tmodrzyk (commented):
Thanks for the changes @tachella ! We could probably have better granularity for the handling of requests errors but I think this is ok for now and will definitely catch flaky downloads.
I have a few minor comments but otherwise this seems ready to go.

Edit: we should also definitely point out the increased allowed import time from 5 to 10% in the PR summary !

deepinv/tests/conftest.py
```diff
@@ -3,10 +3,91 @@
...
+
+    if not call.excinfo.errisinstance(DownloadError):
+        return
+
+    typename = call.excinfo.typename
+    msg = str(call.excinfo.value)
+
+    # user_properties is part of TestReport and is preserved by
+    # pytest-xdist's report (de)serialization, so the controller sees it.
+    report.user_properties.append((_DEEPINV_DOWNLOAD_ERROR_PROP, (typename, msg)))
```
romainvo, line 55: Very neat trick 👍

deepinv/models/utils.py
romainvo: Looks like handling ambigous errors from `urllib` is more tricky than with `requests`. 

Agreed on adding @Tmodrzyk's suggestions, I feel like this is a good start, then we can cover more exceptions in the future if we see this is not robust enough ? (not confident enough to suggest more ahah)

deepinv/utils/io.py
```diff
@@ -45,19 +47,104 @@ def load_torch(
...
+    cache layout mirrors the URL: a file fetched from
+    ``https://huggingface.co/datasets/deepinv/images/resolve/main/celeba_example.jpg``
+    is stored at
+    ``<cache_home>/url_cache/huggingface.co/datasets/deepinv/images/resolve/main/celeba_example.jpg``.
+    Two URLs that differ only in their query string share the same cache
+    entry — fine for the ``?download=true`` query used by HuggingFace.
+
     :param str url: URL of the file to load
     :return: `BytesIO` buffer.
+    :raises DownloadError: if the file cannot be downloaded.
```
romainvo: Add `deepinv.utils.DownloadError` to .rst files for doc building

deepinv/utils/io.py
```diff
@@ -45,19 +47,104 @@ def load_torch(
...
+    """Raised when a network download initiated by deepinv fails.
+
+    Wraps any underlying network error (HTTP status, connection failure,
+    SSL error, DNS failure, timeout, …) so that callers — and the test
+    suite in particular — can detect download failures with a single
+    ``except DownloadError:`` rather than enumerating every exception
+    type that the network stack may raise. The original exception is
+    chained via ``__cause__``.
+
+    Inherits from :class:`requests.exceptions.RequestException` (and
```
romainvo, line 84: Add `requests` to intersphinx-mapping in `docs/source/conf.py` for external link

romainvo (changes requested):
Thanks @tachella for the changes! Looks like we're converging to something clean, I don't have specific comments (minor doc related), handling errors from `torch.hub.load_state_dict_from_url` seems to be the main difficulty now. I'm good with the current state but always open to discuss it further :)

deepinv/models/utils.py
tachella: thanks for pointing out the deprecation, I removed the `socket.timeout`, and added `socket.gaierror`. I think this should be ok for now, we can add more later on

deepinv/utils/io.py
tachella: good catch! I added one :)

deepinv/utils/io.py
tachella: indeed, done!

deepinv/utils/io.py
tachella: done!

tachella:
@romainvo @Tmodrzyk, I applied your suggestions and now tests are passing, let me know if you want to make more changes or we can merge

romainvo (approved):
Looks good to me, thanks @tachella!

Tmodrzyk (approved):
LGTM, don't forget the `changelog.rst` entry !

PR #1235, opened 2026-06-15
Add gradient accumulation option to Trainer

tachella:
- adds the possibility of using gradient accumulation across batches to `Trainer`
- fixes a small ckp saving bug that didn't allow saving the ckp after the first completed epoch

### DIFF
--- a/deepinv/training/trainer.py
+++ b/deepinv/training/trainer.py
@@ -86,6 +86,9 @@ class Trainer:
         using :class:`deepinv.utils.AverageMeter` to deal with uneven batch sizes. Default is :class:`supervised loss <deepinv.loss.SupLoss>`.
     :param float grad_clip: Gradient clipping value for the optimizer. If None, no gradient clipping is performed. Default is None.
     :param bool optimizer_step_multi_dataset: If ``True``, the optimizer step is performed once on all datasets. If ``False``, the optimizer step is performed on each dataset separately.
+    :param int accum_gradients: Number of training steps to accumulate gradients before calling ``optimizer.step()``.
+        A single training step already accumulates gradients from the ``G`` operators when ``optimizer_step_multi_dataset=True``,
+        so ``accum_gradients=1`` preserves the current behavior.
 
     .. note::
 
@@ -285,6 +288,7 @@ class Trainer:
     no_learning_method: str | Reconstructor = "A_adjoint"
     grad_clip: float = None
     check_grad: bool = False
+    accum_gradients: int = 1
     wandb_vis: bool = False
     mlflow_vis: bool = False
     wandb_setup: dict = field(default_factory=dict)
@@ -323,6 +327,9 @@ class Trainer:
                 DeprecationWarning,
                 stacklevel=2,
             )
+        assert (
+            isinstance(self.accum_gradients, int) and self.accum_gradients > 0
+        ), "accum_gradients should be a positive integer."
         # Cache flag for whether model.forward accepts 'update_parameters'
         self._model_accepts_update_parameters = False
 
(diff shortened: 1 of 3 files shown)


deepinv/training/trainer.py
```diff
@@ -1466,9 +1485,9 @@ def train(
                 self.scheduler.step()
 
             if (
-                epoch > 0 and epoch % self.ckp_interval == 0
+                epoch + 1 > 0 and (epoch + 1) % self.ckp_interval == 0
             ) or epoch + 1 == self.epochs:
-                self.save_model(f"ckp_{epoch}.pth.tar", epoch)
+                self.save_model(f"ckp_{epoch + 1}.pth.tar", epoch + 1)
```
Andrewwango, line 1490: Hmm this is not a bug, it's just a different convention (i.e. 0-indexing or 1-indexing). So changing this may bring unnecessary behaviour change to users. If you want to make sure the first ckp saves, then why not do in the `if` block above instead:

```
if epoch == 0 or (
     epoch > 0 and epoch % self.ckp_interval == 0
) or epoch + 1 == self.epochs:
```

deepinv/training/trainer.py
```diff
@@ -86,6 +86,9 @@ class Trainer:
         using :class:`deepinv.utils.AverageMeter` to deal with uneven batch sizes. Default is :class:`supervised loss <deepinv.loss.SupLoss>`.
     :param float grad_clip: Gradient clipping value for the optimizer. If None, no gradient clipping is performed. Default is None.
     :param bool optimizer_step_multi_dataset: If ``True``, the optimizer step is performed once on all datasets. If ``False``, the optimizer step is performed on each dataset separately.
+    :param int accum_gradients: Number of training steps to accumulate gradients before calling ``optimizer.step()``.
```
Andrewwango, line 89: the name `accum_gradients` is a bit vague, since technically `optimizer_step_multi_dataset` is also accumulating gradients. The explanation also alludes to this. Why not `optimizer_step_multi_samples`? Then you don't need the explanation below.

deepinv/training/trainer.py
```diff
@@ -323,6 +327,9 @@ def __post_init__(self):
                 DeprecationWarning,
                 stacklevel=2,
             )
+        assert (
```
Andrewwango, line 330: raise value error

deepinv/training/trainer.py
```diff
@@ -1057,7 +1068,15 @@ def step(
         if self.log_train_batch and train:
             self.log_metrics_mlops(logs, step=train_ite, train=train)
 
-        if train and self.optimizer_step_multi_dataset:
+        if (
+            train
+            and self.optimizer_step_multi_dataset
+            and (
+                train_ite is None
+                or (train_ite + 1) % self.accum_gradients == 0
```
Andrewwango, line 1076: why + 1 here but not above?

Andrewwango (changes requested):
thanks Julian for this useful feature

nucli-vicky (commented):
I like the addition :)

One thought: When the dataloader doesn't have `drop_last=True`, there might be some non-ideal behavior at the end of the epoch:

Let's take a batch_size of 64. With `drop_last=False`, the last batch ($`b_i`$) may be only 3 samples. In the current implementation, the loss (and therefore gradient) of $`b_i`$ is weighted equally as that of $`b_{i-1}`$. Instead, it would make more sense to weigh them according to the number of samples in each.

So (with as in current code, the loss of each batch already being averaged), $`\frac{64 L_{i-1} + 3 L_i}{64 + 3}`$, which would match the gradient produced by a single batch containing the samples of $`b_{i-1}`$ and $`b_i`$.

That way, the gradients generated by each sample are treated equally, instead of giving over-importance to gradients from the few samples in $`b_i`$

PR #1243, opened 2026-06-24
Hotfix pixi

jscanvic:
Pixi just released a new version 2 hours ago which might have broke CI: 536d13d41900bc2e45adb829a34ae5279200bb63 passed 4 hours ago but 536d13d41900bc2e45adb829a34ae5279200bb63 pushed after the release failed

See:
https://github.com/prefix-dev/pixi/releases/tag/v0.71.0

In this PR I pin down the previous version until we figure out a better fix


romainvo (approved):
Thanks for the quick fix @jscanvic !

Issue #1182, opened 2026-05-05
Add Contrast metrics


Tmodrzyk:
### Describe the feature

Add two domain-specific metrics widely used in emission tomography:

- the Contrast to Noise Ratio (CNR) which measures the contrast in reconstructions defined by:
$CNR = \frac{\mu_{roi} - \mu_{bg}}{\sigma_{bg}}$ where $\mu$ is the mean and $\sigma$ the standard deviation, and "bg" stands for background.  
see a python implementation here: https://3dct.github.io/vIQA/nr_metrics/generated/viqa.nr_metrics.cnr.contrast_to_noise_ratio.html#viqa.nr_metrics.cnr.contrast_to_noise_ratio
The metric should expect a reconstruction, and masks for background and regions of interest.

- the Contrast-Recovery (CR) which measures how much of the contrast we recover, and which is defined as:
$CR = \frac{\frac{\mu_{roi}}{\mu_{bg}} - 1}{\frac{a_{roi}}{a_{bg}} - 1}*100$ 
where the numerator is the contrast in the reconstruction, and the denominator the contrast in the ground-truch, with $a$ standing for the ground-truth activity in ROIs and background.
The metric should expect a reconstruction, a ground-truth and masks for background and ROIs.

### Motivation

These metrics are very standard for the evaluation of emission tomography reconstructions and users will need them to evaluate their methods.

nucli-vicky:
If we are going to have mask based metrics: simply relative (or absolute) difference in mean voxel value inside mask (SUVmean) could also be useful, at least in clinical scan evaluation (where the mask can represent a lesion). Same for something like metabolic tumor volume within a mask

Tmodrzyk:
Indeed ! I'm curious if you have another mechanism in mind than using masks ?

nucli-vicky:
@Tmodrzyk Masks seem good, I'm just not sure if the current `trainer.test` method supports providing masks to `metric` functions?

nucli-vicky:
I think being able to pass data aside from measurement, ground-truth, prediction, would be nice to have in a general format, not just masks.

Tmodrzyk:
Ah, I was supposed to refactor the ``Trainer`` last year but the PR has unfortunately gone stale.

However I think this applies more broadly to the library, passing additional data could be useful for physics too: e.g in [this PR](https://github.com/deepinv/deepinv/pull/1099) we can also pass `background` and `attenuation` to the `physics` on top of the measurements.

[...]

PR #1244, opened 2026-06-24
Use dictionary for dataset outputs (#674)

### LINKED ISSUE #674: Improve the format of dataset entries
### Environment (optional)

_No response_

### Describe the feature

Dataset entries are currently (unnamed) tuples following certain conventions (e.g. ground truths always appear before measurements). Those are very cumbersome to deal with for dataset users.

Want to get y? If the tuple has length at least 2 and its 2nd entry has the right type (Tensor or TensorList as opposed to dict) then it's y. Want to get params? If the last entry has the right type, then it's params. Want to get x? It's the 1st element (and it might be NaN!) Rinse and repeat everywhere an arbitrary dataset is used.

I suggest to make it so we can use `y = entry.get("y", None)` or `y = getattr(entry, "y", None)` and so on for x and params.

The main difficulty is to avoid breaking existing code. It should be possible to do so by simply using named tuples instead of unnamed tuples and deprecating positional access.

(Related: #622)

### Motivation

/

### Alternatives considered

_No response_

### Additional context

_No response_

romainvo:
This PR deprecates the tuple format and implements the more flexible dictionary output format for the datasets #674. This is more in line with standard practice in python for passing data, and potentially useful for other features:
- #1182
- allows passing additional data (e.g. costly initial reconstruction) that are not parameters of a `physics`.

### Context

A single class inherits from `torch.utils.data.Dataset`, i.e., `ImageDataset`; all other classes inherit from it or its child classes. 
- In the docstring of `ImageDataset` I'm adding a warning to inform about the deprecated tuple format (while still keeping how the tuple format works). 
- A single `use_dict_output` flag in the constructor of all datasets can be used to opt in to the new format (set to False by default for backward compatibility). 
- A warning is raised if `use_dict_output=False`

### Custom propagation

- Most datasets have custom `__getitem__` implementation so I've basically done the update by hand exhaustively looking at each file in the `datasets` module. Adding a flag to each constructor and duplicating this tiny bit of code is not ideal, but I don't really see another solution, open to ideas! :)
- The `check_dataset` in `deepinv/datasets/base.py` now supports both formats but will raise a warning if a dataset implements the tuple format.
- I've added two utilities in the `datasets/base.py` to handle unpacking of entries regardless of the format across the library, e.g. `unpack_batch` unpacks a dataset output in the canonical `x, y, params` format.

### Functionalities that consume a dataset/dataloader

- I have updated the `Trainer` to seamlessly handle the new format.
- The `GaussianMixtureModel` in `deepinv/optim/utils.py` consumes a dataloader in its `fit` method
- The `MMSE` denoiser in `deepinv/models/mmse.py`

### Tests / examples
-  In `test_trainer.py` I am testing both formats only on a handful of tests to not increase CI time too much, i.e. only (`test_get_samples` and `test_dataloader_formats`), in the rest of the tests I'm using the new `use_dict_output=True` format.
- In `test_datasets.py` I am exhaustively checking for each dataset that both formats work properly.
- For the rest of the tests, I'm making the `use_dict_output=True` the default format whenever a deepinv dataset is used, mostly concerns the `DummyCircles` dataset used throughout the test suite, and a lot of `HDF5Dataset`.
- Same for the examples, I'm using `use_dict_output` as the default output format

### LLM policy


deepinv/datasets/base.py
```diff
@@ -12,6 +12,7 @@
 from PIL.Image import Image as PIL_Image
 from deepinv.utils.tensorlist import TensorList
 from natsort import natsorted
+from collections import OrderedDict
```
Andrewwango: Do we still need ordereddict post python 3.7, since dicts natively preserve insertion order?

deepinv/datasets/base.py
```diff
@@ -95,6 +96,34 @@ def warn_core_types(x):
...
+                raise RuntimeError(
+                    f"{error}, but dict element with key 'y' is type {type(x)}."
+                ) 
+                
+        if "x" not in batch and "y" not in batch:
+            raise RuntimeError(
+                f"{error}, batch is type dict but neither 'x' nor 'y' keys are present"
+            )
+              
+        if "params" in batch and any(
```
Andrewwango: Does it make sense to do a nested dict, or have a flat dict i.e. `{"x": x, "y": y} | params`?

Andrewwango (commented):
minor comments

deepinv/datasets/base.py
Andrewwango: or is to allow something like accessing dict values by index?

Andrewwango:
Maybe we could add a unit test covering typical use cases of how users access data from datasets, and make sure that the new feature covers all the cases, e.g.:

- iterating their own loops i.e. `for (x, y) in iter(loader):`
- using in `dinv.Trainer(dataloader)`
- using in `dinv.test(dataloader)`

deepinv/datasets/base.py
romainvo: I would keep he nested dict, it allows to select `out['params']` without filtering keys manually, and e.g. feed it to a physics in a single go `physics.update(**out['params'])`, which is more in line with the current workflow

deepinv/datasets/base.py
romainvo: Sorry, I'm stuck in the past, I'll switch to `dict`

deepinv/datasets/base.py
```diff
@@ -50,58 +52,157 @@ def warn_core_types(x):
...
+    (nested) dict with optional ``"x"``, ``"y"``, ``"params"`` keys.
+
+    :param batch: a single batch as returned by a dataset's `__getitem__` (or by a
+        `torch.utils.data.DataLoader` wrapping such a dataset).
+    :return: tuple ``(x, y, params)`` where `x` is `None` if absent (the dict-format equivalent
+        of the tuple-format convention for ground-truth-free datasets), `y` is `None` if absent, and
+        `params` is `{}` if absent.
+    """
+    if isinstance(batch, dict):
+        return batch.get("x", None), batch.get("y", None), batch.get("params", {})
```
jscanvic: Are you sure the logic is sound?

According to https://deepinv.org/user_guide/training/datasets.html#base-datasets

1. If x is unspecified, the value should be (torch.nan, y) or (torch.nan, params) or (torch.nan, x, y)
2. If y is unspecified, it can't be None but the value should be x or (x, params)

So I expect we need to actually output different tuple sizes based on what's in the input

If this somehow works with the trainer without following the conventions I'd document it clearly in the docstring

deepinv/optim/utils.py
```diff
@@ -332,10 +332,9 @@ def fit(
         :param bool verbose: Output progress information in the console
         """
         if data_init:
-            first_data = next(iter(dataloader))
+            from deepinv.datasets.base import extract_x_tensor
```
jscanvic: Why a late import?

deepinv/optim/utils.py
```diff
@@ -377,13 +376,16 @@ def _EM_step(self, dataloader: torch.utils.data.DataLoader, verbose: bool = Fals
         :param torch.data.Dataloader dataloader: containing the data
         :param bool verbose: Output progress information in the console
         """
+        from deepinv.datasets.base import extract_x_tensor
```
jscanvic: Why a late import?

deepinv/tests/test_datasets.py
```diff
@@ -466,14 +538,20 @@ def test_hdf5dataset_generate_dataset(tmpdir, physgen, stacked, supervised):
...
+        )
     check_dataset_format(
         test_ds,
         length=1,
-        dtype=None if stacked else tuple,
+        dtype=None if stacked else (dict if use_dict_output else tuple),
         allow_non_tensor=False,
     )
-    x_test, y_test, params_test = test_ds[0]
+    x_test, y_test, params_test = unpack_batch(test_ds[0])
```
jscanvic: I'd do it the other way around, unpack the batch as a dict and pull x y and params from it, that way when we drop support for tuples can simply drop the unpacking function

What do you think?

deepinv/tests/test_datasets.py
```diff
@@ -496,62 +574,84 @@ def test_hdf5dataset_generate_dataset(tmpdir, physgen, stacked, supervised):
...
-    _ = TensorDataset(y=y)
-    _ = TensorDataset(x=x, y=y, params=params)
-    _ = TensorDataset(x=x, params=params)
-    dataset = TensorDataset(y=y, params=params)
-    assert math.isnan(
-        dataset[0][0]
-    ), "Dataset return tuple's first element must be NaN or single-element NaN tensor."
+
+    with pytest.warns(DeprecationWarning, match="use_dict_output=True"):
+        _ = TensorDataset(x=x)
```
jscanvic, line 676: It can be a bit misleading given this checks only that a warning is raised once but we're feeding it a bunch of instantiations

I suggest we leave this block as is and put the with around the line below `dataset = ...` with a condition on the value of use_dict_output using contextlib.nullcontext

Something like:

```python
cm = pytest.warns(DeprecationWarning, match="use_dict_output=True") if not use_dict_output else contextlib.nullcontext()
with cm:
   dataset = TensorDataset(...)
```

jscanvic (changes requested):
Thanks a lot Romain for tackling this very tedious issue

A few comments for now

deepinv/datasets/base.py
```diff
@@ -35,7 +35,9 @@ def check_dataset(dataset: Dataset, allow_non_tensor=True) -> None:
         raise RuntimeError(f"Dataset {dataset} should have length greater than zero.")
 
     batch = dataset.__getitem__(0)
-    error = f"Dataset {dataset} should return either non-nan image `x`, or tuples of either length 2 of 3 of (x, y) or (x, params), or (x, y, params), where x, y are images (either Tensor, TensorList, or nan) and params is a dict"
+    error_tuple = f"Dataset {dataset} should return either non-nan image `x`, or tuples of either length 2 of 3 of (x, y) or (x, params), or (x, y, params), where x, y are images (either Tensor, TensorList, or nan) and params is a dict"
+
+    error_dict = f"Dataset {dataset} should return a dict with at least one of the keys 'x' or 'y', where x, y are images (either Tensor, TensorList) and params is a dict"
```
Andrewwango:
```suggestion
    error_dict = f"Dataset {dataset} should return a dict with at least one of the keys 'x' or 'y' or 'params', where x, y are images (either Tensor, TensorList) and params is a dict"
```

deepinv/datasets/base.py
```diff
@@ -50,58 +52,157 @@ def warn_core_types(x):
...
         if not isinstance(y_or_params, (*image_types, dict)):
             raise RuntimeError(
-                f"{error}, but index 1 of returned tuple is type {type(y_or_params)}."
+                f"{error_tuple}, but index 1 of returned tuple is type {type(y_or_params)}."
             )
         elif isinstance(y_or_params, dict) and any(
             not isinstance(k, str) for k in y_or_params
         ):
-            raise RuntimeError(f"{error}, but params dict has non-string keys.")
+            raise RuntimeError(f"{error_dict}, but params dict has non-string keys.")
```
Andrewwango:
```suggestion
            raise RuntimeError(f"{error_tuple}, but params dict has non-string keys.")
```
this branch is still to do with the old tuple returning? nothing to do with `return {...}`?

deepinv/datasets/base.py
```diff
@@ -50,58 +52,157 @@ def warn_core_types(x):
...
             )
 
         if not isinstance(params, dict):
             raise RuntimeError(
-                f"{error}, but index 2 of returned tuple is not dict but of type {type(params)}."
+                f"{error_tuple}, but index 2 of returned tuple is not dict but of type {type(params)}."
             )
         elif any(not isinstance(k, str) for k in params):
-            raise RuntimeError(f"{error}, but params dict has non-string keys.")
+            raise RuntimeError(f"{error_dict}, but params dict has non-string keys.")
```
Andrewwango: likwise

deepinv/datasets/base.py
```diff
@@ -50,58 +52,157 @@ def warn_core_types(x):
...
-            f"{error}, but returned list or tuple of length {len(batch)}."
+            f"{error_tuple}, but returned list or tuple of length {len(batch)}."
         )
 
     else:
-        raise RuntimeError(f"{error}, but returned batch of type {type(batch)}.")
+        raise RuntimeError(f"{error_dict}, but returned batch of type {type(batch)}.")
+
+
+def unpack_batch(
```
Andrewwango: unrelated, but does your new dataset need a custom collate? or does dataloader batch across nested dicts?

deepinv/datasets/base.py
```diff
@@ -50,58 +52,157 @@ def warn_core_types(x):
...
+    batch: Tensor | TensorList | tuple | list | dict,
+) -> tuple[Tensor | TensorList | float | None, Tensor | TensorList | None, dict]:
+    """Normalize a dataset batch (tuple or dict format) into a canonical ``(x, y, params)`` triple. Used in :class:`deepinv.Trainer` to unpack batches from the dataloader.
+
+    Accepts any of the batch shapes documented in :func:`check_dataset`: a bare image
```
Andrewwango: ` instead of ``"

deepinv/datasets/base.py
Andrewwango: I guess there's two things here: 
- returning batches for trainer - we can ignore the points Jeremy raises above, and we can just refactor trainer to accept e.g. x, None, params i.e. always 3 items. This is only problematic for people with existing trainer child classes (like me lol)
- an external function for people to normalise batches. Yes, then this is a problem. Do we need such a function? If not, we can make this function more internal instead.

deepinv/datasets/base.py
```diff
@@ -50,58 +52,157 @@ def warn_core_types(x):
...
+            return batch[0], batch[1], batch[2]
+        else:
+            raise RuntimeError(
+                f"Batch should be a tensor, or tuple/list of length 1 to 3, or dict, but got tuple/list of length {len(batch)}."
+            )
+
+    return batch, None, {}
+
+
+def extract_x_tensor(batch: Tensor | TensorList | tuple | list | dict) -> Tensor:
```
Andrewwango: why can't this be simply unpack_batch(batch)[0]?

deepinv/datasets/cmrxrecon.py
```diff
@@ -163,6 +165,21 @@ def __init__(
 
             self.samples = samples
 
+        # NOTE: skip FastMRISliceDataset.__init__ (next in MRO) since it would
```
Andrewwango: ?

deepinv/datasets/cmrxrecon.py
```diff
@@ -163,6 +165,21 @@ def __init__(
...
+        # re-scan `root` using FastMRI's own file/metadata format; CMRxRecon
+        # already builds `all_fnames` and `self.samples` itself above using its own format.
+        self.use_dict_output = use_dict_output
+
+        # Putting warning here since not triggered on ImageDataset.__init__
```
Andrewwango: why not?

deepinv/datasets/datagenerator.py
```diff
@@ -455,12 +462,8 @@ def collate_pillow(
             ) -> Tensor:
                 tensors = []
                 for sample in batch:
-                    if isinstance(sample, Image.Image):
```
Andrewwango, line 458: why remove this code from here?

deepinv/datasets/fastmri.py
```diff
@@ -535,7 +559,7 @@ def save_simple_dataset(
         transform = Compose(transform)
 
         xs = [
-            transform(self.__getitem__(i)[0]).squeeze(0)
+            transform(extract_x_tensor(self.__getitem__(i))).squeeze(0)
```
Andrewwango: is it worth having this extra func if it would be a one-liner if else? Happy either way but am always leaning towards something more obvious to read even if it means more verbose

deepinv/datasets/kohler.py
```diff
@@ -194,7 +197,12 @@ def __getitem__(self, index: int) -> tuple[torch.Tensor, torch.Tensor]:
             trajectory_index = index // 12 + 1
         else:
             raise ValueError(f"Unsupported ordering: {self.ordering}")
-        return self.get_item(printout_index, trajectory_index, frames=self.frames)
+        out = self.get_item(printout_index, trajectory_index, frames=self.frames)
+
+        if self.use_dict_output:
+            out = {"x": out[0], "y": out[1]}
+
+        return out
```
Andrewwango:
```suggestion
        x, y = self.get_item(printout_index, trajectory_index, frames=self.frames)

        if self.use_dict_output:
            return {"x": x, "y": y}

        return x, y
```

deepinv/datasets/lidc_idri.py
```diff
@@ -169,4 +173,6 @@ def __getitem__(self, idx: int) -> Any:
         if self.transform is not None:
             slice_array = self.transform(slice_array)
 
-        return slice_array
+        out = {"x": slice_array} if self.use_dict_output else slice_array
+
+        return out
```
Andrewwango:
```suggestion
        return {"x": slice_array} if self.use_dict_output else slice_array
```

deepinv/datasets/lidc_idri.py
```diff
@@ -169,4 +173,6 @@ def __getitem__(self, idx: int) -> Any:
         if self.transform is not None:
             slice_array = self.transform(slice_array)
 
-        return slice_array
+        out = {"x": slice_array} if self.use_dict_output else slice_array
```
Andrewwango: Tbh I like this one-liner `return {x} if self.use_dict_output else x`, it would be nice if other datasets did this instead of doing it in 4 lines, but it's a minor point and if you can't be bothered to change the others then don't worry

deepinv/datasets/patch_dataset.py
```diff
@@ -46,4 +53,8 @@ def __getitem__(self, idx):
         patch = self.all_patches[idx]
         if self.transform:
             patch = self.transform(patch)
-        return patch.reshape(self.shape) if self.shape else patch
+        x = patch.reshape(self.shape) if self.shape else patch
+
+        out = {"x": x} if self.use_dict_output else x
```
Andrewwango:
```suggestion
        return {"x": x} if self.use_dict_output else x
```

deepinv/datasets/satellite.py
```diff
@@ -173,4 +176,10 @@ def __getitem__(self, idx):
         ms = transform_ms(ms)
         pan = transform_pan(pan)
 
-        return TensorList([ms, pan]) if self.return_pan else ms
+        # TODO: need to check images, dont know what to do here
```
Andrewwango: ? this code is correct

deepinv/datasets/skmtea.py
```diff
@@ -107,6 +110,21 @@ def __init__(
         if filter_id is not None:
             self.samples = list(filter(filter_id, self.samples))
 
+        # NOTE: skip FastMRISliceDataset.__init__ (next in MRO) since it would
```
Andrewwango: again, why's this needed? can we just fix fastMRIdataset?

deepinv/datasets/skmtea.py
```diff
@@ -183,8 +201,17 @@ def __getitem__(self, idx):
         y = self.from_torch_complex(y)  # (1, N, H, W) complex -> (1, 2, N, H, W) real
         y = y.squeeze(0) * mask.unsqueeze(0)
 
-        return (
-            self.from_torch_complex(x).squeeze(0),
-            y,
-            {"mask": mask, "coil_maps": maps.moveaxis(-1, 1).squeeze(0)},
-        )
+        if self.use_dict_output:
```
Andrewwango: this if/else is replicating some logic, so looks a bit messy. In other datasets you did it nicer of only doing the return type logic once x, y and params have been already constructed and prepared.

deepinv/tests/test_metric.py
```diff
@@ -489,26 +489,34 @@ def test_snr(power_signal, power_noise):
...
+        image = torch.rand(
             (self.num_channels, self.resolution, self.resolution), dtype=self.dtype
         )
+        return {"x": image} if self.use_dict_output else image
 
 
 @pytest.mark.parametrize("n_channels", (1, 3))
 @pytest.mark.parametrize("dtype", (torch.float16, torch.float32, torch.float64))
-def test_niqe_fit(n_channels: int, dtype: torch.dtype):
+@pytest.mark.parametrize("use_dict_output", [True, False])
```
Andrewwango: not sure this needs to test both

deepinv/tests/test_trainer.py
```diff
@@ -540,6 +569,7 @@ def forward(self, y, physics, **kwargs):
 @pytest.mark.parametrize("online_measurements", [True, False])
 @pytest.mark.parametrize("generate_params", [True, False])
 @pytest.mark.parametrize("batch_size", [1, 2])
+@pytest.mark.parametrize("use_dict_output", [False, True])
 def test_dataloader_formats(
```
Andrewwango, line 573: just to check: there's somewhere in this test to test that a nested dict {params: {}} is batched by a dataloader?

deepinv/training/trainer.py
```diff
@@ -662,26 +662,28 @@ def get_samples_online(self, iterators, g):
 
         Assumes the dataloader returns ground truth `x`, or tuples of (`x`, `params`). Note that `params` are ignored if a physics generator is provided.
 
+        The dataloader batch can be a bare ``x``, a tuple (optionally ``(x, params)``), or a dict with
+        ``"x"`` and optional ``"params"`` keys (see :func:`deepinv.datasets.check_dataset`).
+
```
```suggestion
       Assumes the dataloader returns ground truth `x`, or tuples of (`x`, `params`), or a dict with `"x"` key and optional `"params"` key. Note that `params` are ignored if a physics generator is provided.
```

docs/source/user_guide/training/datasets.rst
```diff
@@ -28,7 +28,20 @@ or any combination of these, in one of the following ways:
 
   If you have a dataset of measurements only `(y)` or `(y, params)` you should modify it such that it returns `(torch.nan, y)` or `(torch.nan, y, params)`
 
-If you have your own dataset (e.g. a PyTorch `Dataset`), check that it is compatible using the function :func:`deepinv.datasets.check_dataset` 
+Alternatively, all datasets accept a ``use_dict_output=True`` constructor argument, which makes them return a dict
+``{"x": x, "y": y, "params": params}`` instead of a tuple, with any key omitted if not applicable (e.g. `x` is omitted rather
+than set to `torch.nan` if your dataset is ground-truth-free). This is recommended for better readability and flexibility, and
+is understood everywhere a dataset is consumed in ``deepinv`` (e.g. :class:`deepinv.Trainer`). It defaults to `False` for
+backward compatibility with the tuple format above.
```
Andrewwango:
```suggestion
Alternatively, set `use_dict_output=True` (default False) in the dataset which makes them return a dict of the format `{"x": x, "y": y, "params": params}` with any key omitted if not applicable. This dict format is recommended over tuple for better readability and flexibility.
```

Andrewwango: in general, user guide should be imperative ("Set ...") rather than descriptive as it was. Also `` -> `.

examples/adversarial-learning/demo_gan_imaging.py
```diff
@@ -80,13 +80,19 @@
 
 train_dataloader = DataLoader(
     dinv.datasets.HDF5Dataset(
-        dataset_path, train=True, load_physics_generator_params=True
+        dataset_path,
+        train=True,
+        load_physics_generator_params=True,
+        use_dict_output=True,
```
Andrewwango, line 86: why the change ?

examples/models/demo_training.py
```diff
@@ -20,6 +20,7 @@
 
 import deepinv as dinv
 import torch
+from deepinv.datasets.base import unpack_batch
```
Andrewwango: Maybe if this func is to be used everywhere, it should be importable from deepinv.datasets?

deepinv/training/trainer.py
```diff
@@ -662,26 +662,28 @@ def get_samples_online(self, iterators, g):
...
         :param list iterators: List of dataloader iterators.
         :param int g: Current dataloader index.
         :returns: a tuple containing at least: the ground truth, the measurement, and the current physics operator.
         """
         data = next(iterators[g])
+        x, _y, params = unpack_batch(data)
```
Andrewwango: why _y?

deepinv/training/trainer.py
```diff
@@ -662,26 +662,28 @@ def get_samples_online(self, iterators, g):
...
 
-        params = {}
-        if isinstance(data, (tuple, list)):
-            x = data[0]
-
-            if len(data) == 2 and isinstance(data[1], dict):
-                params = data[1]
-            else:
+        if isinstance(data, dict):
+            pass  # dict is unambiguous and self-describing, params already extracted
```
```suggestion
            pass
```

deepinv/training/trainer.py
```diff
@@ -709,42 +711,48 @@ def get_samples_offline(self, iterators, g):
...
-            raise ValueError(
-                "If online_measurements=False, the dataloader should output a tuple (x, y) or (x, y, params)"
-            )
 
-        if len(data) == 2:
-            x, y, params = *data, None
-            if isinstance(y, dict):  # x,params offline
+        x, y, params = unpack_batch(data)
+
+        # self.get_samples_offline is called in self.get_samples only when self.online_measurements is False, so y should be provided in the dataloader output
```
Andrewwango: don't need these obvious comments

deepinv/training/trainer.py
```diff
@@ -709,42 +711,48 @@ def get_samples_offline(self, iterators, g):
...
-            )
 
         batch_size_y = y[0].size(0) if isinstance(y, TensorList) else y.size(0)
-        batch_size_x = x[0].size(0) if isinstance(x, TensorList) else x.size(0)
 
-        if batch_size_x != batch_size_y:  # pragma: no cover
-            raise ValueError(
-                f"Data x, y must have same batch size, but got {batch_size_x}, {batch_size_y}"
-            )
+        # NOTE: if dataset output format is a dict, `unpack_batch` will return x=None, so no complicated checks are needed.
```
Andrewwango: NOTEs are useful for non-obvious functionality, but I think the logic here is pretty self-explanatory.

deepinv/training/trainer.py
Andrewwango: move this line after the data check

deepinv/training/trainer.py
Andrewwango: this branch isn't needed

deepinv/training/trainer.py
```diff
@@ -662,26 +662,28 @@ def get_samples_online(self, iterators, g):
...
-            x = data[0]
-
-            if len(data) == 2 and isinstance(data[1], dict):
-                params = data[1]
-            else:
+        if isinstance(data, dict):
+            pass  # dict is unambiguous and self-describing, params already extracted
+        elif isinstance(data, (tuple, list)):
+            if not (len(data) == 2 and isinstance(data[1], dict)):
+                data_type = type(data[1]) if len(data) > 1 else None
```
Andrewwango: put this inline

deepinv/training/trainer.py
```diff
@@ -662,26 +662,28 @@ def get_samples_online(self, iterators, g):
...
                 warnings.warn(
-                    f"Generating online measurements requires dataloader to return tensor `x` or (tensor `x`, dict `params`) but got ({type(x)}, {type(data[1])}, ...). Discarding all data after `x`."
+                    f"Generating online measurements requires dataloader to return tensor `x` or (tensor `x`, dict `params`) but got ({type(x)}, {data_type}, ...). Discarding all data after `x`."
                 )
-        else:
-            x = data
+                params = {}
+        # else: bare x, params already {}
```
```suggestion
```

deepinv/training/trainer.py
```diff
@@ -709,42 +711,48 @@ def get_samples_offline(self, iterators, g):
...
-            x, y, params = *data, None
-            if isinstance(y, dict):  # x,params offline
+        x, y, params = unpack_batch(data)
+
+        # self.get_samples_offline is called in self.get_samples only when self.online_measurements is False, so y should be provided in the dataloader output
+        if y is None:
+            if isinstance(data, dict):
                 raise ValueError(
-                    "If online_measurements=False, measurements y must be provided as a tensor."
+                    "If online_measurements=False, the dataloader should output a dict with keys 'x' and 'y' (and optionally 'params')"
```
Andrewwango: x is not compulsory?

deepinv/training/trainer.py
```diff
@@ -709,42 +711,48 @@ def get_samples_offline(self, iterators, g):
...
+        # NOTE: if dataset output format is a dict, `unpack_batch` will return x=None, so no complicated checks are needed.
+        # If dataset output format is a tuple, the first element x can be torch.nan, which will be collated to a tensor of shape (batch_size,) with all values being torch.nan. In this case, we set x=None to indicate that there is no ground truth in the deepinv convention.
+        if isinstance(x, (tuple, list)) or x is not None:
+            batch_size_x = x[0].size(0) if isinstance(x, TensorList) else x.size(0)
 
-        if torch.isnan(x).all() and x.ndim <= 1:
-            x = None  # Batch of NaNs -> no ground truth in deepinv convention
-        else:
+            if batch_size_x != batch_size_y:  # pragma: no cover
+                raise ValueError(
```
Andrewwango, line 737: why is this check only done when data is tuple?

deepinv/training/trainer.py
```diff
@@ -709,42 +711,48 @@ def get_samples_offline(self, iterators, g):
...
+                    f"Data x, y must have same batch size, but got {batch_size_x}, {batch_size_y}"
+                )
+
+            if torch.isnan(x).all() and x.ndim <= 1:
+                x = None  # Batch of NaNs -> no ground truth in deepinv convention
```
Andrewwango: remove comment

deepinv/tests/test_datasets.py
```diff
@@ -988,20 +1176,26 @@ def mock_lidc_idri():
...
+            )
+            check_dataset_format(
+                LidcIdriSliceDataset(
+                    root=mock_lidc_idri,
+                    transform=totensor,
+                    hounsfield_units=hounsfield_units,
+                    use_dict_output=use_dict_output,
+                ),
+                length=2036,
+                dtype=(dict if use_dict_output else Tensor if totensor else np.ndarray),
```
Andrewwango, line 1236: Is dtype not used in this test?

Andrewwango (changes requested):
Thanks @romainvo for this exciting PR! Lots of comments incoming...

deepinv/datasets/base.py
romainvo: The `default_collate` from torch applies the strategy recursively (https://docs.pytorch.org/docs/2.13/data.html#torch.utils.data.default_collate), i.e. `Mapping[K, V_i] -> Mapping[K, default_collate([V_1, V_2, …])]` so nested dicts are processed properly

deepinv/tests/test_trainer.py
Andrewwango: update: not sure if it's needed or not, but given that unbatched tensors can cause headaches (#1330) maybe it could help!

deepinv/datasets/base.py
romainvo: I feel like this util makes the original goal unclear, i.e. we want to move every internal to the "output as dict" format, I implemented it for the trainer originally (hence the unclear doc) but really it's useful across the whole codebase. I agree with @jscanvic I'm doing things the other way around by packing batches as tuple. What I'm proposing is this 

```python
def batch_as_dict(
        batch: Tensor | tuple | list | dict,
) -> dict:
    """Packs a dataset batch (tuple or dict format) into a canonical dict with keys "x", "y", and "params"."""
    
    # If the batch is already a dict, return it as is
    if isinstance(batch, dict):
        return batch
    
    elif isinstance(batch, Tensor):
        out = {"x": batch}
        
    elif isinstance(batch, (tuple, list)):
        if len(batch) == 1:
            out = {"x": batch[0]}
        
        elif len(batch) == 2:
            x, second = batch
            out = {}
            if not torch.isnan(x).all():
                out = {"x": x}
            if isinstance(second, dict):
                out["params"] = second
            else:
                out["y"] = second
                
        elif len(batch) == 3:
            x, y, params = batch
            out = {}
            if not torch.isnan(x).all():
                out = {"x": x}
            
            out["y"] = y
            out["params"] = params
            
        else:
            raise RuntimeError(
                f"Batch should be a tensor, or tuple/list of length 1 to 3, or dict, but got tuple/list of length {len(batch)}."
            )
            
    return out
```

We can use it on every dataloader output (internally), and unpack the batch as a dict (regardless of whether the dataset was still created with the tuple format). There shouldn't be any friction for the user (every old format like "x, None, params" is still supported) , even in the case of child Trainer classes (unless `get_samples_offline` is overridden, but then we can't take that into account)

deepinv/datasets/satellite.py
romainvo: Yep it was correct, forgot to remove the comment :')

deepinv/datasets/skmtea.py
romainvo: There isn't really anything to fix in `FastMRISliceDataset`, the thing is SKMTEA has its own constructor logic as well as FastMRI, which is not a problem in itself. The problem is SKMTEA inherits from FastMRI, and when inheriting an object, the best practice is to call the parent constructor. However the logic of the FastMRI constructor is different than SKMTEA. And because we cannot call the FastMRI constructor, the `use_dict_output` warning won't be raised.

Now that I'm looking at it again, the only factored logic is the `_retrieve_metadata` and `SliceSampleID` logic. What we can do is not inherit from FastMRI and define these two functions outside FastMRI? or keep it as is

deepinv/tests/test_trainer.py
romainvo: `DummyDataset` below returns a nested dict, so this use-case is tested, if that's what you had in mind?

deepinv/training/trainer.py
romainvo: Nope, at least that's what I deduced from the implementation?

examples/adversarial-learning/demo_gan_imaging.py
romainvo: The idea was to switch to the dict format as default everywhere, e.g. if someone copy-pastes the code, it uses the proper output format.

deepinv/tests/test_datasets.py
romainvo: I'm only factoring the most recurrent patern in `def image_output_type`, for specific cases I stick to custom one-liners

deepinv/tests/test_datasets.py
romainvo: Do you mean it doesn't catch the warnings raised after the second `TensorDataset` is created? If we leave the block as is, we will end up with uncaught warnings, won't we?

deepinv/optim/utils.py
romainvo: I end up with circular import errors if I don't do the late import... I could not resolve why :'(

romainvo:
Thanks, @jscanvic and @Andrewwango, for your thorough reviews! I have tried to apply your suggestions.

The biggest change is how I’m handling the format conversion; we now have a single utility, batch_as_dict, which converts a tuple output format into a dict output (with “x”, “y”, “params” keys omitted if applicable, e.g., “x” full of NaNs, or “params” an empty dict, etc.). The standard pattern is the following:

```python
batch = next(iter(dataloader))
batch = batch_as_dict(batch) # converts batch to dict if tuple format, else no-op

# do your stuff, e.g
x, y = batch["x"], batch["y"]
```

This means that internally (tests, trainer, demos), all the logic is converted to “output dict” format. In the future, when deprecating the tuple format, the only thing to do (almost) will be to remove the following batch = batch_as_dict(batch) line.
I still had to propagate the update in logic, so lot of diffs to check again, sorry.

--- 

I haven't dealt with the parent constructor issue in SKMTEA and CMRx, I feel like the code duplication is minor, and the issue is orthogonal to the already big changes here

deepinv/datasets/cmrxrecon.py
Andrewwango: Ok I understand now, and you're right it's an unrelated minor issue. Please can you just remove the comments? They're unnecessary. Here + in skmtea

deepinv/datasets/skmtea.py
Andrewwango: remove comments then

Andrewwango (approved):
LGTM except minor comment issue. Thanks @romainvo!

romainvo:
@Andrewwango I removed the useless comments. I've added the necessary logic to catch all the `use_dict_output` related warnings in the tests, should be good now. 

I hadn't dealt with two remaining datasets: `RandomPhantomDataset` and `SheppLoganDataset`. They do not inherit from `ImageDataset` but from `torch.utils.data.Dataset`. I wanted to switch the inheritance to factor the `self.use_dict_output` logic but both classes live in the `deepinv.utils` module so I could not do it without having circular import errors popping everywhere.

I'm duplicating the `use_dict_output` attribute/warning for these two classes for the moment (similar issue than with skmtea and cmrx in the end), until we clean the inheritance.

PR #1245, opened 2026-06-25
hotfix: download_archive produces unwanted verbose in doctests

romainvo:
In #1234 we introduce a new mechanism for caching and checking if datasets were already downloaded. `download_archive` now produces unwanted verbose when a dataset is cached, which makes some doctests fail, i.e. `CMRxReconSliceDataset` and potentially `FastMRISliceDataset`, see the failed CI runs:
- https://github.com/deepinv/deepinv/actions/runs/28114899204/job/83251673610
- https://github.com/deepinv/deepinv/actions/runs/28131556571/job/83308734519?pr=1220

I'm adding [ELLIPSIS ](https://docs.python.org/3/library/doctest.html) to match any output of the `download_archive` lines in the concerned doctests.

### DIFF
--- a/deepinv/datasets/cmrxrecon.py
+++ b/deepinv/datasets/cmrxrecon.py
@@ -73,7 +73,6 @@ class CMRxReconSliceDataset(FastMRISliceDataset, MRIMixin):
     ...     get_cache_home() / "CMRxRecon.zip",
     ...     extract=True,
     ... )
-    ...
     >>> dataset = CMRxReconSliceDataset(get_cache_home() / "CMRxRecon")
     >>> x, y, params = next(iter(DataLoader(dataset)))
     >>> x.shape # (B, C, T, H, W)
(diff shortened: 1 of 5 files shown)


tachella:
maybe we could add a verbose option? it seems that the ... are not solving the problem

romainvo:
> maybe we could add a verbose option? it seems that the ... are not solving the problem

Yes, just noticed it :'( . I'm trying a few debugs, but we can fall back to the verbose option if I can't resolve it. 

There is this line in the DDRM doctest:
`denoiser = dinv.models.DRUNet(pretrained="download").to(device)  # doctest: +IGNORE_RESULT` 

but it does not look like it's supported by doctest?

jscanvic:
This is a recurring issue I don't know how we did the previous times but for reference see: https://github.com/deepinv/deepinv/pull/1059#discussion_r2817055483

jscanvic:
**Update:** I rolled it back in favor of printing to stderr which is more standard and requires less maintenance

I implemented a way to ignore doctest errors due to irrelevant unexpected outputs. It only works when run through pytest and not using doctest directly but I think that's what we do everytime in CI already so it should be fine.

The syntax is: `# deepinv: +IGNORE_OUTPUT`.

jscanvic:
After discussing with Romain, I rolled back the previous change and switched `print` to `sys.stderr.write` since doctests ignore stderr outputs and it is generally more standard to write debug / verbose information there instead of to the standard output

jscanvic (approved):

romainvo:
Thanks @jscanvic! Looks fine locally, we can merge once CI passes, I think

PR #1247, opened 2026-06-26
Unpin pixi version

romainvo:
I'm reverting the hotfix in #1243, which pinned the Pixi version due to a breaking update on Pixi's side:
- I'm adding a new dependency group: `physics-gpu` with only `astra-toolbox` in it for the moment.
- I'm also adding wheels specific to the cpu/gpu dependency group for `torchvision` (switching for pypi dependency instead of conda as it was easier to find the wheels)

The 0.71.1 pixi version seems to break some of the legacy `system-requirements`  syntax (https://github.com/prefix-dev/pixi/issues/6462). Specifically mentioning the supported platform for the `cpu` group seems to solve this problem, i.e.
```
[tool.pixi.feature.cpu]
platforms = ["linux-64", "win-64"]
```

I'm not using any of the new pixi syntax so it should be compatible with old pixi versions (works locally at least for me)

---

I'm also adding a fallback mechanism to a "not locked" pixi install in case a PR modifies `pyproject.toml` or if pixi can no longer resolve some environments. If `pixi install -e <env> --locked` fails it retries without `--locked` which automatically updates the lock file.

### DIFF
--- a/.github/workflows/test_cpu.yml
+++ b/.github/workflows/test_cpu.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Setup Pixi
         uses: prefix-dev/setup-pixi@v0.9.3
         with:
-          pixi-version: v0.70.2
+          pixi-version: latest
           environments: full-cpu test-cpu
 
       - name: Restore .testmondata
--- a/.github/workflows/test_cpu.yml
+++ b/.github/workflows/test_cpu.yml
@@ -74,14 +74,14 @@ jobs:
       - name: Install dependencies
         shell: bash
         run: |
-          echo "pixi install -e ${{ matrix.environment }} --locked"
+          set -x
           if ! pixi install -e ${{ matrix.environment }} --locked --manifest-path pyproject.toml --color always; then
             echo "pixi install with --locked option failed, run install without lock"
             pixi install -e ${{ matrix.environment }} --manifest-path pyproject.toml --color always
           fi
 
-          echo "pixi list -e ${{ matrix.environment }}"
           pixi list -e ${{ matrix.environment }} --manifest-path pyproject.toml --color always
+          set +x
 
       - name: Restore deepinv URL cache # shared across matrix environments
         uses: actions/cache/restore@v4
--- a/.github/workflows/test_cpu.yml
+++ b/.github/workflows/test_cpu.yml
@@ -67,7 +67,6 @@ jobs:
         uses: prefix-dev/setup-pixi@v0.9.3
         with:
           pixi-version: latest
-          environments: ${{ matrix.environment }}
           run-install: false  # we drive installation manually below
 
       # Install dependencies with pixi, using the restored lock if available. If the
--- a/.github/workflows/test_cpu.yml
+++ b/.github/workflows/test_cpu.yml
@@ -52,12 +52,11 @@ jobs:
       # full re-run). Restored *before* Setup Pixi so the action installs from
       # this lock. Not restored on push to main: main solves fresh to track
       # latest deps (and re-runs the full suite anyway).
-      - name: Restore pixi.lock + .testmondata
+      - name: Restore .testmondata
         if: github.event_name == 'pull_request'
         uses: actions/cache/restore@v4
         with:
           path: |
-            pixi.lock
             .testmondata
           key: testmon-cpu-main-${{ matrix.environment }}-${{ matrix.os }}-${{ github.run_id }}
           restore-keys: |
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -89,13 +91,16 @@ denoisers = [
     "scipy",
     "diffusers",
     "transformers",
-    "numba>=0.60.0",
+    "numba",
```
jscanvic: Why?

pyproject.toml
```diff
@@ -89,13 +91,16 @@ denoisers = [
...
 ]
 
 # optional dependencies for advanced forward operators
+physics-gpu = [
+    "astra-toolbox"
```
jscanvic: Why removing the version constraint?

jscanvic (changes requested):
Looks sound!

Just minor questions:
- Why changing the version constraints on numba and astra-toolbox?
- Is it backwards compatible with pixi 0.70.2? I'm asking for users that might be using pixi to install the library and might not update pixi for a while

pyproject.toml
romainvo: Forgot that we were also using this dependecy groups for the uv installs, I put them back.

romainvo:
> Looks sound!
> 
> Just minor questions:
> 
>     * Why changing the version constraints on numba and astra-toolbox?

Sorry it was mainly typos from my first iterations. 

>     * Is it backwards compatible with pixi 0.70.2? I'm asking for users that might be using pixi to install the library and might not update pixi for a while

I've made sure the pyproject update is backward compatible.

.github/workflows/test_cpu.yml
```diff
@@ -66,8 +66,22 @@ jobs:
...
-          environments: full-cpu test-cpu
+          pixi-version: latest
+          run-install: false  # we drive installation manually below
+
+      # Install dependencies with pixi, using the restored lock if available. If the
+      # lock is unavailable or fails to solve, fall back to a fresh solve. 
+      - name: Install dependencies
+        shell: bash
+        run: |
+          echo "pixi install -e ${{ matrix.environment }} --locked"
```
jscanvic: Why this echo?

.github/workflows/test_cpu.yml
```diff
@@ -66,8 +66,22 @@ jobs:
...
+          if ! pixi install -e ${{ matrix.environment }} --locked --manifest-path pyproject.toml --color always; then
+            echo "pixi install with --locked option failed, run install without lock"
+            pixi install -e ${{ matrix.environment }} --manifest-path pyproject.toml --color always
+          fi
+
+          echo "pixi list -e ${{ matrix.environment }}"
```
jscanvic: Why this one too?

.github/workflows/test_cpu.yml
```diff
@@ -66,8 +66,22 @@ jobs:
...
+          run-install: false  # we drive installation manually below
+
+      # Install dependencies with pixi, using the restored lock if available. If the
+      # lock is unavailable or fails to solve, fall back to a fresh solve. 
+      - name: Install dependencies
+        shell: bash
+        run: |
+          echo "pixi install -e ${{ matrix.environment }} --locked"
+          if ! pixi install -e ${{ matrix.environment }} --locked --manifest-path pyproject.toml --color always; then
+            echo "pixi install with --locked option failed, run install without lock"
```
jscanvic, line 79: I believe you need to remove the lock file to avoid it being wrongfully used with testmon

.github/workflows/test_gpu.yml
```diff
@@ -55,11 +55,26 @@ jobs:
...
           post-cleanup: true
-          environments: full
           pixi-bin-path: /local/jtachell/pixi/bin/pixi
+          run-install: false  # we drive installation manually below
 
+      # Install dependencies with pixi, using the restored lock if available. If the
+      # lock is unavailable or fails to solve, fall back to a fresh solve. 
+      - name: Install dependencies
+        run: |
+          echo "pixi install -e full --locked"
```
jscanvic: Same echo comment

.github/workflows/test_gpu.yml
```diff
@@ -55,11 +55,26 @@ jobs:
...
+          if ! pixi install -e full --locked --manifest-path pyproject.toml --color always; then
+            echo "pixi install with --locked option failed, run install without lock"
+            pixi install -e full --manifest-path pyproject.toml --color always
+          fi
+
+          echo "pixi list -e full"
```
jscanvic: Same echo comment

.github/workflows/test_gpu.yml
```diff
@@ -55,11 +55,26 @@ jobs:
...
+          run-install: false  # we drive installation manually below
 
+      # Install dependencies with pixi, using the restored lock if available. If the
+      # lock is unavailable or fails to solve, fall back to a fresh solve. 
+      - name: Install dependencies
+        run: |
+          echo "pixi install -e full --locked"
+          if ! pixi install -e full --locked --manifest-path pyproject.toml --color always; then
+            echo "pixi install with --locked option failed, run install without lock"
+            pixi install -e full --manifest-path pyproject.toml --color always
```
jscanvic, line 70: Same .lock removal comment

jscanvic (changes requested):
Thanks Romain!

Two comments:

Instead of using echo to display the commands that are run please use set -x at the beginning and set +x at the end (it toggles displaying commands that are run). The benefit is it shows the actual commands run and won't get out of sync if we change the commands themselves

Also, I believe you need to actually remove the lock file otherwise testmon might assume we're running the cached dependencies while we're not

jscanvic (approved):
LGTM

For reference, the lock file is not actually used by testmon so no need to delete it

Issue #734, opened 2025-09-08
`pyproject` optional deps should be separated by application


Andrewwango:
### Documentation location

pyproject.toml

### Describe the issue or suggestion

This will allow users to selectively install op deps e.g. `pip install deepinv[mri]` without installing everything else like bm3d.

Whilst we're here, the external libs examples should all tell you which op deps you need to install. As well as in each class that requires an op dep (e.g. `estimate_coil_maps`)...

Reported by @chaithyagr

PR #1249, opened 2026-06-26
Example for low-intensity STED fluorescence microscopy denoising + load tiff files for microscopy + view in napari

Andrewwango:
- New example with real data
- raw microscopy data
- RAM denoising
- tiff loader
- view napari function for visualising 2D and 3D data


Andrewwango:
Note: successful doc run is here https://github.com/deepinv/deepinv/actions/runs/28250547779/job/83700675808, you can download the example from there

examples/external-libraries/demo_microscopy_denoising.py
```diff
@@ -0,0 +1,118 @@
...
+# We compare the ground truth, the noisy low-intensity measurement, and the
+# denoised reconstruction.
+
+dinv.utils.plot(
+    {
+        "8 µW excitation": x,
+        "1.5 µW excitation": y,
+        "1.5 µW excitation\ndenoised": x_net,
+    },
+    figsize=(7, 7),
```
Tmodrzyk: We could make this smaller in height

docs/source/_static/demo_microscopy_denoising_screenshot.png
Tmodrzyk, line 1: Is there a way to avoid adding this screenshot to the sources ? For instance on the huggingface.
1Mb seems like a lot considering the whole repo is around 10Mb.

docs/source/changelog.rst
Tmodrzyk, line 1: Also add a mention of the demo file 😄

deepinv/datasets/utils.py
```diff
@@ -59,7 +59,8 @@ def download_archive(
 
     :param str url: URL of archive.
     :param str, pathlib.Path save_path: path where file should be saved.
-    :param bool extract: if ``True``, attempt to extract zipfile or tarball into parent dir.
+    :param bool extract: if ``True``, attempt to extract zipfile, tarball or RAR archive into parent dir.
+        Extracting RAR archives requires `rarfile`, install it with ``pip install rarfile``.
```
Tmodrzyk, line 63: Given the small size of `rarfile`, we might want to add it to the base set of dependencies ? 
I feel like this would be annoying to have to download all dependencies just to have this feature.

deepinv/datasets/utils.py
```diff
@@ -117,6 +120,28 @@ def extract_tarball(file_path: str | Path, extract_dir: str | Path) -> None:
...
+        import rarfile
+    except ImportError:  # pragma: no cover
+        raise ImportError(
+            "Extracting RAR archives requires rarfile, which is not installed. "
+            "Please install it with `pip install rarfile`. Note that rarfile also "
+            "requires the `unrar` or `bsdtar` command-line tool to be installed."
+        )
+
+    with rarfile.RarFile(file_path) as rar_ref:
+        # Files may vary in size so maybe nonlinear progress bar
```
Tmodrzyk: remove comment

deepinv/utils/napari.py
```diff
@@ -0,0 +1,86 @@
...
+
+def plot_napari(*x: Tensor, screenshot: bool = False):  # pragma: no cover
+    """View 2D images or 3D volumes in napari.
+
+    Opens an interactive `napari <https://napari.org>`_ viewer displaying the
+    provided tensors side by side in a grid. This is useful for inspecting
+    volumetric data such as 3D microscopy stacks, which cannot be easily
+    rendered with :func:`deepinv.utils.plot`.
+
+    Requires `napari` to be installed. Install it with ``pip install "napari[all]"``.
```
Tmodrzyk: Add this as a warning or tip.

deepinv/utils/io.py
```diff
@@ -35,6 +35,49 @@ def load_np(
...
+def load_tiff(fname: str | Path, dtype: torch.dtype | None = None) -> torch.Tensor:
+    """Load image or volume from a TIFF file as a torch tensor.
+
+    Integer images are normalized to the range ``[0, 1]`` by dividing by the
+    maximum value representable by their dtype; floating point images are
+    loaded as-is. 2D images of shape `(H, W)` are loaded with a single channel,
+    and 3D arrays of shape `(H, W, C)` are converted to channel-first `(C, H, W)`.
+    In both cases a leading batch dimension is added.
+
+    Requires `tifffile` to be installed. Install it with `pip install tifffile`.
```
Tmodrzyk: Add this as a warning or tip

deepinv/utils/napari.py
```diff
@@ -0,0 +1,86 @@
...
+    is_3d = 5 in ndims
+
+    for a in x:
+        if a.shape[0] != 1 or a.shape[1] != 1:
+            raise ValueError(
+                f"Expected batch dim 1 and channel dim 1, got shape {tuple(a.shape)}"
+            )
+
+    arrays = [a.squeeze(1).squeeze(0).detach().cpu().numpy() for a in x]
+    arrays = [arr / arr.max() for arr in arrays]
```
Tmodrzyk: This should become an argument / option, following the `deepinv.utils.plot` API (rescale_mode, vmin, vmax, etc).
I guess this can be changed once the GUI opens but we should also have some control over it in the code.

Tmodrzyk (commented):
Hi @Andrewwango thank you for this very nice example. 
The notebook is pretty much perfect to me as is.
To me we should not put the screenshot you display at the end in the source files though, as it would make our sources significantly heavier.
I also have a few minor comments regarding the new features, see my specific comments.

examples/external-libraries/demo_microscopy_denoising.py
Andrewwango: done

deepinv/datasets/utils.py
Andrewwango: hmm, I'm going with existing convention to keep all io dependencies which are only used for specific functions, as optional. e.g. similar to pydicom, nibabel etc.

I agree though that it would be annoying, and it should be fixed in #734

deepinv/datasets/utils.py
Andrewwango: done

docs/source/changelog.rst
Andrewwango: following the convention of the changelog, I don't think the new demos should be added to changelog

deepinv/datasets/utils.py
Tmodrzyk: We can keep this solution for now

docs/source/changelog.rst
Tmodrzyk: Does this mean we don't keep track of new documentation, demos, and the people who contributed to them?
Also, this could be an issue since it seems we use the changelog to advertise new versions of the library.
Of course, this is non-blocking for this PR, but we could open an issue to revisit this. What do you think?

Andrewwango: True, it speaks to what the purpose of a changelog is!

examples/external-libraries/demo_microscopy_denoising.py
```diff
@@ -0,0 +1,118 @@
...
+
+dinv.utils.plot(
+    {
+        "8 µW excitation": x,
+        "1.5 µW excitation": y,
+        "1.5 µW excitation\ndenoised": x_net,
+    },
+    figsize=(7, 5),
+)
+
```
Tmodrzyk, line 93: Idk how it's possible but when running this example localy, the interpreter tells me `x` is not defined.
I guess it is lively only in the for loop in the previous cell, but I checked and the docs is built correctly with this cell running. 
Please double check on your end.

deepinv/datasets/utils.py
```diff
@@ -117,6 +118,27 @@ def extract_tarball(file_path: str | Path, extract_dir: str | Path) -> None:
             tar_ref.extract(file_to_be_extracted, extract_dir)
 
 
+def extract_rarfile(file_path: str | Path, extract_dir: str | Path) -> None:
+    """Extract a local RAR archive.
+
+    Requires `rarfile`, install it with ``pip install rarfile``. Note that
+    `rarfile` itself relies on an external ``unrar`` or ``bsdtar`` command-line
+    tool being available on the system.
+    """
```
Tmodrzyk, line 127: This lacks parameters docstring, even though type / description are obvious. 
Other functions in this file seem to lack it as well (e.g `extract_zipfile`) so I'm ok with merging as is.

Tmodrzyk (approved):
LGTM, could you run the example notebook locally to check if you also have an error in the last cell with the plot function ?

examples/external-libraries/demo_microscopy_denoising.py
Andrewwango: strange, it definitely does work and is defined!

Andrewwango: (otherwise the plots would fail. here the plots don't fail)

PR #1252, opened 2026-06-29
Remove/defer torchvision imports

### LINKED ISSUE #936: Defer `torchvision` import?
In the quest to minimise `deepinv` overhead as much as possible, we still require the eager import of `torchvision` in deepinv as it's used in many places.

However, torchvision is still quite a heavy import:

Image

It would be good to review how much we use torchvision/assume the user is already using torchvision. If a lot, then no point in removing the eager import. If less so, then we could try to remove it.

Andrewwango:
Closes #936 . See that issue for motivation

This removes all top-level eager torchvision imports from the main library. Now, import deepinv does not import torchvision eagerly anymore.

All torchvision imports in examples and tests are unchanged.

This PR also breaks some long import chains by forcing lazy imports of some external libs like PIL, h5py as well as lazy imports of physics inside models etc.

NOTE for future reference: removing torchvision altogether is also an option, in order to reduce the install dependencies, and remove some layers of abstraction. However, this is out of scope.

TLDR: import time cut in half, with minimal overhead on top of torch!

image

cc @nucli-vicky 

### DIFF
--- a/deepinv/__init__.py
+++ b/deepinv/__init__.py
@@ -1,6 +1,5 @@
 from .__about__ import *
 import torch
-import torchvision as _torchvision  # early loading
 
 __all__ = [
     "__title__",
--- a/deepinv/datasets/kohler.py
+++ b/deepinv/datasets/kohler.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 import torch
-from torchvision.datasets.utils import download_and_extract_archive
 from PIL import Image
 
 from urllib.parse import urlparse
@@ -154,6 +153,8 @@ class Kohler(ImageDataset):
                 from deepinv.datasets import Kohler
                 Kohler.download("datasets/Kohler")
         """
+        from torchvision.datasets.utils import download_and_extract_archive
+
         root = resolve_root(root, "Kohler")
         for url in cls._archive_urls:
             archive_name = url_basename(url)
--- a/deepinv/datasets/kohler.py
+++ b/deepinv/datasets/kohler.py
@@ -1,15 +1,17 @@
 from __future__ import annotations
 import torch
-from PIL import Image
 
 from urllib.parse import urlparse
 from os.path import basename, join
-from typing import Callable
+from typing import Callable, TYPE_CHECKING
 from types import MappingProxyType
 from pathlib import Path
 
 from deepinv.datasets.base import ImageDataset
 from .utils import resolve_root
+
+if TYPE_CHECKING:
+    from PIL import Image
 
 
 def url_basename(url: str) -> str:
@@ -260,6 +262,8 @@ class Kohler(ImageDataset):
     def get_sharp_frame(
         self, printout_index: int, trajectory_index: int, frame_index: int
     ) -> torch.Tensor | Image.Image | any:
+        from PIL import Image
+
         path = join(
             self.root,
             f"Image{printout_index}",
@@ -274,6 +278,8 @@ class Kohler(ImageDataset):
     def get_blurry_shot(
         self, printout_index: int, trajectory_index: int
     ) -> torch.Tensor | Image.Image | any:
+        from PIL import Image
+
         path = join(self.root, f"Blurry{printout_index}_{trajectory_index}.png")
         blurry_shot = Image.open(path)
         if self.transform is not None:
--- a/deepinv/loss/sup.py
+++ b/deepinv/loss/sup.py
@@ -6,7 +6,6 @@ from torch import Tensor
 from deepinv.loss.loss import Loss
 from deepinv.loss.metric.metric import Metric
 from deepinv.models.base import Reconstructor
-from deepinv.physics.forward import Physics
 
 if TYPE_CHECKING:
     from deepinv.physics.forward import Physics
--- a/deepinv/models/bm3d.py
+++ b/deepinv/models/bm3d.py
@@ -3,7 +3,6 @@ import torch
 from torch import Tensor
 from .base import Denoiser
 from .utils import array2tensor, tensor2array
-from deepinv.physics.functional import dct
 from deepinv.utils import image_to_patches
 
 
@@ -168,6 +167,8 @@ class BM3D(Denoiser):
     def _get_dct_matrix(
         self, n: int, norm: str = "ortho", device: torch.device | None = None
     ) -> tuple[Tensor, Tensor]:
+        from deepinv.physics.functional import dct  # lazy import at init
+
         matrix = dct(torch.eye(n, dtype=torch.float32, device=device), norm=norm).T
         inverse = matrix.T if norm == "ortho" else torch.linalg.inv(matrix)
         return matrix, inverse
--- a/deepinv/models/equivariant.py
+++ b/deepinv/models/equivariant.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from torch import Tensor
 from deepinv.transform import Transform, Rotate, Reflect
 from .base import Denoiser, Reconstructor
-from deepinv.physics.virtual import VirtualLinearPhysics
 import torch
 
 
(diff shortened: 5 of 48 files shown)


Andrewwango:
@nucli-vicky @jscanvic would appreciate a review please! I went a bit obsessive over reducing import time.

pyproject.toml
```diff
@@ -27,8 +27,8 @@ dependencies = [
     "matplotlib",
     "tqdm",
     "torch>=2.2.0",
-    "torchvision",
     "torchmetrics",
+    "torchvision",
```
jscanvic, line 31: Why?

deepinv/__init__.py
```diff
@@ -1,6 +1,5 @@
 from .__about__ import *
 import torch
-import torchvision as _torchvision  # early loading
```
jscanvic, line 3: Nice

deepinv/utils/io.py
```diff
@@ -347,6 +351,8 @@ def _process_array(arr: np.ndarray) -> torch.Tensor:
         )  # Remove single-band dimension if exists
         if arr.ndim == 2:
             if arr.is_complex():
+                from deepinv.utils.mixins import MRIMixin
```
jscanvic, line 354: Why?

deepinv/utils/io.py
```diff
@@ -71,21 +69,24 @@ def get_cache_home() -> Path:
     return path
 
 
-class DownloadError(requests.exceptions.RequestException):
-    r"""Raised when a network download initiated by deepinv fails.
+class DownloadError(IOError):
```
jscanvic, line 72:
```suggestion
class DownloadError(OSError):
```

OSError is the canonical alias of IOError

https://docs.python.org/3/library/exceptions.html#IOError

jscanvic, line 72: Are we sure this is a minor/negligible breakage to change the inheritance chain? I guess it is but I'm just making sure

deepinv/utils/__init__.py
```diff
@@ -7,7 +7,6 @@
     plot_inset,
     plot_videos,
     save_videos,
-    make_grid,
```
jscanvic, line 10: Put it as breaking in the changelog?

deepinv/transform/rotate.py
```diff
@@ -57,6 +61,10 @@ def __init__(
             else interpolation_mode
         )
 
+        from torchvision.transforms.functional import rotate as torchvision_rotate
+
+        self.torchvision_rotate = torchvision_rotate  # lazy import
```
jscanvic, line 66:
```suggestion
        self._torchvision_rotate = torchvision_rotate  # lazy import
```

deepinv/transform/rotate.py
```diff
@@ -86,7 +94,7 @@ def _transform(
         """
         return torch.cat(
             [
-                rotate(
+                self.torchvision_rotate(
```
jscanvic, line 97:
```suggestion
                self._torchvision_rotate(
```

deepinv/optim/data_fidelity.py
```diff
@@ -350,6 +349,10 @@ def __init__(self, sigma: float = 1.0, threshold: float = 1.0):
         self.norm = 1 / (sigma**2)
         self.modulo_round = lambda x: x - threshold * torch.round(x / threshold)
 
+        from deepinv.physics.functional import dct_2d, idct_2d
+
+        self.dct_2d, self.idct_2d = dct_2d, idct_2d  # lazy import
```
jscanvic, line 354:
```suggestion
        self._dct_2d, self._idct_2d = dct_2d, idct_2d  # lazy import
```

deepinv/optim/data_fidelity.py
```diff
@@ -539,14 +542,14 @@ def prox(
                 - (torch.cos(torch.pi * I / MX) + torch.cos(torch.pi * J / NX))
             )
 
-        dct_psi = dct_2d(psi, norm="ortho")
+        dct_psi = self.dct_2d(psi, norm="ortho")
```
jscanvic, line 545:
```suggestion
        dct_psi = self._dct_2d(psi, norm="ortho")
```

deepinv/optim/data_fidelity.py
```diff
@@ -539,14 +542,14 @@ def prox(
...
 
         denom = denom.to(psi.device)
         denom[..., 0, 0] = 1  # avoid division by zero
 
         dct_phi = dct_psi / denom
 
-        phi = idct_2d(dct_phi, norm="ortho")
+        phi = self.idct_2d(dct_phi, norm="ortho")
```
jscanvic, line 552:
```suggestion
        phi = self._idct_2d(dct_phi, norm="ortho")
```

deepinv/datasets/kohler.py
```diff
@@ -154,6 +155,8 @@ def download(cls, root: str | Path = None, remove_finished: bool = False) -> Non
                 from deepinv.datasets import Kohler
                 Kohler.download("datasets/Kohler")
         """
+        from torchvision.datasets.utils import download_and_extract_archive
```
nucli-vicky, line 158: We can probably have the same function with existing `deepinv` functions in `datasets/utils.py` quite easily: https://github.com/pytorch/vision/blob/da446275eccae6cd00923080d9cc91f1f3a73744/torchvision/datasets/utils.py#L374-L391

deepinv/loss/sup.py
```diff
@@ -6,7 +6,6 @@
 from deepinv.loss.loss import Loss
 from deepinv.loss.metric.metric import Metric
 from deepinv.models.base import Reconstructor
-from deepinv.physics.forward import Physics
```
nucli-vicky, line 9: good catch :)

deepinv/models/bm3d.py
```diff
@@ -168,6 +167,8 @@ def _get_wavelet_matrix_torch(
     def _get_dct_matrix(
         self, n: int, norm: str = "ortho", device: torch.device | None = None
     ) -> tuple[Tensor, Tensor]:
+        from deepinv.physics.functional import dct  # lazy import at init
```
nucli-vicky, line 170: I'm a bit confused: in some classes, you added that we import during the `__init__` of the class, and in others (like here), import within the class method that needs the import.

Is there a consistent logic for this right now?

deepinv/models/equivariant.py
```diff
@@ -141,6 +140,12 @@ def __init__(
         self.transform = transform
         self.eval_transform = eval_transform
 
+        from deepinv.physics.virtual import VirtualLinearPhysics
+
+        self.virtual_linear_physics = (
+            VirtualLinearPhysics  # lazy import to break import chains
+        )
+
```
nucli-vicky, line 148: For example, this import can also be done at `forward`. I don't mind either (although importing within the method that needs it is less code), just wondering what the logic is right now :)

deepinv/utils/__init__.py
nucli-vicky, line 10: this would break user code doing 'from deepinv.utils import make_grid`.

I think it makes sense to remove, but we should either (i) clearly note this in changelog or (ii) write `make_grid` within `deepinv`

nucli-vicky (commented):
Thanks @Andrewwango 

I especially like the `TYPE_CHECKING` gated imports when a class isn't needed besides type checking :)

I know there is some disagreement in styleguides w.r.t the use of lazy imports, but personally I'm fine with them here as it is for performance reasons.

Some additional comments, aside from the bugs causing current test errors

deepinv/models/bm3d.py
Andrewwango: good point. I was not consistent, I was just doing whatever felt most elegant per class, which is purely subjective. I think the logic should be: external imports (torchvision, PIL) should be imported at least once a init, and keep a reference to the object (otherwise ruff will get angry we don't use the import) - or, even don't keep a reference, and suppress ruff. Internal deepinv imports, we can do it in the methods

deepinv/models/equivariant.py
Andrewwango: agreed, can do at forward

jscanvic (changes requested):
Thanks Andrew!

A bit of a messy review (sorry!) I did back then and wasn't 100% happy with - feel free to ignore some of the comments

Once you've been through it & CI passes happy to approve

PR #1254, opened 2026-07-01
Fix coverage not uploaded to codecov

romainvo:
Small fix to `test_cpu.yml`to bring back codecoverage upload.

### LLM policy

### DIFF
--- a/.github/workflows/test_cpu.yml
+++ b/.github/workflows/test_cpu.yml
@@ -114,7 +114,7 @@ jobs:
           key: deepinv-url-cache-${{ matrix.os }}-${{ github.run_id }}
 
       - name: Upload coverage to Codecov
-        if: runner.os == 'Linux' && matrix.environment == 'full'
+        if: runner.os == 'Linux' && matrix.environment == 'full-cpu'
         uses: codecov/codecov-action@v5
         with:
           fail_ci_if_error: true
--- a/.github/workflows/test_cpu.yml
+++ b/.github/workflows/test_cpu.yml
@@ -96,6 +96,7 @@ jobs:
       # (pytest-cov does not take into account skipped tests by testmon)
       # If .coverage exists, testmon skips the non-related tests and --cov-append updates coverage with respect to related changes.
       - name: Test with pytest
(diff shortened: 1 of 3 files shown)


Tmodrzyk (approved):
Well, I guess it works now since the CI complains again aha.

romainvo:
> Well, I guess it works now since the CI complains again aha.

I've checked locally, coverage is not that low... `testmon` is messing with the coverage, which does not take into account skipped tests because of unchanged logic

EDIT: we can add an initial full `.coverage` to the cache (I'm adding a mechanism to not use `--testmon` if no `.coverage` exists yet) and compute a running update of the coverage with `--cov-append`. This should give us proper coverage results in each PR.

.github/workflows/test_cpu.yml
```diff
@@ -114,7 +124,7 @@ jobs:
           key: deepinv-url-cache-${{ matrix.os }}-${{ github.run_id }}
 
       - name: Upload coverage to Codecov
-        if: runner.os == 'Linux' && matrix.environment == 'full'
+        if: runner.os == 'Linux' && matrix.environment == 'full-cpu'
```
tachella, line 127: I'd keep the 'full' environment too, since this is the one that runs when we merge to main (which then feeds the coverage badge)

tachella (changes requested):

.github/workflows/test_cpu.yml
romainvo: I'm adding a codecov upload to test_gpu.yml (`full`environment). I'm using the flag syntax from codecov (https://docs.codecov.com/docs/flags) to differentiate between the coverage results of the two environments (`full` vs `full-cpu`)

romainvo: I'm adding a codecov upload to test_gpu.yml (`full`environment). I'm using the flag syntax from codecov (https://docs.codecov.com/docs/flags) to differentiate between the coverage results of the two environments (`full` vs `full-cpu`)

tachella (approved):

PR #1255, opened 2026-07-01
Add subsets for Tomography physics and OSEM

### LINKED ISSUE #975: Tomography baseline iterative algorithms

### LINKED ISSUE #1181: Add NRMSE metric
### Environment (optional)

_No response_

### Describe the feature

The Normalized Root Mean Squared Error (NRMSE) metric is standard in emission tomography to measure pixel by pixel distorsion: https://en.wikipedia.org/wiki/Root_mean_square_deviation

### Motivation

While Deepinv already has MSE and NMSE classes, not computing the squared root inside the ``forward`` makes it harder to interface with other classes in the library.

### Alternatives considered

_No response_

### Additional context

_No response_

Tmodrzyk:
This PR aims at adding an important strategy for reconstruction algorithms used in emission tomography called *ordered-subsets*, with the modified MLEM algorithm being often referred to as OSEM ([see the original paper here](https://ieeexplore.ieee.org/document/363108)). Towards #975.

The idea is basically to split the forward operator $A$ and the measurements $y$ into non-overlapping blocks, and then to apply these smaller updates to compute the complete image $x$. This strategy speeds up reconstruction significantly, and is commonly used both in a clinical and research setting.

The feature relies on two helper functions converting a `Tomography` or `TomographyWithAstra` physics to a `StackedLinearPhysics` containing the subsetted physics. The MLEM algorithm can then be called with `num_subsets>1` to apply OSEM instead. By default `num_susbets=1`.

The MLEM demo file `demo_poisson_mlem.py` has also been updated to showcase the subsetting mechanism and associated speed-ups. 
The PR also adapts MLEM to the new PET physics, and the subset mechanism also works for this operator.
To make it more convenient I had to adapt the PET physics slightly, especially to easily pass attributes to the subsetted physics.

Other related stuff fixed in this PR: 
- The 3D PET demo was adding noise differently than through `gain`, and did not normalize the operator. 
- This is now the case to match the 2D example.
- The plot utility did not share colorbars between the different images when `rescale_mode="clip" ` so this was also fixed.

### LLM policy


deepinv/physics/pet.py
```diff
@@ -82,6 +82,7 @@ class PET(LinearPhysics):
     :param bool normalize_counts: If `False` the :math:`\gamma` normalization term in front of the Poisson noise is removed,
         so that the measurements :math:`y` are true integer counts.
     :param str | torch.device device: device to run the computations on, e.g. `"cpu"` or `"cuda"`
+    :param None, torch.Tensor views: PET sinogram views to project. If ``None``, all views are projected.
```
romainvo: I would specify the expected dtype here and what "views" mean, it refers to view indexes I suppose ?

deepinv/physics/tomography.py
romainvo, line 1: Should we put these utilities in the `physics/functional` module or some other file, to not bloat the already large tomography file ?

deepinv/physics/tomography.py
```diff
@@ -21,6 +26,206 @@
...
+        raise ValueError(f'Unknown subsetting strategy "{strategy}".')
+    if num_angles % num_subsets != 0:
+        raise ValueError(
+            "The default subsetting strategy requires num_angles to be divisible "
+            "by num_subsets."
+        )
+
+    subset_size = num_angles // num_subsets
+    indices = torch.arange(num_angles, device=device)
+    return list(indices.reshape(num_subsets, subset_size))
```
romainvo: Shouldn't we do the splitting across the full range with equally spaced angles, i.e. for indices from 0 to 10, do (0,2,4,6,8) + (1,3,5,7,9)? I never bothered to look how it was coded in other projects but I assumed it was like that, as neighbor positions carry lot of redundance.

deepinv/physics/tomography.py
```diff
@@ -21,6 +26,206 @@
...
+def split_measurements(
+    y: torch.Tensor,
+    physics: LinearPhysics,
+    num_subsets: int,
+    strategy: str = "default",
+) -> TensorList:
+    r"""
+    Splits tomography measurements into angular subsets.
+
+    :param torch.Tensor y: full measurement tensor.
```
romainvo: I would specify the potentially encountered shapes of y (different for Tomography, TomographyWithAstra, also 2D vs 3D)

deepinv/physics/tomography.py
```diff
@@ -21,6 +26,206 @@
...
+    :param deepinv.physics.LinearPhysics physics: tomography physics.
+    :param int num_subsets: number of subsets.
+    :param str strategy: subsetting strategy. Currently only ``"default"`` is supported.
+    :return: measurements as a :class:`deepinv.utils.TensorList`.
+    """
+    if isinstance(physics, Tomography):
+        num_angles = len(physics.theta)
+        dim = -1
```
romainvo: I think measurements for Tomography have a shape [B,C,A,N] with A the angle dimension, or am I mistaken ? if yes dim should be -2

deepinv/physics/tomography.py
```diff
@@ -21,6 +26,206 @@
...
+                angles=theta_subset,
+                normalize=False,
+                device=physics.device,
+                **physics._subset_kwargs,
+            )
+            subset.normalize = physics.normalize
+            if physics.normalize:
+                subset.register_buffer(
+                    "operator_norm", physics.operator_norm.detach().clone()
+                )
```
romainvo: The spectral norm of the subset matrix A_s is going to be smaller than the original operator, for Tomography it's approximately scaled by 1/sqrt(num_subset). 

If all the subsets have exactly the same size, I think it's fair to compute it only once and assign the same values to others, otherwise we need to compute it at least once per subset size.

deepinv/physics/tomography.py
```diff
@@ -21,6 +26,206 @@
...
+        )
+        astra_angles = (
+            None
+            if astra_geometry_vectors is not None
+            else -torch.rad2deg(
+                torch.as_tensor(
+                    physics.projection_geometry["ProjectionAngles"],
+                    device=physics.device,
+                )
+            )
```
romainvo: Maybe add this as TomographyWithAstra property ? seems legit to be able to access the set of angles

deepinv/physics/tomography.py
```diff
@@ -21,6 +26,206 @@
...
+                geometry_vectors=geometry_vectors,
+                normalize=False,
+                device=physics.device,
+                **physics._subset_kwargs,
+            )
+            subset.normalize = physics.normalize
+            if physics.normalize:
+                subset.register_buffer(
+                    "operator_norm", physics.operator_norm.detach().clone()
+                )
```
romainvo: Same comment than above about the spectral norm

deepinv/physics/tomography.py
```diff
@@ -21,6 +26,206 @@
...
+                normalize=False,
+                normalize_counts=bool(normalize_counts.item()),
+                device=physics.background.device,
+                **physics._subset_kwargs,
+            )
+            subset.normalize = physics.normalize
+            if physics.normalize:
+                subset.register_buffer(
+                    "operator_norm", physics.operator_norm.detach().clone()
+                )
```
romainvo: Same comment about spectral norm here

deepinv/physics/tomography.py
```diff
@@ -21,6 +26,206 @@
...
+        num_angles, num_subsets, strategy=strategy, device=y.device
+    )
+    return TensorList([y.index_select(dim, idx) for idx in indices])
+
+
+def split_physics(
+    physics: LinearPhysics,
+    num_subsets: int,
+    strategy: str = "default",
+) -> StackedLinearPhysics:
```
romainvo: Overall I feel like this could be an instance method of LinearPhysics as splitting a physics will always be specific and involve physics-related logic. For future refactoring this will avoid having an overly large `split_physics` method that takes care of every config in the library (this will also remove the need for _subset_kwargs).

By default any LinearPhysics.split_physics would raise a NotImplementedError (or a greedy split along group of lines of the underlying matrix A, but not sure this is very useful)

And obviously in this PR we implement in each Tomography class the "per-angle" split that comes naturally for these type of operators.

The same logic applies to `split_measurements`.

We could also keep these two functions exposed on the `deepinv.physics` module which would call the specific `LinearPhysics.split_physics/split_measurements` methods.

--- 

The important question is, do we consider the splitting logic generic enough to add this to the LinearPhysics API. Maybe this is early, but I feel like there are many many powerful applications (ordered subsets can be applied for any gradient/proximal base strategy, also in other flavors, eg. stochastic). 

It will also provide a prefered interface to distribute a linear physics. This is one of the distributed framework limitation for the moment, that the user needs to split the base LinearPhysics and build a StackedLinearPhysics on their own.

deepinv/utils/plotting.py
```diff
@@ -450,8 +450,13 @@ def plot(
 
     for i, row_imgs in enumerate(imgs):
         for r, img in enumerate(row_imgs):
+            display_kwargs = imshow_kwargs.copy()
+            is_rgb = img.ndim == 3 and img.shape[-1] in (3, 4)
+            if not is_rgb and "norm" not in display_kwargs:
+                display_kwargs.setdefault("vmin", 0.0)
+                display_kwargs.setdefault("vmax", 1.0)
```
romainvo: Are we sure we want to do that? Some signal can have values below zero, or also a pixel dynamic very low compared to 1

deepinv/tests/test_physics.py
```diff
@@ -1473,6 +1473,49 @@ def test_tomography(
     ), f"error: {error} > {r_tol}, fanbeam={fan_beam}, circle={circle}, fbp_interpolate_boundary={fbp_interpolate_boundary}, normalize={normalize}, adjoint_via_backprop={adjoint_via_backprop}, parallel_computation={parallel_computation}, fbp_pseudo_inverse={fbp_pseudo_inverse}"
 
 
+def test_tomography_subset_utilities(device):
```
romainvo: I feel like we should also check that properties of the base `LinearPhysics` are preserved when converted into a `StackedLinearPhysics`, e.g. spectral norm computation, adjoint, forward, etc ?

deepinv/optim/optim_iterators/mlem.py
```diff
@@ -31,18 +35,56 @@ def forward(
...
+            measurements = (y,)
+            physics_list = (physics,)
+            prior_scale = 1.0
+
+        x = x_prev
+        for cur_y, cur_physics in zip(measurements, physics_list, strict=True):
+            # E-step/MM correction: compare measured counts to predicted counts.
+            sensitivity = cur_physics.A_adjoint(ones_like(cur_y))
+            if hasattr(cur_physics, "background"):
+                proj = cur_physics.A(x, add_background=True)
```
romainvo: I would specify which operator it concerns in the comment, i.e PET. 

Naive question, why isn't `add_background=True` the default mode for PET?

deepinv/optim/optim_iterators/mlem.py
```diff
@@ -31,18 +35,56 @@ def forward(
...
+            numerator = x * cur_physics.A_adjoint(cur_y / proj.clamp(min=self.eps))
             denom = sensitivity
 
-        x = x / denom.clamp(min=1e-15)
+            # Optional OSL prior contribution in the multiplicative denominator.
+            if cur_prior is not None:
+                if cur_prior.__class__.__name__ == "TVPrior":
+                    dx = cur_prior.nabla(x)
+                    norm = torch.linalg.vector_norm(dx, ord=2, dim=-1, keepdim=True)
+                    prior_grad = cur_prior.nabla_adjoint(dx / norm.clamp_min(self.eps))
```
romainvo: It feels a bit over-specific, is there a way around it? Naive question maybe but why don't we just call cur_prior.grad when it's a TVPrior, is it a matter of efficiency?

deepinv/optim/optimizers.py
```diff
@@ -2359,6 +2382,31 @@ def __init__(
             **kwargs,
         )
 
+    def forward(self, y, physics, *args, **kwargs):
```
romainvo: Should we authorize the user to pass an already splitted physics, like for any reasons they already did it because it's part of the pipeline and we don't want to do it again here?

Could you also add typing please.

deepinv/optim/optimizers.py
```diff
@@ -2359,6 +2382,31 @@ def __init__(
...
+                    "TomographyWithAstra or PET."
+                )
+
+            kwargs["subset_physics"] = split_physics(
+                physics, self.num_subsets, strategy=self.subset_strategy
+            )
+            kwargs["y_subsets"] = split_measurements(
+                y, physics, self.num_subsets, strategy=self.subset_strategy
+            )
+        return super().forward(y, physics, *args, **kwargs)
```
romainvo: What do you think of instead doing this 
```suggestion
    if self.num_subsets > 1:
        # above code ..

        subset_physics = split_physics(
            physics, self.num_subsets, strategy=self.subset_strategy
        )
        subset_measurements = split_measurements(
            y, physics, self.num_subsets, strategy=self.subset_strategy
        )

        return super().forward(subset_measurements, subset_physics, *args, **kwargs)
    else:
        return super().forward(y, physics, *args, **kwargs)
```

That way it is clear what is used for each case. That means we also dont need to query from kwargs in `MLEMIteration.forward` and jungle between the two format of measurements. 

This will require adding a `bool: subset` flag or a `num_subsets` argument to MLEMIteration.forward though, in order to make it explicit what should be an OS-computation and what should not (some physics are assymetric StackedLinearPhysics by default, eg PanSharpen). I don't feel it's a big deal, it should also increase readibility. 

This is a design choice, but I feel like it improves clarity and maintainability, happy to discuss it further :) !

romainvo (commented):
Hi @Tmodrzyk for the very cool PR, it opens up some many avenues! My comments are mostly on design, as I feel the new `split_physics` functionality poses major questions for potentially new interesting features (and also existing, e.g distributed).

If you feel like it's out of scope to deal with this here, we can merge as is (without the design comments) and deal with it in another PR/issue. Happy with whatever you choose and to discuss further :)

---

(I did not do a pass on the examples yet but they run fine on my local setup at least!)

deepinv/physics/tomography.py
Tmodrzyk: Yes I think you are right, I'll check and point to a paper if possible.
Here I just implemented the "easiest" one, but we should definitely add more.
I know there are many many division and selection strategies, some can get really fancy. 
For instance check [this challenge paper from Schramm](https://scholar.google.com/citations?view_op=view_citation&hl=de&user=_txZ90cAAAAJ&sortby=pubdate&citation_for_view=_txZ90cAAAAJ:S16KYo8Pm5AC)

Tmodrzyk:
Hi @romainvo thank you for the careful review ! I'll answer the specific comments in their threads.

For the broader design choices: 
I agree the subsetting mechanism could be made more general and apply to other physics. 
However I'm not sure / not familiar with other communities (than emission tomography people) using this kind of splitting.
It seems that each implementation of `split_physics` would be very specific, with barely any code reusable across physics ? 
Splitting across "lines" of the forward operator seems non-trivial to me: e.g for `Blur` the actual matrix is a huge block-Toeplitz matrix, but the Blur is applied through a convolution, not through matrix multiplication. So we'd have to map each line of the block-Toeplitz matrix to an equivalent convolution. 
For other physics it could be easier ofc.

The closest common abstraction would be (as you mentioned) stochastic methods (SGD, SAGA etc) but I don't think we support these yet ? I'm not very familiar with these either.
Supporting these in general, I feel, is beyond the scope of this PR, but we should definitely do it.

For now, I feel more confident keeping it limited to the tomography operators. I like your idea of getting it out of `tomography.py` though !

Also, one thing I don't like with the current implementation is to have both `split_physics` and `split_measurements`... 
It feels to me that `split_measurements` could be somehow removed but I don't have a solution in mind. What do you think ?  

Edit: I was also thinking about making a brand new algorithm class `OSEM` which only works on subsetted data to keep the implementation cleaner and avoid bloating `MLEM`, but there will be a bit of code duplication. Again, not sure what's best here.

deepinv/optim/optim_iterators/mlem.py
Tmodrzyk: Totally agree this is over-specific, what we should do is make it an option to use an explicit choice of subgradient  (see https://github.com/deepinv/deepinv/issues/1259).
TV is non-smooth so using autograd on it is very unstable, which is what `.grad` is currently doing.
I've also added this to the Blind PR (https://github.com/deepinv/deepinv/pull/988) so I guess we should just complete another PR for the subgradient thingy before merging subsets/blind ? :sweat_smile:

deepinv/optim/optim_iterators/mlem.py
Tmodrzyk: I'm not sure, I think it has to be initialized with patient-specific data so by default it would be None and so it wouldn't add anything.

romainvo:
> It seems that each implementation of `split_physics` would be very specific, with barely any code reusable across physics ? 

That's the reason I had in mind when I said `split_physics` could be a LinearPhysics instance method

> Splitting across "lines" of the forward operator seems non-trivial to me: e.g for `Blur` the actual matrix is a huge block-Toeplitz matrix, but the Blur is applied through a convolution, not through matrix multiplication. So we'd have to map each line of the block-Toeplitz matrix to an equivalent convolution. For other physics it could be easier ofc.

Indeed, that would not be efficient at all; it was more to illustrate how general the splitting logic could be made across linear operators. The efficient way is just to return a TypeError if no efficient splitting exists (or isn't known).

> The closest common abstraction would be (as you mentioned) stochastic methods (SGD, SAGA etc) but I don't think we support these yet ? I'm not very familiar with these either. Supporting these in general, I feel, is beyond the scope of this PR, but we should definitely do it.

No, I don't think we do. One thing to be aware of is the compatibility with the distributed framework. The `distribute` method exactly expects a `StackedPhysics` as you've built them. I would just make sure that the output of `split_physics` can be passed to `distribute`, and that the distributed inference works properly. But I agree this is out of scope, we can deal with this later.

> For now, I feel more confident keeping it limited to the tomography operators. I like your idea of getting it out of `tomography.py` though !

Yes, completely agree.

> Also, one thing I don't like with the current implementation is to have both `split_physics` and `split_measurements`... It feels to me that `split_measurements` could be somehow removed but I don't have a solution in mind. What do you think ?

It does not bother me to have two separate methods, measurements and physics are very distinct type of objects, I don't know if we can factor that much of the splitting logic. We do pass a `physics` to `split_measurements` so maybe adding a flag, to also split the `physics` if specified, not sure either...  Feels ok to me as is to be honest.

> Edit: I was also thinking about making a brand new algorithm class `OSEM` which only works on subsetted data to keep the implementation cleaner and avoid bloating `MLEM`, but there will be a bit of code duplication. Again, not sure what's best here.

I don't have a strong opinion on that, whichever feels best to you. The brand new OSEM is nice, we can factor the backend later if we find something smart to do. If we do a separate OSEM, maybe I would still authorize the user to pass a non-splitted physics, and OSEM splits it.

deepinv/optim/optim_iterators/mlem.py
Tmodrzyk: I switched to just `.grad()` and when the explicit TV subgradient will be implemented it will have the same behavior !

deepinv/tests/test_physics.py
Tmodrzyk: Agreed I didn't think about that !

deepinv/physics/tomography.py
Tmodrzyk: I implemented the interleaved angles you suggested, we can always add fancier strategies later on :smile:

deepinv/physics/tomography.py
Tmodrzyk: Well I double-checked and the `Tomography` dosctrings was incorrect, fixed it to [B, C, N, A]

deepinv/physics/pet.py
Tmodrzyk: Indeed, I specified that the tensor should contain integers

deepinv/physics/tomography.py
Tmodrzyk: Done, the new version of `split_physics` is in `physics/functional/subsets.py`

deepinv/physics/tomography.py
Tmodrzyk: done !

deepinv/utils/plotting.py
Tmodrzyk: Indeed and I'm not 100% I fixed the bug here, I reverted these changes and the utility will be fixed in a separate PR, also to take into account #1280.

deepinv/physics/tomography.py
Tmodrzyk: So as discussed in the other global comments, I kept the subset mechanism as something specific to tomography, although I would love to make it more general. 
I also moved all the subset logic out of `tomography.py` into `physics/functional/subsets.py`.

Tmodrzyk:
Hi @romainvo I believe this is ready for a second round of review, the CPU CI test failure seems unrelated.
Sorry for the huge changelist, I should have splitted the PR into smaller chuncks.

I got `split_physics` and `split_measurements` out of `tomography.py` and into `physics/functional/subsets.py` .
I removed as much changes as possible from `PET` and `Tomography` physics, e.g `subset_kwargs`.
I defined a separate `OSEM` class to use the subset for tomography, it supports both pre-split and full physics / measurements.
I kept the fix in `MLEM` related to the background term, so that it works for PET.
I also added NRMSE metric to use it in the pet examples, and to take care of #1181.

In the 2D and 3D pet examples I added sections making use of OSEM to show the reconstruction runtime gains.

romainvo:
> Hi @romainvo I believe this is ready for a second round of review, the CPU CI test failure seems unrelated. Sorry for the huge changelist, I should have splitted the PR into smaller chuncks.
> 
> I got `split_physics` and `split_measurements` out of `tomography.py` and into `physics/functional/subsets.py` . I removed as much changes as possible from `PET` and `Tomography` physics, e.g `subset_kwargs`. I defined a separate `OSEM` class to use the subset for tomography, it supports both pre-split and full physics / measurements. I kept the fix in `MLEM` related to the background term, so that it works for PET. I also added NRMSE metric to use it in the pet examples, and to take care of #1181.
> 
> In the 2D and 3D pet examples I added sections making use of OSEM to show the reconstruction runtime gains.

Thanks for the changes, looking into this ASAP!

deepinv/optim/optim_iterators/mlem.py
```diff
@@ -31,9 +49,16 @@ def forward(
         """
         x_prev = X["est"][0]
         k = 0 if "it" not in X else X["it"]
-        sensitivity = physics.A_adjoint(ones_like(y))
 
-        x = x_prev * physics.A_adjoint(y / (physics.A(x_prev).clamp(min=1e-15)))
+        sensitivity = physics.A_adjoint(torch.ones_like(y))
```
Andrewwango: would be nice in future to precompute sensitivity since this is constant

deepinv/optim/optim_iterators/mlem.py
```diff
@@ -1,28 +1,46 @@
...
-from torch import ones_like
+
+if TYPE_CHECKING:
+    from deepinv.optim import DataFidelity, Prior
+    from deepinv.physics import Physics
 
 
 class MLEMIteration(OptimIterator):
-    """
+    r"""
```
Andrewwango, line 15: minor point but somewhere inthe MLEM doc please can you specify the initialisation?
also in future it would be nice to allow different initialisation (e.g. with the sensitivity, or with zeros, or with adjoint) but this isn't urgent. I guess users can already do this with custom_init, maybe add it to the doc? To make life easier to new nuclear medicine users :)

deepinv/optim/optim_iterators/osem.py
```diff
@@ -0,0 +1,133 @@
...
+
+    One iteration corresponds to a complete OSEM epoch, applying one
+    multiplicative update for each measurement and physics subset.
+    More details can be found in the documentation of the
+    :class:`deepinv.optim.optimizers.OSEM` optimizer.
+    """
+
+    def __init__(self, eps: float = 1e-6, cost_fn=None, **kwargs):
+        self.eps = eps
+        super(OSEMIteration, self).__init__(cost_fn=None, **kwargs)
```
Andrewwango: swap this line with above

deepinv/optim/optim_iterators/osem.py
```diff
@@ -0,0 +1,133 @@
...
+        cur_data_fidelity: DataFidelity,
+        cur_prior: Prior | None,
+        cur_params: dict,
+        y: torch.Tensor | TensorList,
+        physics: Physics,
+        *args,
+        **kwargs,
+    ) -> dict[str, tuple[torch.Tensor, None] | torch.Tensor | int | None]:
+        r"""
+        Perform one Ordered-Subsets Expectation-Maximization epoch.
```
Andrewwango:
```suggestion
        Perform one Ordered-Subsets Expectation-Maximization step.
```

deepinv/optim/optim_iterators/osem.py
```diff
@@ -0,0 +1,133 @@
...
+
+        :param dict X: Dictionary containing the current iterate and the estimated cost.
+        :param deepinv.optim.DataFidelity cur_data_fidelity: Instance of the DataFidelity class defining the current data fidelity.
+        :param deepinv.optim.Prior cur_prior: Instance of the Prior class defining the current prior.
+        :param dict cur_params: Dictionary containing the current parameters of the algorithm.
+        :param torch.Tensor, deepinv.utils.TensorList y: Full input data or pre-split measurements.
+        :param deepinv.physics.Physics physics: Full physics or pre-split stacked physics modeling the data-fidelity term.
+        :param deepinv.utils.TensorList y_subsets: Measurement subsets.
+        :param deepinv.physics.StackedLinearPhysics subset_physics: Physics operators corresponding to the measurement subsets.
```
Andrewwango: why make this disappear in kwargs and not an explicit kwarg?

deepinv/optim/optim_iterators/mlem.py
```diff
@@ -1,28 +1,46 @@
...
     Iterator for the Maximum-Likelihood Expectation-Maximization (MLEM) algorithm for Poisson inverse problems.
 
     Class for a single iteration of the MLEM algorithm :footcite:t:`sheppMaximumLikelihoodReconstruction1982`,
```
Andrewwango:
```suggestion
    Class for a single iteration of the MLEM algorithm :footcite:p:`sheppMaximumLikelihoodReconstruction1982`,
```

deepinv/optim/optim_iterators/osem.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class OSEMIteration(OptimIterator):
+    r"""
+    Iterator for the Ordered-Subsets Expectation-Maximization (OSEM) algorithm.
+
+    One iteration corresponds to a complete OSEM epoch, applying one
+    multiplicative update for each measurement and physics subset.
+    More details can be found in the documentation of the
+    :class:`deepinv.optim.optimizers.OSEM` optimizer.
```
Andrewwango, line 20:
```suggestion
    Performs a single iteration of the OSEM algorithm :footcite:p:`hudsonAcceleratedImageReconstruction1994`,
    which is a classic baseline reconstruction method for inverse problems with Poisson noise statistics. Note that :class:`deepinv.optim.optim_iterators.MLEMIteration` is a special case with one subset only.
    More details on the algorithm can be found in the documentation of the :class:`deepinv.optim.optimizers.OSEM` optimizer.
```

deepinv/optim/optimizers.py
```diff
@@ -2251,10 +2253,11 @@ class MLEM(BaseOptim):
     At each iteration, the algorithm performs a multiplicative update of the form:
 
     .. math::
-        x_{k+1} = \frac{x_k}{A^T \mathbf{1}} \odot A^T \left(\frac{y}{A x_k}\right)
+        x_{k+1} = \frac{x_k}{A^T \mathbf{1}} \odot A^T \left(\frac{y}{A x_k + b}\right)
 
     where :math:`A` is the forward operator, :math:`y` is the observed data,
-    :math:`\mathbf{1}` is a tensor of ones, and :math:`\odot` denotes element-wise multiplication.
+    :math:`b` is an optional additive background, :math:`\mathbf{1}` is a tensor of ones,
```
Andrewwango:
```suggestion
    :math:`b` is an optional additive background (useful in PET), :math:`\mathbf{1}` is a tensor of ones,
```

deepinv/optim/optimizers.py
```diff
@@ -2360,6 +2369,224 @@ def __init__(
...
+class OSEM(BaseOptim):
+    r"""
+    Ordered-Subsets Expectation-Maximization (OSEM) algorithm for Poisson inverse problems.
+
+    OSEM was proposed by Hudson and Larkin
+    :footcite:t:`hudsonAcceleratedImageReconstruction1994` to accelerate the
+    MLEM algorithm of Shepp and Vardi
+    :footcite:t:`sheppMaximumLikelihoodReconstruction1982` by splitting the
+    measurement into ordered subsets. One iteration of this optimizer is one
+    complete epoch over all subsets:
```
Andrewwango:
```suggestion
    OSEM was proposed in footcite:t:`hudsonAcceleratedImageReconstruction1994` to accelerate MLEM :footcite:p:`sheppMaximumLikelihoodReconstruction1982` by splitting the
    measurement into ordered subsets. Note MLEM is a special case of OSEM with only one subset. At each iteration, the algorithm performs multiplicative updates over all subsets of the form:
```

deepinv/optim/optimizers.py
```diff
@@ -2360,6 +2369,224 @@ def __init__(
...
+
+    .. math::
+        x_{k,l+1} =
+        \frac{x_{k,l}}{A_l^T \mathbf{1}}
+        \odot A_l^T \left(\frac{y_l}{A_l x_{k,l} + b_l}\right),
```
Andrewwango: same line, this is bloating a bit

deepinv/optim/optimizers.py
```diff
@@ -2360,6 +2369,224 @@ def __init__(
...
+
+    where :math:`A_l` and :math:`y_l` are the corresponding physics and measurement
+    subset, and :math:`b_l` is an optional additive background.
```
Andrewwango:
```suggestion
    subset, :math:`b_l` is an optional additive background, and `l` is the subset index.
```

deepinv/optim/optimizers.py
```diff
@@ -2360,6 +2369,224 @@ def __init__(
...
+        The user can provide either the full measurement tensor and full tomography
+        / PET physics, or pre-split measurements and a
+        :class:`deepinv.physics.StackedLinearPhysics`.
+
+        In the first case, OSEM splits the inputs internally using
+        :func:`deepinv.physics.split_physics` and
+        :func:`deepinv.physics.split_measurements`.
+        Pre-split measurements can be provided as a :class:`deepinv.utils.TensorList`
+        or a list of tensors. Lists are converted to a ``TensorList`` before
+        optimization.
```
Andrewwango:
```suggestion
        The user can provide either the full measurement tensor `y` and full tomography
        / PET `physics` (and OSEM will split internally using :func:`deepinv.physics.split_physics` and
        :func:`deepinv.physics.split_measurements`),
        or pre-split measurements passed as a :class:`deepinv.utils.TensorList` and pre-split physics passed as a :class:`deepinv.physics.StackedLinearPhysics`.
```

deepinv/optim/optimizers.py
```diff
@@ -2360,6 +2369,224 @@ def __init__(
...
+
+    See :class:`deepinv.optim.optim_iterators.OSEMIteration` for the details of one
+    epoch. For pre-split inputs, a non-stacked ``data_fidelity`` is evaluated
+    independently on each subset and summed when computing the objective.
```
Andrewwango:
```suggestion
    See :class:`deepinv.optim.optim_iterators.OSEMIteration` for the details of one iteration.
```
in general the docstring is rather verbose and includes some less important info

deepinv/optim/optimizers.py
```diff
@@ -2360,6 +2369,224 @@ def __init__(
...
+
+    A regularization can be included through the ``prior`` argument using the
+    same One-Step-Late (OSL) heuristic of Green
+    :footcite:t:`greenUseEmAlgorithm1990` as :class:`deepinv.optim.MLEM`.
+    Combined with the subset mechanism, this algorithm is sometimes called OS-MAP-OSL.
```
Andrewwango:
```suggestion
    A regularization can be included by specifying a `prior`. This uses One-Step-Late (OS-MAP-OSL)
    :footcite:p:`greenUseEmAlgorithm1990`, similar to in :class:`deepinv.optim.MLEM`.
```

deepinv/optim/optimizers.py
```diff
@@ -2360,6 +2369,224 @@ def __init__(
...
+
```
Andrewwango, line 2431: in docstring below, all `` should be `

deepinv/optim/optimizers.py
```diff
@@ -2360,6 +2369,224 @@ def __init__(
...
+    def __init__(
+        self,
+        data_fidelity: DataFidelity | list[DataFidelity] = None,
+        prior: Prior | list[Prior] = None,
+        lambda_reg: float = 1.0,
+        g_param: float = None,
+        sigma_denoiser: float = None,
+        num_subsets: int = 2,
+        subset_strategy: str = "default",
+        eps: float = 1e-15,
```
Andrewwango: so this is different to MLEM default?

deepinv/optim/optimizers.py
```diff
@@ -2360,6 +2369,224 @@ def __init__(
...
+        Default: ``2``.
+    :param deepinv.optim.DataFidelity, list[DataFidelity] data_fidelity: data fidelity term.
+        If ``None``, defaults to :class:`deepinv.optim.PoissonLikelihood`.
+    :param deepinv.optim.Prior, list[Prior] prior: prior term. If ``None``, no prior is used.
+        Default: ``None``.
+    :param float lambda_reg: regularization parameter :math:`\lambda`. Default: ``1.0``.
+    :param float g_param: parameter for the prior. Default: ``None``.
+    :param float sigma_denoiser: same as ``g_param``. If both ``g_param`` and ``sigma_denoiser`` are provided, ``g_param`` is used. Default: ``None``.
+    :param str subset_strategy: ordered-subsets strategy. Currently only ``"default"``
+        is supported. Default: ``"default"``.
```
Andrewwango: what's the purpose of this arg?

deepinv/optim/optimizers.py
```diff
@@ -2360,6 +2369,224 @@ def __init__(
...
+            data_fidelity = PoissonLikelihood()
+
+        if g_param is None and sigma_denoiser is not None:
+            g_param = sigma_denoiser
+
+        if not isinstance(num_subsets, int) or num_subsets < 2:
+            raise ValueError(
+                "OSEM requires at least two subsets. "
+                "Use deepinv.optim.MLEM instead for a single subset."
+            )
```
Andrewwango:
```suggestion
            raise ValueError("OSEM requires at least two subsets. Use deepinv.optim.MLEM instead for a single subset.")
            )
```
Side question: why can't we allow num_subsets=1?

deepinv/optim/optimizers.py
```diff
@@ -2360,6 +2369,224 @@ def __init__(
...
+        params_algo: dict[str, float] = None,
+        **kwargs,
+    ):
+        if data_fidelity is None:
+            data_fidelity = PoissonLikelihood()
+
+        if g_param is None and sigma_denoiser is not None:
+            g_param = sigma_denoiser
+
+        if not isinstance(num_subsets, int) or num_subsets < 2:
```
Andrewwango:
```suggestion
        if num_subsets < 2:
```
We generally take the convention in deepinv to not validate typed arguments, trusting that the user does it correctly.

deepinv/optim/optimizers.py
```diff
@@ -2360,6 +2369,224 @@ def __init__(
...
+        from deepinv.utils.tensorlist import TensorList
+
+        if isinstance(physics, StackedLinearPhysics):
+            if not isinstance(y, (TensorList, list)):
+                raise TypeError(
+                    "A pre-split deepinv.physics.StackedLinearPhysics requires "
+                    "pre-split measurements as a deepinv.utils.TensorList or "
+                    "list[torch.Tensor]. Use "
+                    "deepinv.physics.functional.subsets.split_measurements to "
+                    "split full measurements."
```
Andrewwango: Please merge this string, sorry for the minor stylistic comments, similar below, it makes it hard to read!

deepinv/optim/optimizers.py
```diff
@@ -2360,6 +2369,224 @@ def __init__(
...
+            raise TypeError(
+                "OSEM requires a full deepinv.physics.Tomography, "
+                "deepinv.physics.TomographyWithAstra, or deepinv.physics.PET, "
+                "or a pre-split deepinv.physics.StackedLinearPhysics. Use "
+                "deepinv.physics.functional.subsets.split_physics and "
+                "deepinv.physics.functional.subsets.split_measurements to "
+                "prepare supported full physics and measurements."
+            )
+
+        kwargs["subset_physics"] = subset_physics
```
Andrewwango: why do we need to do kwargs like this? It's quite obfuscating

deepinv/optim/optim_iterators/osem.py
```diff
@@ -0,0 +1,133 @@
...
+        :param torch.Tensor, deepinv.utils.TensorList y: Full input data or pre-split measurements.
+        :param deepinv.physics.Physics physics: Full physics or pre-split stacked physics modeling the data-fidelity term.
+        :param deepinv.utils.TensorList y_subsets: Measurement subsets.
+        :param deepinv.physics.StackedLinearPhysics subset_physics: Physics operators corresponding to the measurement subsets.
+        :return: Dictionary ``{"est": (x, None), "cost": F, "it": k + 1}`` containing the updated iterate and estimated cost.
+        """
+        x = X["est"][0]
+        k = 0 if "it" not in X else X["it"]
+
+        y_subsets = kwargs["y_subsets"]
```
Andrewwango: The kwargs convention isn't pretty, I would much rather explicitly passing these as kwargs.
Also, I don't understand the logic here. Why do we need separate arguments for y_subsets and physics_subset? We don't need the full y or physics here? I suggest passing y_subsets and physics_subset in directly in OSEM optimizer. Then you also don't need that `y_subsets = y` line.

Now in this iterator we can assume that y and physics are presplit, since no-one is using the iterator without the optimizer algorithm. You can remove the type validation below, and instead specify in the docstring. More generally I think we trust the users quite a lot in deepinv, so as long as we put it clearly in the docstring, we don't necessarily need to worry about users passing in wrong things. Happy to be challenged on this, but if we want more user validation, then that's a much wider issue that would require much more work outside the scope of this PR

deepinv/optim/optim_iterators/osem.py
```diff
@@ -0,0 +1,133 @@
...
+        if isinstance(y, list) or isinstance(y_subsets, list):
+            raise TypeError(
+                "OSEMIteration requires pre-split measurements as a "
+                "deepinv.utils.TensorList."
+            )
+        subset_physics = kwargs["subset_physics"]
+        num_subsets = len(subset_physics)
+        if num_subsets < 2:
```
Andrewwango: Again here, this has already been validated in OSEM

deepinv/optim/optim_iterators/osem.py
```diff
@@ -0,0 +1,133 @@
...
+                "The number of measurement subsets and physics subsets must match."
+            )
+
+        prior_scale = 1.0 / num_subsets
+        for cur_y, cur_physics in zip(y_subsets, subset_physics, strict=True):
+            sensitivity = cur_physics.A_adjoint(ones_like(cur_y))
+            # For deepinv.physics.PET, we need to add the background term
+            if hasattr(cur_physics, "background"):
+                proj = cur_physics.A(x, add_background=True)
+            # Other deepinv.physics.Physics do not have a background term
```
Andrewwango: remove

deepinv/optim/optim_iterators/osem.py
```diff
@@ -0,0 +1,133 @@
...
+            denom = sensitivity
+
+            # Scale the OSL prior so that one full epoch applies its total weight.
+            if cur_prior is not None:
+                prior_grad = cur_prior.grad(x, cur_params["g_param"])
+                denom = denom + prior_scale * cur_params["lambda"] * prior_grad
+
+            x = numerator / denom.clamp(min=self.eps)
+
+        # Since we support both pre-split and full physics / measurements,
```
Andrewwango: `Since we support both pre-split and full physics / measurements,`

do we? I thought OSEM optimizer split full physics/y in the case of user passing in full physics/y.

Andrewwango: If this is the case, we can simplify the logic below to no longer need y, physics

deepinv/physics/pet.py
```diff
@@ -82,6 +82,9 @@ class PET(LinearPhysics):
     :param bool normalize_counts: If `False` the :math:`\gamma` normalization term in front of the Poisson noise is removed,
         so that the measurements :math:`y` are true integer counts.
     :param str | torch.device device: device to run the computations on, e.g. `"cpu"` or `"cuda"`
+    :param None, torch.Tensor views: one-dimensional tensor of integer indices
+        selecting the PET sinogram views to project. If ``None``, all views are
+        projected.
```
Andrewwango: merge one line

deepinv/physics/pet.py
```diff
@@ -151,6 +155,9 @@ def __init__(
             voxel_size = voxel_size + (voxel_size[-1],)
 
         self.img_size = img_size
+        self.voxel_size = voxel_size
+        self.fwhm_data_mm = fwhm_data_mm
+        self.radial_trim = radial_trim
```
Andrewwango, line 158: why?

deepinv/physics/tomography.py
```diff
@@ -616,6 +616,22 @@ def measurement_shape(self) -> tuple[int, ...]:
     def num_angles(self) -> int:
         return self.xray_transform.range_shape[1]
 
+    @property
+    def angles(self) -> torch.Tensor | None:
+        """Acquisition angles in degrees.
+
+        Returns ``None`` for vector-based geometries, for which the acquisition
+        trajectory is fully described by `self.projection_geometry["Vectors"]`.
```
```suggestion
        """Astra projection geometry angles tensor in degrees. If Astra vector geom, return None.
```

deepinv/physics/functional/subsets.py
```diff
@@ -0,0 +1,355 @@
+from __future__ import annotations
+
+import torch
+
+from deepinv.physics.forward import LinearPhysics, StackedLinearPhysics
+from deepinv.physics.pet import PET
+from deepinv.physics.tomography import Tomography, TomographyWithAstra
+from deepinv.utils.tensorlist import TensorList
+
+SUPPORTED_TOMOGRAPHY_PHYSICS = (Tomography, TomographyWithAstra, PET)
```
Andrewwango: Sorry I'm not generally a fan of high level constants like this

deepinv/physics/functional/subsets.py
```diff
@@ -0,0 +1,355 @@
...
+    # second we build the StackedLinearPhysics with one operator per subset
+    if isinstance(physics, Tomography):
+        subset_kwargs = _get_tomography_subset_kwargs(physics)
+        subset_physics = _get_tomography_subset_physics(physics, indices, subset_kwargs)
+    elif isinstance(physics, TomographyWithAstra):
+        subset_kwargs = _get_astra_subset_kwargs(physics)
+        subset_physics = _get_astra_subset_physics(physics, indices, subset_kwargs)
+    else:
+        subset_kwargs = _get_pet_subset_kwargs(physics)
+        subset_physics = _get_pet_subset_physics(physics, indices, subset_kwargs)
```
Andrewwango: refactor to put the code directly in this function; having too many private functions makes it hard to read, especially when they're just single use funcs. the comment also is a bit superfluous

deepinv/physics/functional/subsets.py
```diff
@@ -0,0 +1,355 @@
...
+
+    # Preserve the normalization of the full operator. Measurements are split by
+    # slicing the output of the full normalized operator, so each subset must use
+    # the same normalization factor for its forward model to match those slices.
```
Andrewwango: unnecessary

deepinv/physics/functional/subsets.py
```diff
@@ -0,0 +1,355 @@
...
+    if physics.normalize:
+        subset_operator_norm = physics.operator_norm.detach().clone()
+        for subset in subset_physics:
+            subset.normalize = True
+            subset.register_buffer(
+                "operator_norm", subset_operator_norm.detach().clone()
+            )
```
Andrewwango:
```suggestion
        for subset in subset_physics:
            subset.normalize = True
            subset.register_buffer(
                "operator_norm", physics.operator_norm.detach().clone()
            )
```

deepinv/physics/functional/subsets.py
```diff
@@ -0,0 +1,355 @@
...
+
+    if dim < 0:
+        dim = y.dim() + dim
+    indices = get_subset_indices(
+        num_angles, num_subsets, strategy=strategy, device=y.device
+    )
+    return TensorList([y.index_select(dim, idx) for idx in indices])
+
+
+def split_physics(
```
Andrewwango: since this is quite tomography specific I would rename this function and file to be e.g. tomography_subsets or equivalent or move it with some of the other tomogrpahy related functionals

deepinv/physics/functional/subsets.py
```diff
@@ -0,0 +1,355 @@
...
+def split_measurements(
+    y: torch.Tensor,
+    physics: LinearPhysics,
+    num_subsets: int,
+    strategy: str = "default",
+) -> TensorList:
+    r"""
+    Splits tomography measurements into angular subsets.
+
+    .. warning::
```
Andrewwango: I wouldn't say this is a warning

deepinv/tests/test_optim.py
```diff
@@ -907,36 +907,67 @@ def prior_g(x, *args, **kwargs):
...
-# contrary to e.g mirror descent which can be tested on L2
-def test_MLEM(imsize, dummy_dataset, device):
-    dataloader = DataLoader(dummy_dataset, batch_size=1, shuffle=False, num_workers=0)
-    test_sample = next(iter(dataloader)).to(device)
-
-    physics = dinv.physics.Blur(
-        dinv.physics.functional.gaussian_blur(sigma=(2, 0.1), angle=45.0),
+# MLEM and OSEM are tested separately because they only converge
+# with respect to the Poisson data fidelity.
+# Other data fidelities can be passed but the convergence is not guaranteed.
```
Andrewwango: preserver your previous comment, it's more informative

deepinv/tests/test_optim.py
```diff
@@ -907,36 +907,67 @@ def prior_g(x, *args, **kwargs):
...
-
-    data_fidelity = dinv.optim.PoissonLikelihood()
+    x_true = torch.rand(
+        (1, *imsize), generator=torch.Generator(device).manual_seed(0), device=device
+    ).clamp(min=0.1)
+    y = physics(x_true).clamp(min=1e-6)
+    x_init = physics.A_adjoint(y).clamp(min=1e-6)
+    algorithm_kwargs = {}
+
+    if algorithm is dinv.optim.OSEM:
```
Andrewwango, line 1010: isinstance

examples/physics/demo_pet2d.py
```diff
@@ -212,27 +218,201 @@
...
+    start = time.perf_counter()
+    x_mlem, metrics_mlem = model_mlem(
+        y, physics, init=torch.ones_like(x), compute_metrics=True
+    )
+    _sync()
+    mlem_time = time.perf_counter() - start
+
+print(f"MLEM runtime: {mlem_time:.2f} s for {mlem_iter} iterations")
+
+psnr = dinv.metric.PSNR(max_pixel=x.max().item())
```
Andrewwango:
```suggestion
psnr = dinv.metric.PSNR(max_pixel=None)
```

examples/physics/demo_pet2d.py
```diff
@@ -212,27 +218,201 @@
...
+# OSEM applies an EM update successively with each pair :math:`(y_l,A_l)`. Processing
+# all subsets once is called an **epoch** here. Each subset update is cheaper than a full
+# MLEM iteration and updates the complete image, so OSEM generally reaches a useful
+# reconstruction in fewer passes over the complete acquisition.
+#
+# The subsets must remain matched: :math:`y_l` must contain exactly the views modeled
+# by :math:`A_l`. DeepInverse provides
+# :func:`deepinv.physics.split_measurements` and
+# :func:`deepinv.physics.split_physics` to construct these pairs.
+
```
Andrewwango: unnecessary I think

examples/physics/demo_pet2d.py
```diff
@@ -212,27 +218,201 @@
...
+# Explicitly splitting the PET acquisition
+# ----------------------------------------
+#
+# We first split the measurements and physics explicitly with the use of
+# :func:`deepinv.physics.split_measurements` and
+# :func:`deepinv.physics.split_physics`.
+# You can also implement your own splitting strategy, and pass the resulting
+# :class:`deepinv.physics.StackedLinearPhysics` and list of measurements to :class:`deepinv.optim.OSEM`.
+
+osem_epochs = 5
```
Andrewwango: can we keep consistent notatino and call this same as what we did for MLEM?

examples/physics/demo_pet2d.py
```diff
@@ -212,27 +218,201 @@
...
+num_subsets = 8
+
+_sync()
```
Andrewwango: Do we need to measure split time?

examples/physics/demo_pet2d.py
```diff
@@ -212,27 +218,201 @@
...
+axes[1].plot([100 * v for v in metrics_mlem["nrmse"][0]], label="MLEM")
+axes[1].plot([100 * v for v in metrics_osem_split["nrmse"][0]], label="OSEM")
+axes[1].set_xlabel("Iteration / epoch")
+axes[1].set_ylabel("NRMSE (%)")
+axes[1].yaxis.set_major_formatter("{x:.0f}%")
+axes[1].legend()
+fig.tight_layout()
+
+# %%
+# Letting OSEM split full inputs automatically
```
Andrewwango: I think demoing this separate functionality is a bit unnecessary, most users just care about the most convenient way of doing this

Andrewwango (changes requested):
Hi @Tmodrzyk  thanks for the extensive PR and I love to see standard PET / SPECT recon algos!

Many comments, mostly stylistic, some about the design. I think most of the comments arise from the fact that my preferred style for deepinv code is rather laconic

deepinv/optim/optim_iterators/mlem.py
Tmodrzyk: Sensitivity is not constant in `PET`, it is patient dependent, mostly influenced by the attenuation map (i.e CT scan) passed as input.

deepinv/optim/optim_iterators/mlem.py
Tmodrzyk: Initialisation is indeed already handled by `custom_init` but I'll add it to the documentation for both classes 🙂

Tmodrzyk: Thanks for the comment, the default initialization was actually incorrect

deepinv/optim/optimizers.py
Tmodrzyk: One subset is just `MLEM`, but the code would actually run. I've removed these safeguards.

deepinv/optim/optimizers.py
Tmodrzyk: Oops good catch

deepinv/optim/optim_iterators/osem.py
Tmodrzyk: Indeed it was a mixup between handling the full physics at the `OSEM` level and not the iterator.
I simplified this so that we only use subsetted physics here.

deepinv/optim/optimizers.py
Tmodrzyk: We can actually use very fancy strategies to select subsets, but for now I implemented only the most basic baseline one. 
If you think this is anticipating too much I can remove this both here and in the utilities. 😄

deepinv/physics/pet.py
Tmodrzyk: Required to build the subset physics in `tomography_subsets.py` + this is generally useful info that you'd like to recover

examples/physics/demo_pet2d.py
Tmodrzyk: Agreed, kept only the one which handles the splitting automatically

deepinv/physics/functional/tomography_subsets.py
```diff
@@ -0,0 +1,281 @@
...
+    if isinstance(physics, Tomography):
+        subset_kwargs = {
+            "img_width": physics.img_width,
+            "circle": physics.radon.circle,
+            "parallel_computation": physics.radon.parallel_computation,
+            "adjoint_via_backprop": physics.adjoint_via_backprop,
+            "fbp_interpolate_boundary": physics.fbp_interpolate_boundary,
+            "fan_beam": physics.fan_beam,
+            "fan_parameters": physics.radon.fan_parameters,
+            "normalize": False,
```
Andrewwango: why is this hardcoded False?

deepinv/physics/functional/tomography_subsets.py
```diff
@@ -0,0 +1,281 @@
...
+                / object_geometry["GridRowCount"],
+            )
+            bounding_box = (
+                object_options["WindowMinX"],
+                object_options["WindowMaxX"],
+                object_options["WindowMinY"],
+                object_options["WindowMaxY"],
+            )
+        else:
+            detector_spacing = (
```
Andrewwango: this is quite specific code to TomographyWithAstra.
- Are there any buffers we can store e.g. pixel_spacing in the astra class such that we don't have to try recompute them here?
- Would it be better to put this code inside TomographyWithAstra e.g. a method called `clone` that creates a new physics with the same geometry, but allowing a new set of angles?
- Another possibly cleaner idea is to create deepcopys of the physics, and just mutate (subsample) each copy's geometry. What do you think about that?

deepinv/physics/functional/tomography_subsets.py
```diff
@@ -0,0 +1,281 @@
...
+
+    if dim < 0:
+        dim = y.dim() + dim
+    indices = get_subset_indices(
+        num_angles, num_subsets, strategy=strategy, device=y.device
+    )
+    return TensorList([y.index_select(dim, idx) for idx in indices])
+
+
+def split_physics(
```
Andrewwango, line 85: I think this would need to be tested e.g. the sum of the adjoints of the split physics equals the adjoint of the full physics (or am I mistaken?)

Andrewwango (changes requested):
Looking good! A couple of comments on the split_physics functions.

deepinv/physics/functional/tomography_subsets.py
Tmodrzyk: This is currently `test_physics.py::test_tomography_subset_utilities` along with `split_measurements`.
Happy to change the name if this is confusing.

deepinv/physics/functional/tomography_subsets.py
Tmodrzyk: The logic regarding the norm was factored at the end of the function but I'm moving it to each branch since it is confusing

deepinv/physics/functional/tomography_subsets.py
Tmodrzyk: I think deepcopy might inadvertly duplicate stuff related to astra that we might not want to duplicate.
I'm adding these as new buffers in astra because all three are generally useful information that we might want to get anyway. 
I like the idea of `.clone()` so I'm adding it too !

deepinv/optim/optimizers.py
```diff
@@ -2265,28 +2273,35 @@ class MLEM(BaseOptim):
...
 
     .. math::
-        x_{k+1} = \frac{x_k}{A^T \mathbf{1} + \lambda \partial g(x_k)} \odot A^T \left(\frac{y}{A x_k}\right)
+        x_{k+1} = \frac{x_k}{A^T \mathbf{1} + \lambda \partial g(x_k)} \odot A^T \left(\frac{y}{A x_k + b}\right)
 
     where :math:`\partial g(x_k)` is a subgradient of :math:`g` at point :math:`x_k`.
 
+    .. note::
+
+        By default, the alogrithm is initialized with a tensor of ones with the same
```
romainvo: typo 
```suggestion
        By default, the algorithm is initialized with a tensor of ones with the same
```

deepinv/optim/optimizers.py
```diff
@@ -2359,12 +2384,253 @@ def __init__(
...
+
+    .. math::
+        x_{k,l+1} = \frac{x_{k,l}}{A_l^T \mathbf{1}} \odot A_l^T \left(\frac{y_l}{A_l x_{k,l} + b_l}\right),
+
+    where :math:`A_l` and :math:`y_l` are the corresponding physics and measurement
+    subset, and :math:`b_l` is an optional additive background and `l` is the subset index.
+
+    .. note::
+
+        Only :class:`deepinv.physics.Tomograph`, :class:`deepinv.physics.TomographyWithAstra`
```
romainvo: typo
```suggestion
        Only :class:`deepinv.physics.Tomography`, :class:`deepinv.physics.TomographyWithAstra`
```

deepinv/optim/optimizers.py
```diff
@@ -2359,12 +2384,253 @@ def __init__(
...
+    See :class:`deepinv.optim.optim_iterators.OSEMIteration` for the details of one
+    iteration.
+
+    A regularization can be included by specifying a `prior`.
+    This uses One-Step-Late (OS-MAP-OSL) :footcite:p:`greenUseEmAlgorithm1990`,
+    similar to in :class:`deepinv.optim.MLEM`.
+
+    .. note::
+
+        By default, the alogrithm is initialized with a tensor of ones with the same
```
romainvo: typo
```suggestion
        By default, the algorithm is initialized with a tensor of ones with the same
```

deepinv/optim/optimizers.py
```diff
@@ -2359,12 +2384,253 @@ def __init__(
...
+
+        # Need to update the PoissonLikelihood data-fidelity with the background
+        # for PET physics
+        if hasattr(physics[0], "background"):
+            for stacked_data_fidelity in self.data_fidelity:
+                if not isinstance(
+                    stacked_data_fidelity.data_fidelity_list[0], PoissonLikelihood
+                ):
+                    continue
+                for i, (subset_data_fidelity, subset_physics) in enumerate(
```
romainvo: Naming is confusing here, above `subset_physics` is a `StackedLinearPhysics` but here it is an entry of a `StackedLinearPhysics`, i.e. a `LinearPhysics`. Instead could do `sub_physics` or `physics_item` or something else?

deepinv/optim/optimizers.py
```diff
@@ -2359,12 +2384,253 @@ def __init__(
...
+                        gain=subset_data_fidelity.gain,
+                        bkg=background / subset_data_fidelity.gain,
+                        denormalize=subset_data_fidelity.d.denormalize,
+                    )
+
+        with torch.no_grad():
+            sensitivities = [
+                cur_physics.A_adjoint(torch.ones_like(cur_y))
+                for cur_y, cur_physics in zip(y, physics, strict=True)
+            ]
```
romainvo, line 2613: Use the same naming than above, i.e. same naming for an entry of `physics`. Same for `y`, ideally use the same suffix for readability.

deepinv/physics/functional/tomography_subsets.py
```diff
@@ -0,0 +1,159 @@
...
+) -> list[torch.Tensor]:
+    r"""Return indices that interleave ASTRA vectors into equal subsets.
+
+    :param torch.Tensor geometry_vectors: ASTRA geometry vectors of shape
+        ``(num_views, 12)``.
+    :param int num_subsets: number of subsets.
+    :return: list of index tensors.
+    """
+    indices = torch.arange(len(geometry_vectors), device=geometry_vectors.device)
+    return list(indices.reshape(-1, num_subsets).T)
```
romainvo: They are the same util, can we factorize into a `get_subset_tensor` or something like this?

deepinv/physics/functional/tomography_subsets.py
```diff
@@ -0,0 +1,159 @@
...
+    :return: :class:`deepinv.physics.StackedLinearPhysics` over angular subsets.
+    """
+    if not isinstance(physics, (Tomography, TomographyWithAstra, PET)):
+        raise TypeError(
+            "split_physics is currently supported for deepinv.physics.Tomography, deepinv.physics.TomographyWithAstra, and deepinv.physics.PET physics."
+        )
+
+    if physics.normalize:
+        warnings.warn(
+            "Subsetted physics cannot be normalized. Divide the output of the subsetted physics by the operator norm of the complete physics.",
```
romainvo: Minor remark but should we keep "Split physics" instead of "Subsetted physics"? @Andrewwango, native ear wanted :)

deepinv/tests/test_external_libraries.py
```diff
@@ -201,6 +201,12 @@ def test_tomography_with_astra_logic(
                 else (64, 48, 32)
             )
         assert physics.num_angles == 32 if cubic else 64
+        expected_angles = torch.linspace(
+            *(0, 180) if geometry_type == "parallel" else (0, 360),
+            steps=num_angles + 1,
+            device=device,
+        )[:-1]
+        assert torch.allclose(physics.angles, expected_angles)
```
romainvo, line 228: Nice :)!

romainvo (changes requested):
Thanks @Tmodrzyk for the changes! I like the updated mlem/osem logic :)

My main point is about the `clone()` logic which feels outside the scope of this PR, remarks in my comment. Otherwise some typos.

deepinv/physics/functional/tomography_subsets.py
romainvo: Hi guys! sorry my early comment was not saved. To be honest I'm not very fan of the `clone()` method:
1. The first thing that comes to mind is we already have a `Physics.clone()` method, which exhaustively and automatically traverses all attributes of the object and duplicates them (while also taking into account random generator), so we would be kind of re-implementing stuff but not as cleanly and with a different purpose also.
2. In the end, we end up with `split_physics` still having heterogeneous logic for each type of physics supported, without that much code being factored among different cases, so we may as well keep the original logic

What I think would work well is @Andrewwango's first proposition, exposing some properties to the astra physics so that the messy astra util code does not end up in the `split_physics`, to make things tight we could have for example
```
@property 
def bounding_box(self):
    if self.is_2d:
            bounding_box = (
                object_options["WindowMinX"],
                object_options["WindowMaxX"],
                object_options["WindowMinY"],
                object_options["WindowMaxY"],
            )
    else:
            bounding_box = (
                object_options["WindowMinX"],
                object_options["WindowMaxX"],
                object_options["WindowMinY"],
                object_options["WindowMaxY"],
                object_options["WindowMinZ"],
                object_options["WindowMaxZ"],
            )
```
I would use a property rather than storing self.bounding_box = bounding_box directly, so that `self.bounding_box` always reflect the state of `self.xray_transform` which holds all the physics information necessary for the astra backend.

Last thing is that the `split_physics` is quite new, but has a lot of potential for very useful features, e.g. `distribute` module. We may need to adapt the `split_physics` logic in the future so that the produced `StackedPhysics` is compatible with distributed computation (maybe it already is but didn't do the thorough check). We can avoid future breaking changes by not committing to a specific structure now, and see what's best after.

I think cloning a physics while replacing the value of some attribute is a very useful feature, but feels outside the scope of this PR, happy to discuss it further of course :) !

deepinv/physics/functional/tomography_subsets.py
Andrewwango: Both seem fine to me haha but I have quite a loose grasp on english

deepinv/physics/functional/tomography_subsets.py
Andrewwango: sounds good

deepinv/optim/optimizers.py
```diff
@@ -2585,12 +2585,22 @@ def forward(
...
+                subsetted_physics = split_physics(physics, self.num_subsets)
             y_subsets = split_measurements(y, physics, self.num_subsets)
 
-            physics = subset_physics
+            physics = subsetted_physics
             y = y_subsets
 
+            if operator_norm is not None:
+                # Keep subset projections in the same normalized domain as y.
+                for subset_physic in physics:
```
Andrewwango: i like the creative use of english of a singular physic

deepinv/optim/optimizers.py
```diff
@@ -2579,23 +2573,33 @@ def forward(
...
                 )
-            subset_physics = split_physics(
-                physics, self.num_subsets, strategy=self.subset_strategy
-            )
-            y_subsets = split_measurements(
-                y, physics, self.num_subsets, strategy=self.subset_strategy
-            )
-            physics = subset_physics
+            operator_norm = physics.operator_norm if physics.normalize else None
+            with warnings.catch_warnings():
```
Andrewwango: why need this? If this warning is not useful, then why don't we remove it at source?

deepinv/physics/tomography.py
```diff
@@ -548,6 +548,13 @@ def __init__(
             else n_detector_pixels
         )
         self.geometry_type = geometry_type
+        self.detector_spacing = detector_spacing
+        self.pixel_spacing = pixel_spacing
+        self.bounding_box = bounding_box
```
Andrewwango: as romain said I think some of these need to be propertys so they are recomputed with new geoms

Andrewwango (changes requested):
nice @Tmodrzyk ! Almost there, just a few small things

deepinv/optim/optimizers.py
Tmodrzyk: My reasoning was that the warning is important if you use the `split_physics` on your own but not really if you let OSEM do the splitting for you. 
Agree it adds these annoying catch lines to suppress it. 
Would you prefer I remove the runtime warning and just write this info in the doc ?

deepinv/physics/tomography.py
Tmodrzyk: These are constructor kwargs and the projector is constructed only once at initialization, so I don't think these need to become properties. 
Am I missing something ?

Tmodrzyk:
Hi @Andrewwango I've addressed your final comments, are there any other changes needed ? 
@romainvo any things requiring changes that I might have missed ?

deepinv/physics/tomography.py
romainvo: You're right, the current workflow does not allow updating the geometry "easily"; nevertheless, a user can still update it manually by re-instantiating an astra projector with a different geometry in `self.xray_transform`, and if they do so the attributes won't be updated. It's a very specific edge case, so I don't mind dealing with this later, as you want

deepinv/physics/functional/tomography_subsets.py
```diff
@@ -0,0 +1,193 @@
...
+                img_width=physics.img_width,
+                circle=physics.radon.circle,
+                parallel_computation=physics.radon.parallel_computation,
+                adjoint_via_backprop=physics.adjoint_via_backprop,
+                fbp_interpolate_boundary=physics.fbp_interpolate_boundary,
+                normalize=False,
+                fan_beam=physics.fan_beam,
+                fan_parameters=physics.radon.fan_parameters,
+                device=physics.device,
+                dtype=physics.dtype,
```
romainvo: Sorry, I did not catch this earlier. `physics.device` is deprecated, so I don't think it's wise to do this. I think the proper way is to have a `device` argument in the `split_physics` method, and depending on where it's called, you either grab the current device from the tensor being processed or you let the user decide. 

`physics.dtype` is not deprecated yet, but it's the same logic, in pytorch `device` and `dtype` should be properties of tensors not modules

deepinv/physics/functional/tomography_subsets.py
```diff
@@ -0,0 +1,193 @@
...
+                    angles=angles.index_select(0, idx),
+                    angular_range=physics.angular_range,
+                    n_detector_pixels=physics.n_detector_pixels,
+                    detector_spacing=physics.detector_spacing,
+                    pixel_spacing=physics.pixel_spacing,
+                    bounding_box=physics.bounding_box,
+                    geometry_type=physics.geometry_type,
+                    geometry_parameters=physics.geometry_parameters,
+                    normalize=False,
+                    device=physics.device,
```
romainvo: Same comment about `physics.device`

deepinv/physics/functional/tomography_subsets.py
```diff
@@ -0,0 +1,193 @@
...
+            PET(
+                img_size=physics.img_size,
+                voxel_size=physics.voxel_size,
+                fwhm_data_mm=physics.fwhm_data_mm,
+                scanner=physics.scanner,
+                radial_trim=physics.radial_trim,
+                gain=physics.noise_model.gain.detach().clone(),
+                normalize=False,
+                normalize_counts=bool(physics.noise_model.normalize),
+                device=physics.background.device,
```
romainvo: This also works but not every physics have buffers like this, so if possible I would stick to a uniform approach

docs/source/changelog.rst
```diff
@@ -16,12 +16,16 @@ New Features
 - Add :func:`deepinv.utils.plot_napari` to interactively view 2D images/3D vols with napari (:gh:`1249` by `Andrew Wang`_)
 - Add support for :func:`Liu-Jia padding <deepinv.physics.functional.liu_jia_pad>` (:gh:`934` by `Jérémy Scanvic`_)
 - Add support for TV-L1 priors :class:`deepinv.optim.TVL1Prior` (:gh:`1236` by `Sarra Amiri`_)
+- Add :class:`deepinv.optim.OSEM` algorithm for tomographic reconstruction and update PET demos to showcase OSEM (:gh:`1255` by `Thibaut Modrzyk`_)
+- Add utilities for subsetted tomography physics :func:`deepinv.physics.split_physics` and :func:`deepinv.physics.split_measurements` (:gh:`1255` by `Thibaut Modrzyk`_)
+- Add :class:`deepinv.loss.metric.NRMSE` metric (:gh:`1256` by `Thibaut Modrzyk`_)
```
romainvo: I think it's the wrong issue number?

deepinv/optim/optimizers.py
```diff
@@ -2359,12 +2384,257 @@ def __init__(
...
+            y = y_subsets
+
+            if operator_norm is not None:
+                # Keep subset projections in the same normalized domain as y.
+                for subset_physic in physics:
+                    subset_physic.register_buffer(
+                        "operator_norm", operator_norm.detach().clone()
+                    )
+                    subset_physic.normalize = True
+                    if hasattr(subset_physic, "background"):
```
romainvo: Isn't the background already normalized in `split_physics` (line 182-185) ?

deepinv/optim/optimizers.py
romainvo: I would maybe add a comment above `operator_norm = physics.operator_norm if physics.normalize else None`  for maintainability. 

I agree that not catching the warning may be confusing when it's called within OSEM, but it's still very useful when called independently.

romainvo (changes requested):
Hi @Tmodrzyk , thanks for the changes! Last comments about `physics.device` and scaling of background in `OSEM.forward` (already done in `split_physics` if I'm not mistaken).

deepinv/optim/optimizers.py
Andrewwango: Yes please, but it's not a big issue. I'm always in favour of clearer docs, we can assume the user reads them

Andrewwango (approved):
just my one remaining comment from previous review then lgtm

deepinv/optim/optimizers.py
Tmodrzyk: Done, added a warning to the doc and removed the runtime warning.

deepinv/physics/functional/tomography_subsets.py
Tmodrzyk: Done, everything becomes a property extracted from `self.xray_transform`

deepinv/optim/optimizers.py
Tmodrzyk: It was but the logic was confusing, data-fidelity needed a different rescaling because it didn't apply the background and normalization in the same order.
I reverted to normalizing within `split_physics` using the full physics operator norm, see my global comment below.

Tmodrzyk:
Regarding the normalization of the subset physics, I reverted using no normalization to using the full physics operator norm.
This breaks the contract of `physics.normalize`, i.e. dividing by the actual `operator_norm` estimated through a power method, but it was also the case when using no normalization when the full physics was normalized.
This simplified the code in other places, e.g in `OSEM`. I kept a warning in the doc to warn the user that the subset physics operator are not normalized using the actual `operator_norm`.

romainvo (approved):
Thanks @Tmodrzyk! I understand the struggle with the normalization. I think we partly inherit it from the design of StackedLinearPhysics, which does not deal with this. Otherwise looks good to me, thanks again for the very cool feature!

PR #1257, opened 2026-07-03
Move from uv to pixi for import time benchmark

romainvo:
One remaining test (`import_time.yml`) was still using `uv` instead of `pixi`. I'm implementing the switch to have a uniform setup.

### LLM policy

### DIFF
--- a/.github/workflows/import_time.yml
+++ b/.github/workflows/import_time.yml
@@ -91,6 +91,3 @@ jobs:
             } else {
               core.info('✅ No significant performance regression detected.');
             }
-      
-      - name: Minimize uv cache
-        run: uv cache prune --ci
--- a/.github/workflows/import_time.yml
+++ b/.github/workflows/import_time.yml
@@ -13,74 +13,55 @@ jobs:
     name: benchmark
     runs-on: ubuntu-latest
 
-    env:
-      # Configure a constant location for the uv cache
-      UV_CACHE_DIR: /tmp/.uv-cache
-
     strategy:
       matrix:
         include:
           # Ubuntu tests
-          - extra: dataset,denoisers,test,training
+          - environment: "full-cpu"
             name: "all optional dependencies"
-          - extra: ""
+          - environment: "test-cpu"
             name: "no optional dependencies"
 
     steps:
       - uses: actions/checkout@v5
 
-      - name: Install uv
-        uses: astral-sh/setup-uv@v6
+      - name: Setup Pixi
+        uses: prefix-dev/setup-pixi@v0.9.3
+        with:
+          pixi-version: latest
+          environments: full-cpu test-cpu
 
       - name: "Set up Hyperfine"
         run: |
           sudo apt-get update && sudo apt-get install -y hyperfine
 
-      - name: Restore uv cache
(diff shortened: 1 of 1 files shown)


.github/workflows/import_time.yml
```diff
@@ -13,74 +13,55 @@ jobs:
...
     steps:
       - uses: actions/checkout@v5
 
-      - name: Install uv
-        uses: astral-sh/setup-uv@v6
+      - name: Setup Pixi
+        uses: prefix-dev/setup-pixi@v0.9.3
+        with:
+          pixi-version: latest
+          environments: full-cpu test-cpu
```
Tmodrzyk: We currently have two ways of installing the dependencies with pixi depending on the workflows:
- one like what you did here
- one manually, by setting `run_install: false` and then `pixi install -e ${{ matrix.environment }} --manifest-path pyproject.toml` (as in `test_cpu.yml`)

Is there a reason for this ? Which one is correct here ? 
From what I'm getting the pixi setup action will already try to run the install, and we do it later in the workflow so it seems redundant.

.github/workflows/import_time.yml
```diff
@@ -13,74 +13,55 @@ jobs:
...
-          if [ "${{ matrix.extra }}" = "" ]; then
-            uv pip install .
-          else
-            EXTRAS=""
-            for extra in $(echo "${{ matrix.extra }}" | tr ',' ' '); do
-              EXTRAS="$EXTRAS --extra $extra"
-            done
-            uv sync $EXTRAS
-          fi
+          pixi install -e ${{ matrix.environment }}
```
Tmodrzyk: Should we use something more similar to the `test_cpu.yml` (L75) workflow here ? 
Like `pixi install -e ${{ matrix.environment }} --manifest-path pyproject.toml --color always`

Tmodrzyk (commented):
Hi @romainvo thank you for updating the CI !
I'm not 100% confident on these kind of codes but it seems that we are doing things differently regarding the way we install dependencies with pixi, depending on the workflow files, see my specific comments.
I fear in this new import workflow we might install the dependencies twice, but I might be mistaken.

.github/workflows/import_time.yml
romainvo: We can remove this part

.github/workflows/import_time.yml
romainvo: Thanks for the careful review! In `test_cpu.yml` we do the manual install to deal with the potentially failing `--locked` install. We don't pull cache in this action so we can keep the default behaviour of the pixi action. 

And you're right, there is no need for a second install below (I wrongly kept the uv sections).

Tmodrzyk:
```suggestion
      - name: Setup Pixi
        uses: prefix-dev/setup-pixi@v0.9.3
        with:
          pixi-version: latest
          environments: ${{ matrix.environment }}
```

Tmodrzyk (changes requested):
Just a minor change to avoid installing both environments for each workflow run, otherwise LGTM

Tmodrzyk (approved):
LGTM !

PR #1261, opened 2026-07-04
datetime convention

matthieutrs:
Tiny change to checkpoint timestamp formatting to avoid `:` in directory names.

Although `:` is valid in filenames on most OS, it can fail on some mounted or cloud-backed filesystems (e.g. some gpu servers). 

Suggested output would be:

```bash
"%y-%m-%d-%H{sep}%M{sep}%S"      # 2026-07-04_20:29:10     <- current
"%Y%m%d_%H%M%S"                   # 20260704_202910          <- proposed
```

### LLM policy

### DIFF
--- a/deepinv/utils/logger.py
+++ b/deepinv/utils/logger.py
@@ -92,10 +92,9 @@ class ProgressMeter:
 def get_timestamp() -> str:
     """Get current timestamp string.
 
-    :return str: timestamp, with separators determined by system.
+    :return str: timestamp.
     """
-    sep = "_" if platform.system() == "Windows" else ":"
-    return datetime.now().strftime(f"%y-%m-%d-%H{sep}%M{sep}%S")
+    return datetime.now().strftime(f"%Y%m%d_%H%M%S")
 
 
 @_deprecated_class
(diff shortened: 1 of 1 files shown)


deepinv/utils/logger.py
```diff
@@ -92,10 +91,9 @@ def _get_epoch_fmtstr(self, num_epochs):
 def get_timestamp() -> str:
     """Get current timestamp string.
 
-    :return str: timestamp, with separators determined by system.
+    :return str: timestamp.
     """
-    sep = "_" if platform.system() == "Windows" else ":"
-    return datetime.now().strftime(f"%y-%m-%d-%H{sep}%M{sep}%S")
```
Andrewwango, line 98: what unix systems fail with colon?

the minimum change I think would be sufficient would be to move to `sep = "_"` for every system.

Andrewwango (changes requested):
I'm confused where you're having this problem, I thought this was only a problem on Windows.

deepinv/utils/logger.py
matthieutrs: It's not an OS issue in my case but it's a path saved on a cloud-backed storage (on a GPU server).

Regarding styling, I can move to your suggestion if you prefer, or `2026-07-04_20-29-10`?

PR #1262, opened 2026-07-04
theta -> angles renaming tomography

matthieutrs:
Renaming the `theta` parameter from Tomography operator to `angles` for consistency

### LLM policy


romainvo (changes requested):
Hi @matthieutrs, thanks for the fix! 

Looks good to me. For backward compatibility, could you add a `self.theta` property pointing to angles and indicating that the old attribute is deprecated?

deepinv/physics/tomography.py
```diff
@@ -116,6 +116,9 @@ class Tomography(LinearPhysics):
 
     """
 
+    @_deprecated_alias(
+        theta="angles",
+    )
```
romainvo: I don't think it's the proper use-case for `_deprecated_alias` as we did not change the `__init__`'s signature, i.e. `theta` was not a method argument but an instance attribute.

I just noticed that we already have a util for our case in `utils/decorators.py`, i.e. `_deprecate_attribute`, which already implements the custom property with the warning. So here in the `__init__` we would do something like this

```python
_deprecate_attribute(
    self,
    attr_name="theta",
    attr_underscore_name="_theta",
    attr_initial_value=self.angles,
    deprecation_message="Some message"
) 
```

romainvo (changes requested):
Hi @matthieutrs, sorry for the delay and thanks for the update! I don't think it's the proper use case for `_deprecated_alias`, I think `_deprecate_attribute` would be more appropriate here. I've put details in my comment.

deepinv/physics/tomography.py
matthieutrs: Thanks @romainvo! I've updated

romainvo (approved):
Thanks @matthieutrs, looks good to me! Approving this, as the failing test is completely unrelated to the changes.

Can you update the changelog before merging please?

PR #1267, opened 2026-07-10
Introduce pre-commit

PAUL-BERNARD:
[pre-commit](https://pre-commit.com/) is a tool that runs a selected list of commands for every git commit on a project. Here, we run `ruff-check` and `black`, which takes ~1s on the whole project (but should be faster on individual commits, as only modified files are checked). For it to work, you have to install it by running `pip install pre-commit` and `pre-commit install`.

If the checks fail, the commit is blocked. You can always bypass pre-commit by using `git commit -n`.

- [-] Added an entry to the [changelog.rst](https://github.com/deepinv/deepinv/blob/main/docs/source/changelog.rst).

### LLM policy

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,31 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+    - repo: https://github.com/pre-commit/pre-commit-hooks
+      rev: v3.2.0
+      hooks:
+          - id: check-yaml
+          - id: check-added-large-files
+            args:
+            - --maxkb=2048
+          - id: check-byte-order-marker
+          - id: check-case-conflict
+          - id: check-merge-conflict
+          - id: mixed-line-ending
+    - repo: https://github.com/psf/black
+      rev: 26.5.1
+      hooks:
+          - id: black
+          - id: black-jupyter
+    - repo: https://github.com/astral-sh/ruff-pre-commit
+      # Ruff version.
+      rev: v0.15.20
+      hooks:
+          - id: ruff-check
+          # - id: ruff-format  # 56 files reformatted
+    # - repo: https://github.com/adamchainz/blacken-docs
+    #   rev: "v1.12.1"
+    #   hooks:
+    #     - id: blacken-docs
+    #       additional_dependencies:
+    #       - black==26.3.1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,10 +22,11 @@ repos:
       rev: v0.15.20
       hooks:
           - id: ruff-check
-          # - id: ruff-format  # 56 files reformatted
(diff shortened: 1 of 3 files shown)


jscanvic (changes requested):
Thanks for taking care of this Paul! Many users might find it useful I believe

I don't know if I'm the only one but I purposely don't use pre-commit hooks for automated linting so I would suggest to:
1. Make sure the change has no effect if `pre-commit` is not installed
2. Leave `pre-commit` out of the pyproject file
3. Add an info block at the end of § 4. Code quality to basically say that if the contributor wants they can optionally automate running the automated linting pipeline by simply installing `pre-commit` thanks to the base config we provide for it. (Instead of adding a new bullet point in the main/mandatory steps)

docs/source/contributing.rst
```diff
@@ -137,9 +137,8 @@ Writing good documentation is also crucial for helping other users use your code
 Code quality is important to us. We require that your code is compliant with PEP8, the `black <https://black.readthedocs.io>`_ style and `ruff <https://docs.astral.sh/ruff/>`_ checkers:
 
 1. Add `typing <https://fastapi.tiangolo.com/python-types/>`_ to your code and docstrings. Typing rules such as PEP585 are automatically checked using ruff.
-2. Run ``black .`` in the root directory of your repository. This will automatically fix all formatting issues.
-3. Run ``ruff check``, which will check all linting options we've enabled. If it fails, follow the suggestions to make a fix!
-4. Push your code. The automatic checkers will run in GitHub actions, along with other actions that we have in place.
```
romainvo, line 142: Revert this block to its original state, and add a paragraph about the optional pre-commit instal

pyproject.toml
```diff
@@ -76,7 +76,8 @@ doc = [
 
 lint = [
     "black",
-    "ruff"
+    "ruff",
+    "pre-commit",
```
romainvo: Remove dependency from pyproject

.pre-commit-config.yaml
```diff
@@ -0,0 +1,32 @@
...
+      hooks:
+          - id: ruff-check
+          # - id: ruff-format  # 56 modified files
+    # - repo: https://github.com/adamchainz/blacken-docs  # 17 modified files
+    #   rev: "1.20.0"
+    #   hooks:
+    #     - id: blacken-docs
+    #       args: [-l 140]
+    #       additional_dependencies:
+    #       - black==26.5.1
```
romainvo: Can we remove these commented lines?

romainvo (changes requested):
Thanks @PAUL-BERNARD! Agree with @jscanvic that we don't need to make the pre-commit install mandatory. Nevertheless, providing the config is still very valuable :)

PAUL-BERNARD:
Thanks for the review!
Yes pre-commit is totally optional, if you don't run `pre-commit install` nothing will happen.

jscanvic (approved):
Thanks Paul! LGTM

romainvo (approved):

PR #1271, opened 2026-07-14
add explicit TV subgradient

### LINKED ISSUE #1259: Explicit subgradient for TV prior
### Environment (optional)

_No response_

### Describe the feature

Currently, `TVPrior` does not override the `.grad` method of the base class `Potential`.

When calling `TVPrior.grad(x)`, the library therefore uses autodiff through the explicit TV functional. This can be unstable because the isotropic TV norm is nonsmooth when local image gradients vanish.

For

```math
g(x) = \|Dx\|_{1,2},
```

we could instead use the explicit TV subgradient:
```math
-\mathrm{div}\left(\frac{Dx}{|D x|}\right),
```
and setting it explicitly to zero in flat regions (where Dx is very small).

This would avoid relying on autodiff at nondifferentiable points and make the behavior of `TVPrior.grad` explicit and stable.

### Motivation

While the TV prior is non-smooth, its implementation still offers a gradient computation so we might as well make it more stable for optimization algorithms.

### Alternatives considered

_No response_

### Additional context

_No response_

Tmodrzyk:
This PR implements an explicit subgradient choice for `TVPrior` instead of using autodiff and closes #1259.
This aims at making the use of `.grad` more stable for this prior, so that it can be used in optimization algorithms such as `MLEM`. 

I checked the `demo_poisson_mlem.py` notebook and the results are similar.
Not sure if we need more testing for this as people mostly use TV through the prox.

### LLM policy

### DIFF
--- a/deepinv/optim/prior.py
+++ b/deepinv/optim/prior.py
@@ -552,13 +552,16 @@ class TVPrior(Prior):
 
         At locations where the finite difference vanishes, the zero element
         of the subdifferential is selected.
+
+        :param torch.Tensor x: Variable :math:`x` at which the subgradient is computed.
+        :return: (:class:`torch.Tensor`) subgradient at :math:`x`.
         """
         dx = self.nabla(x)
         norm_dx = torch.linalg.vector_norm(dx, dim=-1, keepdim=True)
 
         nonzero = norm_dx > 0
 
-        # Do not evaluate an unsafe division by zero, even in the masked branch.
+        # Do not evaluate an unsafe division by zero
         safe_norm = torch.where(nonzero, norm_dx, torch.ones_like(norm_dx))
         normalized_dx = torch.where(
             nonzero,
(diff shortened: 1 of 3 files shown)


romainvo (commented):
Hi @Tmodrzyk! Looks good to me, maybe we can do a test when we check against the autograd computation at points where the finite difference is non-zero? That way, we're double-sure nothing breaks in the stable cases, or is it too much?

Tmodrzyk:
Hi @romainvo I added a small test to check that the new subgradient has the expected behavior i.e:
- finite value everywhere 
- in regions with non-zero finite differences, same value as the autodiff
- in regions with zero finite differences, zero value

romainvo (approved):
Thanks @Tmodrzyk, looks good to me! Approving, as failing tests are completely unrelated. Can you update the changelog before merging please ?

PR #1278, opened 2026-07-15
Fix kwargs applications in parent constructor calls

### LINKED ISSUE #1273: Some DiffusionSDE classes forward *kwargs instead of **kwargs to .super()
### Environment

Python 3.12, Deepinverse 0.4.1

### Describe the bug

In https://github.com/deepinv/deepinv/blob/main/deepinv/sampling/diffusion_sde.py three classes (`EDMDiffusionSDE`, `SongDiffusionSDE`, and `VariancePreservingDiffusion`) are incorrectly forwarding keyword arguments to their parent class through the call to `.super()`. 
These classes use `*kwargs` instead of `**kwargs`, which forwards only the keys.

### Steps to reproduce

Check the classes in https://github.com/deepinv/deepinv/blob/main/deepinv/sampling/diffusion_sde.py

### Expected behavior

Should pass `**kwargs`

_No response_

### Additional context

_No response_

jscanvic:
Fixes #1273

I replace `*kwargs` with `**kwargs` in 3 parent constructor calls

I also checked that there are no other single-star kwargs application using: `git grep -E "[^\\*]\\*kwargs"`

### DIFF
--- a/deepinv/sampling/diffusion_sde.py
+++ b/deepinv/sampling/diffusion_sde.py
@@ -424,7 +424,7 @@ class EDMDiffusionSDE(DiffusionSDE):
             dtype=dtype,
             device=device,
             *args,
-            *kwargs,
+            **kwargs,
         )
 
     def score(self, x: Tensor, t: Tensor | float, *args, **kwargs) -> torch.Tensor:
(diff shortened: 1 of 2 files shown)


Tmodrzyk (approved):
Thanks for handling this PR @jscanvic, LGTM !

PR #1279, opened 2026-07-15
New noisy data-fidelities : PIGDM and Moment Matching 

### LINKED ISSUE #1274: DPSDataFidelity implementation / documentation mismatch
### Environment

Python 3.12 Deepinverse 0.4.1

### Describe the bug

The `DPSDataFidelity` class documents a division by a factor $\sqrt{m}$ but implements a multiplication by this factor instead. See [the doc](https://deepinv.org/api/stubs/deepinv.sampling.DPSDataFidelity.html) and [the implementation](https://github.com/deepinv/deepinv/blob/main/deepinv/sampling/noisy_datafidelity.py).
Moreover, the actual DPS paper cited as reference does not mention nor implements this factor, so a note should explicit why it is there, or preferably the factor should be removed.

### Steps to reproduce

Check the documentation and implementation

### Expected behavior

Documentation matches implementation + a comment explaining the mystery weight.

_No response_

### Additional context

_No response_

samuro95:
Implementation of the PiGDM and Moment Matching noisy data-fidelity terms, for posterior sampling with diffusion models.

### LLM policy


deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,307 @@ def forward(
...
+    parameter = torch.as_tensor(parameter, device=reference.device, dtype=dtype)
+    if parameter.numel() == 1:
+        return parameter.squeeze()
+    if (
+        parameter.shape[0] == reference.shape[0]
+        and all(size == 1 for size in parameter.shape[1:])
+        and parameter.ndim < reference.ndim
+    ):
+        return parameter.view(parameter.shape[0], *([1] * (reference.ndim - 1)))
+    return parameter
```
Tmodrzyk: Could we use Denoiser._handle_sigma() instead ? To avoid duplicating similar logic

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,307 @@ def forward(
...
+    dtype = reference.real.dtype if reference.is_complex() else reference.dtype
+    parameter = torch.as_tensor(parameter, device=reference.device, dtype=dtype)
+    if parameter.numel() == 1:
+        return parameter.squeeze()
+    if (
+        parameter.shape[0] == reference.shape[0]
+        and all(size == 1 for size in parameter.shape[1:])
+        and parameter.ndim < reference.ndim
+    ):
+        return parameter.view(parameter.shape[0], *([1] * (reference.ndim - 1)))
```
Tmodrzyk: This is a bit obscure, is there a way to rewrite this with simpler code ?

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,307 @@ def forward(
...
+    .. math::
+
+        p(x_0|x_t)
+        \approx \mathcal N\!\left(
+            x_0;D(x_t,\sigma_t),\Sigma_t(x_t)
+        \right),
+        \qquad
+        \Sigma_t(x_t)=r_t^2\mathrm{Id},
+        \qquad
+        r_t^2=\frac{\sigma_t^2}{1+\sigma_t^2}.
```
Tmodrzyk: could we inline this ?

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,307 @@ def forward(
...
+        clip: tuple = None,
+        cg_max_iter: int = 3,
+        cg_tol: float = 1e-4,
+        verbose: bool = False,
+    ):
+        super().__init__(weight=weight)
+        self.denoiser = denoiser
+        if clip is not None:
+            if len(clip) != 2:  # pragma: no cover
+                raise ValueError(f"clip must be None or length 2, but got {clip}")
```
Tmodrzyk: We can assume that the user reads the doc and remove this test.

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,307 @@ def forward(
...
+            measurement = physics.A(denoised)
+            difference = measurement - y.to(
+                device=measurement.device, dtype=measurement.dtype
+            )
+
+            def operator(v):
+                adjoint = physics.A_adjoint(v).to(denoised.dtype)
+                covariance_product = denoiser_vjp(adjoint)[0]
+                return v + physics.A(covariance_product)
+
```
Tmodrzyk: Here don't we need to scale `physics.A(covariance_product)` by `sigma**2` ?

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,307 @@ def forward(
...
+            )
+            if self.clip is not None:
+                denoised = torch.clip(denoised, self.clip[0], self.clip[1])
+            measurement = physics.A(denoised)
+            difference = measurement - y.to(
+                device=measurement.device, dtype=measurement.dtype
+            )
+            sigma_t = torch.as_tensor(
+                sigma, device=difference.device, dtype=difference.dtype
+            )
```
Tmodrzyk: To avoid polluting the core logic, we can put this input conversion at the top of the function

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,307 @@ def forward(
...
+    :param float cg_tol: Relative conjugate-gradient tolerance. Default: ``1e-4``.
+    :param bool verbose: If ``True``, print conjugate-gradient convergence
+        information. Default: ``False``.
+    """
+
+    def __init__(
+        self,
+        denoiser: Denoiser = None,
+        weight: float = 1.0,
+        clip: tuple = None,
```
Tmodrzyk: If the parameter can take `None` value the type should be e.g `Denoiser | None`

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,307 @@ def forward(
...
+        :param torch.Tensor, float sigma: Diffusion noise standard deviation.
+        :return: PiGDM gradient, with the same shape and dtype as ``x``.
+        """
+        if not isinstance(physics, dinv.physics.LinearPhysics):
+            raise ValueError("PiGDMDataFidelity only supports linear physics.")
+        if self.denoiser is None:
+            raise ValueError("PiGDMDataFidelity requires a denoiser.")
+
+        input_dtype = x.dtype
+        with torch.enable_grad():
```
Tmodrzyk: Is this really required ?

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,307 @@ def forward(
...
+        :return: Moment-matching gradient, with the same shape and dtype as
+            ``x``.
+        """
+        if not isinstance(physics, dinv.physics.LinearPhysics):
+            raise ValueError("MomentMatchingDataFidelity only supports linear physics.")
+        if self.denoiser is None:
+            raise ValueError("MomentMatchingDataFidelity requires a denoiser.")
+
+        input_dtype = x.dtype
+        with torch.enable_grad():
```
Tmodrzyk: Same here, do we really need to compute the graph ?

Tmodrzyk (changes requested):
Thanks for the contribution @samuro95, the demo looks very good ! 
A few major comments: 
- we should double-check that the computation for MM is correct, it seems a scaling by the noise variance is missing, but I might be mistaken
- both MM and PiGDM use `torch.enable_grad()` which consumes a lot of VRAM. While I understand it's required for the DPS data-fidelity, I don't see how it is used for these data-fidelities. I was not able to run the demo with the current implementation on my laptop using 8Go VRAM, changing to `torch.no_grad()` the VRAM consumption was ok.
- In the moment matching paper they explicitely make use of \Sigma_y the noise variance of the measurement, which is not the case here. Should we always whiten the input before applying the algorithm ? 

Otherwise I have minor style comments.

deepinv/sampling/noisy_datafidelity.py
samuro95: You are right that it does mostly the same but it's a bit wierd to use _handle_sigma() when it's not a sigma.

deepinv/sampling/noisy_datafidelity.py
samuro95: Indeed, I thought that rescaling by sigma_y was equivelent to changing the weight parameter but indeed it is not, thanks !

samuro95:
Thanks for the review :)

I handled your comments. 
- The sigma_y is now there. Indeed that was a big mistake, well spotted :)
- torch.enable_grad() is indeed not necessary. I put a @torch.no_grad() in order to avoid unnecessary grad calculations in CG iterations.
- I updated the denoiser._handle_sigma() function to take as input a reference tensor and then use it at different parts.

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,297 @@ def forward(
         if self.clip is not None:
             x0_t = torch.clip(x0_t, self.clip[0], self.clip[1])  # optional
 
-        out = (self.d(physics.A(x0_t), y) * y.numel() / y.size(0)).sqrt() * self.weight
+        difference = physics.A(x0_t) - y
+        out = self.weight * torch.linalg.vector_norm(
```
mh-nguyen712: so remove `self.d` it is not used anymore

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,297 @@ def forward(
...
+    For a linear forward operator and Gaussian measurement noise with standard
+    deviation :math:`\sigma_y`,
+    integrating this approximation gives a Gaussian approximation of
+    :math:`p_t(y|x_t)`. Its negative log-likelihood gradient is
+
+    .. math::
+
+       -\nabla_{x_t} \log p_t(y|x_t) \approx \lambda J_D(x_t, \sigma_t)^\top A^\top
+        \left(r_t^2 A A^\top + \sigma_y^2\mathrm{Id}\right)^{-1}
+        \left(A D(x_t, \sigma_t) - y\right).
```
mh-nguyen712, line 556: There is a sign error here I think?
```suggestion
       \nabla_{x_t} \log p_t(y|x_t) \approx \lambda J_D(x_t, \sigma_t)^\top A^\top
        \left(r_t^2 A A^\top + \sigma_y^2\mathrm{Id}\right)^{-1}
        \left(A D(x_t, \sigma_t) - y\right).
```

mh-nguyen712: image

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,297 @@ def forward(
...
+        return conjugate_gradient(
+            operator,
+            u,
+            max_iter=self.cg_max_iter,
+            tol=self.cg_tol,
+            verbose=self.verbose,
+        )
+
+    @torch.no_grad()
+    def grad(
```
mh-nguyen712, line 638: I think we can add `get_model_output` here, then reuse it to compute the unconditional (similar to DPS) to avoid calling the denoiser twice

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,297 @@ def forward(
...
+        :return: PiGDM gradient, with the same shape and dtype as ``x``.
+        """
+        if not isinstance(physics, dinv.physics.LinearPhysics):
+            raise ValueError("PiGDMDataFidelity only supports linear physics.")
+        if self.denoiser is None:
+            raise ValueError("PiGDMDataFidelity requires a denoiser.")
+        input_dtype = x.dtype
+        x_denoiser = x.detach().to(torch.float32)
+        sigma_denoiser = self.denoiser._handle_sigma(sigma, reference_tensor=x_denoiser)
+        if not isinstance(physics.noise_model, dinv.physics.GaussianNoise):
```
mh-nguyen712, line 672: Maybe can fallback to a small value of sigma here instead of raising error?

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,297 @@ def forward(
...
+        p(x_0|x_t) \approx \mathcal{N} \left(
+        x_0;D(x_t,\sigma_t),\Sigma_t(x_t) \right),
+        \qquad
+        \Sigma_t(x_t)=\sigma_t^2J_D(x_t,\sigma_t)^\top.
+
+    The resulting negative log-likelihood gradient is
+
+    .. math::
+
+        -\nabla_{x_t} \log p_t(y|x_t) \approx \lambda J_D(x_t, \sigma_t)^\top A^\top
```
mh-nguyen712, line 715: a sign error?

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,297 @@ def forward(
...
+        \left(\sigma_t^2 A J_D(x_t, \sigma_t)^\top A^\top
+        + \sigma_y^2\mathrm{Id}\right)^{-1}
+        \left(A D(x_t, \sigma_t) - y\right).
```
mh-nguyen712, line 718:
```suggestion
        \nabla_{x_t} \log p_t(y|x_t) \approx \lambda J_D(x_t, \sigma_t)^\top A^\top
        \left(\sigma_t^2 A J_D(x_t, \sigma_t) A^\top
        + \sigma_y^2\mathrm{Id}\right)^{-1}
        \left(A D(x_t, \sigma_t) - y\right).
```

deepinv/sampling/noisy_datafidelity.py
```diff
@@ -214,9 +227,297 @@ def forward(
...
+            if len(clip) != 2:  # pragma: no cover
+                raise ValueError(f"clip must be None or length 2, but got {clip}")
+            clip = sorted(clip)
+        self.clip = clip
+        self.cg_max_iter = cg_max_iter
+        self.cg_tol = cg_tol
+        self.verbose = verbose
+
+    @torch.no_grad()
+    def grad(
```
mh-nguyen712, line 763: Same here, I think we can return the denoised output to compute the unconditional score in PosteriorDiffusion

examples/sampling/demo_noisy_data_fidelity.py
```diff
@@ -0,0 +1,368 @@
...
+Noisy data-fidelity terms for diffusion posterior sampling
+==========================================================
+
+This example compares three approximations of the measurement-matching term
+used by diffusion posterior samplers: Diffusion Posterior Sampling (DPS) :footcite:t:`chung2022diffusion`,
+Pseudoinverse-Guided Diffusion Models (PiGDM) :footcite:t:`song2023pseudoinverse`, and Moment Matching :footcite:t:`rozet2024learning`. These
+methods belong to the broader family of explicit posterior approximations reviewed in
+*A Survey on Diffusion Models for Inverse Problems*
+:footcite:t:`daras2024survey`, which also presents other ways of incorporating
+measurements into diffusion models.
```
mh-nguyen712: Maybe list them as in itemize is easier to read
```suggestion
used by diffusion posterior samplers: 

- Diffusion Posterior Sampling (DPS) :footcite:t:`chung2022diffusion`,
- Pseudoinverse-Guided Diffusion Models (PiGDM) :footcite:t:`song2023pseudoinverse`
- Moment Matching :footcite:t:`rozet2024learning` 

These methods belong to the broader family of explicit posterior approximations reviewed in
*A Survey on Diffusion Models for Inverse Problems*
:footcite:t:`daras2024survey`, which also presents other ways of incorporating
measurements into diffusion models.
```

examples/sampling/demo_noisy_data_fidelity.py
```diff
@@ -0,0 +1,368 @@
...
+    norm = torch.linalg.vector_norm(gradient).item()
+    print(f"{name:>15s} gradient norm: {norm:.3e}")
+
+dinv.utils.plot(
+    {
+        f"{name} gradient magnitude": gradient.abs()
+        for name, gradient in gradients.items()
+    },
+    figsize=(9, 3),
+)
```
mh-nguyen712, line 399:
```suggestion
dinv.utils.plot(
    {
        f"{name}": gradient.abs()
        for name, gradient in gradients.items()
    },
    figsize=(9, 3),
    suptitle="Gradient magnitude"
)
```

mh-nguyen712 (changes requested):
Thanks! I have some comments.

deepinv/sampling/noisy_datafidelity.py
samuro95: Thanks for noticing this but actually it is the paper this is wrong ! the sign should be + there

mh-nguyen712: Sorry my mistake!

deepinv/sampling/noisy_datafidelity.py
samuro95: well not really, it really requires Gaussian observation to have a valid Gaussian*Gaussian calculation

mh-nguyen712 (approved):
LGTM! Thanks @samuro95 !

Tmodrzyk:
I'll try to review this on monday @samuro95 !

Tmodrzyk (approved):

PR #1283, opened 2026-07-17
Preserve iterations on failed backtracking checks

### LINKED ISSUE #1272: Failed backtracking increases the global iteration counter in `FixedPoint`
### Environment

Python 3.12, DeepInverse 0.4.1

### Describe the bug

Backtracking is a method to estimate an hyper-parameter such as the descent step size (learning rate) of an algorithm based on a given criterion. Often this criterion uses the current iterate, current objective value or current first order information. To estimate the parameter one typically starts with a large value then iteratively checks the criterion and adjust the parameter value.

In the current implementation, `FixedPoint` starts a complete new outer loop when the criterion fails. 
It does return the previous iterate if the criterion has failed so at least the behavior of the complete optimization algorithm with backtracking seems (roughly) correct.
However in both cases, failure or success of the criterion, it increments the global iteration counter `it`.

This essentially burns global iterations without valid reasons, if the algorithm is supposed to stop with `max_iter` it will stop prematurely.
This also means the `it` used to select the current state of the algorithm is invalid and will advance faster than the real current iterate.
I did not check yet but this may cause some optimization algorithms relying on these internal states to output incorrect reconstructions.

### Steps to reproduce

Check the `FixedPoint` class.
Here is a minimum example to reproduce the bug:
```python
import torch
import deepinv as dinv


# Identity forward operator and observation y = 1.
physics = dinv.physics.LinearPhysics(
    A=lambda x: x,
    A_adjoint=lambda x: x,
)
y = torch.ones(1, 1, 1, 1)
x0 = torch.zeros_like(y)

model = dinv.optim.GD(
    data_fidelity=dinv.optim.L2(),
    stepsize=4.0,
    max_iter=2,
    backtracking=dinv.optim.BacktrackingConfig(
        gamma=0.1,
        eta=0.5,
        max_iter=20,
    ),
    verbose=True,
)

x = model(y, physics, init=x0)

print("Returned x:", x.item())
print("Final stepsize:", model.params_algo["stepsize"][0])
```
Output:
```
Backtracking : new stepsize = 2.000000
Backtracking : new stepsize = 1.000000
Returned x: 0.0
Final stepsize: 1.0
```

### Expected behavior

The snippet should update the input at least twice but instead does not modify it because the step size is updated instead.

_No response_

### Additional context

_No response_

ShreyanshGoyal:
Closes #1272.

Failed backtracking checks now retry the same fixed-point iteration instead of consuming the global iteration budget. Only accepted iterations update metrics, convergence state, and the progress bar.

The regression test reproduces the reported identity-operator example and verifies that two accepted iterations still run after the stepsize is reduced.

- [x] Updated docstrings related to the changes (not applicable).
- [x] Added an entry to the changelog.

### Test

- `python -m pytest deepinv/tests/test_optim.py::test_backtracking_does_not_consume_global_iterations -q`
- `ruff check deepinv/optim/fixed_point.py deepinv/tests/test_optim.py`
- `black --check deepinv/optim/fixed_point.py deepinv/tests/test_optim.py`

### LLM policy

- [x] An LLM tool wrote all of the code.
- [x] An agent submitted the PR and wrote the description.

### DIFF
--- a/deepinv/optim/fixed_point.py
+++ b/deepinv/optim/fixed_point.py
@@ -321,10 +321,12 @@ class FixedPoint(nn.Module):
         if self.anderson_acceleration_config is not None:
             self.init_anderson_acceleration(X)
 
-        for it in tqdm(
-            range(self.max_iter),
+        it = 0
+        progress_bar = tqdm(
+            total=self.max_iter,
             disable=(not self.verbose or not self.show_progress_bar),
-        ):
+        )
+        while it < self.max_iter:
             X_prev = X
             X = self.single_iteration(X, it, *args, **kwargs)
 
@@ -338,12 +340,15 @@ class FixedPoint(nn.Module):
                 )
 
                 # Convergence check
-                if (
+                converged = (
                     self.early_stop
                     and self.check_conv_fn is not None
                     and it > 1
                     and self.check_conv_fn(it, X_prev, X)
-                ):
+                )
+                it += 1
+                progress_bar.update()
+                if converged:
                     break
 
             else:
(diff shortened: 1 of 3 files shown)


jscanvic:
Closing for LLM policy

PR #1286, opened 2026-07-18
PET BrainWeb dataset

### LINKED ISSUE #1178: Add Brainweb phantom dataset
### Environment (optional)

_No response_

### Describe the feature

A popular choice of phantom dataset for PET is the Brainweb dataset [https://brainweb.bic.mni.mcgill.ca/brainweb/](https://brainweb.bic.mni.mcgill.ca/brainweb/).
The users should be able to easily load a test volume representing an emission map for PET, as well as options for simulating higher and lower intensity lesions. (see for instance [https://github.com/casperdcl/brainweb](https://github.com/casperdcl/brainweb)).
Ideally, a test volume should be hosted on Deepinv's huggingface for demos.

### Motivation

We are _very_ close to supporting PET acquisitions and reconstructions (see #1099.
While the PR already adds a few phantoms, they are still fairly simple.

### Alternatives considered

_No response_

### Additional context

_No response_

Tmodrzyk:
Closes #1178 by adding Brainweb phantoms specifically for PET, with emission and attenuation maps, along with the option to insert random or user-defined lesions.
This PR is mostly based on Casper da Costa-Luis' code to convert the original T1 data: https://github.com/casperdcl/brainweb

### LLM policy


deepinv/datasets/brainweb.py
```diff
@@ -0,0 +1,379 @@
...
+    :param float activity: Activity inside the lesion.
+    :param tuple[float, ...], None center_voxel: Optional centre in voxel coordinates.
+    """
+
+    diameter_mm: float
+    activity: float
+    center_voxel: tuple[float, ...] | None = None
+
+
+class BrainWebDataset(ImageDataset):
```
Andrewwango: Seems like a lot going on here. I've previously used brainweb for MRI (ok no additional lesions) with [brainweb-dl](https://github.com/paquiteau/brainweb-dl) @paquiteau which is a light wrapper - doesn't make sense for us to rewrite things.

I use this code (which I put here https://github.com/deepinv/deepinv/issues/804#issuecomment-3357511497)

1. `pip install brainweb-dl`
2. download with `brainweb-dl --all --contrast=T1 --brainweb-dir /path/to/data` 
3.
```python
from brainweb_dl import get_mri
dataset = dinv.datasets.ImageFolder(
    "/path/to/data",
    x_path="*.nii.gz",
    loader=lambda f: get_mri(int(f.stem[7:9]), contrast="T1", brainweb_dir="/path/to/data"),
    transform=lambda x: torch.from_numpy(x).unsqueeze(0) / 4095
)
```

Gives you instant access to the 3D vols.

Andrewwango: I know there will be some PET specific things, but the base dataset should be generalised, because brainweb inherently is.

Tmodrzyk: Hi @Andrewwango I agree that the T1 volumes can be loaded through the utility you mention, thanks for your comment !

However there's quite a big pre-processing pipeline for the volumes to be used in PET.
First the T1 is converted to CT by extracting bones and defining densities for these bones. 
Then the T1 is also converted to an emission map by extracting soft tissues, and affecting pre-defined activities to these tissues. These activities correspond to different choices of radiotracers, the main one being FDG.
Finally, we always put circular lesions to define ROIs used later to compute metrics, with predefined activity levels and diameters. I believe this mimics the tumors we're supposed to monitor IRL. For the metrics to be convenient to compute, you need the masks corresponding to those lesions.

I've added a link in the PR description to the existing code I'm porting here, from Da Costa-Luis and Kris Thielemans. :smile:
Edit: I'll also change the name of the class to BrainwebPET or something, so that there's no confusion with the straight-forward use of Brainweb for MRI.

paquiteau: Hello there :wave: 

I am not familiar with the PET pipeline, but glancing at the code, there at some overlap with what brainweb_dl already provide, notably you can download the "crisp" tissues segmentation either via the CLI or the `get_mri` function with `contrast="crisp"` (you also have the `"fuzzy"` mask which gives you higher-resolution quantitative masking for each tissue class).

Tmodrzyk: Hi there ! :smiley: 
Awesome, we could add `brainweb-dl` in the `datasets` extra dependencies, so that I can use it in this PET version.
The code I mentioned for PET (https://github.com/casperdcl/brainweb) does contain other features like visualization or adding random structures. 
I'm not sure if I should keep the current re-implementation or also add this as a dependency (I don't think the PET version is still maintained).
@Andrewwango what do you think ?

gschramm:
this seems super usefule. I think it would be even better to have more generic way of defining the contrast between GM/WM and skin instead of "just" having to presets (FDG and amyloid). 

https://github.com/Tmodrzyk/deepinv/blob/2328f67959dbfb40fae39bc4f215863a03da40a0/deepinv/datasets/brainweb.py#L100

Why not allowing users to specify GM/WM/skin values as distributions from which the values are sampled?

Tmodrzyk:
Hi @gschramm thanks, your comment means a lot ! 
I was mainly aiming at reproducing the code in the PR description, but surely we can integrate more features.
If you think something else would be useful we'd be more than happy to integrate your recommendations. 😄

gschramm:
I think having a few pre-set that (make sense) are very useful for new users. And for more advanced users, having the possibility to specify a distribution to sample from, provides enough flexibility to adapt the phantoms without the need to rewrite a lot of code.

One thing to be "careful" about are units / absolute values. In principle, what matters is the regional contrast (not absolute values). For the SNR of the simulated data, what matters is the combination of: absolute image value, fwd model sensitivity and gamma factor in the Poisson noise model.
I would make sure to document this.

casperdcl (changes requested):
Hi there - thanks for asking at https://github.com/casperdcl/brainweb/issues/36 - I think the `brainweb` licence ([MPL-2.0](https://www.mozilla.org/en-GB/MPL/2.0/)) is not compatible with vendoring in a [BSD-3-Clause](https://github.com/deepinv/deepinv/blob/main/LICENSE) project such as `DeepInv`.

Can't you simply add to deps? https://github.com/deepinv/deepinv/blob/cc4ae37c203afb573567c90aef17ee157e730bef/pyproject.toml#L25

Then you could do any extra transforms required here without worrying about licences. You're also welcome to open issues/PRs upstream. I just updated it in https://github.com/casperdcl/brainweb/pull/37 so should be easy to develop going forward.

deepinv/datasets/brainweb_pet.py
```diff
@@ -0,0 +1,235 @@
...
+import torch
+
+from deepinv.datasets.base import ImageDataset
+from deepinv.datasets.utils import resolve_root
+
+if TYPE_CHECKING:
+    import brainweb
+
+
+class BrainWebPET(ImageDataset):
```
Andrewwango, line 16: please can you address the docstring using my comments from the BrainWebMRI PR? I observe similar issues:

- the description should be concise and give information of descending importance. Most important is for non-expert users to know what the dataset is about and what it's for. We have generally quite an opinionated doc style in DeepInverse, which is what users like. Hence my quite strict comments on docstrings as we risk losing this opinionation
- Generally the preferred style for docs about configuring the dataset should be imperative, e.g. "Disable the ..." or "Pass..."
- The prose should be concise. "The absolute emission values alone do not determine simulation SNR; forward-model sensitivity and the Poisson noise gain also contribute." This sentence can be rewritten in much fewer words. Also not sure if this sentence contributes much in the first place, why is the forward model relevant here?

deepinv/datasets/brainweb_pet.py
```diff
@@ -0,0 +1,235 @@
...
+    ``t1_noise``, and ``t2_noise`` to zero. The noise amplitudes and smoothing
+    scales can be configured independently, as can the PET activity class and
+    output resolution. The absolute emission values alone do not determine
+    simulation SNR; forward-model sensitivity and the Poisson noise gain also
+    contribute.
+
+    Passing ``lesion_diameters`` adds lesions with ``brainweb.add_lesions`` and
+    includes a binary ``lesion_mask`` in the returned parameters.
+
+    :param str, pathlib.Path, None root: Dataset directory. Defaults to the
```
Andrewwango: This may be controversial but I think I would rather break the line limit than to have every arg broken into multiple lines. This is quite hard to read

deepinv/datasets/brainweb_pet.py
```diff
@@ -0,0 +1,235 @@
...
+        DeepInv cache.
+    :param int, collections.abc.Sequence[int] subject_ids: Subjects to expose.
+        Default is subject 4.
+    :param bool download: Download missing subjects. Default is ``True``.
+    :param collections.abc.Callable, None transform: Optional transform applied
+        jointly to the channel-first emission, attenuation, and requested T1/T2
+        tensors, as well as the lesion mask when present. The voxel size is
+        unchanged.
```
Andrewwango: is it important that this is unchanged? if not remove

deepinv/datasets/brainweb_pet.py
```diff
@@ -0,0 +1,235 @@
...
+    :param bool return_t1: If ``True``, include the T1 volume in the returned
+        parameters. Defaults to ``False``.
```
Andrewwango: use ` instead of ``

deepinv/datasets/brainweb_pet.py
```diff
@@ -0,0 +1,235 @@
...
+    :param str, pathlib.Path, None root: Dataset directory. Defaults to the
+        DeepInv cache.
+    :param int, collections.abc.Sequence[int] subject_ids: Subjects to expose.
+        Default is subject 4.
+    :param bool download: Download missing subjects. Default is ``True``.
+    :param collections.abc.Callable, None transform: Optional transform applied
+        jointly to the channel-first emission, attenuation, and requested T1/T2
+        tensors, as well as the lesion mask when present. The voxel size is
+        unchanged.
+    :param bool return_t1: If ``True``, include the T1 volume in the returned
```
Andrewwango: feels like lots of parameters here. Can't this be combined with T2 to be e.g. contrasts=["T1", "T2"]

deepinv/datasets/brainweb_pet.py
```diff
@@ -0,0 +1,235 @@
...
+        random intensity field. Defaults to ``0``.
+    :param float t1_sigma: Gaussian smoothing scale, in voxels, of the T1
+        random intensity field. Defaults to ``0``.
+    :param float t2_sigma: Gaussian smoothing scale, in voxels, of the T2
+        random intensity field. Defaults to ``0``.
+    :param str outres: BrainWeb output resolution. Valid values are ``"mMR"``
+        (default), ``"MR"``, and ``"brainweb"``.
+    :param type[brainweb.Act], None pet_class: BrainWeb PET activity class. Defaults to
+        ``brainweb.FDG``. Zero-argument callables used as activities are sampled
+        when the dataset is initialized.
```
Andrewwango: In DeepInverse we have nice concisely written docs, so I have to pick up on docstring problems that risk this, sorry for all the docstring comments! `Zero-argument callables used as activities are sampled when the dataset is initialized.` adds very little

docs/source/api/deepinv.datasets.rst
```diff
@@ -68,6 +68,7 @@ Image Datasets
     deepinv.datasets.FMD
     deepinv.datasets.Kohler
     deepinv.datasets.NBUDataset
+    deepinv.datasets.BrainWebPET
```
Andrewwango, line 71: user guide

deepinv/datasets/brainweb_pet.py
```diff
@@ -0,0 +1,235 @@
...
+
+        if self.lesion_kwargs is not None:
+            original = emission.copy()
+            if self.seed is not None:
+                self._brainweb.seed(self.seed + subject)
+            emission = self._brainweb.add_lesions(emission, **self.lesion_kwargs)
+            params["lesion_mask"] = torch.from_numpy(emission != original).unsqueeze(0)
+
+        emission = torch.from_numpy(emission).unsqueeze(0)
+        if self.transform is not None:
```
Andrewwango, line 196: feel like this can be simplified

deepinv/datasets/brainweb_pet.py
```diff
@@ -0,0 +1,235 @@
...
+        self.brainweb_kwargs: dict[str, object] = {
+            "petNoise": pet_noise,
+            "t1Noise": t1_noise,
+            "t2Noise": t2_noise,
+            "petSigma": pet_sigma,
+            "t1Sigma": t1_sigma,
+            "t2Sigma": t2_sigma,
+            "outres": outres,
+        }
+        self.brainweb_kwargs["PetClass"] = pet_class
```
Andrewwango: why is this separate

deepinv/datasets/brainweb_pet.py
```diff
@@ -0,0 +1,235 @@
...
+        lesion_blurs: list[float] | None = None,
+        lesion_threshold: float | None = None,
+        seed: int | None = 0,
+    ) -> None:
+        try:
+            import brainweb
+        except ImportError as error:  # pragma: no cover
+            raise ImportError(
+                "BrainWebPET requires brainweb. Install it with "
+                "`pip install deepinv[dataset]`."
```
Andrewwango: explicit install casper's lib

Andrewwango (changes requested):
Hi @Tmodrzyk , thanks for proposing this! Many comments here are similar to BrainwebMRI PR

deepinv/datasets/brainweb_pet.py
Tmodrzyk: Hi @Andrewwango thanks for the feedback, I think the issue mainly comes from me not spending enough time checking the docstrings produced by LLMs. I'll try to be more careful !

deepinv/datasets/brainweb_pet.py
```diff
@@ -2,78 +2,43 @@
...
-from typing import TYPE_CHECKING, Literal
 
-import numpy as np
 import torch
 
 from deepinv.datasets.base import ImageDataset
 from deepinv.datasets.utils import resolve_root
 
-if TYPE_CHECKING:
+try:
```
Andrewwango: I think we prefer optional dependency imports localised to their classes where possible?

deepinv/datasets/brainweb_pet.py
```diff
@@ -2,78 +2,43 @@
...
-        subject-specific seed is used for deterministic lesions by default.
+    Loads synthetic 3D volumes from BrainWeb dataset :footcite:p:`collinsDesignConstructionRealistic1998`,
+    of shape `(1, 127, 344, 344)`.
+    The dataset has been adapted to emission tomography, and returns an emission and attenuation map, at the Siemens Biograph mMR isotropic resolution of 2.0863 mm per voxel.
+
+    Passing `lesion_diameters` adds high activity lesions with `brainweb.add_lesions` and includes a binary `lesion_mask` in the returned parameters.
+
+    This dataset relies on the original implementation of Casper da Costa-Luis:
+    <https://github.com/casperdcl/brainweb>`_.
+    See the original implementation for a detailed description of the keyword arguments.
```
Andrewwango, line 27: add a pip install command here

deepinv/datasets/brainweb_pet.py
```diff
@@ -2,78 +2,43 @@
...
+
```
Andrewwango, line 28: see also brainweb MRI (and same in MRI PR too)

deepinv/datasets/brainweb_pet.py
```diff
@@ -164,22 +121,23 @@ def __init__(
...
             self.files.append(path)
 
+        random_degradations_kwargs = {
+            "petNoise": 0.0,
+            "t1Noise": 0.0,
+            "t2Noise": 0.0,
+            "petSigma": 0.0,
+            "t1Sigma": 0.0,
+            "t2Sigma": 0.0,
+            **(random_degradations_kwargs or {}),
```
Andrewwango: can random_degradataion_kwargs default be {}?
same with contrast default maybe to (,)?
Or is that bad python?

deepinv/datasets/brainweb_pet.py
Andrewwango: thanks @Tmodrzyk ! I missed your human touch.

Andrewwango (approved):
hi Thibaut, LGTM subject to final minor comments, thanks!

deepinv/datasets/brainweb_pet.py
Tmodrzyk: Agrred, it is there for the typing of brainweb.Act to be resolved, but I'll find a workaround

deepinv/datasets/brainweb_pet.py
Tmodrzyk: I think it's ok, and avoids the annoying `| None ` typing

deepinv/datasets/brainweb_pet.py
Tmodrzyk: I think it's cleaner if we merge this first so that we can have proper :class:`deepinv.datasets...` references here and in the MRI dataset too.

deepinv/datasets/brainweb_pet.py
Tmodrzyk: Actually the tuple one is ok but the mutable empty dict was flagged by ruff. It would be shared across all instances of the dataset.

PR #1307, opened 2026-07-31
Fix rendering issue in the docs

jscanvic:
There's a rendering bug in the docs https://deepinv.org/api/stubs/deepinv.physics.functional.liu_jia_pad.html

I think it's due to using a regular string instead of a raw string


jscanvic:
Okay it seems good now
image

Tmodrzyk (approved):
lgtm

PR #1309, opened 2026-08-03
Remove definitions that are silently overwritten

karpovantonme:
`GaussianNoise` defines `__add__`, `__mul__` and `__rmul__` **twice each** — at lines 260, 275, 311 and again at 327, 342, 378. Python keeps the last binding, so the first three are dead code: unreachable, untestable, and invisible in coverage.

```
$ ruff check --select F811 deepinv/physics/noise.py
noise.py:327:9: F811 Redefinition of unused `__add__` from line 260
noise.py:342:9: F811 Redefinition of unused `__mul__` from line 275
noise.py:378:9: F811 Redefinition of unused `__rmul__` from line 311
```

The two copies are identical, so **nothing changes at runtime**. I checked that rather than assuming it, by comparing the normalised AST of every method in the class before and after:

```
methods before: ['__add__', '__init__', '__mul__', '__rmul__', 'forward']
methods after:  ['__add__', '__init__', '__mul__', '__rmul__', 'forward']
  __add__:  body identical to the original — True
  __init__: body identical to the original — True
  __mul__:  body identical to the original — True
  __rmul__: body identical to the original — True
  forward:  body identical to the original — True
```

What it does remove is the trap: the next person to fix a bug in `__mul__` has a 50% chance of editing the copy that never runs, and the tests will keep passing while the behaviour stays broken.

I kept the **second** definition of each and deleted the first, so the order of methods in the class is unchanged.

Also in the same sweep, two duplicate imports: `import math` twice in `models/diffunet.py`, `import warnings` twice in `physics/forward.py`.

## What I did not run

The test suite needs torch, which I do not have installed here, so I did not run it. Given the AST comparison above the change is a pure deletion of unreachable definitions, but say the word if you would like me to set that up and report a run.

`ruff format` reports two files as needing reformatting both before and after this change, so that drift is pre-existing and I left it alone.


Andrewwango:
Hi @karpovantonme , thanks for the PR, as per the LLM policy https://deepinv.org/contributing.html#llm-policy, please fill in the LLM usage information as per the pull request template https://github.com/deepinv/deepinv/blob/main/.github/pull_request_template.md

karpovantonme:
Filled it in, thanks for pointing at the template rather than just closing it

Ticked "an LLM tool wrote all of the code" -- the duplicate definitions were found by a checker I run across repos, the patch and the description were drafted by Claude, and I read both before opening this. Did not tick the agent box because the PR itself was not agent-submitted, but tell me if you read that line differently and I will change it

Andrewwango (approved):

PR #1311, opened 2026-08-05
New examples: ultrasound despeckling, multispectral demosaicing

Andrewwango:
Two new real-world raw-data examples!!

- ultrasound despeckling with BM3D, using data from Speckle2Self paper https://www.sciencedirect.com/science/article/pii/S1361841525003020
- multispectral 16-band demosaicing with DIP, using data from https://www.nature.com/articles/s41597-024-04099-x

other small fixes
- DIP no longer patch module add
- fastMRI loading more robust at init time
- MRI tiny error message bug
- non_blocking_transfer forced to False in trainer on non-cuda devices (see [here](https://docs.pytorch.org/tutorials/intermediate/pinmem_nonblock.html#other-copy-directions-gpu-cpu-cpu-mps))

here's the ultrasound example final output
image

here's the demosaicing one (ok we need some more DIP iters)

t

### LLM policy


Tmodrzyk:
Hi @Andrewwango, looking at this ASAP :smile:

examples/optimization/demo_multispectral_demosaicing_dip.py
```diff
@@ -0,0 +1,128 @@
...
+model = dinv.models.DeepImagePrior(
+    dinv.models.ConvDecoder(img_size=mask.shape[1:]).to(device),
+    img_size=(256, 4, 4),
+    verbose=True,
+    re_init=True,
+    iterations=10 if dinv.utils.devices_equal(device, "cpu") else 1000,
+).to(device)
+
+with torch.no_grad():
+    x_net = model(y, physics)
```
Tmodrzyk: On my laptop with a 8Go VRAM GPU, this takes around 10 minutes. This will probably be too much for the CI. Is there another way to get decent results than to go to 1000 steps ?

Tmodrzyk (changes requested):
Hi @andrewwango thanks for the cool demos with real data !

For the ultrasound one:
I know the goal of the demo is to showcase the use of deepinv on real data, but it feels quite succinct at the moment.
Could we add a first step with another dataset containing ground-truth, just so that we can have metrics for the reconstruction, then keep your real-data inference ?

We could also compare other denoising methods here. I know supervised ones will probably be out-of-domain but there are probably other reconstructors that could work ? 

For the demosaicing one:
same, it feels a bit short but at least here we are comparing two methods. Unfortunately, DIP is taking a very long time to output meaningful reconstructions. With 100 steps it already takes a minute on my GPU, and the reconstruction is not looking good.
Could we swap it for another method ? Maybe PnP ? It looks like denoiser trained on natural images wouldn't be too out-of-domain here.

Tmodrzyk:
Hi @Andrewwango can you make the doc build pass before I give another look ?

deepinv/models/dip.py
Tmodrzyk, line 1: Is this still needed even if we don't use DIP in the example anymore ?

examples/optimization/demo_multispectral_demosaicing.py
Tmodrzyk, line 1: The TV reconstruction looks good !

examples/optimization/demo_multispectral_demosaicing.py
```diff
@@ -0,0 +1,151 @@
...
+Multispectral demosaicing from raw sensor data
+==============================================
+
+This example reconstructs a full-resolution multispectral image from raw snapshot
+mosaiced measurements using various reconstruction algorithms.
+
+Snapshot multispectral cameras cover the sensor with a mosaic of spectral filters (multispectral filter array),
+so that each pixel records only one band. Demosaicing recovers every band at the full sensor resolution.
+
+Demosaicing is a critical part of any ISP pipeline, and using better algorithms results in higher quality images for downstream tasks.
```
Tmodrzyk: Can we make this acronym explicit ?

examples/optimization/demo_multispectral_demosaicing.py
```diff
@@ -0,0 +1,151 @@
...
+
+We use raw oral-tissue data from the MODID dataset :footcite:p:`chand2024modid` acquired in-vivo for screening of oral diseases such as oral squamous cell carcinoma.
```
Tmodrzyk, line 13: Add the data link here also

examples/optimization/demo_multispectral_demosaicing.py
```diff
@@ -0,0 +1,151 @@
...
+# The file is a raw stream which we reshape to the pixel array of shape 1088x2048 and crop a 768x768 region of interest.
+
+H, W = 1088, 2048
+TILE = 4
+CROP = 768
+r0 = ((H - CROP) // 2 // TILE) * TILE
+c0 = ((W - CROP) // 2 // TILE) * TILE
+
+raw = torch.frombuffer(
+    bytearray(dinv.utils.load_url(dinv.utils.get_image_url("modid.raw")).getvalue()),
```
Tmodrzyk, line 38: can we use `load_example` instead ?

examples/optimization/demo_multispectral_demosaicing.py
```diff
@@ -0,0 +1,151 @@
...
+mask = torch.zeros_like(y)
+for i in range(4):
+    for j in range(4):
+        c = i * 4 + j
+        y[:, c, i::4, j::4] = raw[i::4, j::4]
+        mask[:, c, i::4, j::4] = 1
+
+y = y / y.max()  # normalise
+
+physics = dinv.physics.Inpainting(img_size=mask.shape[1:], mask=mask, device=device)
```
Tmodrzyk, line 57: Why don't we simply use the dedicated `physics.Demosaicing` then ?

examples/optimization/demo_multispectral_demosaicing.py
```diff
@@ -0,0 +1,151 @@
...
+#
+#         model = dinv.models.DeepImagePrior(
+#             dinv.models.ConvDecoder(img_size=mask.shape[1:]),
+#             img_size=(256, 4, 4),
+#             iterations=1000,
+#         ).to(device)
+#
+#
+# Instead of training models from scratch, you could also fine-tune a foundation model such as :class:`deepinv.models.RAM`
+# without ground truth as done in :footcite:t:`wang2026perspective`.
```
Tmodrzyk, line 147: If you have open weights ready and it's the same RAM implementation as the one currently in deepinv, we might as well show a reconstruction with your method here no ? 
I think it would be great to have a learning-based method here too

Tmodrzyk (commented):
Hi @Andrewwango thanks for updating the examples. I have a few minor comments, and we could maybe add your method with RAM in the demosaicing example ?

examples/optimization/demo_multispectral_demosaicing.py
Andrewwango: The code to initialise RAM with 16ch is quite large and I think it's out of scope to include here unfortunately!

examples/optimization/demo_multispectral_demosaicing.py
Andrewwango: deepinv Demosaicing is a thin wrapper over Inpainting anyway where mask = the RGB (i.e. 3ch) filter. Here it's a generalisation with a 4x4 filter. But good point, we should generalise Demosaicing. I added a new issue to do it: #1329

deepinv/models/dip.py
Andrewwango: The fix is not related to DIP usage, actually DIP works fine. It's a separate fix which is needed because the current code messes up your imported torch library such that importing deepinv can create bugs in completely unrelated torch module code.

examples/optimization/demo_multispectral_demosaicing.py
Andrewwango: done

examples/optimization/demo_multispectral_demosaicing.py
Andrewwango: unfortunately not, since load_example accepts only numpy files, torch files or PIL openable files, whereas this needs to be directly read from bytes

Tmodrzyk (approved):
LGTM !

Issue #1253, opened 2026-06-30
`BlurFFT` initialised with `filter=None` gives an unbound local error on `device`


Andrewwango:
Because of the code below, when BlurFFT physics is initialised with no filter, the device variable is unset so gives an unbound local error.

Solution: set device as None as fallback

https://github.com/deepinv/deepinv/blob/c815f3a94c7e90b0bf00e3577d6d3aeff398425d/deepinv/physics/blur.py#L702-L728

romainvo:
Hi @Andrewwango, thanks for reporting this! Do you have this minimal example in mind?

```python
import torch
import deepinv as dinv

physics = dinv.physics.BlurFFT(
    img_size=(64,64),
    device="cuda"
)

x = torch.rand((1,1,64,64), device="cuda")
y = physics(x)
```

We could add a line `device=None` above line 702; this would not solve the problem though, as `self.mask` is initialized with `None` if no filter is provided. Do you expect `BlurFFT` to compute the identity if no filter is provided during init? Otherwise, we could throw a more explicit error message. Both are fine with me

Issue #1221, opened 2026-06-07
Improving the documentation of prox_l2


jscanvic:
In the docs, LinearPhysics.prox_l2 is described as the prox of the L2 data fidelity of the operator scaled by a scalar factor gamma. In DecomposablePhysics.prox_l2 however, gamma can actually take vector values, which is not mentioned in the docs and what it is computing is not made explicit either. I suggest we clarify this.

1. Which docstrings to change?

We shouldn't only change the docstring of DecomposablePhysics.prox_l2 but also the docstring of LinearPhysics.prox_l2 to let the user know it is overloaded with additional semantics in DecomposablePhysics.prox_l2.

It might be worth also changing the other docs, i.e., that of Downsampling.prox_l2 and LinearPhysics.prox_l2.

2. What is it actually computing?

As far as I can tell, the extra semantics in DecomposablePhysics.prox_l2 don't qualify as a prox. It is not clear to me what they're meant to be computing in general.

It's still possible to rephrase it as the solution to a minimization problem close to the original one of the prox but it is not entirely clear why it is of interest (see rendered equation below "General case").

There are two subcases of interest. First, if gamma is constant, then it does indeed compute the regular proximal operator. (See below, "Special case gamma = cst")

Second, if z = 0, then it computes the solution to the minimization problem associated with the L2 data fidelity with Tikhonov regularization in the dual domain given by the SVD decomposition. For BlurFFT for instance, the regularization is done in the Fourier domain making the whole algorithm compute Wiener deconvolutions (see #1199). (See below, "Special case z = 0")

<img width="652" height="270" alt="Image" src="https://github.com/user-attachments/assets/d41b0673-5536-4774-98bf-601953ac6171" />

3. What to put in the docs?

It would be nice to have the general equation in addition to the prox equation. It would also be nice to mention the special case z = 0.

Issue #965, opened 2025-11-30
Stand-alone learning-free reconstructors


jscanvic:
### Describe the feature

Learning-free reconstructors are used in the trainer to allow comparison with, for instance, IDFTs and FBPs for MRI and CT. At first, they were hard-coded without implementing the Reconstructor interface which made them inconvenient to use with model-dependent metrics (see #788). There is now a (private) class `_NoLearningModel` subclassing `Reconstructor` which is not made available outside the Trainer. I suggest we make it a fully-fledged member of `deepinv.models`.

Use cases include faster debugging cycles by having a standard way to make `Physics.A_adjoint` and `Physics.A_dagger` behave like a reconstructor. Also, making learning-free reconstructors  equivariant using the `EquivariantReconstructor` that will come with #881.

### Motivation

/

Andrewwango:
This would be a nice framework that would cover #975

Issue #897, opened 2025-11-06
Tools from Spectral Analysis


jscanvic:
### Describe the feature

It is important that we strive for an extensive support of traditional solvers as practitioners might tend to be more familiar with them than with more recent and elaborate methods. It lowers the barriers to entry, sends a positive signal and makes comparisons easier. At the moment, we have a limited support of traditional methods based on tools from spectral analysis. I suggest we improve that.

Features that would be nice to have include:

* Nonparametric power sectral density (PSD) models
	* Periodogram, Bartlett's method, Welch's method
	* Support for estimation from a single image or from a list of images
* Parametric PSD models
	* Power law, white noise
	* Support for hard-coded parameters (exponent, noise level)
	* Support for fitting parameters from a single image or from a list of images
* Wiener filtering from pre-computed PSDs
	* Examples for image denoising and deblurring

### Motivation

/

PR #1315, opened 2026-08-08
Add Wiener deconvolution

### LINKED ISSUE #1199: Add Wiener deconvolution
### Environment (optional)

_No response_

### Describe the feature

Wiener deconvolution is a common one-step baseline for deconvolution, similar to the role of Filtered Back Projection (FBP) in tomography: https://en.wikipedia.org/wiki/Wiener_deconvolution
It could be an option of `Blur.A_dagger` and `BlurFFT.A_dagger` similarly to what is done with FBP in `Tomography` and `TomographyWithAstra`.

### Motivation

This can be useful as a pre-processing for some applications, or as a non-learning baseline deconvolution algorithm, which would help us getting towards #965.

### Alternatives considered

_No response_

### Additional context

_No response_

ashishv-git:
Fixes #1199.

## Summary

Adds Wiener deconvolution in closed form as `deepinv.models.WienerDeconvolution(lambda_reg=1.0, prior="laplacian")`, a `Reconstructor`. It solves

$$\hat{x} = \arg\min_x \; \frac{1}{2} \|Ax - y\|_2^2 + \frac{1}{2} \|\lambda^{1/2} \odot F x\|_2^2$$

via `DecomposablePhysics.prox_l2`, so no iterative solver is involved.

`lambda_reg` and `prior` select one of three cases:

- **a scalar** with `prior=None` or `"flat"` — Tikhonov regularisation, the same weight at every frequency;
- **a scalar** with `prior="laplacian"` — the penalty weighted by $|H_L(f)|^2$, penalising high frequencies more than low ones;
- **a tensor** — the noise-to-signal ratio $S_n(f)/S_x(f)$ given per frequency, as requested in #897.

The default Laplacian kernel is identical to `skimage.restoration.uft.laplacian`, the reference implementation cited in the issue.

Following the discussion in this thread, `A_dagger` keeps its meaning as the pseudo-inverse and `WienerDeconvolution` is the single public interface, so `deepinv/physics/` is unchanged apart from the `prox_l2` guard described below.

A gallery example, `examples/models/demo_wiener_deconvolution.py`, compares the three choices of $\lambda$ for deblurring. It then does the same for denoising, expressed as a deconvolution with a unit-impulse filter.

## Scope: `BlurFFT` only

`WienerDeconvolution` accepts `BlurFFT` and rejects everything else. Denoising is covered as a deconvolution with a unit-impulse filter, which is a superset of what `deepinv.physics.Denoising` could offer: `Denoising` has the identity as its SVD basis, so it cannot express frequency-dependent regularisation, and neither the Laplacian prior nor a tensor `lambda_reg` would work through it.

`Blur` is rejected too, although #1199 names `Blur.A_dagger` alongside `BlurFFT.A_dagger`. `Blur` is a `LinearPhysics` with no SVD, so `prox_l2` there runs an iterative least-squares solve. For a scalar `lambda_reg` it reaches the same answer as the closed form, agreeing with `BlurFFT` to around 1e-06 with CG at `tol=1e-10`, so the flat case is already available through `Blur`. The frequency-dependent cases are not available through `Blur`: weighting the spectrum needs the Fourier SVD, and passing a half-spectrum `gamma` to `Blur.prox_l2` raises a shape error. For a circular convolution the equivalent `BlurFFT` provides both frequency-dependent cases, and the error message points there.

## Change to `DecomposablePhysics.prox_l2`

This is the only change under `deepinv/physics/`.

`DecomposablePhysics.prox_l2` computed `b = self.A_adjoint(y) + 1 / gamma * z` unconditionally; it now skips that term when `z` is `None` or a zero scalar. Expanding `z` into a full tensor instead, as `LinearPhysics.prox_l2` does with `torch.full_like`, would not work here: that gives a tensor of full spatial width `W`, while `gamma` from an rfft-based subclass is a half-spectrum of width `W // 2 + 1`, so the product fails to broadcast.

#1221 identifies `z = 0` as one of two subcases of interest, and connects it to this issue:

> if z = 0, then it computes the solution to the minimization problem associated with the L2 data fidelity with Tikhonov regularization in the dual domain given by the SVD decomposition. For BlurFFT for instance, the regularization is done in the Fourier domain making the whole algorithm compute Wiener deconvolutions (see #1199).

No existing behaviour changes. The only call site that supplies a scalar `z` is `Trainer.no_learning_inference`, which calls `prox_l2(0.0, y, 5.0)` when `no_learning_method="prox_l2"`; every other site passes a tensor and takes the unchanged path. I compared its output against the previous code path and the maximum difference is zero, as expected, since adding a float `0.0` to a tensor does not change its value.

The `z` parameter's documented type was widened to match `LinearPhysics.prox_l2`, which already accepts `None`. `:param float gamma:` is also inaccurate, since vector `gamma` is supported, but that is the first point #1221 raises, so it belongs with that issue, not this PR.

## Question

**Placement in the API docs.** I added a `Learning-Free Reconstructors` heading to `deepinv.models.rst` and listed `WienerDeconvolution` under it. None of the existing headings fits a `Reconstructor` that does no learning, since it is neither a classical denoiser nor a deep architecture. Its `.. userguide::` points at the `Pseudoinverse` page, where I added a Wiener deconvolution subsection next to the existing least-squares material, so no new user-guide page was needed.

This is also the first learning-free reconstructor exposed as a `deepinv.models` class, since FBP lives on the physics operator, as `Tomography.fbp()` and `A_dagger(fbp=True)`, and not as a model. The placement may therefore set a precedent for #965, so I would welcome your view on where it should live.

## Notes

- **`physics/blur.py` is untouched by this PR.** The conversion helpers are private methods on `WienerDeconvolution`: `_lambda_to_gamma` and `_build_laplacian_gamma`.

- **A filterless `BlurFFT` stays unusable, as in #1253**, where `A` and `A_adjoint` raise `UnboundLocalError`. Such an operator leaves `mask` as `None`, and `_build_laplacian_gamma` reads `physics.mask.device`, so it raises `AttributeError: 'NoneType' object has no attribute 'device'` here. I have not guarded it, since guarding one entry point would leave the rest of the operator broken.

- **`WienerDeconvolution` is not registered in `REST_MODEL_LIST`.** The contributing guide asks for new classes to be added to the existing shared tests, and `test_restoration_models` is the harness for `(y, physics)` reconstructors. I left it out because the parametrisation over that list is currently commented out in favour of `["ram"]`, so only `ram` runs, and because `LINEAR_OPERATORS` has nine entries while this model supports one of them, `fftdeblur`, so registering it would need skips for eight of nine combinations plus edits to `choose_restoration_model`. I would be glad to register it if you think it belongs there.

- **The example adds one entry to `docs/source/refs.bib`.** It is a pure append at the end of the file, so it cannot disturb existing entries.

- **Testing.** One test function, `test_wiener_deconvolution`: 1 case on CPU and 2 when a GPU is present, since the `device` fixture covers both. All pass locally on CPU and on CUDA. The example builds with `make html-pattern`, which runs `sphinx-build -W`.

This description was drafted with LLM assistance and edited by me.

- [X] `python3 -m pytest deepinv/tests` runs successfully.
- [X] `black .` and `ruff check .` run successfully.

### LLM policy

- [X] I, a human, wrote code, and LLM tools helped me to write part of the code.


jscanvic:
Thank you for handling this issue Ashish!

I really like how you handled things and made them fit in the library

I'll dig into the implementation details soon but for now a few high-level remarks.

**Regarding `BlurFFT.A_dagger`**

At the moment `Physics.A_dagger` sometimes serves as a pseudo-inverse and sometimes as a learning-free reconstructor. This somewhat makes sense for well-conditioned forward operators for which the pseudo-inverse is typically a good learning-free reconstructor, but it is very confusing for ill-conditioned operators like blur and tomography.

My personal take is we should refrain from making existing `A_dagger` compute anything other than the true pseudo-inverse, and I would suggest we drop the changes you made to `BlurFFT.A_dagger` and keep `WienerDeconvolution` as the single public interface to your implementation.

@Tmodrzyk What do you think of it? If we really want to have a hard link from physics to associated learning-free reconstructors, I'm happy to introduce a new base method on `Physics` to deal with it. Something like:

```python
physics = dinv.physics.BlurFFT(...)
model: dinv.models.Reconstructor = physics.get_model(kind="wiener")
```

**Other remarks**

- The way you integrate with `BlurFFT.prox_l2` is neat!
- I wouldn't expose a specific method `wiener` on `BlurFFT`
- Please create an example showing the results you get with Wiener filtering

ashishv-git:
Hi Jérémy,

Thank you for the review.

**`BlurFFT.A_dagger`**

Happy to go either way, and I will make the change promptly once it is decided.

One thing that may be useful: the six private helpers in `physics/blur.py` are used only by `BlurFFT.A_dagger`. If it goes, they can move into `models/wiener.py`, the only other file that uses them, and `physics/blur.py` returns to defining no private module-level names. The change to `DecomposablePhysics.prox_l2` would stay, since `WienerDeconvolution` depends on it.

**`BlurFFT.wiener()`**

Agreed, and the PR does not add one.

**Example**

Will do. I will follow up here once the example is ready.

**On the failing check**

`Test windows all optional dependencies` fails on `test_optim.py::test_least_squares_implicit_backward`, which looks like #1168 rather than anything from this PR. The other checks have passed.

Tmodrzyk:
Hi @ashishv-git @jscanvic, I was initially in favor of an argument to have an alternative reconstructor in `A_dagger` but I've changed my stance on this.
I'm now strongly in favor of keeping physics and optim / reconstructor separated. `physics.A_dagger` happens to be a reconstructor, but that should remain a coincidence, and it should just implement the pseudo-inverse.
The physics part of the library is currently in a much more stable and clean state than the reconstructors or the optim part, so we should not pollute it with exceptions. :smile: 

Also, I'm not sure where non-iterative non-learning reconstructors should be implemented in the library at the moment.

ashishv-git:
Thanks, Thibaut.

With that agreed, I will remove the `A_dagger` changes and keep `WienerDeconvolution` as the only public interface. That leaves `physics/blur.py` untouched by this PR; the only remaining change under `deepinv/physics/` is the guard in `DecomposablePhysics.prox_l2`, which `WienerDeconvolution` depends on.

On where non-iterative non-learning reconstructors should be implemented: `WienerDeconvolution` currently lives in `deepinv.models`, where #965 also proposes putting `_NoLearningModel`. Happy to move it if a different module suits better.

I will push the `A_dagger` removal together with the Wiener filtering example, and update the PR description.

jscanvic (changes requested):
It looks good overall!

Just a few changes to make it fit our conventions better:
- Drop the changes made to `A_dagger` as well as all the changes in `test_physics.py`
- Move all of your changes in `blur.py` to `wiener.py`
- Inline `_is_frequency_dependent` and `_is_zero_lambda` for better cohesion with the rest of the code base (one or two-liners are generally inlined)
- Validate the value for `prior` in `_lambda_to_gamma` but not in the constructor of `WienerDeconvolution` 
- Inline `_WIENER_VALID_PRIORS` and `_WIENER_EPS_CLAMP`
- Move `_lambda_to_gamma` and `_build_laplacian_gamma` to private class methods in `WienerDeconvolution`
- Drop `test_wiener_rejects_invalid_prior_at_construction` and `test_wiener_rejects_unsupported_physics` for better cohesion with the rest of the code base
- Merge all the new tests in `test_models.py` into a single test function and inline `wiener_blur_1ch` and `wiener_denoise_impulse`

Once it's done I'll do a final check of the example before approving the changes

ashishv-git:
Hi Jérémy,

Thank you for the review. All eight changes are made, and the example is ready for your final check.

**Scope**

`physics/blur.py` and `tests/test_physics.py` are back to a zero diff against `main`. The only remaining change under `deepinv/physics/` is the `prox_l2` guard that `WienerDeconvolution` depends on.

**Notes on the changes**

*Validation of `prior`.* An invalid value now raises on the first `forward`, not at construction. The check applies to a scalar `lambda_reg` only, since a tensor of rank >= 1 returns before the `prior` branches are reached. The `ValueError` message itself is unchanged.

*The Laplacian epsilon.* `_WIENER_EPS_CLAMP` is inlined at the clamp as requested. `_build_laplacian_gamma` also had an `eps` parameter that nothing passed; I dropped the parameter and kept `eps` as a local, since it is a different quantity from the clamp and the docstring reasons about it as $\varepsilon$. Happy to make it a bare literal too if you would prefer that.

*Two guards removed.* While converting `_build_laplacian_gamma` into a method I dropped its `isinstance(physics, BlurFFT)` safeguard, which its own comment described as unreachable, and its check for an operator with no filter. A filterless `BlurFFT` is already unusable per #1253, where `A` and `A_adjoint` raise `UnboundLocalError`; without a filter `mask` is `None`, so this path now fails with an `AttributeError` where the guard used to raise a `ValueError`.

*Coverage.* The patch coverage check may flag three `ValueError` branches that are no longer exercised: the two physics-type checks in `forward`, and the invalid-`prior` check in `_lambda_to_gamma`. Happy to fold a `pytest.raises` case into `test_wiener_deconvolution` if you would like them covered.

*Non-square image grids.* Moving `_build_laplacian_gamma` meant rewriting the lines that read `img_size` and pass the two spatial dimensions to `filter_fft`, where transposing them is an easy slip. A square grid cannot detect that, since both orderings then give the same result, so I made the grids non-square while merging the tests.

**Example**

`examples/models/demo_wiener_deconvolution.py` compares the three choices of `lambda_reg` for deblurring, then does the same for denoising expressed as a deconvolution with a unit-impulse filter. It appends one entry to `docs/source/refs.bib`, at the end of the file.

**Testing**

`test_wiener_deconvolution` is now a single test function: one case on CPU, and a second on CUDA when a GPU is available. It passes on both. `test_physics.py`, `test_optim.py` and `test_trainer.py` pass as well, and the example builds with `sphinx-build -W`.

I have also updated the PR description.

jscanvic:
Thanks Ashish! I'll have a loop asap

jscanvic:
Sorry for the delay

The code looks good to me - I'll approve once the CI passes

Thanks again!

jscanvic (approved):

ashishv-git:
Thanks for the review and the approval, Jérémy, and for syncing the branch with main!

PR #1317, opened 2026-08-09
Compute per-sample statistics for norm_inputs="standardize" (fixes #908)

### LINKED ISSUE #908: norm_inputs == "standardize" uses batch-level statistics instead of sample-level statistics
### Environment

all

### Describe the bug

The docs for `Metric` say "`standardize` standardizes to same mean and std as ground truth". However, for batch size > 1, this should be interpreted as same mean and std across the gt **batch**, not image:

```python
        if self.norm_inputs == "standardize":
            if x_net is None or x is None:
                raise ValueError(
                    "Both x and x_net must not be None in order to use standardize."
                )
            x_net = (x_net - x_net.mean()) / x_net.std() * x.std() + x.mean()
```

### Steps to reproduce

/

### Expected behavior

Based on the docs, I would expect passing "standardize" would use mean and std on a per sample basis. The current method is not deterministic if the batching / dataloading is non-deterministic, and in general it makes more sense to either 

(i) Standardize using the entire dataset's statistics, which will not be well approximated in the current implementation for smaller batch size, or might not make too much sense for datasets where intensity values across samples are not distributed in a "nice" way.

(ii) Standardize on a per sample basis.

_No response_

### Additional context

We could also provide to the user the option to specify if mean & std should be calculated on a batch or sample level (or even channel level, although I'm not familiar with use-cases where this would be useful)

itsDigvijaysing:
Fixes #908.

`Metric.forward` computed mean and std over the whole tensor, batch dimension included,
so a sample's score changed depending on which other samples were in its batch. The
docstring says it standardizes to the same mean and std as the ground truth, which
implies per sample.

Statistics are now computed per sample (all dims except the batch dim), the same way
`normalize_signal` already does it.

```python
x_hat = torch.tensor([[1.0, 2.0], [30.0, 40.0], [-5.0, 5.0]])
x     = torch.tensor([[0.0, 1.0], [2.0,  3.0], [ 4.0, 5.0]])

MSE(norm_inputs="standardize")(x_hat, x)
# before: tensor([ 1.0241,  5.4739, 10.5409])
# after:  tensor([0., 0., 0.])
```

Passing those samples one at a time already gave `[0, 0, 0]`, so the batched and
unbatched paths disagreed. Added a test that checks they now match.

This changes scores for anyone using `standardize` with batch size > 1, so I marked it
`(Breaking)` in the changelog. Nothing else in the library uses `standardize`.

The issue also mentions making the reduction level configurable (batch / sample /
channel). I kept this to the minimal fix, but happy to add the option if you prefer.

Ran pytest deepinv/tests/test_metric.py (179 passed); did not run the full suite locally."

### LLM policy

### DIFF
--- a/deepinv/loss/metric/metric.py
+++ b/deepinv/loss/metric/metric.py
@@ -28,7 +28,7 @@ class Metric(Module):
         the data must either be of complex dtype or have size 2 in the channel dimension (usually the second dimension after batch).
     :param bool train_loss: if higher is better, invert metric. If lower is better, does nothing.
     :param str reduction: a method to reduce metric score over individual batch scores. ``mean``: takes the mean, ``sum`` takes the sum, ``none`` or None no reduction will be applied (default).
-    :param str norm_inputs: normalize images before passing to metric. ``l2`` normalizes by :math:`\ell_2` spatial norm, ``min_max`` normalizes by min and max of each input, ``clip`` clips to :math:`[0,1]`, ``standardize`` standardizes to same mean and std as ground truth, ``none`` or None no normalization will be applied (default).
+    :param str norm_inputs: normalize images before passing to metric. ``l2`` normalizes by :math:`\ell_2` spatial norm, ``min_max`` normalizes by min and max of each input, ``clip`` clips to :math:`[0,1]`, ``standardize`` standardizes each sample to the same mean and std as its corresponding ground truth sample (statistics are computed per sample, not over the batch), ``none`` or None no normalization will be applied (default).
     :param int, tuple[int], None center_crop: If not `None` (default), center crop the tensor(s) before computing the metrics.
         If an `int` is provided, the cropping is applied equally on all spatial dimensions (by default, all dimensions except the first two).
         If `tuple` of `int`, cropping is performed over the last `len(center_crop)` dimensions. If positive values are provided, a standard center crop is applied.
@@ -234,7 +234,13 @@ class Metric(Module):
                 raise ValueError(
                     "Both x and x_net must not be None in order to use standardize."
                 )
-            x_net = (x_net - x_net.mean()) / x_net.std() * x.std() + x.mean()
+            # Stats per sample, so batching does not change the result
+            non_batched_dims = list(range(1, x_net.ndim))
+            mean = x.mean(dim=non_batched_dims, keepdim=True)
+            std = x.std(dim=non_batched_dims, keepdim=True)
+            mean_net = x_net.mean(dim=non_batched_dims, keepdim=True)
+            std_net = x_net.std(dim=non_batched_dims, keepdim=True)
+            x_net = (x_net - mean_net) / std_net * std + mean
 
         x_net = self.normalizer(x_net)
         x = self.normalizer(x) if x is not None else None
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -44,6 +44,7 @@ Fixed
 - Fix inversion in :class:`deepinv.transform.Reflect` (:gh:`1236` by `Sarra Amiri`_)
 - (Breaking) Have `x_shift` represent horizontal shifts and `y_shift` vertical shifts in :class:`deepinv.transform.Shift` (:gh:`1236` by `Sarra Amiri`_)
 - Force trainer non_blocking_transfers=False on MPS and CPU (:gh:`1311` by `Andrew Wang`_)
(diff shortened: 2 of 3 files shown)


Tmodrzyk:
Hi @itsDigvijaysing thank you for tackling this issue ! 
Don't hesitate to ping me if you need another workflow run or a review when the CI passes. :smile: 
Maybe @nucli-vicky will want to take a look ?

deepinv/loss/metric/metric.py
```diff
@@ -28,7 +28,7 @@ class Metric(Module):
         the data must either be of complex dtype or have size 2 in the channel dimension (usually the second dimension after batch).
     :param bool train_loss: if higher is better, invert metric. If lower is better, does nothing.
     :param str reduction: a method to reduce metric score over individual batch scores. ``mean``: takes the mean, ``sum`` takes the sum, ``none`` or None no reduction will be applied (default).
-    :param str norm_inputs: normalize images before passing to metric. ``l2`` normalizes by :math:`\ell_2` spatial norm, ``min_max`` normalizes by min and max of each input, ``clip`` clips to :math:`[0,1]`, ``standardize`` standardizes to same mean and std as ground truth, ``none`` or None no normalization will be applied (default).
+    :param str norm_inputs: normalize images before passing to metric. ``l2`` normalizes by :math:`\ell_2` spatial norm, ``min_max`` normalizes by min and max of each input, ``clip`` clips to :math:`[0,1]`, ``standardize`` standardizes each sample to the same mean and std as its corresponding ground truth sample (statistics are computed per sample, not over the batch), ``none`` or None no normalization will be applied (default).
```
nucli-vicky, line 31: I'm not sure the `(statistics are computed per sample, not over the batch)` offers additional information. 

Would leave it out

itsDigvijaysing:
@Tmodrzyk 
rebased to fix the changelog conflict, should be clean now. 
Could you kick off a workflow run when you get a chance? Happy to make the reduction level configurable if you'd rather have that than the minimal fix.

itsDigvijaysing:
The Windows failure is test_anscombe_transform. [ unrelated to this PR ](Ubuntu passes the same test, and this change only touches loss/metric/).

deepinv/loss/metric/metric.py
```diff
@@ -234,7 +234,13 @@ def forward(
...
                     "Both x and x_net must not be None in order to use standardize."
                 )
-            x_net = (x_net - x_net.mean()) / x_net.std() * x.std() + x.mean()
+            # Stats per sample, so batching does not change the result
+            non_batched_dims = list(range(1, x_net.ndim))
+            mean = x.mean(dim=non_batched_dims, keepdim=True)
+            std = x.std(dim=non_batched_dims, keepdim=True)
+            mean_net = x_net.mean(dim=non_batched_dims, keepdim=True)
+            std_net = x_net.std(dim=non_batched_dims, keepdim=True)
+            x_net = (x_net - mean_net) / std_net * std + mean
```
nucli-vicky, line 243: we need to use some form of `eps` to guard against a 0 or near-0 standard deviation. 

I would also name it something like `mean_gt` instead of `mean`, for clarity

docs/source/changelog.rst
```diff
@@ -44,6 +44,7 @@ Fixed
 - Fix inversion in :class:`deepinv.transform.Reflect` (:gh:`1236` by `Sarra Amiri`_)
 - (Breaking) Have `x_shift` represent horizontal shifts and `y_shift` vertical shifts in :class:`deepinv.transform.Shift` (:gh:`1236` by `Sarra Amiri`_)
 - Force trainer non_blocking_transfers=False on MPS and CPU (:gh:`1311` by `Andrew Wang`_)
+- (Breaking) Compute per-sample statistics in :class:`deepinv.loss.metric.Metric` when ``norm_inputs="standardize"``, instead of statistics over the whole batch, so that a sample's score no longer depends on the other samples it is batched with (:gh:`1317` by `digvijaysing`_)
```
nucli-vicky, line 47: I think `Breaking` is a bit strongly worded?

Also, _so that a sample's score no longer depends on the other samples it is batched with_ seems a bit redundant / already implied by the rest of the sentence

nucli-vicky (changes requested):
Thanks @itsDigvijaysing for your first contribution!

Some comments here. Additionally, I'm not sure standardizing over all channels is appropriate compared to per-channel? We should at least document in the docstring this is what's being done.

PR #1318, opened 2026-08-10
Bilevel hyperparameter learning with MAID

MohammadSadeghSalehi:
# Bilevel hyperparameter learning with MAID

Adds the Method of Adaptive Inexact Descent to `deepinv.optim.bilevel`, with a
prior-agnostic interface so any convex, twice-differentiable regulariser can be
learned through it. Implements Salehi, Mukherjee, Roberts and Ehrhardt, *SIAM
Journal on Mathematics of Data Science* 2025 (arXiv:2308.10098), with the
saddle-point instantiation of Bogensperger, Ehrhardt, Pock, Salehi and Wong
(arXiv:2412.06436).

## Why this is needed

DeepInverse already has strong reconstruction algorithms and a rich library of
priors. What it does not have is a principled way to *choose the parameters
those priors depend on*. Today a user picks a regularisation weight by hand or
by grid search, and anything with more than two or three parameters is out of
reach: a grid over a 7533-parameter filter bank does not exist.

The alternatives each give something up.

**Grid search** does not scale beyond a handful of parameters, and gives no
gradient to follow.

**Unrolling** differentiates the *algorithm* rather than the solution, so the
learned parameters are tied to the iteration count and the solver used at
training time. Change either and the result is no longer the one that was
learned. It also stores every iterate, so memory grows with the number of
iterations.

**Implicit differentiation at fixed accuracy** solves the right problem, but
must pick a lower-level tolerance in advance. Too loose and the hypergradient
is wrong in a way nothing detects; too tight and most of the compute is spent
polishing solutions that a coarse step would have discarded anyway.

MAID resolves that trade-off by *adapting* the accuracy: it solves the lower
level loosely while far from a solution, tightens only when its descent test
demands it, and certifies each reconstruction with an a posteriori bound so the
inexactness is quantified rather than hoped away. That combination, adaptive
accuracy with a certificate, is what neither grid search nor unrolling
provides.

## What this brings to DeepInverse

**Any prior becomes learnable.** Supply a per-sample energy and everything else
follows by autograd. That turns parameter selection from a per-prior
engineering exercise into a property of the library: the same code path learns
3 parameters or 7533, a hand-designed regulariser or a neural one.

**Reconstructions come with a guarantee.** For a lower level that is
mu-strongly convex, every reported result carries

    ||x* - x|| <= ||grad_x h(x)|| / mu,

computable from the returned iterate alone. A reconstruction is then a bounded
distance from the true minimiser rather than "whatever the solver returned
after N iterations". This matters most where it is hardest to check: on real
data with no ground truth.

**It runs where users run.** float32 on CUDA and MPS, verified unbiased against
a float64 reference, with batched solves and memory-aware batch sizing. Bilevel
learning is expensive, and an implementation that only works in float64 on CPU
is not usable for the problems people care about.

**It is a bridge to the variational literature.** Learned convex regularisers,
input-convex networks and adaptive regularisation all need exactly this
machinery. Having it in DeepInverse makes those methods reproducible here
rather than living in separate research codebases.

## Learning any prior

A prior supplies one function, the energy of a batch of images. Everything else
follows by automatic differentiation: the lower-level gradient is a backward
pass in `x`, the Hessian-vector product a second one, and the mixed Jacobian a
backward pass in `x` then in `theta`. No derivative is written by hand, and the
model that is solved and the model that is differentiated are the same object.

```python
class MyPrior(ParametricPrior):
    def __init__(self, channels=3):
        self.n_params = channels

    def energy(self, x, theta):            # -> (B,)
        w = torch.exp(theta).view(1, -1, 1, 1)
        return (w * x.abs()).flatten(1).sum(dim=1)

    def init_theta(self, *, dtype=torch.float64, device="cpu", seed=0):
        return torch.zeros(self.n_params, dtype=dtype, device=device)
```

Three priors ship with it, chosen to span the range rather than to compete:
a multi-convolution ridge regulariser (7533 parameters), total variation with
learned per-channel weights (3), and an input-convex network (5968).

## Results

Denoising at sigma = 0.05, trained on one 32x32 crop from each of 16 distinct
CBSD68 images, evaluated on 8 unseen images. Same oracle, solver and accuracy
rules throughout; only the energy differs.

| Prior | Parameters | PSNR (dB) | SSIM |
|---|---|---|---|
| noisy input | | 26.05 | 0.5992 |
| total variation, grid-tuned | 1 | 31.28 | 0.8215 |
| convex ridge regulariser | 7533 | **31.64** | **0.8499** |
| learned total variation | 3 | 30.82 | 0.8154 |
| input-convex network | 5968 | 29.20 | 0.7452 |

This is evidence that the interface carries priors across a 2500x range in
parameter count, not a ranking: the budgets differ and none of the runs is
converged. The guide records why each stopped. The learned total variation
halts after 25 iterations at the accuracy floor because it is initialised at
the weight grid search finds. The input-convex network begins with an energy
far below the data term and must grow the prior before its hypergradient
carries much information.

A prior learned on 32x32 patches transfers to whole 256x256 images unseen at
any crop, where the ridge regulariser reaches 32.62 dB against 30.81 for
grid-tuned TV. The regulariser is convolutional, so it carries no notion of
image size.

## Design decisions worth review

**Hyperparameters are derived, not fixed.** `eps0`, `delta0` and `alpha0`
default to `None`, meaning derive at `theta0`: the accuracy from the
stationarity residual at `A* y`, the step from the hypergradient norm. No
absolute constant serves every prior, because both references are properties of
the regulariser. Measured at the same initialisation, `eps0 = 1e-1` leaves a
denoising problem below its noisy input (24.70 dB against 26.01) while sitting
inside a plateau four orders wide for inpainting; and `||z0||` ranges from 3.7
for the ridge prior to 6.9e4 for the input-convex network, so a step suited to
one moves the other by roughly 6900 on its first trial. Explicit floats are
still honoured, and oracles that cannot supply the references fall back with a
warning.

**Tolerances are dimensionless and dtype-aware.** `eps` is a per-element rms
gradient tolerance, so it means the same accuracy at 32x32 and 256x256, and the
floor comes from `torch.finfo(dtype).eps`. Falling short of `eps` is not an
error: the certificate holds at any residual, so a solve that stalls at the
precision floor returns a valid, wider bound rather than raising. The floor
itself is found by detecting that the residual has stopped improving, since it
depends on conditioning and dtype and cannot be predicted in advance.

**float32 is supported and verified.** Against a float64 reference on identical
inputs the float32 hypergradient is inexact but not biased: cosine 0.99999713,
relative norm 2.4e-03, and MPS float32 matches CPU float32 to seven significant
figures. The angle is what matters, since inexactness shortens the
hypergradient while bias would rotate it.

**Batched lower-level solves.** Samples are independent given `theta`, so the
batched Hessian is block diagonal and every operator applies to all of them at
once; only the scalars stay per sample. Batch size is chosen from measured
memory. Accumulation is exact, so splitting a sample set changes peak memory
and nothing else (1.9e-15 in float64). Any DeepInverse physics works, including
operators with per-sample parameters, and construction refuses a physics that
couples samples across the batch.

## Testing

114 tests across `deepinv/tests/test_maid*.py`, covering the lower-level
solvers, the certificate, the oracles, the accelerated path, minibatch
accumulation, the prior interface and the batched path.

The last two were added after codecov flagged `batched.py` and `priors.py`
at 17-18% patch coverage: they had no tests, and the coverage shown was
import-time execution alone. `test_maid_priors.py` and `test_maid_batched.py`
close that. Both were checked by mutation rather than assumed: flipping the
sign of the hypergradient fails all three finite-difference tests, and
dropping a batch from the accumulation fails both exactness tests.

The central check is a finite difference on the hypergradient, taken per
parameter block and per coordinate against
`g(x*(theta))`. It is the only test that can detect a
disagreement between the model that is solved and the model that is
differentiated, since automatic differentiation reproduces whatever the
forward code expresses, including a mistake. The per-block split matters: an
error confined to one block is invisible in a single whole-vector comparison.

Two checks are recommended for any new prior, and both are cheap:

- assemble the Hessian on a small image and confirm its smallest eigenvalue is
  non-negative, establishing convexity in `x`
- compare the hypergradient against a central finite difference along a random
  direction

Both are now run for all three supplied priors as part of the suite, in
`test_maid_priors.py`, rather than quoted from a one-off script.

Run locally on Apple silicon: 114 MAID tests pass, and `test_optim.py` gives 14
failed, 221 passed, 32 skipped. Every failure is on the MPS device, and the
same suite on unmodified `main` at 522bba93 gives 14 failed, 221 passed, 32
skipped with a test-for-test identical failure set, so none of them is
introduced here. Eight are `test_least_squares_implicit_backward`, the rest
`test_tvprior_gradient`, `test_linear_system`, `test_MLEM`,
`test_CP_datafidsplit`, `test_CP_K` and
`test_least_squares_implicit_backward_nonleaf_buffer_grad`. They look like
gaps in MPS operator coverage rather than anything specific to this branch.

`test_datasets.py` was not run locally: it downloads full public datasets
(Flickr2K alone exceeded 3 GB before being stopped). CI covers both.

## Breaking change

`MAIDConfig.eps0`, `delta0` and `alpha0` change from `float` to `float | None`
and now default to `None`. Code passing explicit values is unaffected. The
previous default of `eps0 = 1e-1` was harmful under the per-element tolerance
introduced here, so anyone relying on it was already getting poor behaviour.

## Documentation

- `docs/source/user_guide/reconstruction/bilevel.rst`, covering the oracle
  interface, certified and non-certified paths, the accelerated heuristics with
  an ablation, numerical precision and GPU backends, batching, and learning a
  new prior
- three gallery examples under `examples/optimization/`
- API entries and a changelog entry

### LLM policy


MohammadSadeghSalehi:
A few changes since the last CI run, and one request.

**Request:** the workflows need approval again to run. CI last ran on `6c78e16`; the commits since have no check runs, so the Codecov report above is measured against that commit and predates the tests mentioned below.

**Coverage.** Codecov flagged `batched.py` at 17.5% and `priors.py` at 18.3%, and it was right: those two modules had no tests at all. They now do, 30 tests across `test_maid_batched.py` and `test_maid_priors.py`, covering the batched hypergradient accumulation, the block-diagonal check, per-sample CG, and for every supplied prior a Hessian-eigenvalue convexity check and a finite-difference check on the hypergradient. Measured locally over `deepinv/optim/bilevel`, `priors.py` is at 94% and `batched.py` at 75%, the remainder there being the memory-probe and CUDA/MPS branches that a CPU run cannot reach. The suite is 116 tests.

**Docs build.** The failure on the previous run was real and is fixed: two examples used `Path(__file__)`, which is undefined when sphinx-gallery execs a file rather than importing it.

**The CRR example has been rebuilt.** It trained on two Set3C images and held out one, stopped at 30 outer iterations, and ran through the sequential oracle rather than the batched one. It is now CBSD68, sixteen 32x32 training crops and eight held out, on the batched path. Sixteen is the smallest set that generalises here; with eight the regulariser fits the training crops and loses to TV on held-out data, and raising the budget to 100 outer steps does not close that.

Worth flagging explicitly, since it changes what the example claims: on a level comparison, adaptive accuracy alone does **not** beat a well-chosen fixed accuracy on PSNR at this budget. The two land within 0.01 dB and MAID spends about twice the lower-level iterations. What clears the grid-tuned TV baseline is the accelerated pair, Barzilai-Borwein initialisation and the nonmonotone test, at 31.75 dB against 31.24 held out. This matches what the user guide already says, that fixed accuracy wins on a cheap lower level and that BB carries almost all of the benefit; the example had been asserting otherwise. It now reports all three arms so each switch can be attributed separately.

The value of adapting the accuracy is robustness rather than PSNR: the fixed arm only matches because its tolerance was picked after seeing which values fail. Given MAID's own starting accuracy of 1e-3, the same arm cannot certify descent, exhausts its line search every iteration and never takes a step.

**Also fixed:** `pack_init_theta` defaulted to `weight_scale=0.05` while `ConvexRidgePrior2.init_theta` used 1.0. The scale leaves the energy exactly invariant, since the features are Lipschitz-normalised, but moves the hypergradient by 30x, so the mismatch silently changed step-size behaviour depending on which entry point you came through. Both are 1.0 now, with the invariance documented.

MohammadSadeghSalehi:
@Andrewwango you marked this WIP on 12 August, which was fair: my comment that morning described the example being rebuilt and the test coverage being filled in, so it read as a moving target. That work is finished, so I have taken it out of draft.

What changed since you looked:

- The CRR example was rebuilt. It had trained on two Set3C images and held out one, stopped at 30 outer iterations well short of convergence, and ran through the sequential oracle rather than the batched one this PR adds. It is now CBSD68, sixteen 32x32 training crops and eight held out, on the batched path, and it is 371 lines rather than 1415.
- The example's conclusion changed and is now stated plainly rather than asserted. On a matched comparison, adaptive accuracy alone does not beat a well-chosen fixed accuracy on PSNR at this budget: the two land within 0.01 dB and MAID spends roughly twice the lower-level iterations. What clears the grid-tuned TV baseline is the accelerated pair, Barzilai-Borwein initialisation and the nonmonotone test, at 31.75 dB against 31.24 held out. This agrees with what the user guide already said about cheap lower levels, so the example had been contradicting our own documentation. It now reports all three arms so each switch can be attributed.
- Codecov was right that `batched.py` and `priors.py` were untested. They now have 30 tests, including a Hessian-eigenvalue convexity check and a finite-difference hypergradient check for every supplied prior. The suite is 116 tests.
- The docs build failure was real and is fixed: two examples used `Path(__file__)`, which is undefined when sphinx-gallery execs rather than imports.
- Merged main four times since; currently conflict-free.

One request: the workflows still need approval to run. CI has only ever run on `6c78e16`, so the Codecov report on this PR predates the tests above.

Happy to split anything out if the diff is too large to review in one piece.

Tmodrzyk:
Hi @MohammadSadeghSalehi,

Thanks for the work you put into this. Unfortunately, at over 12 000 lines, this PR is too large for us to review or maintain and exceeds our approximate 3 000 lines limit (see our guidelines with regard to PR scope https://deepinv.org/contributing.html#code-scope).
We’re therefore going to close this PR. 

If you would like to continue working on this feature, we suggest:
- opening an issue to motivate the addition of this feature, and for us to discuss the possible integration and which existing abstractions can be reused  
- dividing the work into smaller PRs
- be more concise with respect to the documentation added, you can refer to the doc guidelines for more details: https://deepinv.org/contributing.html#docstring-guidelines  

We would be happy to discuss this plan with you before you begin. Thanks for your understanding and for your interest in contributing !

MohammadSadeghSalehi:
Thanks, that is fair. I have opened #1390 with the plan you suggested: what the feature adds, which existing abstractions it reuses instead of reimplementing, a four PR split each under 3000 lines, and two design questions I would like settled before opening the first one.

PR #1328, opened 2026-08-14
better adjointness test

tachella:
Small PR improving the meaning of `physics.adjointness_test`

It modifies the definition such that we compute $\|A A^{\top} - (AA^{\top})^{\top}\|_F$, which is always positive and can be related to the operator norm: if the error is much smaller than the norm, the lack of adjointness is less severe.

Keeps backwards compatibility with the previous API, but offers a new way of computing the adjointness error by passing `img_size` only

### LLM policy

### DIFF
--- a/deepinv/physics/forward.py
+++ b/deepinv/physics/forward.py
@@ -723,17 +723,17 @@ class LinearPhysics(Physics):
         )
 
         if u is not None:
-            u_in = randn_like(u)
-        else:
-            if img_size is None:
-                if self.img_size is not None:
-                    img_size = self.img_size
-                else:
-                    raise ValueError(
-                        "Either img_size or u must be provided for the adjointness test."
-                    )
-            # note: this does not support tensor lists
-            u_in = torch.randn((num_samples,) + img_size, device=device)
+            img_size = u.shape[1:]
+        elif img_size is None:
+            if self.img_size is not None:
+                img_size = self.img_size
+            else:
+                raise ValueError(
+                    "Either img_size or u must be provided for the adjointness test."
+                )
+
+        # note: this does not support tensor lists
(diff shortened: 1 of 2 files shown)


deepinv/physics/forward.py
```diff
@@ -693,16 +693,48 @@ def compute_sqnorm(
...
+        elif img_size is None:
+            if self.img_size is not None:
+                img_size = self.img_size
+            else:
+                raise ValueError(
+                    "Either img_size or u must be provided for the adjointness test."
+                )
+
+        # note: this does not support tensor lists
+        u_in = torch.randn((num_samples,) + img_size, device=device)
```
Andrewwango, line 736: dtype=u.dtype?

deepinv/physics/forward.py
```diff
@@ -693,16 +693,48 @@ def compute_sqnorm(
...
+            img_size = u.shape[1:]
+        elif img_size is None:
+            if self.img_size is not None:
+                img_size = self.img_size
+            else:
+                raise ValueError(
+                    "Either img_size or u must be provided for the adjointness test."
+                )
+
+        # note: this does not support tensor lists
```
Andrewwango, line 735: This will break backward compat for users who use tensorlist physics then?

deepinv/physics/forward.py
```diff
@@ -720,7 +752,7 @@ def adjointness_test(self, u, **kwargs):
 
         s2 = (Atv * u_in.conj()).flatten().sum()
 
-        return s1.conj() - s2
+        return (s1.conj() - s2).abs().pow(2).mean().sqrt()
```
Andrewwango, line 755: Do you not think the error should be divided by s2.abs()?

Andrewwango (changes requested):
small comments

PR #1340, opened 2026-08-27
Do not count failed backtracking checks as global iterations in FixedPoint

### LINKED ISSUE #1272: Failed backtracking increases the global iteration counter in `FixedPoint`

MohammadSadeghSalehi:
Closes #1272.

### The reported bug

`FixedPoint` iterates over `range(self.max_iter)`, so a rejected backtracking trial spends a global iteration. A failed sufficient-decrease check discards the trial iterate and shrinks the stepsize, which is part of the search for one iteration rather than an iteration in its own right.

The example from the issue returns `0.0`:

```
Backtracking : new stepsize = 2.000000
Backtracking : new stepsize = 1.000000
Returned x: 0.0
```

Both iterations of `max_iter=2` go on backtracking and the iterate never leaves its initialisation. With this change it returns `1.0`, the minimiser of `0.5 ||x - y||^2` at `y = 1`.

This also fixes the second point in the issue, which was untested there: `single_iteration` was receiving an index running ahead of the iterate it labels, so any algorithm keying internal state off `it` reads the wrong entry.

`it` now counts accepted iterations only. Consecutive failures remain bounded by `backtracking_config.max_iter`, so the loop still terminates, and the convergence check is still evaluated at the index of the iteration just accepted, leaving `check_conv_fn` semantics unchanged.

### A second bug in the same block

The message printed on exhausting the backtracking budget interpolates `self.max_iter_backtracking`. That attribute is defined nowhere in the codebase, so with `verbose=True` the path raised `AttributeError` instead of stopping:

```python
model = dinv.optim.GD(
    data_fidelity=dinv.optim.L2(), stepsize=1e6, max_iter=50,
    backtracking=dinv.optim.BacktrackingConfig(gamma=0.1, eta=0.5, max_iter=2),
    verbose=True,
)
model(y, physics, init=x0)
# AttributeError: 'FixedPoint' object has no attribute 'max_iter_backtracking'
```

It now reports `backtracking_config.max_iter` and returns normally.

### Tests

Two regression tests in `test_optim.py`, one per bug. Both fail on `main` and pass here:

```
without the fix: 2 failed
with the fix:    2 passed
```

`pytest deepinv/tests/test_optim.py -k "not device1"` gives 123 passed, 28 skipped, no failures. The `device1` (MPS) cases are excluded because they fail identically on unmodified `main` on this machine, unrelated to this change.

### DIFF
--- a/deepinv/optim/fixed_point.py
+++ b/deepinv/optim/fixed_point.py
@@ -321,10 +321,19 @@ class FixedPoint(nn.Module):
         if self.anderson_acceleration_config is not None:
             self.init_anderson_acceleration(X)
 
-        for it in tqdm(
-            range(self.max_iter),
+        # ``it`` counts accepted iterations only. A failed backtracking check
+        # rejects the trial iterate and adjusts the stepsize instead, so it is
+        # part of the search for one iteration rather than an iteration in its
+        # own right. Counting it against ``max_iter`` both stops the algorithm
+        # early and hands ``single_iteration`` an index that runs ahead of the
+        # iterate it labels. Consecutive failures remain bounded by
+        # ``backtracking_config.max_iter``, so the loop still terminates.
+        it = 0
+        progress_bar = tqdm(
+            total=self.max_iter,
             disable=(not self.verbose or not self.show_progress_bar),
-        ):
+        )
+        while it < self.max_iter:
             X_prev = X
             X = self.single_iteration(X, it, *args, **kwargs)
 
@@ -337,13 +346,17 @@ class FixedPoint(nn.Module):
                     else None
                 )
 
-                # Convergence check
-                if (
+                # Convergence check, evaluated at the index of the iteration
+                # that has just been accepted.
+                converged = (
                     self.early_stop
                     and self.check_conv_fn is not None
                     and it > 1
                     and self.check_conv_fn(it, X_prev, X)
-                ):
+                )
+                it += 1
+                progress_bar.update(1)
+                if converged:
                     break
 
             else:
@@ -354,9 +367,10 @@ class FixedPoint(nn.Module):
                     if self.verbose:
                         print(
                             f"[Stopping] Reached maximum number of failed backtracking checks "
-                            f"({self.max_iter_backtracking})."
+                            f"({self.backtracking_config.max_iter})."
                         )
                     break
+        progress_bar.close()
 
         return X, metrics
 
--- a/deepinv/tests/test_optim.py
+++ b/deepinv/tests/test_optim.py
@@ -1524,3 +1524,45 @@ def test_sirt(device):
 
         assert sirt.has_converged
         assert x_sirt is not None
+
+
+def test_backtracking_does_not_consume_global_iterations():
+    """Rejected backtracking trials must not count against ``max_iter``.
+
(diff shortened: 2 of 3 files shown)


MohammadSadeghSalehi:
@Tmodrzyk this fixes the backtracking iteration counting you reported in #1272, in case you would like to check it against what you had in mind.

Your example now returns 1.0 rather than 0.0. It also covers the second point you raised but had not tested: `single_iteration` was receiving an index running ahead of the iterate it labels.

One extra thing turned up in the same block. The message printed when the backtracking budget is exhausted interpolates `self.max_iter_backtracking`, which is defined nowhere, so with `verbose=True` that path raised `AttributeError` instead of stopping. Both bugs have a regression test; both fail on main and pass here.

Tmodrzyk:
Hi @MohammadSadeghSalehi,
Thanks for taking the time to investigate this. 
The main issue is already being addressed in #1314, which was linked to #1272 a few weeks before your PR was opened. 
We’re going to close this PR in favor of #1314 to keep things in a single place.

We’re also aware of the additional error you identified, and it is being addressed separately (namely in #1140).
Thanks again for bringing these points to our attention and for your interest in contributing!

PR #1353, opened 2026-08-31
Fixes the incorrect adjoint under the ptychography linear model

ShantanuKodgirwar:
This is a very small PR that fixes the adjoint bug. A probe usually carries phase (experimental case as light passes through lens), so given $p \in \mathbb{C}^n$, it naturally follows $\langle \mathcal F(p \odot x), y\rangle = \langle x, \overline p \odot \mathcal F^{-1}(y)\rangle$. Currently, the code assumes a real probe and never takes care of this and this PR fixes that.  I noticed that by modifying the example deepinv [demo of ptychography](https://deepinv.org/auto_examples/physics/demo_ptychography.html) with a complex probe and also verified it with the [adjointness test](https://deepinv.org/api/stubs/deepinv.physics.LinearPhysics.html#deepinv.physics.LinearPhysics.adjointness_test) from deepinv.

### LLM policy


deepinv/physics/phase_retrieval.py
```diff
@@ -499,7 +499,7 @@ def build_probe(img_size, type="disk", probe_radius=10, device="cpu"):
         x = torch.arange(img_size[1], dtype=torch.float64)
         y = torch.arange(img_size[2], dtype=torch.float64)
         X, Y = torch.meshgrid(x, y, indexing="ij")
-        probe = torch.zeros(img_size, device=device)
+        probe = torch.zeros(img_size, device=device, dtype=torch.complex64)
```
tachella, line 502: Should we set a non zero phase so it catches the complex bug?

tachella (changes requested):

deepinv/physics/phase_retrieval.py
ShantanuKodgirwar: I mean lines 502-503 creates a disk where the inside is ones and outside zeros. So any modification here would alter that disk function for existing use cases. To do as you suggested, we could in principle add a constant phase to the inside part. So something like a pi/4 shift could be induced if we set it to `1 + 1j` instead of `1`. 

However, to avoid any existing (silent) changes to this function, instead I modified the ptychography test to make sure that a complex probe is always initialized. See the latest commit [480c5e6](https://github.com/deepinv/deepinv/pull/1353/commits/480c5e6b5188e1b60681cd751554bcd6e0f740bc) that creates the probe with this function and modifies the imaginary part to non-zero by multiplying probe with `1+1j`. This way, if a real probe is injected, the test would fail. I inspected this already with `uv run pytest deepinv/tests/test_physics.py -k "test_operators_adjointness and ptychography_linear"`.

ShantanuKodgirwar:
@tachella Is there anything else for this PR that must be addressed? Or my earlier comment and this change is sufficient?

tachella (approved):
sorry for the late review, this looks good to go! Many thanks @ShantanuKodgirwar for the bugfix :)

docs/source/changelog.rst
```diff
@@ -35,6 +35,7 @@ Changed
 
 Fixed
 ^^^^^
+- Fix :func: `deepinv.physics.PtychographyLinearOperator.A_adjoint` corrected to consider conjugate of the probe (:gh:`1353` by `Shantanu Kodgirwar`_)
```
tachella:
```suggestion
- Fix :func:`deepinv.physics.PtychographyLinearOperator.A_adjoint` corrected to consider the conjugate of the probe (:gh:`1353` by `Shantanu Kodgirwar`_)
```

tachella (changes requested):
actually I missed a small typo, can you please apply changes?

docs/source/changelog.rst
ShantanuKodgirwar: okay I guess that does it and no worries about the delay :)

tachella:
If tests pass, we can merge :)

ShantanuKodgirwar:
I missed adding author link in the `changelog.rst`. Hopefully the tests pass now.

tachella (approved):

PR #1357, opened 2026-09-01
Add pretrained weights for `FFDNet`

nucli-vicky:
- Add the original `FFDNet` weights, modified from https://github.com/cszn/KAIR/tree/master to fit our implementation
- Fix a bug that makes `FFDNet` init fail when `norm=None`

### LLM policy

### DIFF
--- a/deepinv/models/ffdnet.py
+++ b/deepinv/models/ffdnet.py
@@ -17,23 +17,24 @@ class FFDNet(Denoiser):
 
     The network takes into account the noise level of the input image, which is encoded as an additional input channel.
 
-    Pretrained weights can be downloaded by setting ``pretrained='download'``. They can be initialized as:
+    By default, pretrained grayscale weights are downloaded (``pretrained='download'``). Pretrained weights are
+    also available for RGB images:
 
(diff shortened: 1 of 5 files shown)


deepinv/models/ffdnet.py
tachella: I would set "download" by default, as we do with other denoisers, eg DRUNet

tachella (changes requested):
I have just a minor point, otherwise it should be good to go, thanks a lot :)

deepinv/models/ffdnet.py
nucli-vicky: hmm. I didnt do it because the other default parameters + `pretrained = "download"` would lead to a value error, as default params dont match the pretrained weights right now

nucli-vicky: so either we change the other defaults, so pretrained download default doesnt break, or we keep as is. The first option however would not make user-code backward compatible if they count on the current defaults

tachella: can we set the default parameters to the download ones (I guess this is the best configuration)? 
Even if this is not backwards compatible (we could add a warning)?

nucli-vicky: okay, agree! so default as 3-channels / rgb (greyscale and rgb pretrained weights do not have the same number of filters etc unfortunately)

tachella (approved):
LGTM, thanks

PR #1360, opened 2026-09-01
PIDAL implementation for Poisson inverse problems

sacha-bouchez:
The current ADMM algorithm is not guaranteed to work when the proximal operator of the data fidelity term doesn't have a closed-form solution. By default, deepinv will fall back to gradient descent algorithm to solve the ADMM data fidelity proximal step, which is not guaranteed to work with functions such as Kullback-Leibler divergence. This is the case for Poisson inverse problems.

We propose to implement a splitting strategy that aims to remove $A$ from the proximal operator in order to have a closed-form solution for the latter. We introduce the following scheme proposed in [1]:

$$x^{k+1} = \underset{x}{\text{argmin}}\lbrace \frac{1}{2\gamma}\left\Vert Mx - z^k + u^k\right\Vert^2\rbrace$$
$$z_1^{k+1} = \text{prox}_{\text{KL}(. | y)}(Ax^{k+1} + u_1^k)$$
$$z_2^{k+1} = \text{prox}_{\gamma \lambda g}(x^{k+1} + u_2^k)$$
$$z_3^{k+1} = \text{max}(0, x^{k+1} + u_3^k)$$
$$u^{k+1} = u^k + Mx^{k+1} - z^{k+1}$$

with $M = [A^T, Id, Id]^T$ and $z = [z_1, z_2, z_3]^T$.

We want to add an example with PET.

[1] Figueiredo, M. A. T., & Bioucas-Dias, J. M. (2010). Restoration of Poissonian Images Using Alternating Direction Optimization. IEEE Transactions on Image Processing, 19(12), 3133–3145. https://doi.org/10.1109/tip.2010.2053941

### DIFF
--- a/deepinv/optim/__init__.py
+++ b/deepinv/optim/__init__.py
@@ -28,6 +28,7 @@ from .optimizers import (
     SIRT,
     MLEM,
     OSEM,
+    PIDAL,
 )
 from .fixed_point import FixedPoint
 from .prior import (
@@ -60,6 +61,7 @@ from .optim_iterators import (
     SIRTIteration,
     MLEMIteration,
     OSEMIteration,
+    PIDALIteration,
 )
 from .epll import EPLL
 from .dpir import DPIR
--- a/deepinv/optim/distance.py
+++ b/deepinv/optim/distance.py
@@ -228,9 +228,10 @@ class PoissonLikelihoodDistance(Distance):
         """
         if self.denormalize:
             y = y / self.gain
-        return (-y * torch.log(x / self.gain + self.bkg)).flatten().sum() + (
-            (x / self.gain) + self.bkg - y
-        ).reshape(x.shape[0], -1).sum(dim=1)
+        scaled_x = x / self.gain + self.bkg
+        return (
+            (-y * torch.log(scaled_x) + scaled_x - y).reshape(x.shape[0], -1).sum(dim=1)
+        )
 
     def grad(self, x: torch.Tensor, y: torch.Tensor, *args, **kwargs) -> torch.Tensor:
         r"""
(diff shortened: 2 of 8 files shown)


sacha-bouchez:
Hello @Tmodrzyk, don't forget to commit your (big brain's) fix for that PR ;)

EDIT : It is finally working with homemade changes. The PoissonLikelihood object just doesn't like that the forward operator $A$ is being normalized. I will commit demos soon.

Tmodrzyk:
Ho shoot @sacha-bouchez I did not see the edit to the comment and pushed codex's version first thing in the morning.
Nothing too bad though, you can still revert to your previous version :smile:

sacha-bouchez:
Current implementation uses a homemade trick to solve the x-step in a generic way. Authors of [1] initially used a trick with Fourier transform in the specific context of deblurring. In [2], the algorithm proposed is more generic and is not restricted to deblurring, with an application to emission tomography.

Thus, we should:
- Implement the Fourier trick when the physics define a deconvolution problem.
- Find a more appropriate name (D in PIDAL stands for "deconvolution")
- Rather leave the current implementation as it is or implement the one from [2]. Maybe implementation of [2] requires its own `deepinv.optim` class for the sake of consistentness with sota.

I am interested in other PIDAL enthusiasts thoughts !

[1] Figueiredo, M. A. T., & Bioucas-Dias, J. M. (2009). Deconvolution of Poissonian Images Using Variable Splitting and Augmented Lagrangian Optimization (Version 1). arXiv. https://doi.org/10.48550/ARXIV.0904.4868

[2] Lingenfelter, D. J., & Fessler, J. A. (2010). Augmented Lagrangian methods for penalized likelihood reconstruction in emission tomography. In IEEE Nuclear Science Symposuim &amp; Medical Imaging Conference (pp. 3288–3291). IEEE. 2010 IEEE Nuclear Science Symposium and Medical Imaging Conference (2010 NSS/MIC). https://doi.org/10.1109/nssmic.2010.5874413

Tmodrzyk:
Hi @sacha-bouchez I think it's fine to condition the use of the FFT "trick" on the forward operator being `BlurFFT` or `Blur` with circular boundary conditions, and keep only one algorithm. 
Since the main contribution of PIDAL is not solving the least-squares problem through Fourier (this was already well-known), but rather the specific splitting they use in combination with the augmented Lagrangian, I think this is fine to keep this name :smile:

PR #1374, opened 2026-09-02
More pretrained MRI examples

Andrewwango:

Please read our [contributing guidelines](https://deepinv.org/contributing.html) for instructions on how to contribute, including writing tests, documentation and code style.

### LLM policy


deepinv/models/direct.py
```diff
@@ -0,0 +1,262 @@
...
+        acs_size_h = max(get_acs_size(mask[0, 0, :, w2], h2), 1) // 2
+        acs_size_w = max(get_acs_size(mask[0, 0, h2], w2), 1) // 2
+
+        acs_mask = torch.zeros(y.shape[0], 1, y.shape[2], y.shape[3], 1, dtype=y.dtype)
+        acs_mask[
+            :, :, h2 - acs_size_h : h2 + acs_size_h, w2 - acs_size_w : w2 + acs_size_w
+        ] = 1.0
+
+        # deepinv MRI uses centered FFTs. Models with uncentered operators (Calgary configs) need y pre-shifted into their
+        # convention (a checkerboard modulation, i.e. a half-FOV image shift); centered ones (e.g. vSHARP) take y directly.
```
nucli-vicky: this is already somewhat documented in the docstring I feel like?

deepinv/models/direct_mri.py
nucli-vicky, line 1: I would give this a different filename, e.g. `direct_mri.py`. People browsing through the model files may otherwise think this is a general model file that can be applied to a lot of modalities

deepinv/models/direct.py
```diff
@@ -0,0 +1,262 @@
...
+from deepinv.models.utils import load_state_dict_from_url
+from deepinv.utils.mixins import MRIMixin
+from deepinv.physics.mri import MultiCoilMRI, MRI
+
+
+class DIRECTModel(Reconstructor, MRIMixin):
+    r"""Pretrained DIRECT (NKI-AI) multi-coil MRI reconstruction models.
+
+    Runs a model from `DIRECT <https://github.com/NKI-AI/direct>`_. Calgary-Campinas 12-coil brain models
+    (from `<https://huggingface.co/NKI-AI/direct-calgary-campinas>`_, fixed 5x or 10x acc):
```
nucli-vicky: would personally put quite early in the docstring that these are models to map from k- to image-space. It's documented in the `forward`, but would be better if it was clear at the top of the file

deepinv/models/direct.py
```diff
@@ -0,0 +1,262 @@
...
+
+    - `vsharp_brain` (or `vsharp_cardiac`, `vsharp_knee`, `vsharp_prostate`)
+
+    The wrapped models handle the MRI physics and estimate coil maps themselves.
+
+    .. note::
+        deepinv uses centered FFTs but DIRECT uncentered, so we pre-shift `y` (a checkerboard modulation) into
+        DIRECT's convention. The output scale is not preserved, so its intensity is proportional to but not equal to `y`.
+
+    This model requires DIRECT >=2.2.0 and Python >=3.12. Install it with `pip install deepinv[direct]`.
```
nucli-vicky: would put this as a `.. note::` as well?

deepinv/models/direct.py
```diff
@@ -0,0 +1,262 @@
...
+
+    :param str model_name: model name, see list above.
+    :param bool, str, Path pretrained: `True` or `download` downloads `model_name` weights to `models_dir` (skipped if already present). Or pass checkpoint path directly.
+    :param str, Path models_dir: see above; defaults to torch hub cache (:func:`torch.hub.get_dir`).
```
nucli-vicky: why not default to / just use `deepinv.utils.get_cache_home` ? best to have uniform conventions as to where pretrained weights are saved and how this is controlled.

deepinv/models/direct.py
```diff
@@ -0,0 +1,262 @@
...
+        if pretrained:
+            self.engine.model.load_state_dict(state["model"], strict=False)
+            for name, module in self.engine.models.items():
+                if name in state:
+                    module.load_state_dict(state[name], strict=False)
+
+        self.engine.model.to(device).eval()
+        for m in self.engine.models.values():
+            m.to(device).eval()
+        self.to(device)
```
nucli-vicky: I realize the main priority for the model being added is the demo, but calling `.eval` on everything at `init` seems not ideal for general use.

Not familiar with direct and its engine. From a quick glance, seems like `engine` is not an `nn.Module`.

--> should `DIRECTModel` have its own `.to`, `.eval`, and `.train` methods, so `engine.models` and `engine.model` also get placed / set correctly when these are called? 

And / or alternatively something like 

```python
       self.add_module("engine_model", self.engine.model)
       self.engine_models = nn.ModuleDict(self.engine.models)  
```
to ensure these `nn.Module`s are properly registered as submodules of `DIRECTModel`

deepinv/datasets/calgary.py
```diff
@@ -0,0 +1,110 @@
...
+from typing import Callable
+from deepinv.datasets.fastmri import FastMRISliceDataset, MRISliceTransform
+
+
+class CalgarySliceDataset(FastMRISliceDataset):
+    # TODO: auto-download demo sample for MICCAI tutorial
+    """Dataset for `Calgary-Campinas <https://sites.google.com/view/calgary-campinas-dataset>`_ 12-coil raw brain kspace.
+
+    Thin wrapper over :class:`deepinv.datasets.FastMRISliceDataset` that reads Calgary `h5` volumes,
+    whose only `kspace` array of shape `(num_slices, H, W, 2N)` interleaves the real and imaginary parts of the N coils.
```
nucli-vicky: I find the 'whose only `kspace` array of shape ... interleaves ...' a bit confusing:

maybe something like 'which contains a single `kspace` array'?

May just be me having a bit of interpreting issues, feel free to ignore :)

examples/models/demo_mri_pretrained.py
```diff
@@ -0,0 +1,236 @@
...
+dinv.datasets.download_archive(
+    dinv.utils.get_image_url("CMRxRecon.zip"),
+    dinv.utils.get_cache_home() / "CMRxRecon.zip",
+    extract=True,
+)
+
+dataset = dinv.datasets.CMRxReconSliceDataset(dinv.utils.get_cache_home() / "CMRxRecon")
+
+x, y, params = next(iter(DataLoader(dataset)))
+x, y = x[:, :, 5], y[:, :, 5].to(device)
```
nucli-vicky: 5 looks like a bit of a magic number here?

examples/models/demo_prospective_mri.py
```diff
@@ -0,0 +1,72 @@
+r"""
+Reconstruct prospectively-undersampled raw multicoil MRI
+========================================================
+
+This example reconstructs real prospectively undersampled multicoil brain k-space from Yu et al TODO CITE.
+
+This demonstrates the performance of reconstruction models (:class:`deepinv.models.RAM`) in a deployment scenario rather than a typical simulated (retrospective) scenario.
```
nucli-vicky: --> because there is no corresponding fully sampled scan?

examples/models/demo_prospective_mri.py
```diff
@@ -0,0 +1,72 @@
...
+y = y[..., y.shape[-1] // 2, :, :].to(device)
+
+# %%
+# Build the physics
+# -----------------
+#
+# We recover the prospective sampling mask from the kspace zeros
+# We also estimate coil maps using ESPIRiT.
+# We estimate manually the noise level as `sigma=0.02`. Decreasing it increases the noise in the reconstruction, whereas
+# increasing it increases the smoothness in the reconstruction.
```
nucli-vicky, line 52: It would be nice to demonstrate / show the influence of `sigma` by plotting the reconstructions with a set of different `sigma`s

examples/models/demo_prospective_mri.py
```diff
@@ -0,0 +1,72 @@
...
+coil_maps = dinv.physics.MultiCoilMRI.estimate_coil_maps(
+    y, calib_size=24, espirit_crop=0.99
+)  # (1, N, H, W) complex
+
+physics = dinv.physics.MultiCoilMRI(
+    mask=mask,
+    coil_maps=coil_maps,
+    device=device,
+    noise_model=dinv.physics.GaussianNoise(sigma=0.02),
+)
```
nucli-vicky, line 64: what is the reason we choose `GaussianNoise` over `RicianNoise`? Purely because `RAM` was trained only with Poisson and Gaussian noise?

examples/models/demo_prospective_mri.py
```diff
@@ -0,0 +1,72 @@
...
+    dinv.utils.get_cache_home() / "mridata" / "prospective_t2.h5", ifft_slice_dim=True
+)
+y = y[..., y.shape[-1] // 2, :, :].to(device)
+
+# %%
+# Build the physics
+# -----------------
+#
+# We recover the prospective sampling mask from the kspace zeros
+# We also estimate coil maps using ESPIRiT.
```
nucli-vicky: to do: cite ESPIRiT here

examples/models/demo_prospective_mri.py
```diff
@@ -0,0 +1,72 @@
...
+# Reconstruct with RAM
+# --------------------
+
+model = dinv.models.RAM(device=device, pretrained=True)
+
+with torch.no_grad():
+    x_zf = physics.A_adjoint(y)
+    physics.phase_correct_maps(x_zf)
+    x_ram = model(y / x_zf.max(), physics) * x_zf.max()
+    x_sense = physics.A_dagger(y)
```
nucli-vicky: what does `physics.phase_correct_maps` do here? It modifies the state of the physics, but may be unclear to readers.

examples/models/demo_prospective_mri.py
```diff
@@ -0,0 +1,72 @@
...
+
+dinv.utils.plot([x_zf, x_sense, x_ram], titles=["Zero-filled", "SENSE", "RAM"])
```
nucli-vicky: cite SENSE + we need to explain it a little bit as well maybe?

nucli-vicky: at least the fact that this uses least-squares / iterative method under the hood

examples/models/demo_prospective_mri.py
nucli-vicky: I get why this is interesting. But it also makes comparison a bit more difficult, especially 'why should u use deep-learning methods for this data'

nucli-vicky (commented):
Thanks! @Andrewwango 

An initial set of comments / questions / thoughts :)

deepinv/datasets/calgary.py
Andrewwango: sorry, it was leftover from LLM slop

deepinv/models/direct.py
Andrewwango: well spotted! Fixed

examples/models/demo_prospective_mri.py
Andrewwango: rician noise is defined on the magnitude images I think? I.e. it's useful for magnitude-image denoising etc.

Here, we're on the complex data such that y = Ax + e where e~Normal

examples/models/demo_prospective_mri.py
Andrewwango: added fista+wavelet recon

deepinv/datasets/calgary.py
```diff
@@ -0,0 +1,131 @@
...
+class CalgarySliceDataset(FastMRISliceDataset):
+    """Dataset for `Calgary-Campinas <https://sites.google.com/view/calgary-campinas-dataset>`_ 12-coil raw brain kspace.
+
+    Loads Calgary `h5` volumes of shape `(num_slices, H, W, 2N)`, where slice dim is in image domain and `H,W` is kspace.
+    The dataset loads and preprocesses all kspace slices per volume, of shape `(2, N, H, W)`. These are fully-sampled for train/val volumes and masked for the test set.
+
+    Also computes the GT `x`, the magnitude root-sum-square reconstructions of shape `(1, H, W)`, or `torch.nan` for the masked test set.
+
+    The dataset is loaded as tuples `(x, y, params)`, where
+    `params` optionally contains the sampling `mask` and, if desired, estimated `coil_maps`.
```
romainvo, line 15: I think we should mention that the dict format should be preferred? (avoids painful refactoring in the future when we fully deprecate)

deepinv/models/direct_mri.py
```diff
@@ -0,0 +1,258 @@
...
+    - `iterdualnet_5x` (or `_10x`), as above
+    - `kikinet_5x` (or `_10x`), as above
+    - `lpdnet_5x` (or `_10x`), as above
+    - `unet_5x` (or `_10x`), as above
+    - `xpdnet_5x` (or `_10x`), as above
+    - `multidomainnet` (downloaded from https://files.aiforoncology.nl/direct-project, repaired locally, uploaded to https://huggingface.co/Andrewwango/direct), as above
+    - `vsharp_brain` :footcite:p:`yiasemis2024vsharp,yiasemis2025uniform`, trained on a mix of MRI datasets (including brain, cardiac, knee and prostate), downloaded from `here <https://huggingface.co/NKI-AI/direct-uniform>`__
+    - `vsharp_cardiac`: identical model as above, but for cardiac inference
+    - `vsharp_knee`, likewise
+    - `vsharp_prostate`, likewise
```
romainvo, line 33: Can avoid having the "as above" and "likewise" repeated n times? we could do bold sections like this, not sure how it renders though but you get the idea. Also we're providing an deepinv interface so I think we should provide citations for each of the model if possible.

```suggestion
    **Models trained on Calgary-Campinas 12-coil brain** (downloaded from `here <https://huggingface.co/NKI-AI/direct-calgary-campinas>`__)
    - `jointicnet_5x` (or `_10x`) :footcite:p:`jun2021joint`
    - `recurrentvarnet_5x` (or `_10x`),
    - `varnet_5x` (or `_10x`),
    - `conjgradnet_5x` (or `_10x`),
    - `iterdualnet_5x` (or `_10x`),
    - `kikinet_5x` (or `_10x`),
    - `lpdnet_5x` (or `_10x`),
    - `unet_5x` (or `_10x`),
    - `xpdnet_5x` (or `_10x`),
    - `multidomainnet` (downloaded from https://files.aiforoncology.nl/direct-project, repaired locally, uploaded to https://huggingface.co/Andrewwango/direct),

    **Models trained on a mix of MRI datasets (including brain, cardiac, knee and prostate)** (downloaded from `here <https://huggingface.co/NKI-AI/direct-uniform>`__)
    - `vsharp_brain` :footcite:p:`yiasemis2024vsharp,yiasemis2025uniform`,
    - `vsharp_cardiac`,
    - `vsharp_knee`, 
    - `vsharp_prostate`
```

I think the names of the `vsharp` models are self-explanatory?

pyproject.toml
```diff
@@ -216,8 +229,8 @@ deepinv = { path = ".", editable = true}
 default   = { features = ["gpu"], solve-group = "gpu" }
 test      = { features = ["gpu", "test"], solve-group = "gpu" }
 full      = { features = ["gpu", "doc", "test", "dataset", "denoisers", "physics", "physics-gpu", "training"], solve-group = "gpu" }
-docs      = { features = ["gpu", "doc", "dataset", "denoisers", "physics", "physics-gpu", "training", "py312"], solve-group = "gpu-docs" }
-doctest   = { features = ["gpu", "doc", "test", "dataset", "denoisers", "physics", "physics-gpu", "training", "py312"], solve-group = "gpu-docs" }
+docs      = { features = ["gpu", "doc", "dataset", "denoisers", "physics", "physics-gpu", "training", "py312", "direct"], solve-group = "gpu-docs" }
+doctest   = { features = ["gpu", "doc", "test", "dataset", "denoisers", "physics", "physics-gpu", "training", "py312", "direct"], solve-group = "gpu-docs" }
```
romainvo, line 233: Why is "direct" added to these two environments only? `test_models.py::test_direct_model` will always be skipped in the current config.

Similarly, shouldn't `direct-recon` work fine on cpu envs, i.e. `full-cpu`, `docs-cpu` and `doctest-cpu`?

deepinv/utils/io.py
```diff
@@ -517,3 +517,98 @@ def load_blosc2(
...
+    Only the first average, contrast, phase, repetition and set is kept. Noise scans are dropped.
+
+    .. note::
+
+        Readout oversampling is removed by default, since scanners oversample the (fully-sampled) readout (typically 2x) to avoid frequency-encode aliasing,
+        which doubles the readout FOV and is redundant for reconstruction.
+
+    :param str filename: path to the ISMRMRD `.h5` file.
+    :param bool ifft_slice_dim: if `True`, inverse FFT the slice/depth axis D so it is returned in image space, ready to index slices. Note for 2D it is ignored.
+    :return: (:class:`torch.Tensor`) real tensor of shape `(1, 2, N, D, H, W)`, where N = num coils.
```
romainvo, line 546: No need to specify the return type here, sphinx produces a "Return type: <type>" inset using the type hint in the function signature.

examples/models/demo_mri_pretrained.py
```diff
@@ -0,0 +1,275 @@
...
+    model_name="vsharp_cardiac", pretrained=True, device=device
+)
+ram = dinv.models.RAM(device=device, pretrained=True)
+
+with torch.no_grad():
+    x_zf = physics.A_adjoint(y).cpu()
+    x_vsharp = vsharp(y, physics).cpu()
+
+    # Although RAM is scale-equivariant, we bring y into friendlier scale (currently it's very small)
+    x_ram = ram(y / x_zf.max(), physics).cpu() * x_zf.max()
```
romainvo, line 83: Curious, you mean it does not work without this trick?

deepinv/datasets/calgary.py
```diff
@@ -0,0 +1,131 @@
...
+    """Extract params and estimate coil maps for Calgary raw data.
+
+    To be used with :class:`deepinv.datasets.CalgarySliceDataset`.
+    """
+
+    def __call__(self, target, kspace, seed=None, metadata=None, **kwargs):
+        params = {
+            "mask": (
+                self.to_torch_complex(kspace.unsqueeze(0)).abs().sum(1) > 0
+            ).float()
```
romainvo, line 127: The mask estimation here makes sense only for the test set right? I know it's documented in the docstring of CalgarySliceDataset but I would mention it again in this one's docstring just to make it clear

deepinv/physics/mri.py
```diff
@@ -495,6 +495,25 @@ def estimate_coil_maps(
 
         return torch_maps
 
+    def phase_correct_maps(self, x: torch.Tensor, smooth=0.05):
+        """Re-gauge coil maps to the smooth phase of some reasonably smooth-phased image.
```
romainvo, line 499: Would be nice to have a citation motivating this pre-processing?

examples/models/demo_mri_pretrained.py
```diff
@@ -0,0 +1,275 @@
...
+ram = dinv.models.RAM(device=device, pretrained=True)
+
+with torch.no_grad():
+    x_zf = physics.A_adjoint(y).cpu()
+    x_sense = physics.A_dagger(y).cpu()
+    x_vsharp = vsharp(y, physics).cpu()
+    x_jointicnet = jointicnet(y, physics).cpu()
+
+    coil_maps = physics.phase_correct_maps(x_zf)
+    physics.update(coil_maps=coil_maps)
```
romainvo, line 160: Since this preprocessing step is not automatic, I think it would be nice to show results with and without

deepinv/models/direct_mri.py
```diff
@@ -0,0 +1,258 @@
...
+            if not self._sens_in_forward:
+                sample["sensitivity_map"] = self.engine.compute_sensitivity_map(
+                    sample["sensitivity_map"]
+                )
+            x, _ = self.engine.forward_function(sample)
+
+        if isinstance(x, (list, tuple)):  # vSHARP returns one image per unrolled step
+            x = x[-1]
+
+        x *= sample["scaling_factor"].view(-1, *([1] * (x.ndim - 1)))
```
romainvo: Curious but what does the scaling_factor corresponds to ?

deepinv/models/direct_mri.py
```diff
@@ -0,0 +1,258 @@
...
+                "NKI-AI/direct-uniform",
+                "uniform_vsharp.pt",
+                f"uniform_{model_name[len('vsharp_'):]}.yaml",
+            )
+        elif model_name.startswith("multidomainnet"):
+            repo, weights_file, cfg_file = (
+                "Andrewwango/direct",
+                f"{model_name}.pt",
+                f"{model_name}.yaml",
+            )
```
romainvo, line 93: Shouldn't we put these weights on the deeping HF?

deepinv/models/direct_mri.py
```diff
@@ -0,0 +1,258 @@
...
+        )
+        self.engine = setup_engine(
+            cfg,
+            str(device),
+            model,
+            additional,
+            forward_operator=forward_operator,
+            backward_operator=self.backward_operator,
+            mixed_precision=False,
+        )
```
romainvo, line 134: self.engine is not a `nn.Module` right? this `__init__` function looks fairly complicated too me and tough to maintain, I would maybe add a few comments to indicate the type of key objects, e.g. here it's an `Engine` class which is a custom direct object. I like comments in code so I'm a bit biased ahah

deepinv/tests/test_models.py
```diff
@@ -1976,3 +1976,27 @@ def test_srresnet_inputs():
...
+    coil_maps /= coil_maps.abs().pow(2).sum(1, keepdim=True).sqrt() + 1e-8
+    mask = dinv.physics.generator.GaussianMaskGenerator(
+        img_size, acceleration=4, device=device
+    ).step()["mask"]
+    physics = dinv.physics.MultiCoilMRI(
+        img_size=img_size, mask=mask, coil_maps=coil_maps, device=device
+    )
+    y = physics(torch.randn(1, 2, *img_size, device=device))
+
+    model = dinv.models.DIRECTModel("unet_5x", pretrained=True, device=device)
```
romainvo, line 2000: I think we should test for every model specified in the docstring under "Available models", what do you think? The routing in the init of `DIRECTModel` and even the forward is fairly complicated, I fear that we end up with buggy stuff if not tested exhaustively..

deepinv/models/direct_mri.py
```diff
@@ -0,0 +1,258 @@
...
+        # Forward pass
+        sample = {
+            "masked_kspace": kspace,
+            "sampling_mask": mask.to(y.dtype).unsqueeze(-1),
+            "acs_mask": acs_mask,
+        }
+        if kspace.shape[1] == 1:
+            sample["sensitivity_map"] = torch.zeros_like(kspace).index_fill_(
+                -1, torch.tensor([0]), 1.0
+            )
```
romainvo: Is this standard? For single-coil the sensitivity of the imaginary component is zero?

deepinv/models/direct_mri.py
```diff
@@ -0,0 +1,258 @@
...
+        else:
+            sample = EstimateSensitivityMapModule(
+                kspace_key="masked_kspace",
+                backward_operator=self.backward_operator,
+                gaussian_sigma=0.7,
+            )(sample)
+        sample = ComputeScalingFactorModule(
+            normalize_key="masked_kspace", percentile=0.99
+        )(sample)
+        sample = NormalizeModule(keys_to_normalize=["masked_kspace"])(sample)
```
romainvo, line 236: This is quite opaque, I would add a small comment explaining what it does:
- ComputeScalingFactorModule returns the 99th percentile of "masked_space" magnitudes for each batch element
- NormalizedModule applies the scaling.

Or maybe it's obvious, as you want

deepinv/models/direct_mri.py
```diff
@@ -0,0 +1,258 @@
...
+        if kspace.shape[1] == 1:
+            sample["sensitivity_map"] = torch.zeros_like(kspace).index_fill_(
+                -1, torch.tensor([0]), 1.0
+            )
+        else:
+            sample = EstimateSensitivityMapModule(
+                kspace_key="masked_kspace",
+                backward_operator=self.backward_operator,
+                gaussian_sigma=0.7,
+            )(sample)
```
romainvo, line 232: Should we expose the `type_of_map` argument of `EstimateSensivityMapModule` (https://github.com/NKI-AI/direct/blob/490a14631e6a252690c3356e0093dff3f884e12f/direct/data/mri_transforms.py#L1140) and give the users the option to choose between "rss_estimate" or "espirit" (e.g. could be an instance attribute)

deepinv/models/direct_mri.py
```diff
@@ -0,0 +1,258 @@
...
+        sample = {
+            k: v.to(device) if torch.is_tensor(v) else v for k, v in sample.items()
+        }
+
+        with torch.no_grad():
+            if not self._sens_in_forward:
+                sample["sensitivity_map"] = self.engine.compute_sensitivity_map(
+                    sample["sensitivity_map"]
+                )
+            x, _ = self.engine.forward_function(sample)
```
romainvo, line 247: This means the user can never trained or finetune the model? Seems quite restrictive, are there constraints I am missing?

romainvo (changes requested):
Thanks @Andrewwango for these nice additions! The DIRECTModel interface is very cool, it really expands the range of sota models available!!

Remarks in my comments, I haven't done a full pass on the examples, maybe other comments to come when the docs build (use_dict_output stuff I missed in the previous PR :/ )

deepinv/utils/io.py
Andrewwango: ooh nice

examples/models/demo_mri_pretrained.py
Andrewwango: yes :/ 

I think because the intensity is so low, it gets floored by the minimum sigma level the model was trained on. i.e. breaking scale equivariance

examples/models/demo_mri_pretrained.py
```diff
@@ -0,0 +1,275 @@
...
+    # Although RAM is scale-equivariant, we bring y into friendlier scale (currently it's very small)
+    x_ram = ram(y / x_zf.max(), physics).cpu() * x_zf.max()
+
+dinv.utils.plot(
+    {"Fully-sampled": x, "Zero-filled": x_zf, "vSHARP": x_vsharp, "RAM": x_ram},
+    subtitles=[
+        f"{metric(x).item():.1f}",
+        f"{metric(x_zf).item():.1f}",
+        f"{metric(x_vsharp).item():.1f}",
+        f"{metric(x_ram).item():.1f}",
```
romainvo: Add labels to the plots (Sharpness index), at least to the first column if we don't want to repeat it 4 times

examples/models/demo_mri_pretrained.py
```diff
@@ -0,0 +1,275 @@
...
+        "vSHARP": x_vsharp,
+        "Joint-ICNet": x_jointicnet,
+        "RAM": x_ram,
+    },
+    subtitles=[
+        f"{metric(x_zf).item():.1f}",
+        f"{metric(x_sense).item():.1f}",
+        f"{metric(x_vsharp).item():.1f}",
+        f"{metric(x_jointicnet).item():.1f}",
+        f"{metric(x_ram).item():.1f}",
```
romainvo: Same comment, add labels

examples/models/demo_mri_pretrained.py
```diff
@@ -0,0 +1,275 @@
...
+        "SENSE": x_sense,
+        "vSHARP": x_vsharp,
+        "RAM": x_ram,
+    },
+    subtitles=[
+        f"{metric(x).item():.1f}",
+        f"{metric(x_zf).item():.1f}",
+        f"{metric(x_sense).item():.1f}",
+        f"{metric(x_vsharp).item():.1f}",
+        f"{metric(x_ram).item():.1f}",
```
romainvo: Add labels

romainvo (changes requested):
Very few additional comments about labels in the docs

deepinv/physics/mri.py
Andrewwango: Sorry I don;t have one!

deepinv/models/direct_mri.py
Andrewwango: probs

Andrewwango: done

examples/models/demo_mri_pretrained.py
Andrewwango: good idea, done

pyproject.toml
Andrewwango: done, let's see if it works. Summary:
- GPU doc/doctest: kept at current solution, since GPU doc is already at py3.12
- CPU doc/doctest: ignore (example is too heavy to run on CPU anyway)
- GPU test / CPU test: add extra group for DIRECT + all future deps that require py>=3.12. In test_cpu.yml and test_gpu.yml, add an extra action to test tests that contain these extra dependencies

deepinv/models/direct_mri.py
Andrewwango: added more comments

deepinv/models/direct_mri.py
Andrewwango: hmm, not sure why one would want to do rss_estimate, it will invariably lead to worse model performance. If they want RSS, they can just input the RSS themselves

deepinv/models/direct_mri.py
Andrewwango: Not sure but I get the impression it wil add complication. I added a warning to the docstring suggesting to add an issue if needed

examples/models/demo_mri_pretrained.py
Andrewwango: done

pyproject.toml
```diff
@@ -177,8 +177,8 @@ python = "3.12.*"
 [tool.pixi.feature.direct.pypi-dependencies]
 direct-recon = ">=2.2.0"
 
-# Override DIRECT pins since it works fine on older torch
-[tool.pixi.feature.direct.pypi-options.dependency-overrides]
+# Override DIRECT pins since it works fine on older torch. Note: GPU-only
+[tool.pixi.feature.direct-gpu.pypi-options.dependency-overrides]
 torch = { version = "*", index = "https://download.pytorch.org/whl/cu124" }
 torchvision = { version = "*", index = "https://download.pytorch.org/whl/cu124" }
```
romainvo: I don't think we need this, it's exactly the same dependency specified in the `gpu` feature, can we try resolving the environments on the CI without this?

pyproject.toml
```diff
@@ -229,8 +229,11 @@ deepinv = { path = ".", editable = true}
 default   = { features = ["gpu"], solve-group = "gpu" }
 test      = { features = ["gpu", "test"], solve-group = "gpu" }
 full      = { features = ["gpu", "doc", "test", "dataset", "denoisers", "physics", "physics-gpu", "training"], solve-group = "gpu" }
-docs      = { features = ["gpu", "doc", "dataset", "denoisers", "physics", "physics-gpu", "training", "py312", "direct"], solve-group = "gpu-docs" }
-doctest   = { features = ["gpu", "doc", "test", "dataset", "denoisers", "physics", "physics-gpu", "training", "py312", "direct"], solve-group = "gpu-docs" }
+docs      = { features = ["gpu", "doc", "dataset", "denoisers", "physics", "physics-gpu", "training", "py312", "direct", "direct-gpu"], solve-group = "gpu-docs" }
+doctest   = { features = ["gpu", "doc", "test", "dataset", "denoisers", "physics", "physics-gpu", "training", "py312", "direct", "direct-gpu"], solve-group = "gpu-docs" }
+# specific gpu env for dependencies that need py>=3.12 (GPU-only for now)
+min-py312-test     = { features = ["gpu", "test", "py312", "direct", "direct-gpu"], solve-group = "gpu-min-py312" }
```
romainvo: I think if we add `direct` as dependency we need to make sure that it integrates with the the `full` dependencies when python version is pinned to 3.12.*, so I would do something more like

```yaml
full-py312 = { features = ["gpu", "doc", "test", "dataset", "denoisers", "physics", "physics-gpu", "training", "py312", "direct"], solve-group = "gpu-py312" }
```

Also if we agree this is temporary, I think we can add small note in the comment indicating that https://github.com/NKI-AI/direct/pull/328 should solve our current issue and that this should be removed asap? Or we open an issue and then it's even more clear

romainvo (changes requested):
Thanks for clarifying the docs and adding the helpful comments in the code! Remarks only on how to manage the additional dependency.

pyproject.toml
Andrewwango: It's still needed unfortunately :( because it forces pixi to ignore the DIRECT torch pin. Otherwise, pixi will never ignore the DIRECT pin -> problems. This forces it to ignore

Andrewwango: I think it's inevitable to have things like this since we can never constrain the requirements of the optional dependencies and expect them to have the same ethos as us (i.e. minimal pins)

pyproject.toml
Andrewwango: agreed, done. Also added a comment in both pyproject below, and in test_gpu

romainvo (approved):
Thanks @Andrewwango for the last updates! Happy to merge once tests pass

romainvo (approved):

Issue #1224, opened 2026-06-08
Missing cleanup of zip file in BSDS500


Zeppo1994:
### Environment

All

### Describe the bug

We are basically keeping an unnecessary zip file alive.

### Steps to reproduce

-

### Expected behavior

We should change this part of the code:

zip_path = os.path.join(self.base_path, "download.zip")
if download and not os.path.exists(zip_path):   # zip = sentinel
    download_archive(...)
    extract_zipfile(zip_path, self.base_path)
    # ← os.remove(zip_path) missing

if not download and not os.path.exists(zip_path):  # also checked here
    raise NameError(...)

The not download guard is what forces that zip must stay — removing it would make the "dataset not found" check fail on the next run. The most principled fix would be to switch sentinel to BSDS500-master (this is what it extracts to I think) and to remove zip after extract.

jscanvic:
Welcome and thank you for raising this issue!

I agree it would be best to avoid cluttering the disk that way.

A few comments:
- This bug affects other datasets including DIV2K
- torchvision maintainers chose to do this kind of clean up for some datasets and not others, e.g., EMNIST is cleaned up https://docs.pytorch.org/vision/main/_modules/torchvision/datasets/mnist.html#MNIST but not CIFAR10 (the relevant flag of download_and_extract_archive is left unset) https://docs.pytorch.org/vision/main/_modules/torchvision/datasets/cifar.html#CIFAR10
- I'd argue it's best to do the cleanup logic inside the "should we download it" conditional instead of always deleting it if it's detected

Zeppo1994:
The issue is larger than I first thought then and we should indeed handle all datasets in the same way.

For me, the cleanest solution would be to drop the download option entirely, and to check internally if the correct folder exists in root. If this is not the case, it will resort to download and clean up the zipfile directly.

As far as I can tell, most of our datasets should be public and do not need specific login data. Still, one would need to decide how to handle those appropriately.

jscanvic:
Agreed we should treat all the datasets the same ideally

I wouldn't force auto-downloading though - on HPC compute nodes there is typically no internet access and downloading typically takes time before failing

PR #1382, opened 2026-09-05
Add Set5, BSD100, McMaster, and Kodak24 datasets

nucli-vicky:
- Add Set5, BSD100, McMaster, Kodak24
- Add an optional `password` argument to `extract_zipfile`. Some datasets have a password for the zip, even though the password is publicly documented on the source's html (see https://www4.comp.polyu.edu.hk/~cslzhang/CDM_Dataset.htm)

These datasets are all used often as test sets. For instance, `SwinIR` reports performance on all of these in their papers.

Making these available saves users time trying to source them, and makes plugging them into our existing framework zero-cost to them. Additionally, these datasets may be useful to the benchmarks project.

**Questions:**

- *There is a lot of duplicate code between these classes. This was already the case for our other natural imaging datasets, but this PR intensifies the code duplication. A new parent class or some extra functions in `utils.py` to handle e.g. all the downloading and extracting logic after `if not self.check_dataset_exists()` may reduce duplicate code. We can handle this in this PR, or in another one. Thoughts? Same argument can probably be made for tests of these datasets* ==> deferred to a later PR

- We might be able to address #1224 in this PR. Have I interpreted correctly the agreed upon solution is to delete the zip only when we downloaded and extracted the zip in that runtime of the `init`? @jscanvic @Zeppo1994

### LLM policy

### DIFF
--- a/deepinv/datasets/bsd100.py
+++ b/deepinv/datasets/bsd100.py
@@ -29,7 +29,6 @@ class BSD100HR(ImageFolder):
                 |
                 --- xxx
 
-    This dataset wrapper gives access to the 100 high resolution images in the `BSD100_HR` folder.
     Raw dataset source : https://huggingface.co/datasets/eugenesiow/BSD100
 
     :param str root: Root directory of dataset. Directory path from where we load and save the dataset.
--- a/deepinv/datasets/bsd100.py
+++ b/deepinv/datasets/bsd100.py
@@ -1,0 +1,131 @@
+from typing import Callable
+from types import MappingProxyType
+import os
+
+from deepinv.datasets.utils import (
+    calculate_md5_for_folder,
+    download_archive,
+    extract_tarball,
+)
+from deepinv.datasets.base import ImageFolder
+from .utils import resolve_root
+
+
+class BSD100HR(ImageFolder):
+    """Dataset for `BSD100 <https://paperswithcode.com/dataset/bsd100>`_.
+
+    The BSD100 dataset :footcite:p:`martin2001database` is a dataset consisting of 100 images commonly used for testing performance of image reconstruction algorithms.
+    Images have sizes ranging from 240×160 to 480×320 pixels.
+
+    **Raw data file structure:** ::
+
+        self.root --- BSD100_HR.tar.gz
+                |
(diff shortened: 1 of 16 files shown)


deepinv/datasets/bsd100.py
```diff
@@ -0,0 +1,114 @@
...
+                    print("Dataset has been successfully downloaded.")
+                else:
+                    raise ValueError("There is an issue with the data downloaded.")
+            # stop the execution since the dataset is not available and we didn't download it
+            else:
+                raise RuntimeError(
+                    f"Dataset not found at `{self.root}`. Please set `root` correctly (currently `root={self.root}`) OR set `download=True` (currently `download={download}`)."
+                )
+
+        # Initialize ImageFolder
```
Andrewwango: Remove

deepinv/datasets/mcmaster.py
```diff
@@ -0,0 +1,120 @@
...
+)
+from deepinv.datasets.base import ImageFolder
+from .utils import resolve_root
+
+
+class McMaster(ImageFolder):
+    """Dataset for `McMaster <https://www4.comp.polyu.edu.hk/~cslzhang/CDM_Dataset.htm>`_.
+
+    The McMaster dataset :footcite:p:`zhang2011color` is a dataset consisting of 18 images commonly used for testing performance of
+    color demosaicking and image reconstruction algorithms.
```
Andrewwango: Our convention is "demosaicing"

Andrewwango (approved):
LGTM, minor comments

deepinv/datasets/bsd100.py
nucli-vicky: agree, I took this over from all the other datasets (set14, div2k, lsdir, urban100 all have `# Initialize ImageFolder`).

Do I remove this everywhere then?

deepinv/datasets/mcmaster.py
nucli-vicky: okay. i noticed LSDIR docstring also uses 'demosaicking', will replace it there too

nucli-vicky:
Thanks @Andrewwango ! I have addressed your comments :)

PR #1393, opened 2026-09-08
Minor deprecation warnings clean-up

jscanvic:
1. I use `compute_sqnorm` instead of the deprecated `compute_norm` with default `squared=True`
2. Rewrite sigma setting without traversing dir(physics) to avoid traversing `physics.device` which causes warnings

Towards #1007

### DIFF
--- a/deepinv/tests/test_models.py
+++ b/deepinv/tests/test_models.py
@@ -1173,17 +1173,18 @@ def test_restoration_models(
 
     sigma = 0.02
 
-    # Recursively set the noise model sigma in the physics (and sub-physics)
-    for p in physics.modules():
-        if not isinstance(p, dinv.physics.Physics) or not hasattr(p, "noise_model"):
-            continue
+    if physics is not None:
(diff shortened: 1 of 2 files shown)


deepinv/tests/test_models.py
```diff
@@ -1171,27 +1171,21 @@ def test_restoration_models(
...
-            pass
-
     sigma = 0.02
-    _set_sigma_physics(physics, sigma)
+
+    # Recursively set the noise model sigma in the physics (and sub-physics)
+    for p in physics.modules():
+        # Filter out sub-modules that are not sub-physics and sub-physics
+        # without a noise model
+        if not isinstance(p, dinv.physics.Physics) or not hasattr(p, "noise_model"):
```
PAUL-BERNARD:
```suggestion
        if not isinstance(p, dinv.physics.Physics) or not hasattr(p, "noise_model"):
```
The code is explicit enough

PAUL-BERNARD (approved):

PR #1395, opened 2026-09-08
Fix inverse Homography

### LINKED ISSUE #1391: Incorrect inverse in Homography
### Environment

Latest

### Describe the bug

It looks like the inverse in Homography is incorrect

The forward is basically implemented as a series of elementary transformations (rotations, translations, ...), which means the true inverse should either be 1) the inverse of each elementary transformation applied in the reverse order, or 2) the elementary transforms applied in the original order but with inverse params that are no longer independent and depend on each other.

At the moment, it seems to be doing an incorrect mix of the two: keeping the forward order and inverting each elementary transform independently of the others.

It seems to be one of the reasons `test_transform_identity` is unstable and depends heavily on the seed #1238.

**Fix**

The cleanest fix would arguably be to stick with 2) and use the trickier formula for parameter inversion in homographies.

**Offending code**

The different parameters are inverted independently from each other without the pipeline order being inverted.

https://github.com/deepinv/deepinv/blob/596f13fe76df0883853bbee245e79a15a6254113/deepinv/transform/projective.py#L232-L247

### Steps to reproduce

**Reproducing**

The two resulting images should show a square at the same location but the second one is located lower than it should.

```python
import deepinv as dinv
import torch

device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")

img_size = (1, 3, 32, 32)

mask = torch.ones(img_size, device=device)
mask[..., :16] = 0.001

transform = dinv.transform.projective.Homography(
    theta_max=5,
    theta_z_max=20,
    zoom_factor_min=0.85,
    shift_max=0.2,
    skew_max=5,
    x_stretch_factor_min=0.85,
    y_stretch_factor_min=0.85,
    padding="zeros",
    interpolation="bicubic",
    device=device,
    rng=torch.Generator(device=device).manual_seed(0)
)

x = torch.zeros(1, 3, 256, 256, device=device)
h, w = 40, 40
x[..., h : h + 30, w : w + 30] = 1

x_t = transform.identity(x)

dinv.utils.plot([ x, x_t ], titles=["$x$", "$T_g^{-1} T_g x$"], figsize=(8, 4))

```

### Expected behavior

/

_No response_

### Additional context

_No response_

jscanvic:
Fixes #1391 

- Override `inverse` similarly to `ChainTransform` to have the right inverse
- Add a warning that `invert_params` can't be used as is with `transform`

I checked that it makes the example from the issue work and also that it partially fixes the instability of `test_transform_identity` (there's another separate issue and I'll wrap this up in a separate PR)

### DIFF
--- a/deepinv/transform/projective.py
+++ b/deepinv/transform/projective.py
@@ -3,6 +3,7 @@ from dataclasses import dataclass
 
 from typing import Iterable
 
+import warnings
 import numpy as np
 import torch
 from PIL import Image
@@ -10,7 +11,7 @@ from PIL import Image
 from deepinv.transform.base import Transform, TransformParam
 
 
-def rotation_matrix(tx: float, ty: float, tz: float) -> np.ndarray:
+def rotation_matrix(tx: float, ty: float, tz: float, inverse: bool = False) -> np.ndarray:
     """Numpy implementation of ``scipy`` rotation matrix from Euler angles.
 
     Construct 3D extrinsic rotation matrix from x, y and z angles. This is equivalent of using the ``scipy`` function:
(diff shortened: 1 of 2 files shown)


Andrewwango (approved):
nice, thanks Jeremy

Issue #1401, opened 2026-09-10
Utils load_dataset left out of dataset format migration


jscanvic:
Function `deepinv.utils.demo.load_dataset` instantiates `ImageFolder` with `use_dict_output` always left as default

Following the migration in #1244 it should be made to either use use_dict_output by default it turns out not to be breaking, or a param use_dict_output should be added to `load_dataset` and passed to `ImageFolder`

In particular, this causes deprecation warnings in the CI (see #1007)

https://github.com/deepinv/deepinv/blob/cf1c1495de35b42b8e84070eac33081d6644d8d8/deepinv/utils/demo.py#L145

Additionally, `PlaceholderDataset` should also be migrated

PR #1402, opened 2026-09-10
Filter some warnings about new dataset format

jscanvic:
The new dataset format migration resulted in some deprecation warnings in the CI. In this PR, I filter part of them using pytest. The remaining ones are trickier, I opened a separate issue for that #1401

Towards #1007

### DIFF
--- a/deepinv/tests/test_trainer.py
+++ b/deepinv/tests/test_trainer.py
@@ -118,21 +118,6 @@ def get_dummy_physics_generator(rng, device):
     return DummyPhysicsGenerator(rng=rng, device=device)
 
 
-@pytest.fixture
-def use_dict_output(request):
-    # Catch deprecation warnings from previous dataset format
-    _use_dict_output = request.param
-    with (
-        pytest.warns(
-            DeprecationWarning,
-            match="The tuple format for dataset outputs is deprecated",
-        )
-        if not _use_dict_output
-        else contextlib.nullcontext()
-    ):
-        yield _use_dict_output
-
-
 @pytest.mark.parametrize(
     "use_physics_generator", [None, "param", "noise", "param+noise"]
 )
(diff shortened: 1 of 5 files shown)


deepinv/tests/test_trainer.py
```diff
@@ -118,12 +118,27 @@ def step(self, batch_size=1, seed=None, **kwargs):
...
+    _use_dict_output = request.param
+    with (
+        pytest.warns(
+            DeprecationWarning,
+            match="The tuple format for dataset outputs is deprecated",
+        )
+        if not _use_dict_output
+        else contextlib.nullcontext()
+    ):
+        yield _use_dict_output
```
romainvo: Very nice, so much better than what I did in the migration PR ahah. It could come as a replacement for all the code duplication for catching warnings, right? The `with dataset_output_context(use_dict_output):` we use in a lot of tests ?

deepinv/tests/test_trainer.py
```diff
@@ -568,12 +583,18 @@ def forward(self, y, physics, **kwargs):
...
-@pytest.mark.parametrize("online_measurements", [True, False])
+@pytest.mark.parametrize(
+    "ground_truth, measurements, online_measurements",
+    [
+        (True, True, True),
+        (True, True, False),
+        (True, False, True),
+        (False, True, False),
+    ],
+)
```
romainvo, line 579: Good catch

romainvo (approved):
Thanks @jscanvic , LGTM!

deepinv/tests/test_trainer.py
jscanvic: Good catch! Does 0d6ba9c look like what you had in mind?

romainvo: Yes exactly, much more concise :)!

jscanvic:
@romainvo I rewrote some of the import or skip logic to avoid issues with the absence of the expected warning causes an error

I also added use_dict_output=True to 2 other datasets in the test, leaving only 2 tricky cases that I put in a separate issue #1401

Should be good now

romainvo:
Thanks @jscanvic fro pushing this and handling the changes!! I've just factored the fixture in conftest.py to avoid code duplicate, when tests pass we can merge :)

jscanvic:
Thanks!

PR #1408, opened 2026-09-12
Make shifts implementation parallel

jscanvic:
The current implementation of Shift is sequential in the number of transformations. I suggest to make it parallel

I do so by swapping the more specific torch.roll for the more general torch.gather

I also fix a negligible bug flagged by Claude when shift_max = 0.0 - in that case, we end up with float tensors instead of int tensors due to torch.zeros defaulting to floats (very unlikely to be ever triggered)

**Backwards compatibility**

I made sure the change is backwards compatible - notably:
-  I made sure that it is the params that are interleaved instead of the input (transformations of the same image end up far away from each other in memory) which is the current behavior
- I made sure the zip_longest quirk is preserved

### DIFF
--- a/deepinv/transform/shift.py
+++ b/deepinv/transform/shift.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 from typing import Iterable
-from itertools import zip_longest
 import torch
 from deepinv.transform.base import Transform, TransformParam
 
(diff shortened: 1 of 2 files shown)


Tmodrzyk (approved):
I learned many things reviewing this, LGTM !

PR #1415, opened 2026-09-16
Only run the CI on one device

### LINKED ISSUE #1414: Only use "cuda" devices when running the GPU tests
### Environment

All

### Describe the bug

Spotted by @PAUL-BERNARD , the current `device` fixture in `conftest.py` iterates over all available devices, i.e.

```python
@pytest.fixture(
    params=list(
        dict.fromkeys([torch.device("cpu"), dinv.utils.get_device(verbose=False)])
    )
)
def device(request):
    return request.param
```

Which means that on the GPU runners, the GPU workflow spends a significant amount of time running CPU tests.

### Steps to reproduce

--

### Expected behavior

We should not spend GPU time on CPU tasks and deselect the "CPU" device when running the GPU tests

_No response_

### Additional context

_No response_

PAUL-BERNARD:
Closes #1414

This introduces an environment variable DEEPINV_DEVICE_TEST that can be set to "cpu" or "gpu" to restrict the device list to a single CPU or a CUDA/MPS device.
If no GPU device is available and DEEPINV_DEVICE_TEST is set to "gpu", a warning is raised. 

### LLM policy


deepinv/tests/conftest.py
```diff
@@ -90,11 +92,24 @@ def pytest_terminal_summary(terminalreporter, exitstatus, config):
...
+def get_device_list():
+    env = os.environ.copy()
+    cpu = torch.device("cpu")
+    gpu = dinv.utils.get_device(verbose=False)
+
+    if env.get("DEEPINV_TEST_DEVICE") == "gpu":
+        if gpu.type == "cpu":
+            warnings.warn(
+                "DEEPINV_TEST_DEVICE variable environment is set to 'gpu', but no GPU device was found"
+            )
```
romainvo: I think we should make the CI fail in this case, ie raise an error. Warnings are somewhat hard to track in the logs and a "cuda" device not available on the runners is definitely an issue I think

deepinv/tests/conftest.py
```diff
@@ -90,11 +92,24 @@ def pytest_terminal_summary(terminalreporter, exitstatus, config):
...
+        return [gpu]
+    if env.get("DEEPINV_TEST_DEVICE") == "cpu":
+        return [cpu]
```
romainvo:
```suggestion
    elif env.get("DEEPINV_TEST_DEVICE") == "cpu":
        return [cpu]
```

romainvo (changes requested):
Thanks @PAUL-BERNARD! Very minor comments otherwise looks good :)

deepinv/tests/conftest.py
PAUL-BERNARD: Yes, I just changed the warning to an error.

I also removed places where the device parametrization is configured at the file/test level, so that it uses the fixture instead.

romainvo: Oh nice! Thanks for cleaning this

romainvo (approved):

PR #1379, opened 2026-09-04
Adds ptychography geometry for allowing experimental data

ShantanuKodgirwar:
A continuation of Ref #1353, which fixed the adjoint operator to use the complex conjugate probe.

### Why
To work with experimental data, it requires defining the acquisition geometry, i.e., detector size, pixel pitch, wavelength, sample-to-detector distance. This determines the object-plane pixel size, which allows us to map pixels to physical units and subsequently knowing the spatial resolution. The current implementation does not account for this.

### What changed
This PR does the following:
- Adding `PtychographyGeometry` as a base class from which `FarFieldPtychographyGeometry` and `NearFieldPtychographyGeometry` classes are inherited to describe the experimental setup and calculate the object-plane pixel size and field of view. Far field geometry is the more common case in conventional (lensless) ptychography that assumes the well-known Fraunhofer propagation. The later near-field geometry assumes you have an (unscaled) Fresnel transfer-function/Angular spectrum propogator that will be added by @sebaseb98  in a separate PR. 
- Updating `PtychographyLinearOperator` to support probes smaller than the object through the new `extract_patch` and `place_patch` methods, better matching real scanning experiments.
- Updating `Ptychography` and `PtychographyLinearOperator` to accept an optional `geometry` argument. To maintain backwards compatibility, the current geometry-free API is retained. But maybe in future this could be deprecated.
- Adding tests and API documentation for the new geometry classes and ptychography behavior.

### Discussion
Currently the only thing missing is to update the `examples/demo_ptychography.py` file that still needs to be cleaned up a bit from my side and simplified at places. I could also try to add an experimental dataset example, but I am not sure how to do this with `deepinv` infrastructure in mind. This could be postponed for future too. The current PR now anyway allows swapping for experimental data easily.

@vsechaud as we discussed at the Hackathon, you are currently adding #1351 for Fourier ptychography. You suggested that we could potentially subclass a new `ConventionalPtychography` and `FourierPtychography` from a base `PhaseRetrieval` class and maybe use deepinv's multioperator functionality. We could in principle, then also inherit the `PtychographyGeometry` class for this and `NearFieldGeometry` where detector pixel size is the same as object-plane pixel size which is useful for near-field conventional ptychography / edge-enhanced phase imaging (as was done by Sebastian and Gabriel), but more importantly this is the same geometry used by Fourier ptychography. 

[...]

PR #1418, opened 2026-09-17
add Fresnel phase retrieval and demo

sebaseb98:

Please read our [contributing guidelines](https://deepinv.org/contributing.html) for instructions on how to contribute, including writing tests, documentation and code style.

### LLM policy

- [X] I, a human, wrote code, and LLM tools helped me to write part of the code.

### DIFF
--- a/deepinv/physics/propagation.py
+++ b/deepinv/physics/propagation.py
@@ -1,0 +1,68 @@
+from __future__ import annotations
+import torch
+from torch import Tensor
+from torch.nn.functional import pad
+import numpy as np
+
+from deepinv.physics.forward import Physics
+
+class AngularSpectrumPropagation(Physics):
+    def __init__(self, device, **kwargs):
+        super().__init__(device=device, **kwargs)
+
+
+class FresnelPropagation(AngularSpectrumPropagation):
+    r"""
+    Fresnel propagation operator.
+
+    used to propagate a complex-valued wave field in the near-field regime
+    """
+    def __init__(
+            self, 
+            img_size: tuple[int, int],  # H,W
+            fresnel_number: float, 
+            device: torch.device | str = "cpu",
+            **kwargs,
+        ):
+        super().__init__(device=device, **kwargs)
+        self.img_size = img_size
+        self.fresnel_number = fresnel_number
+        self.device = device
+        xi, eta = torch.meshgrid(
+                    torch.fft.fftfreq(self.data_shape, device=self.device),
+                    torch.fft.fftfreq(self.data_shape, device=self.device),
+                    indexing="ij",
+                )
+        self.kernel_func = torch.exp(
+            (-1j * np.pi) / self.fresnel_number * (xi * xi + eta * eta)
+        ).to(self.device)
+
+    def A(self, x: Tensor) -> Tensor:
+        r"""
+        Fresnel propagation operator.
+        :param torch.Tensor x: input image (complex-valued)
+        :return: (:class:`torch.Tensor`) output measurements
+        """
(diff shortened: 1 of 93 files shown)


sebaseb98:
Hi, I think we are on a stage where merging is not too far. However there are still some open points:

1. The data is currently local, I tried some things to download from [here](https://gitlab.gwdg.de/irp/datasets/dataset-holograms-beads/) in the demo but it did not really work. Maybe we can put it to huggingface with the correct reference
2. At the moment we define the Fresnel Propagation via multiple physical parameters (wave_length, distance, pixel_size). This could also be done using only the Fresnel number, i.e. one parameter and is rather a design choice. From a mathematical standpoint I would say using only one parameter makes more sense, but of course the physical interpretation gets lost 
3. The example is quite minimal at the moment, we could maybe extend it a bit and see how e.g. taking 1 or 2 distances only affects the reconstruction. Style-wise I am not sure about this handcrafted modification of the prox. It works and these physical constraints are necessary but maybe there are better ways to incorporate them.

We (@gabs1234 and I) should also have a look over the propagators and the tests, maybe there is still some generated non-sense :D
When there are bigger changes I would maybe wait for Gabriel to be back in Hamburg (should be in 1-2 weeks), then we can meet and work together, the smaller things I can also handle myself

Andrewwango:
@sebaseb98 I uploaded the data here:
- https://huggingface.co/datasets/deepinv/images/blob/main/holograms_beads.npz
- https://huggingface.co/datasets/deepinv/images/blob/main/holograms_multidistance_hotopy.npz

sebaseb98:
Hi @Andrewwango, Thanks a lot. They are both identical as far as I could see, so if memory/minimalistic file set up is important, you could also delete [holograms_beads.npz](https://huggingface.co/datasets/deepinv/images/blob/main/holograms_beads.npz).

sebaseb98:
I now added the physics operators to the test-lists, I assume we don't need the test-file `deepinv/tests/test_propagation.py` right?

deepinv/physics/propagation.py
```diff
@@ -0,0 +1,297 @@
...
+
+        magnitude = self.H.abs()
+        cutoff = rcond * magnitude.amax()
+        invertible = magnitude > cutoff
+        safe_H = torch.where(invertible, self.H, torch.ones_like(self.H))
+        H_pinv = torch.where(invertible, safe_H.reciprocal(), torch.zeros_like(self.H))
+        return self._apply_transfer_function(y, H_pinv)
+
+
+class FresnelPropagation(AngularSpectrumPropagation):
```
ShantanuKodgirwar, line 212: @sebaseb98 Would you consider adding `FraunhoferPropagation` class inheriting a base class such as `Propagation` or perhaps could also be `AngularSpectrumPropagation`? This could then be unified with #1379 and used for standard ptychography. This way near-field ptycho and far-field ptycho. both would be usable sooner or later and we won't need to use hard-coded FFTs there.

sebaseb98: Hi, I am not to sure if this is really useful, some thoughts:

- is it in the scope of the current PR (what do you think @Andrewwango)?
  in my opinion it could also be added separately as this PR also focusses a lot on the example rather than creating a general and complete wave propagation operator collection.
- are the hard-coded FTs really a problem? It might be unified, but we also add code that is not really needed because the forward operator is simple.

Overall I think the architecture of the propagation is not really solved at this point I think and therefore I am a bit hesitant. There is still this open question of having the physical parameters or rather the Fresnel number and I would appreciate feedback from the maintainers what is the best way forward.

PR #1427, opened 2026-09-22
fix: allow Trainer gradient clipping without gradient checking

### LINKED ISSUE #1387: Trainer crashes with AttributeError when grad_clip is set and check_grad=False

mvanhorn:
Keep `check_clip_grad()` and its clipping/AMP unscale behavior intact, call it as before, and guard the two reads of `check_grad_val.avg` with `self.check_grad` rather than the returned norm. This preserves the public helper return value and makes gradient logging follow the documented opt-in flag without disabling clipping or creating an unnecessary meter. Extend gradient tests in `deepinv/tests/test_trainer.py` to exercise clipping and checking independently, including a small real-backprop CPU training regression and the deferred multi-dataset path where supported; no new production helper is needed.

Setting `Trainer.grad_clip=0.5` with the default `check_grad=False` crashes on the first training step. `check_clip_grad()` returns the clipping norm, but `compute_loss()` treats any returned norm as evidence that `check_grad_val` exists. That meter is initialized and updated only when gradient checking is enabled. The same incorrect assumption appears in `step()` for deferred clipping during scaled multi-dataset training, so both production call sites need correction.

Parameterize clipping (`None`, `0.5`) and checking (`False`, `True`): training completes in all four cases, and `gradient_norm` is emitted only when checking is enabled.
Run a tiny deterministic CPU model and synthetic online denoising data through actual backward/optimizer steps with `check_grad` omitted and clipping enabled. Choose a loss yielding gradients above the threshold; assert nonzero gradients are bounded after clipping and model parameters update, preventing a false pass from disabling clipping.
Preserve existing `test_gradient_norm` assertions for checked, unclipped gradient norms; with checking plus clipping, verify logging still reports the pre-clipping norm and final gradients respect the threshold.
Exercise `optimizer_step_multi_dataset=True` with a real enabled GradScaler on supported hardware, clipping enabled, and each checking value. Assert completion, bounded gradients, and logging behavior; retain normal hardware skips when unavailable. Also cover the unscaled multi-dataset CPU path without introducing production test seams.
Run focused gradient regressions first, then `python -m pytest deepinv/tests/test_trainer.py`; run Black check and Ruff on the two changed Python files. Use installed dependencies only during offline validation, and record any unavailable GPU coverage rather than claiming it passed.

  Not run in this workspace, so this is left for CI to confirm.
  Not run in this workspace, so this is left for CI to confirm.
  Not run in this workspace, so this is left for CI to confirm.

### LLM policy

Fixes #1387


PAUL-BERNARD:
Closing for LLM policy

